Issue #2, opened 2026-08-06
Part I implementation plan


ralexstokes:
The Part I implementation plan: a working companion to `SPEC.md`,
sufficient to start building. Its prerequisite is satisfied: the repository
is bootstrapped, both temporary D.3 guardrails are active on the near-empty
workspace, and the def-construction coherence and trait-spelling compiler
findings are recorded in `SPEC.md` §8 and §4.1. This file begins where code
begins: milestones M1–M5.

`SPEC.md` is the behavioral reference and wins any disagreement; this
file adds *order, module boundaries, and per-milestone obligations*, not
new semantics. Section references (`§n`) and invariant numbers (`§13.n`)
point into `SPEC.md`. The milestone structure and exit criteria are
normative; the module topology and work breakdowns are informative —
reshape them freely if the spike teaches better shapes.

## Ground rules

These apply from the first commit; they are the §1 constraints most easily
lost by drift, restated as working agreements:

1. **Layers in order** (§1): L1 engine → L2 raw actor + mailbox → L3
   handler actor → L4 observation. Each layer complete — including its
   invariant tests — before the next begins. No actor *execution or
   mailbox mechanics* in L1 — §1's boundary is execution, not
   vocabulary: L1 models the child kinds and carries their declared
   policy (mailbox settings included) as plain data.
2. **Pure core, mutable shell** (§1): decision modules are synchronous
   state machines — `step(&mut state, event, &mut impl EffectSink)`; no
   awaits, clock reads, randomness, channel ops, or spawns; `now` and
   jitter samples arrive as data. Only the driver shell touches the `runtime`
   module. During initial development the temporary D.3 clause-2 lint enforces
   this boundary, and M1's exit criterion requires the ladder, funnel,
   intensity, and backoff to be tested by driving the state machine
   directly.
3. **Policies are plain data** (§1): `Clone`/`Debug`/`Eq` structs and
   enums with pure derivation functions only. No trait objects, callbacks,
   or builder side effects.
4. **One primitive per concept** (§1): before writing a second staleness
   check, take-once flag, exit path, or readiness mechanism, find the
   first one. The consuming-owned-value shape (drop = fallback effect) is
   built once and reused: construction closures, cancel guards, readiness
   tokens, exit-report tokens (§4.2, §7).
5. **Spec-first conformance**: every §13 invariant lands as a test in the
   milestone that makes it satisfiable, using the shared test toolkit already
   in the workspace. A milestone is not done because its code exists; it is
[...]

PR #3, opened 2026-08-07
M1: implement supervision engine and tasks

ralexstokes:
Implements M1 of #2 and establishes the base of the M1–M5 stacked PR series.

## What changed

- Added opaque membership/incarnation fencing with fail-closed exhaustion.
- Added validated supervision policies, readiness, shutdown, restart, intensity, and mailbox declaration data.
- Added typed exit classification, the owned report path, pure ladder/funnel/intensity/backoff state machines, and deterministic event arbitration.
- Added the mutable runtime driver, ordered and dynamic tree façades, restartable and one-shot tasks, exact dynamic admission/removal semantics, recursive shutdown, and startup rollback.
- Recorded the resolved §14.2 arbitration order in `SPEC.md`.
- Added the exit-path no-downcast guard and hardened the rustdoc runtime-reachability guard against generated blanket impls.
- Added the complete M1 invariant suite for lifecycle, readiness, dynamic membership, resources, subtree supervision, and task ownership.

## Why

M1 is the runtime-independent supervision and task layer required before mailbox and actor execution can be built. It closes the task/scope portions of the v0 invariants while preserving the runtime boundary and leaving actor mechanics for M2/M3.

## Developer impact

This introduces the initial public tree/task API and the internal engine/driver foundation. Later milestones will stack on this branch: M2 raw actors/mailboxes, M3 handler actors, M4 observation, and M5 acceptance scenarios/docs.

## Validation

- `./scripts/dev just ci` — 80 tests plus fmt, clippy, docs, rustdoc API reachability, runtime-path, exit-path, and nixfmt checks
- `./scripts/dev just ci-nix` — authoritative clean Nix lane


ralexstokes:
## Adversarial review

Reviewed the M1 engine against SPEC (§3, §6, §7, §9, §10, §13). The core design holds up well under adversarial reading — the pure-core/mutable-shell split is genuinely honored (no clocks/awaits/channels in decision modules), the single `Fence` primitive is real (saturating, poisoned terminal, fail-closed `supersedes`), and the arbitration table, intensity edges, readiness gating, and stop ladder all match the spec. The defects found were concentrated in the abnormal-teardown periphery of the driver shell — every one was a path where the driver stops running while shared cells still owe someone a completion.

### Findings (all fixed in 5883050)

1. **Critical — remove() racing a queued split `define` panicked the scope driver.** `handle_admission` consumed the slot via the panicking `take_definition`; a removal that landed between the admission's first poll and the driver dequeuing it left the slot `Lowered`, so the driver died with "a tree was lowered more than once", orphaning the whole scope and hanging the admission future. §13.12 mandates this exact race resolve `NotAdmitting(ReservationEnded)` while the scope keeps admitting. Fixed by consuming via `take_defined` (None on `Lowered` → the mandated rejection). Regression: `removal_of_a_polled_split_definition_keeps_the_scope_admitting` (deterministic; reproduced pre-fix).

2. **Major — hard-aborted scopes stranded their descendants.** `ScopeRuntime::Drop` aborted active children but never terminalized their member cells, and the exit monitor silently discarded its event when the driver's channel was gone. Result: `TaskRef::wait()`/`wait_stopped()` hung forever and `shutdown()` reported success while stranding descendants — §13.7's "exactly one published exit per incarnation, on every path including abort" and §13.2's "terminality is the only hard send failure" both broken. Fixed twice over: `Drop` now terminalizes active children (`Aborted`), and the monitor publishes terminality itself when its send fails (post-join, per §7's two-phase rule). Regression: `hard_aborted_subtree_descendants_still_publish_exits` (hung pre-fix).

3. **Major — lost wakeup for `shutdown_and_wait` on a restarting scope.** Waiters park on the scope signal but exit on *member* terminality; a parent drain that terminalized a scope membership in its restart window pulsed only `member.changed`, never the scope signal → permanent hang. The scope cell now registers its signal for member terminality. Regression: `shutdown_and_wait_wakes_when_a_parent_drain_terminalizes_a_restarting_subtree` (hung pre-fix).

4. **Major — admission/removal responses dropped unresolved.** An `AdmissionRequest` successfully enqueued but never dequeued (driver returned or hard-aborted with events buffered) left its caller awaiting forever. `AdmissionRequest` now resolves its response on drop (`complete` is first-wins, so normal handling is unaffected), and `DynamicControl::close` completes in-flight removals — §13.12's "dynamic mutations resolve on every path".

5. **Minor — incarnation-mint exhaustion wedged startup.** The §3.1 exhaustion path terminalized the membership but never routed the terminal outcome anywhere, so an ordered scope waited forever on its cursor child. Now routed through the same paths as a terminal exit (`fail_startup` / drain advance). Unreachable at u64 scale; fixed because the codebase takes the theoretical contract seriously elsewhere.

6. **Test gaps closed:** `Shutdown::Abort` yielding `Aborted { after_grace: false }` (no grace ever ran), and completion during the tidy beat classifying `Completed` + `cancelled: true` rather than policy-pre-decided `Aborted` (§10/§13.10 edges).

Verified sound (attempted and failed to break): fencing saturation and cross-scope fail-closed comparison, membership-before-placement and reservation terminalization, exit-classification precedence and the report-token consume-or-fallback, intensity window edges (tripping charge counted, trip never spawns), reverse-order teardown with join-before-advance, level-triggered shutdown latches including the pending-incarnation stop latch, per-incarnation readiness re-arm, and Appendix A defaults.

ralexstokes:
Addressed the Codex findings:

- **Publish terminality only after driver cleanup** — real, and it is a facet of the teardown-obligation restructure: tracked on #17 (with an added acceptance criterion: after `shutdown()`/`wait()` returns, no driver task or teardown state remains live). Not patched here because a correct fix is the #17 ordering/join structure, not another ad-hoc sequencing tweak.
- **Mark locally requested subtree shutdowns as cancelled** — checked and the premise doesn't hold on this commit: every self-requested shutdown enters the driver loop through `Pending::Shutdown`, whose intake fires the incarnation's cancel latch before draining (driver.rs, the `Pending::Shutdown` arm), so the `ShutdownRequested → Ok(())` arm in `run_nested_tree` is only ever reached with the latch already fired. Verified empirically: a `ScopeRef::request_shutdown()` stop pre-parent-startup classifies `Completed { cancelled: true }` with no code change. Since nothing pinned this, 8494400 adds `locally_requested_subtree_shutdown_reads_cancelled` as a regression test instead of a second firing site.

PR #4, opened 2026-08-07
M2: add raw actors and membership-owned mailboxes

ralexstokes:
## Summary

- add membership-owned queue/latest mailboxes with restart-transparent send, exact cancellation/timeout withdrawal, and B.3 identity evidence
- add one-budget request/reply calls and consuming reply channels
- add `RawActor`/`RawContext`, raw construction paths, readiness, shutdown drain/discard, and dynamic actor slots
- cover M2 invariants for rebind windows, at-most-once prefixes, ordering/conflation, cancellation/deadline races, and one-shot ownership

## Validation

- `./scripts/dev just ci`
- `./scripts/dev just ci-nix`

Stacked on M1 (#3). Implements the M2 portion of #2.


ralexstokes:
## Adversarial review

Reviewed the M2 raw-actor/mailbox layer against SPEC §5.1, §3, B.3. The mailbox core survived adversarial reading intact: lock ordering is consistent (mailbox → operation, wakers fired outside locks), acceptance/withdrawal is genuinely linearized through the mailbox mutex, the `bind → freeze → close → terminate` lifecycle enforces at-most-once with nothing buffering across incarnation boundaries, capacity accounting has no off-by-one, the full B.3 error identity table is test-pinned, and every send/call future is cancellation-safe at every await point (including the acceptance-wins-at-the-deadline-instant race). All `expect`/`unreachable!` sites are guarded by invariants I could not break.

### Findings (fixed in fecea25 + the M1 merge)

1. **Major — forced teardown left nested memberships permanently non-terminal.** After `force_shutdown` hard-aborted a nested scope, `ScopeRuntime::Drop` froze and closed the mailbox but never terminalized the member, so the mailbox never reached `Terminal`: parked `send`s hung forever and `try_send` reported the *transient* `NotRunning` forever — where B.3 requires `Terminated`, and §3.3 callers legitimately treat `AcceptanceTimedOut` as always-safe-to-retry, retrying a dead membership indefinitely. The driver fix lands via the M1 merge (`MemberCell::terminalize` already cascades to `mailbox.terminate()`, which fails parked waiters correctly); this PR pins the mailbox-visible contract: `forced_teardown_terminalizes_nested_mailboxes` (hung pre-fix).

2. **Test gap — the close path for an accepted-but-undelivered call.** Existing tests covered a handler explicitly dropping a received `Reply` and conflation-away, but not the envelope destroyed when `close` drops the queue at incarnation end. Added `undelivered_call_killed_at_incarnation_close_reports_reply_dropped`, pinning `ReplyDropped` + `Some(accepting)` per §5.1/B.3.

3. **Test gap — withdrawal during the rebind window.** Cancellation of a parked send was only tested against a bound-and-full mailbox, not the unbound restart window where the withdraw/promote race is sharpest. Added `dropping_a_parked_send_in_the_rebind_window_withdraws_it`: a send parked in the frozen window and dropped before the next bind is provably never promoted to the replacement (a later sentinel send proves ordering, not just absence-so-far).

ralexstokes:
Addressed the Codex finding: **run panics no longer unwind through the actor value**. 9d20ccf backports the panic boundary the M3 rework introduces — `RawActor::run` is polled behind `CatchUnwindFuture`, the mailbox freezes, and the actor is dropped under its own `catch_unwind` with the run panic's payload resumed after — so a panicking destructor is a second *contained* panic instead of a process abort. Regression: `raw_run_panic_with_panicking_destructor_publishes_one_report` (SIGABRT'd the test process pre-fix), asserting the published report is `Panicked` with the run panic's payload winning per §7's precedence.

PR #6, opened 2026-08-07
M3: implement handler actors and event resources

ralexstokes:
Stacked on #4. Implements issue #2 milestone M3 (L3 handler actors and the §14.1 spike).

## What changed

- adds the exact `Actor` / `Context` / `StopContext` callback contract plus restartable and one-shot actor definitions across ordered, dynamic, slot, and fused-add surfaces
- implements the public `Handler<A>` raw wrapper with per-incarnation `Args`, declared readiness, `Uninit -> Running` construction, drain/discard behavior, and callback panic containment
- adds the merged RawContext event source: FIFO continuations with fairness, heterogeneous keyed timeouts/intervals with bounded retraction, incarnation-owned offloads and guards, and cancellation-aware blocking work
- wires clean local `ctx.stop()` through the ordinary child stop ladder and freezes intake/resources at the stop boundary
- preserves exact timer/mailbox fire watermarks, teardown-before-actor-state ordering, panic resumption on the actor task, and no cross-incarnation deferred work
- records the accepted §14.1 wrapper/decorator verdict in SPEC.md

## Spike and invariants

Executable coverage includes the shard-store and assistant-control-plane skeletons, raw and handler decorator composition, all four one-shot actor resource paths, declared/manual readiness, exact exit/double-panic behavior, dynamic actor admission, invariant 3 timer/offload fencing, invariant 7 actor panic paths, invariant 12 actor mutation, and invariant 15 for both queue and latest mailboxes.

## Validation

- `./scripts/dev just ci` — passed (140 tests plus fmt, clippy -D warnings, docs, rustdoc/API reachability, runtime-path, exit-path, and nixfmt checks)
- `./scripts/dev just ci-nix` — passed


ralexstokes:
## Adversarial review

Reviewed the M3 handler-actor layer against SPEC §4.3, §5.2–§5.5, §6, §7, B.1. The hard concurrency is right: the fired-timer batch faithfully implements §5.2's priority and anti-starvation rules, timer retraction/rekey is race-free, drain is exactly the frozen accepted prefix on both stop paths, offload fencing across restarts is structural, and panic classification (init vs handle vs continuation vs destructor, double-panic containment) matches §7. Four real defects and one architectural inconsistency, all addressed in d149618:

1. **Major — `Duration::MAX` timers fired immediately.** `now.checked_add(after).unwrap_or(now)` turned clock overflow into deadline-is-now; `set_interval` with a huge period became exactly the "infinite immediate loop" §5.3 exists to forbid (confirmed by executed repro). Deadlines are now `Option<Instant>` with `None` = never, mirroring the offload path. Regression: `overflowing_timer_deadlines_never_fire` (both timeout and interval forms; failed pre-fix).

2. **Major — a `Manual` readiness override on a wrapped `Handler` was silently defeated.** The engine gated on `override.unwrap_or(declared)` while `Handler` consulted its own private field (still `AfterInit`) and auto-fired `mark_ready` after `init` — two unsynchronized sources of truth; `wait_started()` completed with no `mark_ready` ever called (confirmed by executed repro). The engine-resolved mode now rides `RawRunContext` into a public `RawContext::readiness()`, and the blanket loop consults it. Regression: `manual_readiness_override_on_a_wrapped_handler_stays_gated`, including the explicit-release half.

3. **Architectural — `Context::stop()` was built on a `pub(crate)` capability.** §1 principle 5 (a MUST) says the blanket loop may not rely on anything a hand-written raw actor cannot reach, but B.1 marked `stop()` absent from the Raw column — the code resolved the spec's internal tension in the direction that broke the principle. **I resolved it the other way**: `RawContext::stop()` is now public (freeze-at-call semantics, documented as the primitive `Context::stop` is built on) and B.1's row is amended accordingly. ⚠️ This is a deliberate spec + API-surface decision — flagging for maintainer sign-off; the alternative (amending principle 5 with a carve-out) is a one-line revert of the B.1 edit plus re-privatizing the method.

4. **Minor — unbounded offload-resource growth.** Completed offloads' resource entries were only cleared at incarnation end, so a long-lived actor's ledger grew by one entry per offload forever. Finished entries are now pruned on the next `offload` call — O(in-flight) instead of O(ever-issued). (Not separately unit-tested: there is no public resource-count surface, and adding one just for this seemed worse than the fix; happy to add a test-only accessor if you'd rather have it pinned.)

5. **Minor — §5.2's "discarded and reported on exit" for queued continuations was silent, and §5.5's teardown order inverted on the crash path.** Discarded continuations are now reported via `tracing` at the stop freeze with id/incarnation/count. Callback-panic paths now freeze and join incarnation-owned offload work *before* dropping actor state, so §5.5's ordering holds on the crash path too while preserving §7's containment boundary (the destructor still runs outside the callback's unwind). Regression: `incarnation_offloads_are_destroyed_before_actor_state_on_panic`.

ralexstokes:
Addressed the Codex finding: **handler errors now tear down incarnation resources before actor state**. 3377e4f routes the live-loop and drain-loop `Err` outcomes through `fail_after_teardown` — freeze + join of offload work while the actor value is still owned by the frame — so §5.5's teardown order holds on the error path exactly as the earlier fix made it hold on the panic path. Regression: `incarnation_offloads_are_destroyed_before_actor_state_on_error` (recorded actor-before-offload pre-fix).

PR #7, opened 2026-08-07
Implement M4 observation

ralexstokes:
## Summary

Implements issue #2 milestone M4 on top of #6:

- adds restart-stable recursive `ScopeSnapshot` pull/watch observation
- adds per-subscriber bounded lifecycle streams with exact drop-oldest/coalesced `Lagged` accounting
- assigns membership-owned lifecycle sequence watermarks and forwards descendant events with origin identity and subscription-relative paths
- stages snapshot publication before lifecycle delivery, including the zero-snapshot-subscriber pull path
- adds traversal, catch-up watermark lookup, and deadline-bounded `wait_for_child`
- preserves scope streams and sequence continuity across subtree restart while refreshing descendant memberships
- separates terminality from retention pruning so `Removed` is the actual prune edge
- emits lifecycle tracing from one lazy choke point and keeps public runtime reachability clean

## Contract coverage

The M4 tests cover:

- first-start/final-stop event-to-snapshot alignment with zero snapshot subscribers
- two exact lag episodes and fast/slow subscriber isolation
- ancestor forwarding, causal `Added` introduction, per-origin ordering, and recursive watermarks
- subscribe-then-snapshot catch-up, post-lag resync, stale scope-token discard, and post-snapshot scope introduction
- subtree restart windows (`nested: None`, retained `scope_seq`, absolute `restart_at`) and continuous scope sequence
- retained tombstones versus explicit/remove-on-terminal/scope-teardown pruning
- terminal and never-spawned stream/watch closure
- `wait_for_child` later-id, at-or-past, timeout, terminal-child, and terminal-scope outcomes
- undefined reservation invisibility, sequence exhaustion, Arc-shared watch behavior, and public trait bounds

## Validation

- `./scripts/dev just ci` — passed (153 tests, rustdoc reachability and runtime-path guardrails clean)
- `./scripts/dev just ci-nix` — passed


ralexstokes:
## Adversarial review

Reviewed M4 observation against SPEC §12, §3.2, B.4, B.6, §13.14. The observation core is well built — the single `OBSERVATION_GATE` makes state mutation, snapshot projection, watermark stamping, and event staging one atomic edge, which is what makes §13.14's consistent-or-newer guarantee actually hold. Watermark catch-up is exact (subscription, staging, seq stamping, and snapshot computation all under the gate; first seq 1 vs initial watermark 0, so no false already-reflected), lag is drop-oldest with exact coalesced counts and isolated subscribers, origin sequences are gap-free, seq exhaustion fails closed via the shared fence, and pre-admission cells are invisible everywhere. Findings, addressed in fdc74bd:

1. **Major — the hard-abort path bypassed every place terminal edges are emitted.** A hard-aborted scope incarnation (readiness-timeout teardown of an `Abort` subtree, ladder escalation, forced shutdown) left active children with no `Exited`, no `Removed`, a `Stopped` scope snapshot still showing rows with `incarnation: Some` (violating B.6), member cells that never reached terminal (hanging `wait()`), and — on restart — `set_admitted_children` silently `clear()`ing the previous incarnation's residents, so `Added` was never paired (violating §3.2's "both edges or neither"). `ScopeRuntime::Drop` now terminalizes still-active children through `terminalize_child` (publishing `Exited` and closing nested observation) and runs `prune_terminal_members()` (publishing `Removed`) before the scope's final event, mirroring the graceful path; `set_admitted_children` now debug-asserts a clean slate rather than silently clearing. Regression: `hard_aborted_scope_pairs_added_with_exited_and_removed` — asserts the exact pairing, edge order, no-live-rows-under-Stopped, and terminal handle resolution.

2. **Minor — `StartupAborted` misclassification.** (a) Never-ran members surfaced as `StartupAborted { exit: NeverStarted }`, where B.6 places the never-ran terminal under `Stopped` — a member that never started did not have a pre-ready *failure*. (b) The flag was computed per-incarnation (`!was_ready`), so a membership whose first incarnation passed readiness but whose later incarnation was stopped pre-ready during drain terminalized as `StartupAborted` — contradicting this PR's own arm doc. The flag is now `pre_ready` (initial member, startup not complete, initial readiness edge never passed) at the exit site, and `false` on every never-ran path. Regression: `never_ran_members_stop_rather_than_report_startup_abort`.

One observation for a future maintainer thought (no change made): `OBSERVATION_GATE` is a single process-wide mutex, so every transition in every tree serializes with every recursive snapshot computation in the process. It is correctness-sound as written and fine at v0 scale; worth revisiting if multi-tree hosts appear.

ralexstokes:
Addressed both Codex findings in 5a1c13a:

- **descendant() now resolves leaf path endings** — the traversal required a nested snapshot of every component including the last, so any path ending at a task, actor, or restart-window scope returned `None` (even `descendant(["worker"])` for a direct task; existing callers only ever ended paths at scopes, which is why nothing caught it). The nested snapshot is now required only to advance past a component. Regression covers direct-task, scope, through-scope-leaf, missing, and past-a-leaf endings.
- **wait_for_child far-future deadlines stay pending** — the `checked_add` overflow fallback clamped to `now`, expiring a `Duration::MAX` wait immediately; overflow now means a deadline that never arrives, matching the timer/offload rule. Regression reproduced the immediate `TimedOut` pre-fix.

PR #8, opened 2026-08-07
M5: add acceptance scenarios and v0 docs

### LINKED ISSUE #2: Part I implementation plan

ralexstokes:
Closes #2.

## Summary

- port the full core-wave shard-store scenario, including pre/post-commit crash windows, durable reconciliation, idempotent operation ids, incarnation-aware retry, and exact-handle retirement
- port the sidecar scenario with task-first ordered startup, per-child shutdown policy, startup rollback, and two complete host-owned embed/run/stop cycles
- port the assistant control plane with nested dynamic scopes, cyclic slot wiring, two-level panic isolation, latest-value streaming, journal/ack redelivery, lifecycle evidence, idle eviction, and a fenced racing remount
- add v0 operational guides and strengthen public rustdoc for retry, ordering, shutdown/resource ownership, observation catch-up, one-shot construction, blocking work, and embedding

## Core substitutions

Part II-only surfaces are kept out of this core port: peer watches use lifecycle subscriptions, group strategies use `OneForOne`, keyed conflation uses `latest()`/`queue`, and metrics use internal counters.

Every acceptance wait is a bounded lifecycle, snapshot, event, or state poll; none use sleep-and-hope.

## Validation

- `./scripts/dev just ci` — 157 tests passed; Clippy, rustdoc, runtime-path, exit-path, API-reachability, and formatting guardrails clean
- `./scripts/dev just ci-nix` — all authoritative flake checks passed

## Stack

- depends on #7 (M4 observation)
- implements M5 and the v0 documentation exit criteria from #2


ralexstokes:
## Adversarial review

Reviewed the M5 acceptance scenarios and docs against Appendix C, §3.3, B.3, and the implemented API. The strong parts first: every wait in the new tests is a bounded poll/event/snapshot wait (no sleep-based synchronization), the src hunks are rustdoc-only and accurate, docs/observation.md matches `observe.rs` and §12/B.4/B.6 exactly, the retry table matches `CallErrorKind`/B.3 including per-kind `incarnation_observed` rules, and the shutdown/embedding docs match §10/§11/B.9 and the code. The weaknesses were concentrated where Appendix C's obligations are hardest. Addressed in bf649d2:

1. **Major — the shard-store retry loop modeled the anti-pattern §3.3 exists to forbid, and the docs pointed readers at it as the exemplar.** `replace_with_retry` blind-retried in a bare loop: it never awaited a superseding incarnation after `ReplyDropped` (step 2 — the retry relied on exactly the rebind-window parking step 2 says to avoid), folded `ResponseTimedOut` into the same blind-retry arm ("never blindly retry `ResponseTimedOut`" — step 3), and over-asserted `incarnation_observed.is_some()` for `AcceptanceTimedOut` (B.3 allows `None`, a latent CI flake). The helper now hand-rolls the full discipline: one overall deadline, `ReplyDropped` → save the accepting token and `wait_for_child` until a `supersedes`-newer router incarnation runs before resending, `AcceptanceTimedOut` as the only blind-safe retry, and `ResponseTimedOut` a hard failure directing at reconciliation.

2. **Major — C.1's accepted-request quiescence was untested** (the retiring mount received zero traffic before removal). Added `shard_store_retire_waits_for_accepted_requests`: a write accepted and mid-handling when `remove_scope` begins (sequenced via the synchronous `membership_status: Removing` flip from §13.12) completes durably — reply delivered, value persisted — before retirement resolves.

3. **Major — the sidecar startup-failure test couldn't detect an eagerly-torn-down prefix.** Its started-flags latch true at task start, so nothing asserted the C.2 park. It now asserts `config` and `telemetry` are still `Running` in the `StartupFailed` snapshot.

4. **Minor — C.5's "idle eviction" and "cancellable streaming" weren't exercised** (eviction was a bare script-driven `remove_scope`; nothing cancelled a stream mid-flight). Added `assistant_sessions_idle_evict_on_timers_and_streams_cancel_mid_flight` under virtual time: idleness detected by an actor-local keyed timer that activity re-arms (also giving §5.3 replacement semantics acceptance coverage), eviction as session-scope removal, and a stream cut off mid-flight failing senders `Terminated` with no leaked value.

5. **Doc fixes:** retry-and-ordering.md's third `myself()`-deadlock workaround said to "send the request without awaiting it" — a plain `send` future is lazy, so that recipe silently sends nothing; it now prescribes `try_send` (with `Full`/`NotRunning` handling). README's first example discarded the `#[must_use]` `OneShotTaskRef`, greeting new users with a warning; it now binds the tuple.

Note: this PR also received the merged-forward M1–M4 review fixes (driver teardown stranding, mailbox terminality, timer/readiness fixes, observation pairing) — full `just ci` is green at the tip.

ralexstokes:
Follow-up from the review discussion: 97c9e91 adds the spec amendment codifying the teardown-obligation semantics the review fixes implement — §1's third implementation-shape constraint ("Promises are owned completions": destructor discharge, fail-closed, synchronous fallback; residency-emits-`Removed` and one-change-signal-per-cell as normative corollaries), §10's driver-death discharge rule with the coarse kill verdict (`Aborted { after_grace: false }`, `cancelled: true`, first publication wins — post-join precision is deliberately an orderly-path property), and §13.17, the fault-shaped test anchor for the whole class. The structural refactor (shared obligation primitive, RAII residency, single wake signal, abort-user-futures-not-the-driver) is tracked in #17.

ralexstokes:
Addressed all three Codex findings in 62037da:

- `replace_with_retry` now spends every attempt from the one overall budget — remaining time computed before each attempt, the call's acceptance deadline capped to it, loud failure at exhaustion — so no call can run past the logical deadline (§3.3 step 1 taught exactly).
- The quiescence test's handler-entry wait is bounded (1s timeout) so a regression where the accepted write never reaches `handle` fails instead of hanging the run.
- The idle-eviction test now joins root teardown via `system.shutdown(..)` before its runtime drops, honoring the documented runtime-lifetime precondition instead of relying on `System::drop`'s request-only semantics.

PR #21, opened 2026-08-07
Express teardown obligations as owned completions

### LINKED ISSUE #17: Express teardown obligations as owned completions
## Context

The adversarial review of the M1–M5 stack (PRs #3, #4, #6, #7, #8) found one bug class in five coats: **the driver loop stopped running while shared cells still owed someone a completion** — a member cell awaiting terminality, an admission/removal future awaiting resolution, an `Added` awaiting its `Removed`, a waiter parked on the wrong signal. Every fix landed as a patch at a discharge site (a `Drop` impl here, a send-failure fallback there). The invariants enforced by a single owned primitive (`Fence`, `ReportToken`) had no bugs; the invariants enforced by a live event loop remembering to act had all of them.

This issue is the structural follow-up: apply spec principle 3 ("exactly-once effects are expressed by consuming owned values — drop = the fallback effect") to every teardown obligation, so the patch sites become instances of one mechanism instead of hand-maintained special cases. The spec amendment (landing via PR #8) adds the normative language: §1's third implementation-shape constraint ("Promises are owned completions"), §10's driver-death discharge rule with the coarse kill verdict, and §13.17's fault-shaped test anchor.

## The four moves

1. **One `Obligation<T>`-shaped primitive.** An owned completion whose destructor discharges it fail-closed with a **synchronous** fallback (complete/publish/emit/pulse — never await, never join). Retrofit the existing ad-hoc instances onto it:
   - `AdmissionRequest::Drop` → completes `NotAdmitting(Terminal)` (added in review)
   - `DynamicControl::close` completing in-flight removals (added in review)
   - `ReportToken` (pre-existing — the model citizen)
   - per-child terminality: the guard owned by `ScopeRuntime` whose drop terminalizes with the kill verdict — replaces the hand-written active-branch logic in `ScopeRuntime::Drop` and the exit monitor's send-failure fallback

2. **Residency is ownership.** A resident entry in a scope's observed child set emits `Removed` on drop, making §3.2's exact `Added`/`Removed` pairing structural. Replaces the hand-sequenced `terminalize_child` + `prune_terminal_members()` calls in `ScopeRuntime::Drop` and makes the `set_admitted_children` clean-slate `debug_assert` unnecessary (clearing *is* emitting).

3. **One wake primitive per cell.** Terminality is the single condition; every compound wait (`shutdown_scope`, `wait_for_incarnation`) derives from the member cell's one change signal. Replaces the review patch that registers the scope signal as a member terminal-signal — collapse to one signal so there is no second wake path to forget.

4. **Hard abort kills user futures, not framework loops.** The ladder's abort delivers an abort-now command with a hard deadline to the nested driver, whose epilogue is synchronous effect-execution (fire latches, abort children, discharge obligations) — always able to finish promptly because the pure-core/mutable-shell split keeps it await-free. Task-abort of the driver remains only as a backstop for a wedged executor; with moves 1–2 the backstop is safe regardless.

## Acceptance criteria

- Each review patch site is replaced by (or reduced to an instance of) the shared mechanism; no hand-maintained per-obligation `Drop` logic remains in `driver.rs`.
- §13.17's provocations pass: under hard abort of an ancestor with each obligation class provably outstanding, every completion resolves bounded — waits yield `Aborted`, sends fail `Terminated`, add/remove futures yield their enumerated rejections, streams pair every `Added`, no stopped-scope snapshot carries a live incarnation.
- Stretch: a fault-injection harness that destroys the driver at arbitrary await points and runs the same assertions (§13.17 names the fixed provocations as the floor, not the ceiling).

## Regression tests already pinning the behavior

`removal_of_a_polled_split_definition_keeps_the_scope_admitting`, `hard_aborted_subtree_descendants_still_publish_exits`, `shutdown_and_wait_wakes_when_a_parent_drain_terminalizes_a_restarting_subtree`, `forced_teardown_terminalizes_nested_mailboxes`, `dropping_a_parked_send_in_the_rebind_window_withdraws_it`, `hard_aborted_scope_pairs_added_with_exited_and_removed`, `never_ran_members_stop_rather_than_report_startup_abort` — the refactor must keep all of these green; they are the executable form of the invariant.

ralexstokes:
Closes #17.

## What changed

- Added one generic owned `Obligation<T>` completion primitive and moved admission, removal, exit-report, and per-child terminality fallbacks onto it.
- Made observed child residency own its matching `Removed` edge, including restart and driver-death cleanup.
- Collapsed scope and typed one-shot waits onto each member cell's single change signal.
- Routed nested hard abort through an acknowledged recursive driver drain: scheduled framework drivers join their children, while task abort remains the pre-ack deadline backstop.
- Made `System::shutdown` / `System::wait` join root lifecycle teardown and preserved removal obligations until member terminality and residency cleanup.

The issue's four structural moves are represented in issue order; two follow-up commits address adversarial review findings.

## Why

The previous driver loop was itself the guarantor for several teardown completions. If the loop was destroyed with work outstanding, callers, mailboxes, lifecycle streams, and scope waiters could be stranded or observe terminality before teardown had actually finished. Ownership now guarantees the fail-closed fallback, while the event loop remains the precise orderly path.

The recursive abort path also needed an explicit acknowledgement boundary. Without it, a parent could return after task-aborting a deeper framework driver but before that driver and its child futures were actually joined.

## Impact

Hard-aborted or otherwise destroyed drivers discharge outstanding framework obligations. `Added`/`Removed` pairing and member terminality are structural, compound waits have one wake source, removers wake only after the membership is no longer resident, and owning shutdown APIs join recursively acknowledged driver teardown.

## Validation

- `./scripts/dev just ci`
- `./scripts/dev just ci-nix`
- All 180 workspace tests pass, including the seven regressions listed in #17 plus root-driver teardown, removal-ordering, and recursive multi-threaded join regressions.

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -1211,6 +1211,7 @@ impl ScopeCell {
 
 pub(crate) struct SystemRun {
     pub(crate) root: Arc<ScopeCell>,
+    driver: Option<runtime::JoinHandle<(), StopReason>>,
 }
 
 pub(crate) struct AdmissionResponse {
@@ -1427,9 +1428,7 @@ pub(crate) fn start_admission(
         }),
     };
     runtime::spawn((), async move {
-        let _ = runtime::mpsc_send(&control.events, DriverEvent::Admission(request))
-            .await
-            .is_err();
+        let _ = runtime::mpsc_send(&control.events, DriverEvent::Admission(request)).await;
     });
     response
 }
@@ -1530,8 +1529,38 @@ impl SystemRun {
         self.root.request_shutdown();
     }
 
-    pub(crate) async fn shutdown(&self, timeout: Duration) -> Result<(), ShutdownTimeout> {
-        shutdown_scope(Arc::clone(&self.root), timeout).await
+    pub(crate) async fn shutdown(&mut self, timeout: Duration) -> Result<(), ShutdownTimeout> {
+        let result = shutdown_scope(Arc::clone(&self.root), timeout).await;
+        self.join_driver().await;
+        result
+    }
+
+    pub(crate) async fn wait(&mut self) -> StopReason {
+        let reason = self.root.wait_stopped().await;
+        self.join_driver().await;
+        reason
+    }
+
+    async fn join_driver(&mut self) {
+        let Some(driver) = self.driver.take() else {
+            return;
+        };
+        match runtime::join(driver).await {
+            runtime::JoinOutcome::Ok { .. } => {}
+            runtime::JoinOutcome::Panic { message, .. } => {
+                let exit = Exit::new(ExitKind::Panicked { message }, false);
+                self.root.set_startup(Err(StartupError::ShutdownRequested));
+                self.root
+                    .finish_live_root_incarnation(StopReason::ShutdownRequested, exit);
+            }
+            runtime::JoinOutcome::Cancelled { .. } => {
+                self.root.set_startup(Err(StartupError::ShutdownRequested));
+                self.root.finish_live_root_incarnation(
+                    StopReason::ShutdownRequested,
+                    Exit::new(ExitKind::Aborted { after_grace: false }, true),
+                );
+            }
+        }
     }
 }
 
@@ -1614,14 +1643,15 @@ pub(crate) async fn shutdown_scope(
 pub(crate) fn spawn_system(plan: ScopePlan) -> SystemRun {
     let root = Arc::clone(&plan.root);
     let monitor_root = Arc::clone(&root);
-    let handle = runtime::spawn((), async move { run_scope(plan, true, None).await });
-    runtime::spawn((), async move {
-        match runtime::join(handle).await {
-            runtime::JoinOutcome::Ok { .. } => {}
+    let driver = runtime::spawn((), async move { run_scope(plan, true, None).await });
+    let lifecycle = runtime::spawn((), async move {
+        match runtime::join(driver).await {
+            runtime::JoinOutcome::Ok { value, .. } => value,
             runtime::JoinOutcome::Panic { message, .. } => {
                 let exit = Exit::new(ExitKind::Panicked { message }, false);
                 monitor_root.set_startup(Err(StartupError::ShutdownRequested));
                 monitor_root.finish_live_root_incarnation(StopReason::ShutdownRequested, exit);
+                StopReason::ShutdownRequested
             }
             runtime::JoinOutcome::Cancelled { .. } => {
                 monitor_root.set_startup(Err(StartupError::ShutdownRequested));
@@ -1629,10 +1659,14 @@ pub(crate) fn spawn_system(plan: ScopePlan) -> SystemRun {
                     StopReason::ShutdownRequested,
                     Exit::new(ExitKind::Aborted { after_grace: false }, true),
                 );
+                StopReason::ShutdownRequested
             }
         }
     });
-    SystemRun { root }
+    SystemRun {
+        root,
+        driver: Some(lifecycle),
+    }
 }
 
 enum ChildEvent {
@@ -1684,6 +1718,8 @@ struct ActiveChild {
     readiness: ReadinessGate,
     ready_signal: Latch,
     construction_release: Latch,
+    framework_abort: Option<Latch>,
+    framework_abort_deadline: Option<Instant>,
 }
 
 struct ChildTerminality {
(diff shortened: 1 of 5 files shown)


ralexstokes (commented):
## Review

**Overview.** The PR replaces implicit driver-loop guarantees with an owned `Obligation<T>` primitive (drop runs a fail-closed fallback), makes child residency own its `Removed` edge, collapses one-shot/scope waits onto the member's single change signal, and routes nested hard abort through a synchronous `ScopeRuntime` drop epilogue with task abort retained as the deadline backstop. The design is sound and the invariants are well-commented; I verified the load-bearing paths rather than just the diff:

- **Observation-gate discipline** — every place a `ResidentChild` can drop (`prune_child`, `clear_residents`, `begin_children`) holds `OBSERVATION_GATE`, and `clear_residents_locked` correctly releases the child-set mutex before the drops so `emit_locked → snapshot_locked` can't self-deadlock. `ScopeCell` teardown is covered by the weak-parent upgrade failing.
- **Ladder/backstop interaction** — after `HardAbort` the ladder clears its deadline, so `ladder_deadline.or(framework_abort_deadline)` always schedules the framework-abort beat, and the next `advance_ladder` fires `abort_handle.abort()`. No livelock; `force_all` re-arming the ladder can push the backstop by at most one beat.
- **Signal unification** — `MemberCell::terminalize` pulses `changed`, and `OneShotTaskRef::wait` checks both completion state and member terminality against that one signal, so removing `terminal_signals` loses no wakeups.
- **Fail-closed coverage** — queued `DriverEvent::Admission` values dropped on driver death (including the `AncestorAbort` early return) resolve their callers via the obligation fallback, and `discharge_child_terminality` is a no-op on already-terminal members, so stale armed obligations are harmless.

**Finding (fixed in 9423a2f).** Dropping a `DynamicEntry` now *completes its removal response* — a side effect the pre-PR code didn't have — and three paths dropped the entry before the member's observation edges were published:

1. `remove_dynamic`'s terminal-stage path: the bare `state.entries.remove(id)` completed the response while the dynamic-state mutex was held and **before** `prune_child` emitted the `Removed` edge. The old code deliberately completed after pruning; a remover awaiting on another worker could wake and snapshot the child still resident.
2. `prune_terminal`: same pattern, reachable when an in-flight removal races a terminal exit under `Retention::Remove`.
3. `cancel_dynamic_reservation`: entry dropped before the member terminalized (unreachable by waiters today since unadmitted entries can't leak their removal Arc, but fixed for a uniform invariant).

All three now hold the removed entry across `prune_child`/terminalize and drop it last, matching `finalize_removal`'s existing ordering. No deterministic regression test is possible under a current-thread runtime (the window only exists for a concurrently scheduled waiter), so the fix is ordering-only. `just ci` passes: 178/178.

**Minor notes, no action needed:**
- `SystemRun::join_driver`'s `Panic`/`Cancelled` arms duplicate the lifecycle monitor's handling; they only fire if the monitor task itself dies, and `finish_live_root_incarnation` guards re-entry, so it's harmless defense.
- On the synchronous drop epilogue, grandchild scope drivers are still torn down via async task abort (their own epilogues run when tokio drops the futures) — the obligations keep that fail-closed, so the one-level-synchronous bound seems intentional and fine.

ralexstokes (commented):
## Adversarial follow-up review

I found two actionable teardown bugs and one same-class ordering hole; all are fixed in `10d987c`.

1. **Owning shutdown only joined one level of recursive hard-abort teardown.** A nested driver acknowledged ancestor abort by returning immediately. Its `ScopeRuntime::drop` task-aborted deeper scope drivers but could not join them, so the root lifecycle join could complete while a grandchild driver or child future was still executing. I reproduced this on a two-worker runtime with a deeply nested leaf deliberately in-progress inside `poll`: `System::shutdown` returned before that leaf future and its construction were dropped. This violated both the public “every actor future is still joined” contract and #17’s explicit owning-shutdown acceptance criterion.

   The fix adds an abort acknowledgement per nested framework driver. A driver that receives the command now recursively force-drains and joins its children; the parent’s tidy-beat task abort remains armed only until that acknowledgement, so it is still the backstop for a driver that never runs.

2. **Dynamic teardown completed removals before the removal was observable.** `DynamicControl::close` dropped all `DynamicEntry` obligations before `ScopeRuntime::drop` terminalized active members or cleared residency. On a multi-thread runtime, a removal waiter could therefore wake while its member was still live/resident and before the matching `Removed` edge. The close path now returns and holds the entries through child terminality and residency cleanup, then drops them to complete removals.

3. **Queued fused-admission cancellation had the same early-drop shape.** Its unadmitted entry was discarded under the dynamic-state lock before member terminality. That response is not currently exposed to a waiter in this branch, but preserving the invariant prevents the next API change from making the latent ordering observable; the entry is now held until terminal publication.

New regressions cover removal-completion ordering and a multi-threaded, recursively aborted driver join. Validation is green:

- `./scripts/dev just ci` — 180/180 tests plus fmt, clippy, docs, API reachability, and path checks
- `./scripts/dev just ci-nix` — all clean-flake checks pass

I found no further actionable issues in the changed obligation, residency, signal, recursive-abort, or owning-join paths after these fixes.

PR #38, opened 2026-08-07
Use Tokio Notify for runtime latches

### LINKED ISSUE #29: Evaluate replacing the custom Latch with Tokio cancellation primitives
## Context

`driver::Latch` currently implements a one-shot, multi-waiter signal using an `AtomicBool` plus the custom `Signal`, `Waiter`, and `WaiterFuture` machinery. Shelterwood's public `CancellationToken` composes one or two latches to represent supervisor shutdown and optional operation-local cancellation.

PR #28 removes an unrelated, unused `tokio_util::sync::CancellationToken` wrapper and the resulting unused dependency. That cleanup raised the broader question of whether Shelterwood should rely more on Tokio's maintained concurrency primitives instead of carrying bespoke waiter bookkeeping.

## Goal

Evaluate whether the private `Latch` implementation can be replaced by a Tokio-backed abstraction, with `tokio_util::sync::CancellationToken` as the leading candidate. Keep the existing public `shelterwood::CancellationToken` API and Shelterwood's runtime-boundary discipline.

## Required semantics

- firing is idempotent and reports whether this call performed the transition
- all current waiters wake after firing
- future waiters complete immediately after firing
- waiting remains cancellation-safe
- a derived operation token is cancelled by either supervisor shutdown or operation-local cancellation
- cancelling an operation-local token does not cancel its supervisor token
- shutdown, abort, and deadline arbitration behavior remains unchanged
- Tokio and Tokio-util paths remain confined to the runtime module

A direct `tokio::sync::Notify` substitution is not sufficient without additional state: `notify_one` wakes only one waiter, while `notify_waiters` does not retain a notification for future waiters. `tokio_util::sync::CancellationToken::child_token` appears closer to the required parent/child behavior, but its cancellation propagation and ordering must be checked against Shelterwood's lifecycle rules.

## Acceptance criteria

- document the chosen primitive and semantic mapping, including any reason to retain the custom implementation
- if migrating, route the implementation through the runtime boundary without exposing Tokio types in the public API
- preserve the current public `CancellationToken` behavior
- add focused concurrent tests for pre-fire waiters, post-fire waiters, cancelled waits, and parent/local cancellation races
- remove `Signal`, `Waiter`, and `WaiterFuture` only if they have no remaining consumers
- compare dependency and performance costs rather than assuming an improvement
- pass `./scripts/dev just ci`

Related: #28

ralexstokes:
## Summary

- move the private one-shot `Latch` behind the runtime boundary
- replace its custom `Signal` waiter bookkeeping with one shared `AtomicBool + tokio::sync::Notify` state
- preserve the public `CancellationToken` primary/secondary OR-composition and all transition semantics
- add focused concurrency coverage for idempotent fire, pre- and post-fire waiters, dropped waits, and parent/local cancellation races

## Design

`tokio_util::sync::CancellationToken` is not a direct semantic match: it does not report the winning cancellation transition, merges parent and local causes, and would reintroduce cancellation-tree allocation/locking plus the dependency set removed by #28. `tokio::sync::watch<bool>` would put locking on the hot `is_fired` path.

The selected design keeps a linearizable atomic transition and retained fired state while delegating waiter registration, broadcast wakeup, and cancelled-wait cleanup to Tokio's `Notify`. Creating `notified()` before rechecking the atomic closes the lost-wakeup window. Existing `Signal` machinery remains for repeatable generation notifications, where it still has consumers.

This changes no public API and adds no dependency.

## Validation

- `./scripts/dev just ci`
- 200 tests passed
- public API runtime reachability: clean
- runtime module path restriction: clean

Closes #29

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -30,7 +30,7 @@ use crate::{
     },
     policy::{DefaultsInheritance, ResolvedDefaults},
     raw::{RawRunContext, RawSpawn},
-    runtime,
+    runtime::{self, Latch},
     task::{OnceTaskBody, TaskContext, TaskFactory},
     tree::{
         BuilderCore, ChildConstruction, ChildPlan, NotAdmittingCause, RemoveOutcome, ReserveError,
@@ -276,38 +276,6 @@ impl Future for WaiterFuture {
             Poll::Ready(())
         } else {
             Poll::Pending
-        }
-    }
-}
-
-#[derive(Clone, Debug, Default)]
-pub(crate) struct Latch {
-    fired: Arc<AtomicBool>,
-    signal: Signal,
-}
-
-impl Latch {
-    pub(crate) fn fire(&self) -> bool {
-        if self
-            .fired
-            .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
-            .is_ok()
-        {
-            self.signal.pulse();
-            true
-        } else {
-            false
-        }
-    }
-
-    pub(crate) fn is_fired(&self) -> bool {
-        self.fired.load(Ordering::Acquire)
-    }
-
-    pub(crate) async fn fired(&self) {
-        let mut watcher = self.signal.watcher();
-        while !self.is_fired() {
-            watcher.changed().await;
         }
     }
 }
@@ -3363,11 +3331,12 @@ mod tests {
         exit::RecordedOutcome,
         identity::{FenceCounter, ScopeIdentity},
         mailbox::MailboxCell,
+        runtime::Latch,
         tree::{SlotCell, lower_tree_for_test},
     };
 
     use super::{
-        DynamicControl, DynamicEntry, Latch, MemberCell, MemberStage, Obligation, RemovalResponse,
+        DynamicControl, DynamicEntry, MemberCell, MemberStage, Obligation, RemovalResponse,
         ScopeCell, ScopeFlavor, mint_child_incarnation, report_channel, run_scope_incarnation,
     };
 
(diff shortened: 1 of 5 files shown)


ralexstokes:
## Review

Reviewed the full diff, read the surrounding code at head `8dd4c87`, and re-ran `./scripts/dev just ci` locally (200/200 tests, all lanes clean). **No changes requested.**

### Correctness

- `Latch::fired()` follows Tokio's documented `Notify` pattern exactly: create the `Notified` future, then recheck the flag, then await. Combined with `fire()` setting the atomic (AcqRel CAS) *before* `notify_waiters()`, every interleaving is covered — either the recheck observes `fired == true`, or the `Notified` was created early enough that tokio guarantees it observes the subsequent `notify_waiters()` (guarantee present in the pinned tokio 1.53.1).
- The one-shot transition report (`fire() -> bool`) is preserved verbatim, so `CancellationToken` primary/secondary OR-composition and `TaskContext::mark_ready` semantics are unchanged.
- `fired()` is cancel-safe: `notify_waiters` grants no permits, and dropped `Notified` futures unlink from the intrusive queue — the new `cancelled_waits_do_not_consume_the_signal` test pins this down.
- The outer `loop` in `fired()` can never iterate twice today (a wake implies the flag is set), but it is harmless as defense against future notify sources.

### Boundary & conventions

- Moving `Latch` into `runtime.rs` is the right home under the "only boundary to the async runtime" rule; `task.rs` tests correctly go through `#[crate::runtime::test]` and `runtime::{spawn, join, timeout, yield_now}` rather than touching tokio, and the module-path restriction lane stays clean.
- No dead code left behind: `Signal`/`SignalWatcher` retain real consumers (mailbox, observe, task completion), and imports were tightened at every call site (`driver`, `raw`, `task`, `tree`).
- The doc comment honestly records why `tokio_util::CancellationToken` and `watch` were rejected — useful context for the next reader.

### Tests

- Coverage hits the right axes: concurrent-fire linearizability (32 racing firers), pre-fire waiter broadcast (32 parked waiters, first-poll-pending asserted), post-fire immediate completion, cancelled-wait cleanup (1024 dropped waits), and the parent/local cancellation race repeated 128× on a multi-thread runtime. All spin loops yield cooperatively, so none can starve the runtime.

Verified locally with `./scripts/dev just ci` at 8dd4c87 — exit 0.

Issue #37, opened 2026-08-07
Reduce hand-rolled Arc<Mutex> synchronization in favor of standard primitives


ralexstokes:
## Context

#29 evaluates replacing the custom `Latch` waiter machinery with Tokio primitives. The same theme — hand-rolled shared-state synchronization where a maintained primitive models the pattern directly — recurs across the codebase at larger scale: roughly 140 `Mutex` sites, with 83 `lock()` calls in `driver.rs` alone.

Current hygiene is good and this is not a correctness fire: everything is `std::sync::Mutex` (correctly, not `tokio::sync::Mutex`), critical sections are short lock-copy-release, and no lock is held across an `.await`. The motivation is the same as #29: fewer bespoke concurrency invariants to reason about and test, not performance — the supervision control plane is low-traffic.

The usage decomposes into four clusters with different verdicts.

### Cluster 1 — Sync-shim factories

`TaskFactory` (`task.rs:19`), `ArgsFactory` (`actor.rs:447`), `ScopeFactory` (`tree.rs:1435`), `RawFactory` (`raw.rs:1243`), and the factory in `raw.rs:1071` are all `Arc<Mutex<Box<dyn Fn() -> _ + Send>>>`. The `Mutex` exists only because `Box<dyn Fn + Send>` is not `Sync`; it is a shareability shim, not synchronization. Tightening the bound to `Fn + Send + Sync` allows plain `Arc<dyn Fn ...>` — no `Mutex`, one less indirection. The cost is a stricter public bound on user-supplied closures, which nearly all closures satisfy. If accepting non-`Sync` closures was a deliberate API decision, keep the shim and document it at the type alias.

### Cluster 2 — Hand-rolled oneshots

`Mutex<Option<Result<..>>>` plus a latch/obligation for single-delivery results: `driver.rs:1218`, `driver.rs:1257`, and `CompletionState` (`task.rs:341`). These are oneshot channels built by hand; `tokio::sync::oneshot` models them directly, routed through the runtime boundary per the discipline established in #29.

### Cluster 3 — State cell + `Signal` pairs

[...]

PR #40, opened 2026-08-07
Replace synchronization-only factory locks

ralexstokes:
## Summary

- tighten restartable task, actor, raw-actor, and subtree factories from `Fn + Send` to `Fn + Send + Sync`
- store the factories directly as `Arc<dyn Fn + Send + Sync>`
- invoke the shared closures without a `Mutex` or boxed-inner indirection

## Semantic mapping

The removed mutexes did not protect mutable state: they existed only to make `Box<dyn Fn + Send>` shareable. Requiring `Sync` on the factory closure makes that shareability contract explicit and preserves repeatable `Fn` invocation semantics. The issue records why this breaking bound is appropriate now and why thread-per-core support would require a separate `!Send` placement mechanism rather than these mutex shims.

`ActorDef::cloned` consequently requires `Args: Sync`, because its factory closure shares the captured clone source.

## Impact

This removes five synchronization-only factory wrappers and their poison-handling paths. Runtime behavior is unchanged for `Sync` closures; closures using `Send`-only interior mutability must move synchronization into their captured state explicitly.

## Validation

- `./scripts/dev just ci`
- 200 tests passed
- public API runtime reachability: clean
- runtime module path restriction: clean
- exit-path downcast restriction: clean

Part of #37 (Cluster 1).

### DIFF
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -6,7 +6,7 @@ use std::{
     hash::Hash,
     marker::PhantomData,
     panic::{AssertUnwindSafe, catch_unwind, resume_unwind},
-    sync::{Arc, Mutex},
+    sync::Arc,
     time::Duration,
 };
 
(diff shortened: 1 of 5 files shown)


ralexstokes:
Fresh-agent adversarial review completed against `main...6103c12`.

Reviewed all five factory paths, public `Fn + Send + Sync` bounds (including `ActorDef::cloned`), restart/concurrency semantics, documentation, and existing coverage. No actionable findings were found, so no review-fix commit was needed.

Validation: `./scripts/dev just ci` (200 tests plus formatting, clippy, docs, API/path checks) and `git diff --check main...HEAD` both pass.

Issue #35, opened 2026-08-07
Hardening of Part I


ralexstokes:
- [x] **1. Reclaim historical runtime state under dynamic churn** (#53)

  Problem

  Long-lived scopes retain state that is no longer live:

  - Removed dynamic children remain in ScopeRuntime.children.
  - Resolved readiness, stop, and restart deadlines remain in DeadlineQueue until their original deadline.
  - Cancelled Signal::changed() waiters remain registered until another pulse.

  This makes memory and scan costs proportional to historical activity rather than current residents.

  Proposed direction

  Use removable, generation-keyed registrations for children and deadlines, and make waiter cancellation retract its registration.

  Acceptance

  - High-cycle add/remove workloads keep runtime storage bounded.
  - Prompt readiness/removal cancels or compacts stale deadlines.
  - Repeatedly cancelling changed() during a quiet period does not grow waiter storage.
  - Stale events cannot target a reused slot.

  ———

  - [x] **2. Preserve membership supersession across nested scopes** (#44)

  Problem

  Declared children of a nested scope mint memberships from the builder’s temporary identity. Runtime additions mint from the stable parent-owned scope identity.

  Consequently:

  - Re-adding an initially declared ID may produce a replacement that does not supersede the original.
  - Descendants before and after a nested-scope restart can be incomparable.
  - Retry logic can reject a valid current replacement as stale.
  - docs/retry-and-ordering.md does not accurately describe all replacement cases.

  Proposed direction

  Mint or rebase all child memberships from the stable scope identity during lowering.

  Acceptance

  - A nested dynamic replacement supersedes the removed membership.
  - Corresponding descendants after a nested-scope restart supersede their predecessors.
  - Unrelated scopes and different IDs remain incomparable.
  - Documentation and tests describe the resulting ordering consistently.

  ———

  - [x] **3. Keep parked mailbox operations’ incarnation evidence current** (#42)

  Problem

  When a bounded mailbox binds, only waiters promoted into available capacity observe the new incarnation. Overflow waiters remain parked with stale evidence.

  If one times out, send_timeout or call may report an older incarnation or None, despite having remained pending while the new incarnation was bound.

  Proposed direction

  Binding should update observation evidence for every live parked operation independently of promotion.

  Acceptance

  - With capacity one and multiple pre-bind waiters, overflow timeouts report the bound incarnation.
  - The same holds across a rebind.
  - Cover both send_timeout and call.
  - Cancellation and promotion races return consistent evidence.

[...]

PR #42, opened 2026-08-07
Refresh parked mailbox incarnation evidence

ralexstokes:
Part of #35 (item 3)

## Summary

- Record each bind or rebind incarnation on every live parked mailbox operation, including FIFO overflow that cannot yet be promoted.
- Preserve timeout-versus-promotion consistency through the mailbox/operation lock ordering.
- Add deterministic coverage for capacity-one pre-bind overflow, rebind overflow, `send_timeout`, `call`, and deadline/promotion races.

## Why

Binding previously updated incarnation evidence only for waiters promoted into available capacity. Operations that remained parked could therefore time out with `None` or an older incarnation even though they stayed pending through a newer bind.

## Validation

- `./scripts/dev cargo test -p shelterwood --test integration` (161 passed)
- `./scripts/dev just ci` (210 tests passed; formatting, Clippy, docs, API/runtime/exit path checks, and Nix formatting all passed)

### DIFF
--- a/crates/shelterwood/src/mailbox.rs
+++ b/crates/shelterwood/src/mailbox.rs
@@ -932,6 +932,14 @@ impl<M: Send + 'static> MailboxControl for MailboxCell<M> {
         }
         state.status = BindingStatus::Bound(incarnation);
         state.last_bound = Some(incarnation);
+        // Binding is an observation edge for every operation that remained
+        // parked through it, including FIFO overflow that cannot be promoted
+        // into the current capacity. Withdrawal takes the mailbox lock before
+        // the operation lock, so a concurrent timeout sees either the prior
+        // evidence or this incarnation consistently with which edge won.
+        for operation in &state.waiters {
+            operation.observe(incarnation);
+        }
         let promotion = promote_waiters(&mut state);
         drop(state);
         self.changed.pulse();
--- a/crates/shelterwood/tests/deadlines.rs
+++ b/crates/shelterwood/tests/deadlines.rs
@@ -71,6 +71,54 @@ fn boundary_actor(
 }
 
 #[tokio::test(start_paused = true)]
+async fn prebind_overflow_timeouts_observe_the_bound_incarnation() {
+    let gate = ReleaseGate::default();
+    let values = Arc::new(Mutex::new(Vec::new()));
+    let calls = Arc::new(AtomicUsize::new(0));
+    let mut tree = Tree::new();
+    let actor = tree
+        .add_raw_once(
+            "prebind-overflow",
+            RawOnceDef::new(boundary_actor(Some(gate.clone()), &values, &calls, false))
+                .mailbox(Mailbox::queue(1).expect("non-zero capacity")),
+        )
+        .expect("valid actor");
+    let width = Duration::from_secs(10);
+    let mut accepted = Box::pin(actor.send_timeout(Message::Value(1), width));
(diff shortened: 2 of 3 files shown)


ralexstokes:
Adversarial review completed against current `origin/main` and issue #35 item 3.

Scope reviewed: bind/rebind observation semantics; mailbox/operation race linearization and lock order; timeout withdrawal, cancellation, promotion, waker/drop behavior; `send_timeout`/`call` consistency; test determinism and coverage; docs/API impact.

Finding addressed:
- The rebind acceptance test registered its FIFO waiters through spawned tasks followed by a single `yield_now`, making the setup scheduler-dependent. Commit `93e3f7a` replaces that with explicit one-poll registration and direct deterministic observation of promotion and timeout outcomes.

No further actionable findings remain. Production bind/withdraw/promotion paths consistently use mailbox-state → operation-state ordering, release wakers after the mailbox lock, and preserve the correct acceptance-versus-timeout result.

Validation:
- Focused item-3 tests: passed
- `./scripts/dev cargo test -p shelterwood --test integration`: 161 passed
- `./scripts/dev just ci`: 203 passed; formatting, Clippy, doc tests/docs, API reachability, runtime/exit-path checks, and Nix formatting passed

ralexstokes:
Rebased onto current `main` at `7c401f5` (`93e3f7a` → `f780e62`) with no conflicts. Post-rebase validation: `git diff --check`, the 161-test integration suite, and `./scripts/dev just ci` (210 tests plus formatting, Clippy, docs, API/path, and nixfmt checks). The remote branch was updated with an exact `--force-with-lease`.

ralexstokes (commented):
## Review — LGTM, no findings

**What it does:** makes every bind/rebind an observation edge for *all* parked waiters (including FIFO overflow that can't be promoted), so a later timeout reports the newest incarnation the operation actually survived through instead of `None`/stale evidence.

### Correctness

- **Lock ordering claim verified.** The comment in `bind` says withdrawal takes the mailbox lock before the operation lock — confirmed: `withdraw` (mailbox.rs:848→850), `submit`, `bind`, and `promote_waiters` all acquire mailbox → operation, never the reverse. No deadlock, and the bind-vs-timeout race is fully serialized by the mailbox lock: a withdrawn operation is removed from `waiters` under the same lock, so `bind` never observes a stale entry (and `observe` is a no-op on non-`Waiting` outcomes anyway).
- **Idempotence.** Waiters promoted immediately after the loop are observed twice (once in `bind`, again in `promote_waiters`) with the same incarnation — harmless; `observe` only overwrites `newest_observed`.
- **Coverage of all park paths.** Waiters parked while `Unbound`, `Frozen`, or `Bound`-at-capacity all flow through `state.waiters` and are picked up by the loop; `submit` already observes on entry for `Bound`/`Frozen`, so no path is missed.

### Tests

Deterministic, paused-clock coverage of exactly the previously-broken cases: pre-bind overflow (`send_timeout` + `call`), rebind overflow with a superseding incarnation, and the promotion-vs-deadline race. The rebind test's `poll_until` cadence (1 ms interval) can't be auto-advanced past the 10 s waiter deadlines, so it's sound under `start_paused`.

### Validation performed

- `./scripts/dev just ci` on this branch: **pass**.
- Cross-merge with #45 (both touch `mailbox.rs` and `tests/deadlines.rs`): merges cleanly; full `--test integration` suite on the merged tree: **166 passed, 0 failed**.

PR #43, opened 2026-08-07
Simplify duplicated supervision scaffolding

ralexstokes:
## Summary

- make the engine's `RestartDecision` the single source of restart attempt, delay, deadline, intensity, and cumulative-count data consumed by the driver
- move ordered and dynamic builders directly when they become subtrees, without allocating, minting, terminalizing, and discarding a replacement root
- funnel callback-handler live/drain verdicts and `CallFuture` reply verdicts through shared transition helpers
- replace `ScopeRuntime`'s synchronized startup/draining booleans and optional reason with one internal `ScopePhase` model

## Why

Item 9 identified implementation-only duplication that made equivalent transitions easier to drift apart. Restart scheduling was the clearest example: the engine calculated a schedule and emitted effects into a vector, while the driver ignored those values and recomputed the delay/deadline. Tree consumption also created a complete rooted builder solely to satisfy `mem::replace`, spending a scope identity and immediately discarding its cells.

The shared helpers retain the existing callback failure/panic and call timeout/race ordering. The lifecycle phase records whether draining began while startup was pending, complete, or failed, so the refactor preserves admission, startup-abort, and completion behavior without parallel fields.

## Prior merged evidence

This diff intentionally does not repeat cleanup that is already on `main`:

- #28 removed the module-wide dead-code allowance, unused runtime wrappers/exit scaffolding, and the resulting `tokio-util` dependency.
- #22 removed the unused direct `slab` declaration.

Those merged PRs satisfy item 9's dead-code and unused-dependency acceptance. This PR completes the remaining scaffolding work.

## Relationship to merged changes

Rebased onto the mainline through #37, #42, and #45. Restart publication uses #37's watch-backed observation record, retains #45's overflow-safe optional deadline, and consumes the scheduler decision as the authoritative attempt, delay, deadline, intensity, and cumulative count.

## Impact

No public API or intended runtime behavior changes. New unit coverage proves subtree conversion mints no phantom scope and that the consolidated lifecycle phase preserves startup status across drain entry.

## Validation

- `./scripts/dev cargo test -p shelterwood --lib` (54 passed)
- `./scripts/dev cargo test -p shelterwood --test integration` (166 passed)
- `./scripts/dev just ci` (222 tests passed; formatting, Clippy with `-D warnings`, docs, API/runtime/exit path checks, and Nix formatting all passed)
- `./scripts/dev just ci-nix` (all clean-Nix checks passed)

Part of #35 (item 9).

### DIFF
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -368,51 +368,51 @@ impl<A: Actor> RawActor for Handler<A> {
             raw.mark_ready();
         }
 
-        loop {
+        let live_failure = loop {
             let received = CatchUnwindFuture::new(raw.recv()).await;
             let message = match received {
                 Ok(Some(message)) => message,
-                Ok(None) => break,
+                Ok(None) => break None,
                 Err(payload) => resume_unwind(payload),
             };
-            let handled = {
-                let actor = self
-                    .actor
-                    .as_mut()
-                    .expect("initialized handler retains its actor");
-                let mut context = Context::<A>::new(raw, false);
-                CatchUnwindFuture::new(actor.handle(message, &mut context)).await
-            };
-            match handled {
-                Ok(Ok(())) => {}
-                Ok(Err(error)) => return fail_after_teardown(raw, error).await,
-                Err(payload) => resume_unwind(payload),
+            let actor = self
+                .actor
+                .as_mut()
+                .expect("initialized handler retains its actor");
+            if let Err(failure) = handle_message(actor, raw, message, false).await {
+                break Some(failure);
             }
-        }
+        };
 
-        match raw.mailbox_shutdown() {
-            MailboxShutdown::Drain => {
+        let failure = match (live_failure, raw.mailbox_shutdown()) {
+            (Some(failure), _) => Some(failure),
+            (None, MailboxShutdown::Drain) => {
+                let mut failure = None;
                 while let Some(message) = raw.try_recv() {
-                    let handled = {
-                        let actor = self
-                            .actor
-                            .as_mut()
-                            .expect("initialized handler retains its actor");
-                        let mut context = Context::<A>::new(raw, true);
-                        CatchUnwindFuture::new(actor.handle(message, &mut context)).await
-                    };
-                    match handled {
-                        Ok(Ok(())) => {}
-                        Ok(Err(error)) => return fail_after_teardown(raw, error).await,
-                        Err(payload) => resume_unwind(payload),
+                    let actor = self
+                        .actor
+                        .as_mut()
+                        .expect("initialized handler retains its actor");
+                    if let Err(current) = handle_message(actor, raw, message, true).await {
+                        failure = Some(current);
+                        break;
                     }
                 }
+                failure
             }
-            MailboxShutdown::Discard => {
+            (None, MailboxShutdown::Discard) => {
                 while let Some(message) = raw.try_recv() {
                     drop(message);
                 }
+                None
             }
+        };
+
+        if let Some(failure) = failure {
+            return match failure {
+                HandlerFailure::Returned(error) => fail_after_teardown(raw, error).await,
+                HandlerFailure::Panicked(payload) => resume_unwind(payload),
+            };
         }
 
         let mut context = StopContext::<A>::new(raw);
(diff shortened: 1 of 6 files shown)


ralexstokes:
Fresh-agent adversarial review completed against `main` (`14af0e7`) through `cf60b14`.

Reviewed item 9 end to end: authoritative restart scheduling/count publication; all callback-handler live/drain and `CallFuture` reply/timeout/race verdicts; startup/running/failed histories through scope drain; direct builder moves, identity minting/exhaustion, and armed drop behavior; panic/destructor ordering; public bounds; tests/docs; and merged #28 evidence for the dead-code/tokio-util cleanup. I also checked the overlap with #37/PR #40 and found this diff semantically independent.

No actionable findings were found, so no review-fix commit was needed.

Validation: focused actor, mailbox, deadlines, drain, subtree, dynamic, lifecycle, and observation suites passed; `./scripts/dev just ci` passed (202 tests plus formatting, clippy `-D warnings`, docs, API/path checks); `./scripts/dev just ci-nix` passed; and `git diff --check origin/main...HEAD` passed.

ralexstokes:
Rebased onto current `main` at `7c401f5` (`cf60b14` → `65bc7ac`). The driver conflict was resolved by retaining #37's watch-backed observation record while publishing the scheduler decision's authoritative restart total. Post-rebase validation: `git diff --check`, 49 library tests, `./scripts/dev just ci` (209 tests), and `./scripts/dev just ci-nix`. The remote branch was updated with an exact `--force-with-lease`.

ralexstokes:
Rebased again after `main` advanced from `7c401f5` to `c87d783` via merged #42 and #45 (`65bc7ac` → `0b62928`). Driver/engine conflicts were resolved by keeping #45's overflow-safe optional restart deadline inside #43's complete `RestartDecision`; the driver now publishes that decision verbatim and arms a restart only when intensity permits and the deadline is representable. Post-rebase validation: `git diff --check`, 54 library tests, 166 integration tests, `./scripts/dev just ci` (222 tests), and `./scripts/dev just ci-nix`. The remote branch was updated with an exact `--force-with-lease`.

ralexstokes (commented):
## Review — LGTM, no findings

**What it does:** four independent simplifications — `RestartDecision` becomes the single restart verdict (deleting the `EffectSink`/`RestartEffect` layer and the driver's re-derivation of `delay`/`scheduled_at`), `ScopePhase` replaces the three synchronized startup/draining fields, subtree `into_core` moves the builder without minting and discarding a phantom root, and the handler loop / `CallFuture` reply arms are deduplicated.

### Verified equivalent

- **`total_restarts` sourcing.** Publishing `decision.charge.total_restarts` instead of `saturating_add(1)` on the record is equivalent: both `IntensityState` and the published counter are per-scope-incarnation (`set_state(Starting)` resets `record.total_restarts` to 0, `IntensityState::default()` is fresh per `ScopeRuntime`), and each publish corresponds to exactly one `charge` on the single driver task.
- **Ungated `decision.restart_at`.** The tripped-gating moved from `schedule_restart` into the driver, which only pushes the deadline in the non-tripped branch — same spawn behavior, and the record now uniformly publishes the derived schedule (`None` still reserved for an unrepresentable deadline, matching #45's contract and its tests).
- **`ScopePhase`.** Checked every replaced predicate: `startup_complete`/`startup_failed` survive into `Draining` via the captured `StartupPhase`; `progress_startup`'s three-way guard collapses to `phase != Starting` exactly; both `fail_startup` call sites are gated on `!is_draining()`, so the unconditional `phase = StartupFailed` can never erase a drain reason; `begin_drain` is idempotent and the startup-pending capture happens before the transition.
- **Handler loop.** `handle_message` + `HandlerFailure` preserves the old control flow precisely, including skipping the drain phase after a live-loop failure and the teardown-then-resume ordering for panics.
- **`CallFuture::poll_reply`.** All four inlined reply matches collapse faithfully; the one behavior nuance — the `Withdrawal::Accepted` arm previously tolerated `self.reply == None` — is unreachable (reply is always populated at start and never taken), so the `expect` is right.
- **`into_core`.** Moving `self.core` out directly is the real simplification (no phantom root minted, terminalized, and discarded), and the thread-local creation counter makes it testable without touching production types.

### Merge coordination

Conflicts with #44 (`driver.rs`, `identity.rs` — #44 rewrites `ScopeIdentity::new`, which this PR instruments, and changes `with_counter`'s signature) and with #47 (`actor.rs` — both restructure `Handler::run`). #44 and #47 merge cleanly with each other, so landing this one first or last are both fine; whoever is second on each conflicting file needs a manual rebase. On the #44 side note the instrumented counter moves naturally into `fresh_counter()` there, since #44's `ScopeIdentity::new` no longer allocates a lineage.

### Validation

- `./scripts/dev just ci` on the branch: **pass**.

PR #45, opened 2026-08-07
Centralize overflow-safe deadline and duration semantics

ralexstokes:
Part of #35 (item 5)

## Summary

- add one internal overflow-safe `Deadline` policy where unrepresentable deadlines mean “never”
- apply it to readiness, shutdown escalation, restart scheduling, mailbox budgets, actor timers/offloads, and snapshot waits
- capture `call` budgets before invoking the user message constructor
- check the current child snapshot before rejecting an elapsed zero-duration wait
- reapply an exact `Duration` clamp after floating-point backoff conversion and jitter

## Why

Several time paths converted `Instant::checked_add` overflow into `now`, causing very long readiness, shutdown, and restart policies to fire immediately. Calls also started their timer after running user construction code, zero-duration child waits rejected an already-satisfied predicate, and extreme floating-point backoff values could round above their configured maximum.

## User impact

`Duration::MAX` supervision policies now remain pending instead of firing immediately, call deadlines cover message construction, already-satisfied zero-duration waits succeed, and generated backoffs never exceed `max`.

## Validation

- `./scripts/dev cargo test -p shelterwood --test integration deadlines::` (10 passed)
- `./scripts/dev just ci` (217 tests passed; formatting, Clippy, docs, API/runtime/exit path checks, and Nix formatting all passed)

### DIFF
--- a/crates/shelterwood/src/deadline.rs
+++ b/crates/shelterwood/src/deadline.rs
@@ -1,0 +1,43 @@
+//! Overflow-safe internal deadline semantics.
+
+use std::time::{Duration, Instant};
+
+/// An absolute deadline, or one too distant for the clock to represent.
+///
+/// Overflow consistently means that the deadline never arrives. In
+/// particular, it must never be substituted with the budget's start time.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) struct Deadline(Option<Instant>);
+
+impl Deadline {
+    /// Captures a duration budget relative to `started_at`.
+    pub(crate) fn after(started_at: Instant, duration: Duration) -> Self {
+        Self(started_at.checked_add(duration))
+    }
+
+    /// Returns the representable absolute deadline, if there is one.
+    pub(crate) fn instant(self) -> Option<Instant> {
+        self.0
+    }
+
+    /// Reports whether a representable deadline has elapsed.
+    pub(crate) fn is_due(self, now: Instant) -> bool {
+        self.0.is_some_and(|deadline| now >= deadline)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::time::{Duration, Instant};
+
+    use super::Deadline;
+
+    #[test]
+    fn overflow_is_never_due() {
+        let now = Instant::now();
+        let deadline = Deadline::after(now, Duration::MAX);
+
+        assert_eq!(deadline.instant(), None);
+        assert!(!deadline.is_due(now));
+    }
+}
--- a/crates/shelterwood/src/deadline.rs
+++ b/crates/shelterwood/src/deadline.rs
@@ -24,6 +24,14 @@ impl Deadline {
     pub(crate) fn is_due(self, now: Instant) -> bool {
         self.0.is_some_and(|deadline| now >= deadline)
     }
+
+    /// Reports whether a representable deadline is strictly in the past.
+    ///
+    /// Exact-boundary arbitration can still let already-ready work win, but
+    /// work that has not started must not begin after its budget has elapsed.
+    pub(crate) fn is_overdue(self, now: Instant) -> bool {
+        self.0.is_some_and(|deadline| now > deadline)
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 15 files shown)


ralexstokes:
Adversarial review completed for issue #35 item 5.

Scope: the central deadline policy and every migrated readiness, shutdown, restart, mailbox, actor-timer, offload, and snapshot-wait call site; representable/overflowing deadlines; `Duration::MAX`; pending-deadline wake behavior; public `restart_at`; constructor-inclusive call budgets; exact/zero boundary ordering; extreme backoff conversion; virtual-time determinism; and user-facing docs.

Findings fixed in `05bc7ce`:

- A constructor could consume the full call budget and still place a newly created request into an available mailbox. Calls now reject work strictly after the captured deadline while preserving the established exact-boundary winner rule, with a stronger available-mailbox regression test.
- An intensity trip's internal “do not spawn” `None` was being published as `ChildSnapshot::restart_at`, conflating it with an unrepresentable deadline. Snapshot publication now uses the actual derived schedule; `None` retains its overflow meaning.
- Extreme backoff coverage now exercises huge factors, attempts, jitter modes, and non-finite samples; relevant timer/wait overflow tests use paused virtual time; observation and retry docs now state overflow, zero-duration, `restart_at`, and constructor-budget semantics.

Validation:

- focused deadline, engine, backoff, timer, and observation tests passed
- `./scripts/dev just ci` passed (210 tests, clippy, formatting, docs, public API reachability, runtime/exit path checks, Nix formatting)
- GitHub Actions CI passed: run `31212295705`
- final remote diff re-reviewed clean at `05bc7ceea0d9ae4aa914f7cddb9c773508c8dbb6`

PR remains draft and unmerged.

ralexstokes:
Rebased onto current `main` at `7c401f5` (`05bc7ce` → `6ab22f3`) with no conflicts. Post-rebase validation: `git diff --check`, 10 focused deadline tests, and `./scripts/dev just ci` (217 tests plus formatting, Clippy, docs, API/path, and nixfmt checks). The remote branch was updated with an exact `--force-with-lease`.

ralexstokes (commented):
## Review — LGTM, no findings

**What it does:** introduces one internal `Deadline` policy (overflow ⇒ "never fires", never "fires now") and migrates every readiness, shutdown-escalation, restart-scheduling, mailbox-budget, actor-timer/offload, and snapshot-wait site onto it; captures `call` budgets before the user message constructor; lets zero-duration `wait_for_child` observe an already-satisfied snapshot; and re-clamps backoff to the exact `Duration` cap after float conversion/jitter.

### Correctness

- **Migration is complete.** No `checked_add(...).unwrap_or(now)` patterns remain; the only `checked_add` sites left are `Deadline::after` itself and the (unrelated) identity counters. Clock plumbing is consistent: `runtime::now()` (tokio, paused-clock-aware) feeds both capture and `sleep_until_std`.
- **Unrepresentable deadlines compose.** `ReadinessGate::Waiting { deadline: None }` was already the unbounded encoding, and `DeadlineKind::Readiness`/`Restart` entries are only pushed for representable instants, so an overflowing deadline simply never enters the queue — both readiness arming sites (initial and post-restart) handle it.
- **Restart record/queue split is right.** The record publishes the effect's schedule even on an intensity trip (preserving prior observable behavior), while the spawn deadline comes from `decision.restart_at`, which is `None` on trip *or* overflow — so `restart_at: None` + `Restarting` cleanly means "never", as now documented in `observe.rs` and `docs/observation.md`.
- **`StopLadder` can't wedge.** The one spot where an overflowed escalation deadline would be alarming — the hard-abort beat — is `tidy_abort_beat`, clamped to ≤ 10 ms, so the `Escalated` phase always gets a representable deadline. `Graceful { grace: MAX }` correctly parks at `Cooperative` forever, and forced cleanup (`shutdown(ZERO)`) still reclaims the straggler, as the test shows.
- **`AcceptanceTimedOut` contract preserved.** The post-constructor `is_overdue` check fires before any `SendFuture` exists, so the guaranteed-unaccepted meaning holds; `is_overdue` is strict (`>`), keeping exact-boundary arbitration consistent with the acceptance path, and an overflowed budget is never overdue.
- **Backoff clamp covers both shapes.** `delay.min(maximum)` bounds `Fixed` (max = delay) and `Exponential` (max = cap) after every float round-trip; the non-finite-sample fallback keeps the jitter multiplier in range.

### Notes (non-blocking)

- `message_construction_consumes_the_call_budget` white-boxes tokio's `advance` (clock moves before the first yield point). If tokio ever changes that, the test fails loudly (`poll_once` would return `Pending`) rather than silently passing, so this is acceptable — the inline comment already flags it.

### Validation performed

- `./scripts/dev just ci` on this branch: **pass**.
- Cross-merge with #42: clean; full `--test integration` on the merged tree: **166 passed, 0 failed**.

PR #47, opened 2026-08-07
Make offload teardown and sequencing panic-safe

ralexstokes:
Part of #35 (item 6)

## What changed

- move offload polling and disposal outside the shared-state mutex
- catch and retain cancellation/destructor panics while continuing cancellation, abort, and completion signalling for remaining resources
- preserve an already-caught actor panic over secondary teardown failures
- make event sequence allocation, queue insertion, and timer-watermark capture share one lock/linearization point
- add deterministic coverage for direct cancellation, a polled scoped-cancellation race, actor-panic precedence, and concurrent queue/watermark ordering

## Why

A panicking user-future destructor could unwind cancellation, poison shared offload state, strand later cleanup, or replace the actor's original panic. Separately, sequence allocation could race queue insertion and invalidate timer watermark ordering.

## Validation

- `./scripts/dev cargo test -p shelterwood --lib concurrent_pushes_and_timer_watermark_share_one_linearization_point` (1 passed)
- `./scripts/dev cargo test -p shelterwood --test integration offload` (19 passed)
- `./scripts/dev just ci` (217 tests passed; formatting, Clippy, docs, API/runtime/exit path checks, and Nix formatting all passed)

### DIFF
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -1,12 +1,7 @@
 //! Callback-oriented actors layered entirely on the public raw-actor surface.
 
 use std::{
-    fmt,
-    future::Future,
-    hash::Hash,
-    marker::PhantomData,
-    panic::{AssertUnwindSafe, catch_unwind, resume_unwind},
-    sync::Arc,
+    fmt, future::Future, hash::Hash, marker::PhantomData, panic::resume_unwind, sync::Arc,
     time::Duration,
 };
 
@@ -333,6 +328,7 @@ impl<'a, A: Actor> StopContext<'a, A> {
 /// actors. Its declared readiness is read before [`RawActor::run`] is polled.
 pub struct Handler<A: Actor> {
     args: Option<A::Args>,
+    actor: Option<A>,
     readiness: Readiness,
 }
 
@@ -345,6 +341,7 @@ impl<A: Actor> Handler<A> {
     fn with_readiness(args: A::Args, readiness: Readiness) -> Self {
         Self {
             args: Some(args),
+            actor: None,
             readiness,
         }
     }
@@ -362,10 +359,11 @@ impl<A: Actor> RawActor for Handler<A> {
             .args
             .take()
             .expect("handler actor initialization invoked more than once");
-        let mut actor = {
+        let actor = {
             let mut context = Context::<A>::new(raw, false);
             A::init(args, &mut context).await?
         };
+        self.actor = Some(actor);
         if raw.readiness() == Readiness::AfterInit {
             raw.mark_ready();
         }
@@ -375,16 +373,20 @@ impl<A: Actor> RawActor for Handler<A> {
             let message = match received {
                 Ok(Some(message)) => message,
                 Ok(None) => break,
-                Err(payload) => resume_after_teardown(raw, actor, payload).await,
+                Err(payload) => resume_unwind(payload),
             };
             let handled = {
+                let actor = self
+                    .actor
+                    .as_mut()
+                    .expect("initialized handler retains its actor");
                 let mut context = Context::<A>::new(raw, false);
                 CatchUnwindFuture::new(actor.handle(message, &mut context)).await
             };
             match handled {
                 Ok(Ok(())) => {}
                 Ok(Err(error)) => return fail_after_teardown(raw, error).await,
-                Err(payload) => resume_after_teardown(raw, actor, payload).await,
+                Err(payload) => resume_unwind(payload),
             }
         }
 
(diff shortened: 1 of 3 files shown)


ralexstokes:
Fresh adversarial review at `515d162`.

I audited offload polling/cancellation/drop, hard-abort teardown (including already-unwinding paths), primary-vs-cleanup panic precedence, idempotence/completion behavior, multiple cleanup panics, and event FIFO/watermark/signal ordering.

Two correctness findings were fixed in `515d162`:

1. On hard abort, high-level handler state could be destroyed before the raw resource owner. A panicking actor destructor could therefore mask an already-recorded owned-offload panic. Handler ownership now stays within the resource-first teardown boundary, and the original actor/offload panic is retained across asynchronous cleanup.
2. The `u64::saturating_add` event sequence watermark collapsed distinct insertions at `u64::MAX`, allowing post-watermark events into an older timer batch. Watermarks now snapshot the FIFO prefix length under the same mutex used for insertion and drain exactly that prefix.

Regression coverage includes hard-abort panic precedence, cleanup after destructor panic, cancellation polling/drop outside the mutex, concurrent insertion vs timer watermarking, exact FIFO-prefix draining, and event-before-signal wakeup ordering.

Validation:
- `./scripts/dev just ci` — passed (210 tests plus fmt, clippy, docs, API/runtime checks)
- `./scripts/dev just ci-nix` — passed (all 4 flake checks)

Re-review after the fixes: no remaining findings. The PR remains draft; I did not merge or mark it ready.

ralexstokes:
Rebased onto current `main` at `7c401f5` (`515d162` → `32836f4`). The actor conflict was resolved by retaining #37's lock-free `Sync` factory representation while preserving the reviewed panic-precedence fix and outer containment-boundary teardown. Post-rebase validation: `git diff --check`, 19 focused offload tests, the concurrent watermark unit test, and `./scripts/dev just ci` (217 tests). The remote branch was updated with an exact `--force-with-lease`.

ralexstokes (commented):
## Review — LGTM, no findings

**What it does:** makes offload teardown panic-safe (polling and disposal outside the shared mutex, destructor panics caught and retained while cancellation/completion signalling continues), gives an already-caught actor panic precedence over secondary teardown panics, and collapses event sequencing to one lock where FIFO order itself is the sequence.

### Verified sound

- **`SharedOffloadState` protocol.** Poll takes the future out of the mutex; cancel either disposes an idle future (outside the lock) or marks an in-progress poll for the poller to dispose. Every path — complete, cancelled-idle, cancelled-mid-poll, poll panic — fires `finished` exactly where it should and records panics into the shared `PanicSlot` without poisoning the mutex. This also fixes a real stranded-join bug: the old cancel path never fired `finished` at all (the latch was only fired by the operation running to completion), and the old design held the mutex across the user poll, which is exactly what let a destructor panic poison shared state.
- **Cancel-while-polling semantics.** `finished` now fires before the in-flight poll's future is actually dropped. That's safe — `offload`/`offload_scoped` require `'static` futures, so nothing borrows actor state — and it's the documented intent ("cancellation always signals completion"). A destructor panic recorded after the actor's epilogue has already drained the slot is discarded; that's a narrow hard-abort-adjacent race and consistent with the crate's discard-secondary-panic policy.
- **Count-based watermark.** Sound because the consumer is single and the plain `events.pop()` path is only reachable once `fired_batch` is `None`: a batch always drains its counted prefix to zero (or is dropped together with the queue in `freeze`, which clears both) before plain pops resume, so the count can't be made to overshoot onto post-watermark events. The `insert_with`/`push_with_hooks` test hooks prove insertion and watermarking share the one linearization point.
- **Panic precedence.** The epilogue's ordering — record primary from `actor.run`, then freeze/join/drop with every step individually contained, then `resume_preferred_panic(primary, cleanup)` — preserves the actor's original diagnostic over cancellation/destructor panics, including across hard abort (`RawIncarnationOwner::Drop` discards everything while the thread is already panicking, resumes preferred otherwise). The `hard_abort_preserves_owned_offload_panic_over_handler_destructor` test threads the interesting path: the owned offload panic travels through `drop_raw` → `RawResources::drop` → resumed → caught → preferred over the handler's destructor panic.
- **`Handler` restructure.** Moving the actor into `self` and deleting `resume_after_teardown` centralizes §5.5 teardown in the one epilogue, and it *improves* conformance on the normal-exit path: the actor state was previously dropped at the end of `run`, before resources were frozen and joined; now it is consistently dropped last.
- **Freeze reorder** (cancel offloads before clearing continuations/timers) is safe in the single-threaded section, and each `offload.cancel()` contains its own destructor panic so one failure can't strand later resources.

### Merge coordination

Conflicts with #43 in `actor.rs` (both restructure `Handler::run`); merges cleanly with #44. Whichever of #43/#47 lands second needs a manual rebase of that function.

### Validation

- `./scripts/dev just ci` on the branch: **pass**.

PR #48, opened 2026-08-07
Use Tokio watch for projected scope state

ralexstokes:
Stack step 4/4 for #37. This PR is based on #46.

## Audit result

The pre-change audit rejected both a single-writer assumption and one consolidated ScopeCell state lock:

- `control` is an epoch-tagged request plane with concurrent callers.
- `child_identity` and `lifecycle_sequence` mint monotonic identities.
- `current_dynamic` publishes an independent dynamic request route.
- `OBSERVATION_GATE`, not any per-field lock, provides compound consistency for recursive snapshots and lifecycle-event staging.

Accordingly, those four mutexes remain separate and the invariant is now documented on `ScopeCell`.

## Semantic mapping

The four retained fields projected into observation are migrated through runtime-owned Tokio watch wrappers:

- `record` → latest `ScopeRecord`; `wait_started` subscribes directly to this state.
- `config` → latest resolved `ScopeConfig`.
- `current_children` → latest resident set; closure-based reads clone slot handles before recursion, and removals/take release the watch value guard before residency completion can emit `Removed`.
- `parent` → latest weak ancestor link used for recursive snapshot/event forwarding.

`OBSERVATION_GATE` still surrounds every compound observation-visible transition. The watch channels retain independently readable field values but are explicitly not treated as a multi-field transaction.

Tokio implementation types remain confined to `runtime.rs`, and no watch guard crosses an `await`.

## Adversarial review

A fresh reviewer found that direct scope-watch notification could wake `wait_started` before member terminal publication and, for a live incarnation, before `last_stopped_epoch` advanced. Commit `5a4b37f` stages those record updates silently and pulses the scope watch beside the original member/control publication boundary. Deterministic custom-waker tests cover both the live-epoch and no-live-root paths.

## Validation

- `./scripts/dev just ci`
- 206/206 tests passed
- formatting, Clippy with warnings denied, docs, API reachability, runtime-path, exit-path, and Nix formatting gates passed

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -583,16 +583,30 @@ pub(crate) struct ScopeRecord {
     pub(crate) total_restarts: u64,
 }
 
+/// Shared scope state follows two distinct synchronization regimes.
+///
+/// `OBSERVATION_GATE` serializes compound, observation-visible transitions
+/// across `config`, `record`, `current_children`, and `parent`, including
+/// recursive snapshot projection and lifecycle-event staging. Their watch
+/// channels retain the latest independently readable value; they do not make
+/// that compound transition atomic, so every multi-field observation path must
+/// continue to hold the gate.
+///
+/// The remaining mutexes are deliberately not collapsed into one state lock.
+/// `control` is an epoch-tagged request plane with concurrent callers,
+/// `child_identity` and `lifecycle_sequence` mint monotonic identities, and
+/// `current_dynamic` owns the independently published dynamic request route.
+/// Scope state therefore has multiple writers and no single-writer invariant.
 pub(crate) struct ScopeCell {
     pub(crate) member: Arc<MemberCell>,
     pub(crate) flavor: ScopeFlavor,
     pub(crate) child_identity: Mutex<ScopeIdentity>,
-    config: Mutex<crate::tree::ScopeConfig>,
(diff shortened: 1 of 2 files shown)


ralexstokes:
Fresh adversarial review completed by a dedicated reviewer agent.

The review found and directly fixed one observable ordering regression: `wait_started`’s new direct scope-watch notification could fire while terminal scope state was being mutated, before member terminality and—on the live-incarnation path—before `control.last_stopped_epoch` advanced. That weakened the former member-signal publication boundary.

Commit `5a4b37f` now stages startup/terminal scope-record updates with a silent watch mutation, completes the member/control transition, and only then pulses the scope-record watch. The same publication discipline covers normal `set_startup` and never-started terminalization. It also corrects the child-set comment to refer to the watch value guard.

Deterministic custom-waker regressions cover both live-epoch and no-live-root paths and fail on the original PR head.

Validation at reviewed head `5a4b37f`:
- focused driver suite: 19/19
- `./scripts/dev just ci`: 206/206 tests
- formatting, Clippy with warnings denied, docs, API reachability, runtime-path, exit-path, and Nix formatting gates passed
- worktree clean; local and remote heads synchronized
- PR remains open/draft with the correct #46 base and exactly two stack-local commits

No further findings remained after tracing recursive observation, residency/removal ordering, straggler collection, drop paths, watch reentrancy, runtime confinement, and await boundaries.

PR #52, opened 2026-08-07
Remove global and quadratic runtime hot paths

### LINKED ISSUE #39: Scope `OBSERVATION_GATE` per tree
## Context

Follow-up to the [Cluster 4 audit in #37](https://github.com/ralexstokes/shelterwood/issues/37#issuecomment-5220896937): the gate is process-wide, so independent trees serialize their lifecycle transitions against each other.

This records the design for when a workload makes it worth implementing. This is deferred, not planned work.

## Why it is safe to scope the gate

Two facts verified against the code:

1. **A `ScopeCell` never migrates between trees while running.** Nested trees receive their `ScopeCell` from the parent at reservation (`root_override` in `lower()`), parent links are set once at admission under the gate, and `begin_incarnation` runs only after admission (`run_nested_tree`). Tree membership is fixed before a cell's driver starts.
2. **Nothing depends on cross-tree ordering.** Lifecycle sequence numbers are minted per-cell (`lifecycle_sequence`'s `FenceCounter`), not by the gate. The gate's contract — mutation + recursive projection + event staging as one atomic edge, "including cross-scope forwarding" — is within-tree. Cross-tree serialization is pure false sharing.

## Recommended mechanism

Each cell holds an `Arc<Mutex<()>>` gate instead of referencing the static. Acquisition uses the lock-handoff retry loop, which is what makes reassignment safe against concurrent observers:

```rust
loop {
    let gate = self.gate();            // clone the Arc
    let guard = gate.lock(...);
    if Arc::ptr_eq(&gate, &self.gate()) { break (gate, guard); }
    // gate was swapped while we waited — retry with the new one
}
```

Assignment happens where tree identity is already decided:

- a standalone system's cells share the root's gate, stamped in `lower()`;
- subtree adoption stamps the parent's gate onto every descendant cell during `lower(root_override)` / `set_admitted_children`, before the subtree's driver starts. The only concurrent gate-takers at that point are pre-start observers (e.g. an early `subscribe_snapshots`), which the retry loop handles.

If a future change ever re-gates a subtree with live writers, hold both gates in `Arc::as_ptr` order while swapping; per the adoption ordering above, that case does not currently exist.

Everything else is untouched: the atomic observation edge (now per tree, which is all the contract claims), synchronous `snapshot()`, the epoch request plane, and pre-admission single-cell edges (`terminalize_never_started` on unlinked cells uses the cell's own gate). `OBSERVATION_GATE` is one symbol with all uses localized in `driver.rs`, so the refactor is mechanical.

## Alternatives considered

- **Read-side relief only (cheapest, independently useful).** `snapshot()` projects fresh under the gate; `SnapshotHub` already holds the last published snapshot as an `Arc`. Serving plain reads from that (Arc clone, no gate) removes reader/writer contention at the cost of bounded staleness, keeping a gated `snapshot_now()` for callers needing the fresh projection. This helps even within one busy tree, since the critical section is O(subtree) projection plus ancestor fan-out.
- **Per-tree projection task (rejected for now).** Mutators append to an ordered log; one task per tree consumes and publishes. Eliminates the gate but makes the observation edge asynchronous — `set_state` would return before observers can see it — breaking the synchronous-consistency property the gate exists to provide. Only worth revisiting if projection work becomes the bottleneck inside a single tree.

## When to do it

Only pays off in processes hosting many concurrently-churning trees; single-tree deployments gain nothing (a per-tree gate serializes exactly as much as the static does). Trigger on a measured workload, not preemptively.

Split out from #37.

ralexstokes:
Part of #35 (item 8). Closes #39.

## What changed

- Scope the observation publication gate to each running tree. Adopted subtrees inherit the parent gate, complete in-flight edges before handoff, and recursively re-home resident descendants without splitting parent/child projection.
- Replace raw actors' vector-scanned keyed timers with a type-aware hash index plus independently ordered deadline index. Replacement/removal use keyed lookup while equal deadlines retain arming order, heterogeneous key types remain distinct, and item 5's centralized overflow-safe deadline semantics remain authoritative.
- Replace parked mailbox sends' `VecDeque::retain` cancellation with monotonic, directly removable intrusive registrations. FIFO promotion, current incarnation evidence for every waiter from item 3, and the mailbox-lock-before-operation-lock race protocol are preserved.

## Why

Independent systems previously contended on one process-global observation mutex. Keyed timer replacement/removal and parked-send cancellation performed repeated linear scans, making churn and mass cancellation scale quadratically.

## Adversarial review fixes

- Serialize observation-gate adoption with an edge already executing under the former gate, and re-home a resident descendant tree as one atomic handoff.
- Permit cancellation after terminal teardown has detached the waiter queue without a stale-registration panic; monotonic IDs still cannot target later waiters.
- Cover exact-key equality under timer hash collisions and make the 16,384-key scaling assertion select distinct hashes before requiring one probe per removal.
- Exercise head, tail, and interior waiter unlinking in a deterministic operation-count test rather than relying on elapsed time.

## Correctness and regression coverage

- Verifies independent systems make observation progress while another system's gate is held.
- Verifies adopted subtrees and existing descendants share one gate; waiting and already-active pre-admission observers cross handoff safely.
- Verifies heterogeneous/colliding timer keys, replacement ordering, overflow-safe arming, stale-index retraction, and 16,384 keyed removals with one exact-key probe per removal.
- Verifies 16,384 parked sends cancel with one direct unlink each, stale IDs cannot unlink successors, and cancellation can race detached terminal teardown.
- Existing observation forwarding/atomic snapshot, timer retraction/interval/watermark, mailbox FIFO/latest/promotion/cancellation/rebind, item 3 evidence, and item 5 deadline suites remain green.

## Rebase

Rebased onto `c87d783` (merged PRs #42 and #45). The raw-timer conflict was resolved by composing the keyed indexes with `Deadline::after(...).instant()`, and mailbox binding adapts item 3's all-waiter evidence refresh through the intrusive queue.

## Validation

- Focused timer store, mailbox, events, deadlines, rebind, and observation-gate tests
- `./scripts/dev just ci` (231 tests; fmt, clippy `-D warnings`, nextest, docs, API/runtime-path checks)
- `./scripts/dev just ci-nix` (all clean Nix checks passed)

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -698,10 +698,65 @@ impl ScopeCell {
     }
 
     fn adopt_observation_gate(&self, gate: &Arc<Mutex<()>>) {
-        *self
-            .observation_gate
-            .lock()
-            .expect("observation gate handoff mutex poisoned") = Arc::clone(gate);
+        loop {
+            let current = self.observation_gate();
+            if Arc::ptr_eq(&current, gate) {
+                return;
+            }
+
+            // An operation that passed `with_observation_gate`'s pointer
+            // check is allowed to finish its complete observation edge before
+            // the handoff. Conversely, an operation that only captured this
+            // gate will see the replacement after acquiring it and retry.
+            let current_guard = current
+                .lock()
+                .unwrap_or_else(std::sync::PoisonError::into_inner);
+            let mut installed = self
+                .observation_gate
+                .lock()
+                .expect("observation gate handoff mutex poisoned");
+            if Arc::ptr_eq(&current, &installed) {
+                *installed = Arc::clone(gate);
+                drop(installed);
+                self.adopt_descendant_observation_gates_locked(&current, gate);
+                drop(current_guard);
+                return;
+            }
+            drop(installed);
+            drop(current_guard);
+        }
+    }
+
+    /// Re-homes a resident subtree while its former tree gate is held. The
+    /// caller also holds the destination gate, so observers cannot enter
+    /// either half of the tree while the handoff is being installed.
+    fn adopt_descendant_observation_gates_locked(
+        &self,
+        previous: &Arc<Mutex<()>>,
+        gate: &Arc<Mutex<()>>,
+    ) {
+        let descendants = self.current_children.read_with(|children| {
+            children
+                .iter()
+                .filter_map(|resident| resident.slot.scope.as_ref().cloned())
+                .collect::<Vec<_>>()
+        });
+        for descendant in descendants {
+            let mut installed = descendant
+                .observation_gate
+                .lock()
+                .expect("observation gate handoff mutex poisoned");
+            if Arc::ptr_eq(&installed, previous) {
+                *installed = Arc::clone(gate);
+            } else {
+                assert!(
+                    Arc::ptr_eq(&installed, gate),
+                    "one resident tree must share one observation gate"
+                );
+            }
+            drop(installed);
+            descendant.adopt_descendant_observation_gates_locked(previous, gate);
+        }
     }
 
     /// Runs against the cell's current tree gate. Adoption can race an early
@@ -1164,8 +1219,8 @@ impl ScopeCell {
     }
 
     fn set_admitted_children(self: &Arc<Self>, children: Vec<Arc<SlotCell>>) {
-        let gate = self.observation_gate();
         self.with_observation_gate(|| {
+            let gate = self.observation_gate();
             self.clear_residents_locked();
             for child in children {
                 if let Some(scope) = &child.scope {
@@ -1187,8 +1242,8 @@ impl ScopeCell {
     }
 
     fn admit_child(self: &Arc<Self>, child: &Arc<SlotCell>) {
-        let gate = self.observation_gate();
         self.with_observation_gate(|| {
+            let gate = self.observation_gate();
             if let Some(scope) = &child.scope {
                 scope.adopt_observation_gate(&gate);
                 scope.parent.replace(Some(Arc::downgrade(self)));
@@ -3438,9 +3493,10 @@ mod tests {
 
     fn isolated_scope(id: &'static str, flavor: ScopeFlavor) -> Arc<ScopeCell> {
         let mut identity = ScopeIdentity::new().expect("scope identity available");
+        let id = ChildId::from(id);
         let member = MemberCell::new(
-            ChildId::from(id),
-            identity.mint_membership().expect("membership available"),
+            id.clone(),
+            identity.mint_membership(&id).expect("membership available"),
         );
         ScopeCell::new(
             member,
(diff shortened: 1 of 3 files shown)


ralexstokes:
Adversarial review complete after rebasing this draft from base `7c401f5` onto `c87d783` (merged #42 and #45).

Scope reviewed:

- per-tree observation-gate ownership, deep adoption, active/waiting observer handoff, ancestor projection, and unrelated-system independence;
- type-erased timer hash/equality, replacement/removal, deadline plus arming order, stale deadline entries, overflow and interval rearming, collision handling, and high-volume lookup work;
- parked-send O(1) unlinking, FIFO/latest promotion, cancellation/accept/termination races, stale IDs, lock order, detached teardown, and deterministic scaling coverage.

Findings fixed in `870ea47`:

1. Adoption could replace a child gate while an observation edge was already executing under the old gate, splitting an atomic parent/child edge. Handoff now waits on the former gate and recursively re-homes resident descendants under the destination gate.
2. Terminal teardown detached the waiter queue before clearing operation registrations; cancellation in that window could try to unlink from the replacement empty queue and panic. Terminal-state withdrawal now recognizes the detached registration safely, while teardown clears it when discharging the detached entry.
3. The mass-cancellation test only exercised repeated tail removal, and timer probe counting depended on randomized hashes not colliding. Coverage now explicitly exercises head/interior/tail unlinking and selects distinct hashes before asserting one exact probe per removal.
4. Added regression coverage for stale waiter IDs and exact erased equality within a timer hash-collision bucket.

Rebase resolution in `a679d78` preserves item 3 by refreshing every intrusive waiter on bind and preserves item 5 by using `Deadline::after(...).instant()` for initial and interval timer arming.

Validation at `870ea47`:

- focused timer store, mailbox, events, deadlines, rebind, and observation-gate tests: pass;
- `./scripts/dev just ci`: pass (231 tests plus fmt, clippy `-D warnings`, docs, API/runtime-path checks);
- `./scripts/dev just ci-nix`: pass (all clean Nix checks).

No further actionable findings remain. The PR is intentionally still a draft.

ralexstokes:
Rebased onto current `main` (`441ab7d`) and force-pushed as `6230cd7` with an exact lease. Conflict resolution preserves the item-8 per-scope observation-gate handoff, recursive descendant re-homing, mailbox waiter/cancellation behavior, and indexed timer semantics while retaining merged-main restart accounting, raw cleanup coverage, and membership identity rules. Local validation passed: `./scripts/dev just ci` (250 tests) and `./scripts/dev just ci-nix`.

ralexstokes (commented):
Reviewed all three subsystems adversarially; no correctness findings. This also implements #39 in full (now linked from the description), including the live-writer re-gating case #39 deferred: instead of `Arc::as_ptr`-ordered dual locking, adoption holds the destination gate (via the admission edge) plus the old tree gate, and since multi-gate acquisition only ever follows admission edges — which are acyclic — no deadlock cycle is possible.

**What was verified**

*Per-tree observation gate (driver.rs)*
- The double-check in `with_observation_gate` plus adoption holding the old gate through descendant re-homing covers every interleaving I could construct: an in-flight edge blocks the handoff; a waiter that acquired the obsolete gate re-reads the field and retries; a reader that picked up the new gate blocks on the admission edge until the parent link is set, so no observer ever sees a half-adopted tree.
- `set_admitted_children`/`admit_child` are the only `parent.replace` sites and both adopt first; `publish_snapshot_chain_locked`/`emit_locked` walk parent links that are only ever set under the shared gate, so cross-scope forwarding stays atomic within a tree.
- The gate read inside the admission critical section is stable: swapping a cell's gate requires holding its current gate, which the admission edge owns.

*Mailbox waiter queue (mailbox.rs)*
- Lock order is uniformly mailbox → operation across `submit`, `withdraw`, and promotion; `Termination::drain` holds no mailbox lock but operates on a detached queue.
- The one hazard of `mem::take` resetting `next_id` to 0 — a stale `WaiterId` from the detached queue aliasing a new registration — is unreachable: `Terminal` is absorbing (`bind` refuses it) and `submit` terminates rather than parks under it, so the fresh queue never receives entries. `close()` deliberately leaves `waiters` in place, so IDs stay monotonic across unbind/rebind.

*Timer store (raw.rs)*
- `take_due` over `(Instant, arming_order)` reproduces the old `sort_unstable` firing order exactly.
- Overflow-safe deadlines are preserved: a `None` deadline never enters the `BTreeSet`, on arming or interval re-arm.
- Cancel-between-fire-and-deliver is benign — `deadlines.remove` on an already-consumed index is a no-op, and `deliver_timer` tolerates a missing arming.

**Validation**: merged current `origin/main` (c7ff8b4, no semantic overlap — its raw.rs changes are readiness plumbing) into the PR head and ran `./scripts/dev just ci`: 258/258 passed, fmt/clippy/docs clean.

**Optional nits, non-blocking**
- `TimerStore::entry_mut` uses `?` on `keyed.get_mut(&hash)` where the `armings` invariant guarantees presence — an `expect` (as in `remove_arming`) would surface an invariant break instead of silently skipping delivery.
- `deliver_timer`'s interval path does two `entry_mut` lookups (interval check, then mutate); collapsible to one, though both are O(1).

PR #50, opened 2026-08-07
Make readiness and completion verdicts authoritative

ralexstokes:
Part of #35 (item 4)

## What changed

- wait for terminal membership publication before a one-shot task claim releases its typed value
- return every competing terminal `Exit` (failure, panic, abort, never-started, or readiness timeout) instead of an early `Ok(T)`
- resolve each raw incarnation's effective readiness once, bypassing `RawActor::readiness` entirely when the definition supplies an override
- pass that single readiness value to both the driver gate and `RawContext`

## Why

The typed value channel previously became observable before the driver had classified the final exit, so a value returned after a readiness deadline could incorrectly win over `ReadinessTimedOut`. Raw actor readiness was also evaluated independently by the driver and runner, allowing stateful implementations to disagree.

## Impact

One-shot completion now reflects the authoritative terminal verdict. Raw actors receive the exact readiness mode used by supervision, with one actor readiness evaluation per incarnation and no unnecessary evaluation under an override.

## Validation

- focused task and raw-readiness integration tests
- `./scripts/dev just ci` (212 tests plus formatting, Clippy, docs, and repository guards)
- `./scripts/dev just ci-nix` (all clean Nix checks passed)

PR #53, opened 2026-08-07
Reclaim dynamic runtime state

ralexstokes:
Part of #35 (item 1)

## What changed

- Store scope children in a reusable generational arena and carry generation-checked child keys through driver events and deadlines.
- Reclaim dynamic child registrations after explicit removal and `Retention::Remove`, while preserving Removed-before-completion ordering.
- Give readiness, stop, and restart deadlines removable handles; cancelled registrations release their payloads immediately and compact heap tombstones so storage stays proportional to live deadlines.
- Back internal change signals with Tokio watch channels, eliminating the quiet-period weak-waiter tombstones left after the #37 channel migration.
- Preserve exact readiness/deadline tie semantics by allowing same-clock-edge producers one turn to publish their retained readiness latch before due-deadline collection.

## Why

The scope driver retained every historically admitted child, deadline payloads lived until their original due time, and the remaining internal signal implementation kept cancelled weak waiters until a later pulse. Long-lived dynamic scopes therefore paid memory and scan costs for historical churn, and numeric slot reuse would have made direct reclamation unsafe for late events.

The new keys and handles make those registrations explicitly removable without letting stale events address a replacement resident.

## Validation

- `./scripts/dev just ci`
- `./scripts/dev just ci-nix`
- 1,000-cycle dynamic add/remove stress with one reused child slot and prompt far-future readiness-deadline reclamation
- 10,000-cycle deadline cancellation/compaction stress
- 10,000 repeatedly cancelled quiet-period `changed()` waits
- stale-generation slot reuse test

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -4,13 +4,13 @@ use std::{
     collections::HashMap,
     fmt,
     future::Future,
+    ops::{Index, IndexMut},
     pin::Pin,
     sync::{
         Arc, Mutex, Weak,
         atomic::{AtomicBool, AtomicU64, Ordering},
         mpsc,
     },
-    task::{Context, Poll, Waker},
     time::{Duration, Instant},
 };
 
@@ -19,9 +19,9 @@ use crate::{
     ScopeState, ShutdownStraggler, ShutdownTimeout, StartupFailure, StartupFailureCause,
     deadline::Deadline,
     engine::{
-        ArbitrationClass, DeadlineQueue, ExitDispatch, IntensityState, MembershipMode,
-        ReadinessEffect, ReadinessEvent, ReadinessGate, RestartState, ScopeMode, StopAction,
-        StopLadder, arbitrate, dispatch_exit, schedule_restart,
+        ArbitrationClass, DeadlineHandle, DeadlineQueue, ExitDispatch, IntensityState,
+        MembershipMode, ReadinessEffect, ReadinessEvent, ReadinessGate, RestartState, ScopeMode,
+        StopAction, StopLadder, arbitrate, dispatch_exit, schedule_restart,
     },
     exit::{JoinVerdict, RecordedOutcome, classify_exit},
     identity::{FenceCounter, ScopeIdentity},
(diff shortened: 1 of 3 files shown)


ralexstokes:
Adversarial review completed against `main` at `c87d783` and the published PR head `8e98452`.

Finding addressed:

- The first deadline implementation made registrations removable, but used a never-reused `u64` handle and panicked when that counter exhausted. That did not satisfy the proposed generation-keyed lifecycle and exposed a driver-fatal edge.
- Replaced it with a reusable generational slot arena. Cancelled registrations drop their payload immediately, old handles fail closed after slot reuse, exhausted generations retire their slot, and arming-order exhaustion rebases live heap entries without changing FIFO order for equal deadlines.

Coverage strengthened:

- proves prompt payload destruction, stale-handle rejection, cancelled-earliest wake recomputation, bounded heap tombstones and registration slots over 10,000 cycles, generation exhaustion, and equal-deadline FIFO across order rebasing;
- extends the 1,000-cycle dynamic stress test to cover both explicit removal and automatic `Retention::Remove`, with exact live/slot/deadline storage bounds each cycle;
- forces child-generation exhaustion and verifies the exhausted slot is retired safely.

I also re-reviewed dynamic admission/removal, never-started reservations, ordered startup, stale `DriverEvent`/`DeadlineKind` delivery, readiness/stop/restart cancellation, overflow-as-never deadlines, exact-deadline arbitration under virtual time, and the Tokio-watch signal migration. No additional actionable finding remained.

Validation at `8e98452`: `./scripts/dev just ci` (229 tests) and `./scripts/dev just ci-nix` both pass.

ralexstokes:
Rebased onto `main` at `441ab7d` (`c888bfb`). Resolved the driver overlap by retaining the generation-checked child/deadline arenas and reclamation coverage alongside the merged scope lifecycle/scaffolding changes. Local `./scripts/dev just ci` passes (248 tests), and `./scripts/dev just ci-nix` passes.

ralexstokes:
Follow-up rebase onto `main` at `c7ff8b4` after PR #50 merged (`d05b751`). The construction callback overlap preserves PR #50’s authoritative readiness argument and PR #53’s generation-checked child key. Local `./scripts/dev just ci` passes (256 tests), and `./scripts/dev just ci-nix` passes.

ralexstokes (commented):
Reviewed at `d05b751`; one finding, resolved directly on the branch in `f5c9092`. `./scripts/dev just ci` passes before and after.

## What I verified

- **Generational keys close the slot-reuse hazard end to end.** Every consumer of `ChildKey`/`DeadlineHandle` re-resolves through `get`/`get_mut` (or a guarded `Index` on a key proven live in the same handler), and every removal bumps the slot generation, so a late event or fired deadline can miss but never address a replacement resident. Generation exhaustion (`u64::MAX`) retires the slot instead of recycling it, in both arenas.
- **Ordered-scope positional invariants hold.** The `key_at(...).expect("ordered children are never reclaimed...")` calls in `progress_startup`/`fail_startup` are safe: reclamation only happens via `finalize_removal` (requires dynamic control, which only `ScopeFlavor::Dynamic` gets) or `prune_terminal` (explicitly gated on `Dynamic`).
- **No leaked deadline registrations.** All four `deadlines.push` sites store the returned handle, and every path that invalidates an incarnation (ready, exit, stop, restart-spawn, reclaim) cancels the corresponding handle. Handles that went stale via `pop_due` cancel as harmless no-ops. `handle_constructed` correctly cancels the just-pushed handle if the incarnation raced away.
- **Removal responses can't hang on a dead driver.** The terminal-resident removal path now depends on driver liveness, but `Obligation<RemovalResponses>` + `DynamicControl::close()` in `ScopeRuntime::Drop` complete outstanding responses on every driver-exit path, and the removal scan runs even while draining.
- **Signal-on-watch is sound.** `WatchReceiver::changed` swallows `Err(closed)` into an immediate return, which would spin — the `_signal` clone held by each `SignalWatcher` is what prevents closure, and the comment documents exactly that. Conflation of pulses is fine because the driver loop rescans all sources per wake.
- **Heap tombstone bounds.** Cancel-time compaction keeps `entries ≤ 2 × live` at every cancel, so storage is bounded by 2× peak live registrations; head tombstones are additionally pruned by `next`/`pop_due`. The `next_arming_order` rebase at sequence exhaustion preserves relative order of live entries and keeps handles valid.

## Finding (resolved in `f5c9092`)

- `remove_dynamic`'s terminal-resident branch became byte-for-byte identical to the general path below it once both were routed through the driver — the special case was dead duplication. Collapsed into one path, keeping the explanatory comment.

## Notes (no change made)

- The `yield_now()` on `ScopeWake::Deadline` is a best-effort turn-yield: deterministic under the current-thread test runtime, heuristic under a multi-threaded runtime where the same-instant producer may sit on another worker. That matches the stated intent ("one turn to publish"), just noting the tie is not a hard guarantee off the test runtime.
- Storage stays proportional to *peak* live deadlines rather than current live (tombstone ratio is only re-checked at cancel time, not as live entries pop due). Bounded and fine; mentioning in case the PR text's "proportional to live deadlines" was meant more strictly.
- Test coverage is strong: stale-generation, exhaustion, compaction bounds, watch-registration stability, and the 1k-cycle storage-reuse stress all pin the new invariants.

Issue #56, opened 2026-08-08
Refactor-first: fix layering inversion and collapse duplicated machinery to retire bug classes


ralexstokes:
## Summary

A second file-by-file and cross-module review at baseline `de197e6bb8db0c371cac8c445f0e741543b2955e` (same baseline as #55, independent pass). This issue is deliberately **refactor-first**: the deep structural problems below are each the *cause* of a class of bugs, and several concrete bugs found in this review fall out for free once the structure is fixed. Standalone bugs that no refactor subsumes are listed after, then test follow-ups.

Overlap with #55 is cross-referenced per item rather than repeated; where a refactor here subsumes a #55 checklist item, that is noted.

Estimated reduction from the refactor program: **~2.5–3k of ~12k production lines, with no public API change.**

## Phase 1 — Fix the layering inversion around `driver.rs`

The intended layering (leaf types → mailbox/raw → actor/task → driver → tree) is inverted at every driver edge: `mailbox.rs`, `raw.rs`, `task.rs`, and `tree.rs` all import *upward* from the 5.2k-line driver. Two causes, both mechanical to fix:

- [x] **Move the driver's runtime facade into `runtime.rs`.** [`driver.rs:44–233`](https://github.com/ralexstokes/shelterwood/blob/de197e6bb8db0c371cac8c445f0e741543b2955e/crates/shelterwood/src/driver.rs#L44-L233) is pure re-wrapping: `driver::now` = `runtime::now()`, `Selected`/`select` = renamed `runtime::Either`/`select_two`, `Signal` wraps `runtime::watch(())` + `pulse`, `runtime_available` = `runtime::is_available`, `sleep_deadline`/`sleep_until` wrap `runtime::sleep_until_std`. The crate is not even consistent — `tree.rs:1730` and the driver loop itself reach the same primitives directly. Consumers to re-point: mailbox.rs (10 sites), raw.rs (8 sites), task.rs, tree.rs.
[...]

PR #58, opened 2026-08-08
Fix restart-window shutdown and offload wakeups

ralexstokes:
Stack 1 of the issue #56 refactor program.

## What changed

- Expedite a nested scope's pending restart incarnation when shutdown lands during restart backoff.
- Preserve SPEC B.9 semantics: pre-spawn requests still park, and the shutdown timeout still arms only once teardown begins.
- Handle finite and unrepresentable restart deadlines without waiting for the configured backoff.
- Wake the parent driver through the member-record control plane and make duplicate same-batch restart starts inert.
- Couple every escaped offload-destructor panic recording with the actor event signal.
- Add public coverage for the shutdown-versus-startup race producing `StartupError::ShutdownRequested`.

## Why

A scope-level shutdown request issued between nested-scope incarnations only pulsed the stopped nested scope. Its parent remained asleep until the restart deadline, so a long backoff delayed shutdown and an unrepresentable deadline could hang forever. Separately, an offload future whose destructor panicked after cancellation recorded the panic without waking an otherwise idle actor.

## User impact

`ScopeRef::shutdown_and_wait` now promptly begins the pending incarnation's teardown during restart backoff, including `Duration::MAX`, while retaining incarnation-targeted and pre-spawn behavior. Escaped offload destructor panics are promptly supervised without requiring unrelated mailbox or timer activity.

## Validation

- `./scripts/dev just ci`
  - 306 nextest tests
  - 3 doctests
  - rustdoc and public API reachability
  - runtime/exit architecture checks
  - packaged-crate verification
  - Nix formatting check

Part of #56.

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -1143,6 +1143,7 @@ impl ScopeCell {
 
     pub(crate) fn request_shutdown(&self) -> u64 {
         let mut control = self.control.lock().expect("scope control mutex poisoned");
+        let pending_incarnation = !control.live;
         let target = if control.live {
             control.current_epoch
         } else {
@@ -1159,7 +1160,24 @@ impl ScopeCell {
         }
         drop(control);
         self.member.record.pulse();
+        // A nested scope has no live driver while its parent is retaining it
+        // in restart backoff. Wake that parent as well so it can start the
+        // pending incarnation immediately; that incarnation consumes the
+        // stop and begins the timeout-bounded teardown required by B.9.
+        if pending_incarnation
+            && let Some(parent) = self.parent.read_cloned().as_ref().and_then(Weak::upgrade)
+        {
+            parent.member.record.pulse();
+        }
         target
+    }
+
+    fn has_pending_incarnation_shutdown(&self) -> bool {
+        let control = self.control.lock().expect("scope control mutex poisoned");
+        !control.live
+            && control.shutdown.is_some_and(|request| {
+                !request.consumed && request.epoch > control.last_stopped_epoch
+            })
     }
 
     fn take_shutdown_request(&self, epoch: u64) -> bool {
@@ -2270,6 +2288,22 @@ enum SpawnBody {
 }
 
 impl ScopeRuntime {
+    fn pending_restart_shutdowns(&self) -> Vec<ChildKey> {
+        self.children
+            .keys()
+            .filter(|key| {
+                let child = &self.children[*key];
+                child.active.is_none()
+                    && matches!(child.slot.member.record().stage, MemberStage::Restarting)
+                    && child
+                        .slot
+                        .scope
+                        .as_ref()
+                        .is_some_and(|scope| scope.has_pending_incarnation_shutdown())
+            })
+            .collect()
+    }
+
     #[cfg(test)]
     fn record_storage(&self) {
         *self
@@ -2288,7 +2322,11 @@ impl ScopeRuntime {
         let Some(child) = self.children.get(key) else {
             return;
         };
-        if self.phase.is_draining() || child.is_terminal() || child.is_disposing() {
+        if self.phase.is_draining()
+            || child.active.is_some()
+            || child.is_terminal()
+            || child.is_disposing()
+        {
             return;
         }
         let child = &mut self.children[key];
@@ -3752,6 +3790,12 @@ async fn run_scope_incarnation(
         if root.take_shutdown_request(epoch) {
             pending.push((ArbitrationClass::ScopeShutdown, Pending::Shutdown));
         }
+        for child in scope.pending_restart_shutdowns() {
+            pending.push((
+                ArbitrationClass::ScopeShutdown,
+                Pending::RestartShutdown(child),
+            ));
+        }
         if !scope.ancestor_shutdown_seen
             && scope
                 .ancestor_shutdown
@@ -3870,6 +3914,7 @@ async fn run_scope_incarnation(
                     }
                     scope.begin_drain(StopReason::ShutdownRequested);
                 }
+                Pending::RestartShutdown(child) => scope.spawn_child(child),
                 Pending::AncestorShutdown => {
                     scope.begin_drain(StopReason::ShutdownRequested);
                 }
@@ -3946,6 +3991,7 @@ async fn run_scope_incarnation(
 
 enum Pending {
     Shutdown,
+    RestartShutdown(ChildKey),
     AncestorShutdown,
     AncestorAbort,
     Force,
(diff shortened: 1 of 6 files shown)


crates/shelterwood/src/driver.rs
```diff
@@ -3752,6 +3790,12 @@ async fn run_scope_incarnation(
         if root.take_shutdown_request(epoch) {
             pending.push((ArbitrationClass::ScopeShutdown, Pending::Shutdown));
         }
+        for child in scope.pending_restart_shutdowns() {
+            pending.push((
+                ArbitrationClass::ScopeShutdown,
```
ralexstokes: [P1] Preserve removal/teardown precedence for this synthetic restart. `remove_dynamic` synchronously sets `record.removing`, fires the removal latch, and pulses this same driver, but `pending_restart_shutdowns()` still selects a `Restarting` member and this assigns it `ScopeShutdown`, ahead of `MembershipRemoval`. The stable batch then runs `spawn_child` first, invoking the subtree factory and starting an incarnation even though SPEC §11 says the removal latch suppresses restarts from the instant it is set; the ordinary `BackoffDue` path cannot do this. The same insertion order puts this event before an already-fired `AncestorShutdown`/`AncestorAbort` (and ahead of `Force` by class), so parent teardown can likewise spawn-then-immediately-stop the child. Recheck `removing`/parent teardown at execution or arbitrate this after those controls, and add deterministic same-batch tests asserting the factory count does not advance for removal and ancestor shutdown.

ralexstokes (commented):
Blocking arbitration finding: the new pending-incarnation fast path bypasses existing suppression precedence for removal and parent teardown. The normal restart-deadline path is `BackoffDue`, so removals and shutdown controls win before `spawn_child`; the new path is inserted as the highest-priority `ScopeShutdown` event and can invoke user construction first. Please preserve those suppressions and add deterministic same-batch coverage. I also checked SPEC B.9’s delayed timeout arming, finite/overflowing backoffs, duplicate deadline/start handling, the raw offload signal handoff/lost-wakeup window, and the new public tests; I found no other actionable issue in those paths.

crates/shelterwood/src/driver.rs
ralexstokes: Fixed in 1d0c23c. The expedited path now rechecks suppression both while collecting and immediately before execution: current scope shutdown/force requests (including consumed same-batch requests), ancestor shutdown/abort latches, draining/hard-force state, synchronous `record.removing`/member removal, and fused dynamic cancellation/removal state all prevent the spawn. Added a deterministic paused-time same-batch removal + nested-shutdown regression that asserts the subtree factory/start counters remain at one. `./scripts/dev just ci` is green with 307 tests plus all docs/API/architecture/package checks.

ralexstokes: Re-review at `1d0c23c`: the execution-time checks fix the level-triggered cases from this thread. The paused removal regression genuinely latches removal and nested shutdown before yielding and proves no second factory/start. `has_stop_request(epoch)` intentionally still sees consumed same-batch shutdown/force records; the ancestor shutdown/abort latches are read directly, so those paths are covered by the code without needing separate integration fixtures.

One P1 remains: `RestartShutdown` is still classified as `ScopeShutdown`, ahead of every queued `ChildExit`. A same-wake exit from another child can transition this scope into `IntensityTripped` drain (directly in `handle_exit`), or complete a terminal pre-ready path into startup-failure drain, but `restart_is_suppressed()` cannot see that unprocessed event. The expedited handler therefore invokes the pending subtree factory first and only then processes the exit and drains it. This is exactly the ordering SPEC §14.2 protects when it says child exits precede backoff work so all newly observed terminal facts apply before restart. Please arbitrate this spawn as restart/backoff work (or otherwise process child-exit drain transitions first) and add a same-batch exit/intensity-trip regression that keeps the pending subtree’s factory count unchanged.

ralexstokes: Addressed the remaining same-wake ordering hole in e5a7268. `RestartShutdown` is now classified as `BackoffDue`, so queued `ChildExit` work runs first and can transition the scope into intensity/startup-failure drain; the existing execution-time suppression check then prevents construction. Added `same_batch_intensity_exit_suppresses_expedited_restart_work` to pin that ordering, and `./scripts/dev just ci` is green (308 tests).

ralexstokes: Re-review at `e5a7268`: the production ordering fix is correct. Classifying `RestartShutdown` as `BackoffDue` restores SPEC §14.2 ordering: every collected `ChildExit` runs first, and an intensity/startup-failure transition makes `restart_is_suppressed()` observe `phase.is_draining()` before construction. Together with the level-triggered checks from `1d0c23c`, the functional P1 is resolved.

One test-coverage action remains: `same_batch_intensity_exit_suppresses_expedited_restart_work` is a synthetic restatement, not a regression over the fixed path. It manually assigns the expedited event `BackoffDue`, manually flips a local `draining` bool, and manually suppresses a counter; it never exercises `run_scope_incarnation`'s production classification, `handle_exit` intensity tripping, `expedite_restart_shutdown`, or the subtree factory. It would stay green if the production line reverted to `ScopeShutdown`, and the existing `arbitration_order_is_explicit_and_stable` already proves `ChildExit < BackoffDue`. Please replace/augment it with a test coupled to the production mapping and suppression path—ideally the requested same-batch intensity fixture asserting the pending subtree factory count remains unchanged, or at minimum a shared production classification helper plus a runtime-level suppression test. CI for this head is still pending as of this re-review.

ralexstokes: Replaced the synthetic regression in 73e3b2d. Production now shares `restart_shutdown_work` and `driver_event_class` between the live loop and the test. The driver-level test constructs a real `ScopeRuntime`, makes a retained subtree eligible for expedited restart, feeds a real failing `Exited` event through `handle_exit` to trip intensity, then executes the production expedited path and observes the actual factory counter after scheduler turns. I also mutation-checked it locally: changing the shared production classifier back to `ScopeShutdown` makes this test fail (factory count 1 vs 0); restored `BackoffDue` passes. Full `./scripts/dev just ci` is green at 308 tests.

ralexstokes: Final re-review at `73e3b2d`: resolved. `restart_shutdown_work` is now the single production classifier used by the live loop and the test, and `driver_event_class` likewise supplies the real `ChildExit` class. The replacement test builds a real `ScopeRuntime`, makes the nested member genuinely eligible for expedited restart, processes a real failing `Exited` event through `handle_exit` to trip the configured zero-budget intensity, then runs the actual `expedite_restart_shutdown` path and observes the real subtree factory after scheduler turns. Reclassifying the shared helper to `ScopeShutdown` necessarily reverses these pending events and explains the reported mutation failure (`1` factory vs expected `0`), so the test now pins both the production mapping and the runtime suppression effect. The original removal/ancestor/force arbitration P1 and the follow-up coverage concern are fully addressed; I found no remaining actionable issue in this PR. GitHub Actions for `73e3b2d` was still pending when I posted this conclusion.

PR #59, opened 2026-08-08
Refactor supervision state into proper layers

ralexstokes:
Stacked on #58. Implements Phase 1 of #56.

## Summary

- move the runtime facade (`Signal`, watchers, actor/blocking work, deadline sleeps) into `runtime.rs` and repoint lower-layer consumers
- extract restart-stable member/scope state and observation projection into `cells.rs`
- extract declaration lowering IR and construction ownership into `plan.rs`
- move `StopReason`/`StartupError` to `exit.rs` and `ScopeFlavor` to `policy.rs`
- replace `ChildArena`'s generational vector slots with monotonic `u64` keys in a `BTreeMap`
- make `ScopeCell` the single source of truth for scope flavor
- document and assert mailbox configure/bind/close ordering
- add a repository check preventing upward driver/tree imports across the new boundaries

## Invariants preserved

- mailbox termination publishes before terminal member wakeups
- observation updates share one tree-wide gate, including dynamically admitted subtrees
- declaration/plan/runtime Drop paths retain exactly one terminality owner
- child keys are monotonic, ordered, never reused, and poison on exhaustion

## Validation

- `./scripts/dev just ci` (309 tests)
- `./scripts/dev just ci-nix`

Part of #56.

### DIFF
--- a/tools/check-layering-paths.sh
+++ b/tools/check-layering-paths.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+readonly lower_layers=(
+  crates/shelterwood/src/mailbox.rs
+  crates/shelterwood/src/raw.rs
+  crates/shelterwood/src/task.rs
+)
+readonly cells_path="crates/shelterwood/src/cells.rs"
+
+check_forbidden() {
+  local message="$1"
+  local forbidden="$2"
+  shift 2
+
+  local matches
+  local status
+  set +e
+  matches="$({ rg --line-number "$forbidden" "$@"; } 2>&1)"
+  status=$?
+  set -e
+
+  case "$status" in
+    0)
+      echo "$message" >&2
+      echo "$matches" >&2
+      exit 1
+      ;;
+    1) ;;
+    *)
+      echo "$matches" >&2
+      exit "$status"
+      ;;
+  esac
+}
+
+check_forbidden \
+  "upward driver references found below the driver layer:" \
+  '\bdriver::' \
+  "${lower_layers[@]}"
+
+check_forbidden \
+  "upward driver or tree references found in the shared cells layer:" \
+  '\b(driver|tree)::' \
+  "$cells_path"
+
+echo "shared-cell layering restrictions: clean"
--- a/tools/check-layering-paths.sh
+++ b/tools/check-layering-paths.sh
@@ -1,12 +1,21 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-readonly lower_layers=(
+readonly below_driver_layers=(
+  crates/shelterwood/src/actor.rs
+  crates/shelterwood/src/cells.rs
+  crates/shelterwood/src/deadline.rs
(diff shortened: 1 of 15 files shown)


tools/check-layering-paths.sh
```diff
@@ -0,0 +1,53 @@
...
+      echo "$matches" >&2
+      exit "$status"
+      ;;
+  esac
+}
+
+check_forbidden \
+  "upward driver references found below the driver layer:" \
+  '\bdriver::' \
+  "${lower_layers[@]}"
```
ralexstokes: **[P2] Cover the other extracted below-driver boundaries.** `driver.rs` now imports `plan`, so `plan.rs` is part of the below-driver DAG, and this PR also fixes the old `observe -> tree` edge by moving `StopReason` to `exit`. Neither `plan.rs` nor `observe.rs` is scanned here, and the lower-layer set is forbidden from importing `driver` but not `tree`. Consequently `plan.rs` can add `use crate::driver::...`/`tree::...`, or `observe.rs` can restore a tree dependency, while this check still reports clean—recreating the cycles Phase 1 is meant to retire. Please include those modules in the appropriate checks and forbid both upper modules where the intended DAG requires it; matching aliased/root-reexport imports as well would keep the guard from being syntactically bypassed.

ralexstokes (commented):
Found one actionable Phase 1 gap (P2) in the new architecture guard. I otherwise found no correctness or invariant regression in the stacked diff.

Audited: runtime-facade relocation; `MemberCell`/`ScopeCell` extraction; observation-gate adoption and recursive projection; `ResidentChild`/`ScopePlan`/`ChildRuntime` Drop ownership; mailbox configure/bind/close and terminal-publication ordering; type-erased dynamic routing; scope-flavor deduplication with nested root overrides; monotonic `ChildArena` ordering, staleness, reclamation, and poison behavior; public re-exports; and the complete Phase 1 checklist in #56.

Validation observed/run: GitHub CI run 31275953017 succeeded; focused driver tests (32/32) and mailbox tests (6/6) passed; the layering script and `git diff --check` passed.

tools/check-layering-paths.sh
ralexstokes: Resolved at `0f8e96c`. The guard now enumerates every production module below the driver—including `plan.rs` and `observe.rs`—and forbids both explicit `driver::` and `tree::` references across that set; `driver.rs` separately forbids `tree::`. I reran `./tools/check-layering-paths.sh` and independently searched the covered paths, both clean. The alias/root-reexport note was optional hardening beyond the concrete coverage gap, so the P2 is fully addressed.

ralexstokes (commented):
Re-reviewed at `0f8e96c`: no actionable findings remain. The prior P2 is fully resolved by applying the `driver`/`tree` prohibition to all production modules below the driver and retaining the separate `driver -> tree` prohibition. The expanded check passes, the covered paths are clean under an independent search, `git diff --check` passes, and the reported full `just ci` run is green. I replied to and resolved the original thread.

PR #60, opened 2026-08-08
Consolidate mailbox deadline operations

ralexstokes:
Stacked on #59. Implements Phase 2 of #56.

## Summary

- replace the three hand-rolled deadline scaffolds in `ReplyReceive`, `SendTimeout`, and `CallFuture` with one internal `Deadlined<F>` combinator
- move replies onto the existing runtime oneshot abstraction and make reply-vs-timeout arbitration atomic with close-then-receive
- share timed-send/call withdrawal routing and capture timeout incarnation evidence under the mailbox → operation lock order
- let an immediately ready send win a zero-width send deadline while preserving zero-width call construction short-circuiting
- replace the intrusive waiter linked map with monotonic never-reused `u64` keys in `BTreeMap`, including permanent poison on exhaustion
- document Drop-as-completion and mailbox/operation lock ordering

## Semantics pinned

- operation completion/acceptance wins at the exact deadline
- call construction consumes the single overall budget
- cancellation before acceptance recovers the message
- reply send racing timeout has one atomic winner
- timed-out sends report the latest bound incarnation consistently
- stale waiter IDs cannot remove later registrations

## Validation

- `./scripts/dev just ci` (312 tests)
- `./scripts/dev just ci-nix`

Part of #56.

### DIFF
--- a/crates/shelterwood/src/runtime.rs
+++ b/crates/shelterwood/src/runtime.rs
@@ -23,6 +23,12 @@ use crate::deadline::Deadline;
 #[cfg(test)]
 pub(crate) use tokio::test;
 
+/// Advances a paused test clock, keeping timer control in this module.
+#[cfg(test)]
+pub(crate) async fn advance(duration: Duration) {
+    time::advance(duration).await;
+}
+
 pub(crate) fn now() -> std::time::Instant {
     time::Instant::now().into_std()
 }
--- a/crates/shelterwood/src/runtime.rs
+++ b/crates/shelterwood/src/runtime.rs
@@ -8,7 +8,7 @@ use std::{
     pin::Pin,
     sync::{
         Arc, Mutex,
-        atomic::{AtomicBool, AtomicUsize, Ordering},
+        atomic::{AtomicBool, AtomicU8, AtomicUsize, Ordering},
     },
     time::Duration,
 };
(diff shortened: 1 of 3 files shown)


crates/shelterwood/src/runtime.rs
```diff
@@ -395,6 +395,23 @@ impl<T> OneShotSender<T> {
...
+    ) -> std::task::Poll<Option<T>> {
+        Pin::new(&mut self.0).poll(context).map(Result::ok)
+    }
+
+    /// Closes the receive side and returns a value whose send won the race.
+    ///
+    /// `tokio::sync::oneshot` makes `close` atomic with `send`; checking the
+    /// slot after closing therefore gives deadline users one transition point
+    /// at which an already-completed delivery wins and every later send loses.
+    pub(crate) fn close_and_try_receive(&mut self) -> Option<T> {
```
ralexstokes: **[P2] Preserve the sender-drop winner at the deadline boundary.** `close()` is atomic with a value-bearing `send`, but `try_recv().ok()` collapses Tokio's empty completed channel and receiver-closed channel into the same `None`. A valid interleaving is: the elapsed poll sees `poll_receive == Pending`; another thread drops the unanswered `Reply` (Tokio marks the channel complete with no value); then `close_and_try_receive` runs and returns `None`. Both `ReplyOperation` and `CallOperation` classify that as timeout even though sender drop won first, violating the documented `Dropped`/`ReplyDropped` distinction (and the latter's retry discipline). The reverse ordering yields the same `None`, so the receiver alone cannot recover the winner. Please retain an explicit atomic terminal kind alongside the oneshot (or otherwise linearize drop vs timeout) and cover this exact-boundary drop race.

ralexstokes (commented):
Adversarial Phase 2 review at `9dc6f33` found one actionable correctness issue:

- **P2 — sender-drop vs deadline is not linearized, so the public error kind can be wrong.** The inline comment has the concrete interleaving. The value-send/close race is handled correctly, but `Reply` drop uses Tokio oneshot completion with an empty slot, and this helper erases whether that completion or receiver closure won. As a result, an unanswered capability dropped before the deadline transition can be surfaced as `ReplyError::Timeout` / `CallErrorKind::ResponseTimedOut` instead of `Dropped` / `ReplyDropped`. Those variants carry different documented retry semantics.

I also audited first-poll budget capture, exact-boundary value/acceptance precedence, zero-duration send versus call behavior, send cancellation/message recovery, timeout incarnation evidence, call construction budget consumption, public future `Send` conformance, waiter FIFO/stale-key/exhaustion behavior, and mailbox→operation lock ordering. I found no other actionable issues.

Focused validation passed:
- `./scripts/dev cargo test -p shelterwood --test integration deadlines::`
- `./scripts/dev cargo test -p shelterwood --test integration api_trait_conformance::documented_identity_handle_token_and_owned_value_bounds_compile`
- `./scripts/dev cargo test -p shelterwood --lib mailbox::tests::`

ralexstokes (commented):
Re-review at exact head `0a6426528db641873139e8e80ceb6854be6d66a0`: **no remaining actionable findings**.

The prior P2 is fully fixed. The shared transition word now linearizes all three relevant contenders before the receiver classifies the outcome:
- sender drop wins as `SenderClosed` → `Dropped` / `ReplyDropped`;
- receiver close wins as `Empty` → timeout;
- value send wins as `Value`, including the exact deadline boundary.

I also checked the intermediate `SENDING` state: the receiver re-polls the Tokio channel with the operation waker already registered, so a send that won but has not yet published cannot lose its completion wake. The generic runtime-oneshot consumers retain their previous send/drop behavior and public future `Send` conformance remains intact.

Focused re-validation passed:
- `./scripts/dev cargo test -p shelterwood --lib runtime::tests::`
- `./scripts/dev cargo test -p shelterwood --test integration deadlines::`
- `./scripts/dev cargo test -p shelterwood --test integration mailbox::`
- `./scripts/dev cargo test -p shelterwood --test integration delivery::`
- `./scripts/dev cargo test -p shelterwood --test integration tasks::one_shot`
- `./scripts/dev cargo test -p shelterwood --test integration api_trait_conformance::documented_identity_handle_token_and_owned_value_bounds_compile`
- `./scripts/dev cargo test -p shelterwood --lib mailbox::tests::`

The prior inline thread is resolved.

crates/shelterwood/src/runtime.rs
ralexstokes: Fixed and verified at `0a6426528db641873139e8e80ceb6854be6d66a0`. The runtime oneshot now carries an atomic transition state that distinguishes sender drop, value send, and receiver close; deadline classification follows that winner, with a registered-waker `Pending` state only while a winning send publishes its payload. Focused runtime, deadline, mailbox, delivery, one-shot task, and API conformance tests pass. Resolving this thread.

PR #62, opened 2026-08-08
Consolidate definition source state and setters

ralexstokes:
Part of #56. This draft is stacked on #61.

## What changed

- replace the separate raw, scope, and one-shot source state machines with one private `DefinitionSource<R, O>`
- generate the shared option setters across all eight definition families from one private macro while retaining each family's special surfaces
- extend layering enforcement for the new module
- add ownership/transition tests and exact setter API coverage

The implementation/tooling code is net 61 lines smaller; the commit is net +12 only because it adds 73 lines of integration coverage.

## Validation

- `./scripts/dev just ci` — 320/320 tests plus Clippy, docs, API reachability, layering, and package checks
- `./scripts/dev just ci-nix` — all clean Nix checks passed

Please review the single commit above `agent/issue-56-contexts`.

PR #63, opened 2026-08-08
Collapse tree builder method surfaces

ralexstokes:
Part of #56. This draft is stacked on #62.

## What changed

- generate the common `Tree` and `DynamicTree` builder surfaces from one private macro
- preserve the two nominal public builder types, every documented signature and return type, and flavor-specific spawn behavior
- keep `Tree::strategy` explicit where the surfaces intentionally differ
- add exact-type API conformance calls for every generated method on both builders

Tree implementation code is net 72 lines smaller.

## Validation

- `./scripts/dev just ci` — 321/321 tests plus Clippy, docs, API reachability, layering, and package checks
- `./scripts/dev just ci-nix` — all clean Nix checks passed

Please review the single commit above `agent/issue-56-definitions`.

### DIFF
--- a/crates/shelterwood/src/tree.rs
+++ b/crates/shelterwood/src/tree.rs
@@ -61,6 +61,138 @@ fn attach_actor_slot<M: Send + 'static>(slot: Arc<SlotCell>) -> ActorSlot<M> {
     ActorSlot { slot, mailbox }
 }
 
+macro_rules! impl_common_builder_surface {
+    (
+        reserve_actor: $reserve_actor_doc:literal,
+        add_actor: $add_actor_doc:literal,
+        add_actor_once: $add_actor_once_doc:literal,
+        add_raw: $add_raw_doc:literal,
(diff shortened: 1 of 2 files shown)


ralexstokes (commented):
Adversarial review of the single commit at `31e08a59b3a8fa729ba4689ba4fa4bf2bc4bcf88`: no actionable findings.

I compared normalized rustdoc JSON for every inherent `Tree` and `DynamicTree` method against base `38dea71daaabc81459dda2bb16dccce836685fc9`; documentation, visibility, signatures, generic parameters, bounds, and return types are identical. I also checked each generated body against the removed implementation: reservation flavor selection, actor/raw/task/subtree restartable vs. one-shot dispatch, initial dynamic membership semantics, and typed handle/claim returns are preserved. `Tree::strategy` and both flavor-specific `spawn` methods remain explicit and unchanged, and the targeted API conformance test passes.

PR #66, opened 2026-08-08
Centralize never-started slot terminalization

ralexstokes:
Part of #56. This draft is stacked on #65.

## What changed

- make `SlotCell` the single implementation of member-plus-nested-scope never-started publication
- combine unlowered definition claiming with that publication for reservation/removal paths
- route builder drop, plan drop, dynamic close, reservation cancellation, pre-admission removal, and admission rejection through the shared slot ritual
- retain the existing isolated disposal funnel so dynamic removal completion still follows terminal publication and user-definition destruction

This removes all structural copies of the slot-level ritual and is net 6 production lines smaller.

## Validation

- focused dynamic module — 22/22 tests
- focused disposal module — 16/16 tests
- `./scripts/dev just ci` — 326/326 tests plus Clippy, docs, API reachability, layering, and package checks
- `./scripts/dev just ci-nix` — all clean Nix checks passed

Please review the single commit above `agent/issue-56-child-planning`.

PR #67, opened 2026-08-08
Route terminal outcomes after joined disposal

ralexstokes:
## Summary

- route incarnation exhaustion through the same joined-disposal terminal funnel as ordinary exits
- keep dynamic removal, retention pruning, and ordered-scope progression pending until retained factory disposal finishes
- remove synchronous terminalization from restart-window shutdown and ordered startup rollback
- publish never-started nested scope state without stealing terminal ownership from the supervising scope

## Why

Issue #56 identified a divergent incarnation-exhaustion ladder that pruned retention incorrectly and completed dynamic removal before isolated disposal. The same early-completion shape was also reachable while stopping between restart incarnations and while terminalizing an ordered startup suffix.

This change leaves disposal-panic ownership unchanged: only a concrete exited incarnation can be reclassified by its retained factory destructor. Never-started and between-incarnation children keep their authoritative verdict while still joining disposal before terminal routing.

## Verification

- focused incarnation-exhaustion unit coverage
- blocking-disposal regressions for restart-window removal and ordered startup rollback
- all 18 disposal integration tests
- `./scripts/dev just ci` (329 tests plus clippy, docs, API/layering checks, and packaging)
- `./scripts/dev just ci-nix`

Part of #56. Stacked on #66.

### DIFF
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -690,6 +690,15 @@ impl ScopeCell {
             if matches!(record.stage, MemberStage::Terminal(_)) {
                 return false;
             }
+            // Nested publication and closure are reached only through this
+            // residency lookup, so a supervised child must still be resident
+            // here when it is terminalized. That holds because residency is
+            // installed before the child can be spawned (`set_admitted_children`
+            // for a planned incarnation, `admit_child` before dynamic
+            // `spawn_child`) and is only withdrawn by pruning, which always
+            // follows terminality. A child that has already left residency
+            // owns its nested scope through `SlotCell::terminalize_never_started`
+            // instead.
             let nested = self.current_children.read_with(|children| {
                 children
                     .iter()
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -686,18 +686,9 @@ impl ScopeCell {
         startup_aborted: bool,
     ) -> bool {
         self.with_observation_gate(|| {
-            if matches!(member.record().stage, MemberStage::Terminal(_)) {
+            let record = member.record();
+            if matches!(record.stage, MemberStage::Terminal(_)) {
                 return false;
-            }
-            member.update_locked(|record| record.startup_aborted = startup_aborted);
-            member.terminalize(exit.clone());
-            if let Some(incarnation) = exited_incarnation {
-                self.emit_locked(LifecycleEventKind::Exited {
-                    id: member.id().clone(),
-                    membership: member.membership(),
-                    incarnation,
-                    exit: exit.clone(),
-                });
             }
             let nested = self.current_children.read_with(|children| {
                 children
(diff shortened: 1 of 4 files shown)


ralexstokes:
Adversarial review of exact head `fd3b192140389598b62626a73e85e21c5d34c243` against `f3ddfc6b90d857881ac73b3f3d1b2a1b00513f8a`:

**[P1] Keep startup failure behind the joined rollback barrier** — `crates/shelterwood/src/driver.rs:2013-2028`

`fail_startup` publishes `root.set_startup(Err(...))` before it starts the ordered suffix disposals. Before this patch, the suffix was terminalized synchronously in the same driver turn; now terminality is correctly deferred to the asynchronous `ConstructionDisposed` event. Consequently `wait_started()` can return while a retained never-ran suffix is still `Admitted`, so the public startup result no longer means startup rollback has reached terminal state.

This is observable with an existing test at this exact head:

```text
target/debug/deps/integration-1d0d0e61a3e6b8c9 observation::never_ran_members_stop_rather_than_report_startup_abort --nocapture
...
a never-ran terminal is Stopped, not StartupAborted: Some(ChildSnapshot { ... state: Admitted, ... })
```

The full focused observation filter produced 16 passes and this failure; running the single test again failed identically. CI happened to pass because disposal can win the scheduling race. Holding terminality until disposal is right, but the startup error/`StartupFailed` publication must also wait until the rollback suffix has completed joined disposal and terminal routing (for example via an explicit rollback-pending state/count).

**[P2] Make the no-incarnation terminal transition one observation-gated publication** — `crates/shelterwood/src/driver.rs:1970-1985`, `crates/shelterwood/src/cells.rs:1211-1217`

`prepare_never_started_terminal()` emits the nested `Stopped` event and publishes ancestor snapshots, then releases the shared observation gate before the parent member is terminalized. At that publication point the parent child is still `Admitted`/`Restarting` and its nested snapshot is omitted. The following `terminalize_child(..., exited_incarnation: None, ...)` mutates the member and closes nested observation but emits no `Exited` event and calls no `publish_snapshot_chain_locked()`, so a retained child can leave existing parent snapshot subscribers permanently on that pre-terminal value. Incarnation exhaustion after a previous incarnation has the same missing-publication problem even without `prepare_never_started_terminal`; the removed explicit `transition_child` used to provide that publication.

Please perform nested never-started state, parent member terminality, final snapshot publication, and nested observation closure under one shared observation-gate transition, and ensure every `exited_incarnation == None` terminal route publishes the parent snapshot even when retention is `Retain`.

The 18 disposal integration tests passed; the findings are in the observation/rollback boundary rather than the blocking-disposal mechanics.

ralexstokes:
Fresh adversarial re-review of exact amended head `ac2a67119c9548fa3f20c02b376f91664cc1778a` against `f3ddfc6b90d857881ac73b3f3d1b2a1b00513f8a`: **clean; no remaining actionable P0-P3 findings.**

The prior findings are resolved as follows:

- The former P1 is retracted. `System::wait_started()` is documented to wait for readiness or terminal startup failure, not full rollback. Publishing that error before joined suffix disposal is necessary so `start_or_shutdown(timeout)` can initiate shutdown and bound a blocking destructor; delaying it behind rollback would defeat that API contract. The amended observation test now waits for the never-ran child terminal edge before asserting its classification, which tests the intended asynchronous boundary.
- The former P2 is fixed. `ScopeCell::terminalize_child` now makes parent member terminality, nested never-started state/event publication, final parent projection publication, and nested observation closure under the shared observation gate. The explicit no-incarnation snapshot publication also covers retained incarnation-exhaustion/restart-window outcomes that have no `Exited` event. The new parent `SnapshotReceiver` regression exercises this directly.

I re-audited dynamic removal/retention routing, ordered rollback and reverse-drain progression, hard-force detachment and late completion idempotency, destructor-panic ownership, and nested observation ordering. Focused exact-head results: 17/17 observation tests and 18/18 disposal tests passed; `git diff --check` is clean and the worktree is clean. GitHub CI was still pending at the time of this comment.

PR #68, opened 2026-08-08
Preserve authoritative task exit evidence

ralexstokes:
## Summary

- make exit precedence explicit and table-driven: panic, readiness timeout, application failure, abort, completion, then never-started
- preserve an already-recorded application error when the task join is later cancelled by forced shutdown
- keep panic authoritative and retain the first verdict on equal precedence
- make `OneShotTaskRef::wait` fail loudly on the impossible `Completed`-without-value invariant instead of returning `Err(Completed)`

## Why

Issue #56 identified two contradictory diagnostics: forced cancellation could discard an application error, and a completed one-shot could surface its completed exit through the error branch. The new classifier separates authoritative task evidence from join fallback and exhaustively tests the precedence matrix.

## Verification

- 102 library tests
- `./scripts/dev just ci` (330 tests plus clippy, docs, API/layering checks, and packaging)
- `./scripts/dev just ci-nix`

Part of #56. Stacked on #67.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1232,14 +1232,36 @@ One classification, produced at one point, used by every consumer.
   already says nothing ran.
 - **Verdict precedence.** When one incarnation's end admits several
   readings, classification picks the highest of: `Panicked` >
-  `ReadinessTimedOut` > `Aborted` > `Failed` > `Completed`; `cancelled` is
+  `ReadinessTimedOut` > `Failed` > `Aborted` > `Completed`; `cancelled` is
   orthogonal and never competes. Concretely: a panic is never masked,
   wherever it lands (`run`, `on_stop` — superseding the run's outcome,
-  §4.1 — or a destructor, via the fallback report token); a
+  §4.1 — or a destructor, via the fallback report token); and a
   readiness-deadline expiry names the *cause* even when the teardown it
-  triggers ends in a grace-expiry abort (the mechanism); and `Aborted`
-  describes a future destroyed before yielding an outcome, so it cannot
-  genuinely conflict with `Failed`/`Completed`, which require one.
+  triggers ends in a grace-expiry abort (the mechanism).
+- **`Aborted` genuinely competes with a recorded outcome, and the rule is
+  asymmetric.** `Aborted` describes a future destroyed before yielding an
+  outcome, so it reads as if it could never conflict with
+  `Failed`/`Completed`, which require one. It conflicts anyway, by race:
+  the body can record its result and the destruction still land before
+  the join retires, and §10's ladder can arm a supervisor-forced abort
+  verdict against a membership that has already recorded one. Precedence
+  resolves that race in one direction only, and the asymmetry is
+  deliberate rather than an artifact of the ordering:
+  - A recorded `Failed(error)` **survives** a later abort. Destruction
+    proves only that teardown ended the task; it must not erase the
+    structured application error the task already produced, which is the
+    only evidence naming *why* the child was failing.
+  - A recorded `Completed` does **not** survive: cancellation overrides
+    it and the exit reads `Aborted { after_grace }` with
+    `cancelled: true`. A supervisor that destroyed a child before its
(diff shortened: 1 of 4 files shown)


ralexstokes:
Adversarial review of exact head `e6c53c030744eccf23159ee536bd71720a06e02f` against base `fd3b192140389598b62626a73e85e21c5d34c243`.

**[P2] Route forced outcomes through the same precedence fold** — `crates/shelterwood/src/driver.rs:1815`

`active.forced_outcome.or(recorded)` throws away the task's recorded evidence before `classify_exit` sees it. For a nested scope, the hard-abort ladder installs `RecordedOutcome::Aborted` at `driver.rs:1554-1557`. If that scope had already entered an intensity-trip or startup-failure drain when its ancestor forces it, `run_nested_tree` still returns the authenticated structured `ExitError` at `driver.rs:2392-2395`, but the parent selects the lower-ranked forced abort and publishes `ExitKind::Aborted`. That contradicts this PR's documented `Failed > Aborted` ordering and continues to lose the structured diagnostic in a reachable forced-shutdown race (restart dispatch remains failure-equivalent, but the public exit evidence changes).

Please fold `forced_outcome`, the task-reported outcome, and the join verdict together under one precedence rule instead of using `Option::or`. A regression should cover a nested scope that has already tripped/fails startup and is then hard-forced by its parent before its drain completes. Extending the unit table to the full Cartesian matrix would also cover currently omitted cases such as recorded panic vs cancellation and readiness timeout vs join panic.

Focused verification run: exit-classifier unit test, completed-without-value invariant test, all `tasks::` integration tests (11), and all `one_shot_resource_ownership::` integration tests (15). All executed tests passed.

ralexstokes:
Re-review of exact amended head `ec0c584a1ef2ead789772f5542d57a94aa39192b` against base `ac2a67119c9548fa3f20c02b376f91664cc1778a`: **no findings**.

The prior P2 is resolved. `handle_exit` now reconciles task-produced and supervisor-forced evidence before folding in the join verdict. The ordering remains `Panicked > ReadinessTimedOut > Failed > Aborted > Completed`; consequently, an authenticated nested-scope startup/intensity failure survives a forced abort, while a higher-ranked forced readiness timeout still supersedes a returned outcome. The `>=` comparison keeps the task-produced value on equal rank, preserving the established left/earlier-evidence tie behavior and its payload/provenance. Join panic and cancellation then compose with the reconciled outcome under the same ordering, and the independent `cancelled` observation remains intact.

I also rechecked the one-shot publication path and restart dispatch implications: typed values remain gated on authoritative `Completed`, and the `Failed`-versus-`Aborted` correction does not change failure-restart eligibility.

Targeted verification passed:

- all three `exit::tests`
- all 11 `tasks::` integration tests
- `readiness::nested_dynamic_startup_failure_rolls_back_and_preserves_inner_cause`
- all 20 `subtrees::` integration tests

ralexstokes:
Adversarial re-review of exact amended head dcf0fc0654fbe33ea3b96f9ca8b04d36c4edf6d1 against base ac2a67119c9548fa3f20c02b376f91664cc1778a: **no findings**.

The OutcomeRank refactor directly expresses the diagnostic precedence through derived Ord: NeverStarted < Completed < Aborted < Failed < ReadinessTimedOut < Panicked. Both provisional task-versus-forced reconciliation and final recorded-versus-join classification map into that one private rank type, so the two folds cannot silently drift through unrelated numeric literals. The >= comparisons preserve the established tie rule: earlier task-produced evidence retains its payload and provenance when ranks match. A forced readiness timeout or panic still supersedes weaker task evidence, while a forced abort cannot erase an authenticated application or nested-scope failure.

OutcomeRank, recorded_rank, rank, and reconciliation remain crate-private; the public Exit and ExitKind surface is unchanged. The repository public runtime-API reachability check is clean.

Verification at the exact head passed:

- all 104 shelterwood unit tests
- all 226 shelterwood integration tests
- all 3 shelterwood doctests
- just runtime-api-check

PR #72, opened 2026-08-08
Simplify deadline queue registration arena

ralexstokes:
Part of #56. Stacked on #71.

## Summary
- replace generational slot reuse with monotonic deadline handles
- keep active registrations in one keyed arena and queued ordering in the deadline heap
- make stale cancellation fail closed while releasing payloads and bounding heap storage
- retain equal-deadline FIFO ordering and explicit poison/exhaustion behavior

## Verification
- ./scripts/dev just ci (335/335 tests)
- ./scripts/dev just ci-nix
- git diff --check

PR #73, opened 2026-08-08
Move remaining scope policy into engine

ralexstokes:
Part of #56. Stacked on #72.

## Summary
- make readiness configuration, transitions, deadline effects, and tie arbitration engine-owned
- move scope startup, drain, finish, epoch liveness, and stable-run restart settling into pure policy types
- prevent stale scope drivers from publishing over newer incarnations
- split child construction dispatch from task spawning and document latch/report liveness
- make ordered immediate startup iterative and poison scope-epoch exhaustion rather than aliasing generations

## Verification
- focused engine tests: 18/18
- focused driver tests: 35/35
- 4,096-child iterative-startup regression
- scope-epoch exhaustion regression
- ./scripts/dev just ci (339/339 tests)
- ./scripts/dev just ci-nix
- git diff --check

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -2951,6 +2951,36 @@ mod tests {
     }
 
     #[test]
+    fn a_declined_epoch_still_publishes_its_owned_terminal_exit() {
+        let scope = isolated_scope("scope", ScopeFlavor::Ordered);
+        let epoch = scope.begin_incarnation().expect("scope epoch is available");
+        // The orderly finisher retires the epoch without a terminal exit, so
+        // a second owner still holds the only membership verdict.
+        scope.finish_incarnation(epoch, StopReason::Finished);
+        assert!(!matches!(
+            scope.member.record().stage,
+            MemberStage::Terminal(_)
+        ));
+
+        scope.finish_root_incarnation(
+            epoch,
+            StopReason::ShutdownRequested,
+            Exit::new(ExitKind::Aborted { after_grace: false }, true),
+        );
+        assert!(matches!(
+            scope.member.record().stage,
+            MemberStage::Terminal(_)
+        ));
+        assert_eq!(
+            scope.record().state,
+            ScopeState::Stopped {
+                reason: StopReason::Finished,
+            },
+            "a declined epoch must not rewrite the retired stop reason"
+        );
+    }
+
+    #[test]
     fn admitted_subtrees_share_their_parent_observation_gate() {
         let root = isolated_scope("root", ScopeFlavor::Ordered);
         let nested = isolated_scope("nested", ScopeFlavor::Dynamic);
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -14,8 +14,9 @@ use crate::{
     deadline::Deadline,
     engine::{
         ArbitrationClass, DeadlineHandle, DeadlineQueue, ExitDispatch, IntensityState,
-        MembershipMode, ReadinessEffect, ReadinessEvent, ReadinessGate, RestartState, ScopeMode,
-        StopAction, StopLadder, arbitrate, dispatch_exit, schedule_restart,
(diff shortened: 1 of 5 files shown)


crates/shelterwood/src/driver.rs
```diff
@@ -1288,158 +1409,107 @@ impl ScopeRuntime {
...
+                self.root.transition_child(
+                    &child.slot.member,
+                    |record| record.stage = MemberStage::Running,
+                    Some(LifecycleEventKind::Ready {
+                        id: child.slot.member.id().clone(),
+                        membership: child.slot.member.membership(),
+                        incarnation,
+                    }),
+                );
+                self.progress_startup();
```
ralexstokes: [P1] Keep ordered immediate-readiness startup iterative. `spawn_child` now applies `BecameReady` synchronously, and this call re-enters `progress_startup`; the ordered loop is itself still inside `spawn_child` for the prior key and has not advanced `next_ordered_start` yet. Each default-`Immediate` ordered child therefore nests another `progress_startup -> spawn_child -> BecameReady -> progress_startup` frame. A sufficiently large ordinary ordered tree can overflow the runtime worker stack and abort the process. The pre-refactor path marked immediate children ready without re-entering the pump, so its `while` loop stayed constant-stack. Please defer/pump startup progress iteratively (or otherwise prevent synchronous re-entry) and add a many-immediate-children regression test.

crates/shelterwood/src/engine.rs
```diff
@@ -291,52 +309,327 @@ pub(crate) fn schedule_restart(
...
+impl Default for ScopeEpochs {
+    fn default() -> Self {
+        Self::Idle { last_stopped: 0 }
+    }
+}
+
+impl ScopeEpochs {
+    pub(crate) fn begin(&mut self) -> u64 {
+        let (current, last_stopped) = match *self {
+            Self::Idle { last_stopped } => (last_stopped.saturating_add(1), last_stopped),
```
ralexstokes: [P2] Do not alias exhausted scope epochs. With `last_stopped == u64::MAX`, `begin()` returns a live `current == u64::MAX` while retaining `last_stopped == u64::MAX`; `finished(u64::MAX)` is then already true during that live incarnation, so `shutdown_scope` can return before it stops. A second `begin()` at saturation also mints the same epoch, allowing the older driver's `finish(MAX)` to stop the newer projection—the stale-driver guarantee this PR is meant to establish. The new saturation test misses both cases. Please poison/fail closed on exhaustion (as the other monotonic domains do) rather than saturating into a reusable live id, and test `finished` plus repeated `begin` at the boundary.

ralexstokes (commented):
Adversarial review of exact head `0b4eba749cfaebb0208867bb1c8995911919b2be` against `6297beff142371c9391ce7516cf60dad524549ed` found two correctness issues. Focused engine tests (18/18), driver tests (35/35), `git diff --check`, and the GitHub `ci` check all pass; the findings are boundary/control-flow gaps not exercised by those tests.

ralexstokes (commented):
Exact-head adversarial re-review of `b4c8c2c407908040318dc09a544272632d6fcd15` against base `5a27d1a8f2209677963786c21a8e2b769e2512ec`: no remaining actionable findings.

Both prior findings are resolved: readiness effects no longer synchronously re-enter the ordered startup pump (with a 4,096-child regression), and scope epochs now reject live replacement and permanently poison before the reserved `u64::MAX` value. I also audited all readiness-effect and epoch Option callers, including deadline signal-wins, restart/startup, nested exhaustion, and exhausted shutdown paths.

Evidence: corrected patch is range-diff equivalent to reviewed `cc0632c`; engine tests 18/18; driver tests 35/35; large ordered startup regression 1/1; `git diff --check` clean; GitHub CI passed.

ralexstokes (commented):
Final exact-head adversarial verdict for `da960bddb7e7a456a08e53085b8ba895a01a0015` against `88b0e3d1a1b7a31701c0e151094f2532449c7b2b`: no actionable findings. The patch is range-diff identical to the corrected/reviewed `b4c8c2c`, both prior regressions remain fixed, `git diff --check` is clean, and GitHub CI passed.

PR #74, opened 2026-08-08
Make dynamic admission an explicit state machine

ralexstokes:
Part of #56. Stacked on #73.

## Summary
- replace Admission correlated Options and booleans with Immediate, Unpolled, InFlight, and Done states
- guard runtime availability at reservation and first poll, returning ReserveError::NoRuntime
- cancel and terminalize failed first polls synchronously so IDs are immediately reusable
- preserve fused cancellation and split detachment semantics under select and timeout
- make post-completion polling safely fused
- document NoRuntime timing, error precedence, ownership, terminalization, reuse, and fused polling in the normative SPEC

## Verification
- dynamic integration tests: 25/25 before stacking
- focused NoRuntime/error-precedence regressions
- ./scripts/dev just ci after final stacking (341/341 tests)
- ./scripts/dev just ci-nix
- git diff --check

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1526,8 +1526,9 @@ Rules (normative):
   point for the id errors — `EmptyId` and `DuplicateId` on either flavor
   (a retained tombstone counts as a duplicate — retention semantics
   above); on dynamic scopes additionally `RemovalInProgress` (same id
-  mid-removal, §11) and `NotAdmitting` (stage rule below). From success the slot's
-  handles resolve through the cell like any other handle; the slot is
+  mid-removal, §11), `NoRuntime` (runtime boundary below), and
+  `NotAdmitting` (stage rule below). From success the slot's handles
+  resolve through the cell like any other handle; the slot is
   parameterized by exactly what a handle needs (`M` for the wire type,
   `T` for the subtree's ref dispatch) so refs exist before any actor type
   or factory is named.
@@ -1551,10 +1552,11 @@ Rules (normative):
   and cannot fail — spec-level validation already happened eagerly at
   spec construction (§9.3). On a dynamic scope, `define` *is* the
   admission call: a future resolving at admission to the receipt (B.8).
-  Its only error is `NotAdmitting`: the id errors were already spent at
+  Its operation errors are `NotAdmitting` and the first-poll
+  `NoRuntime` rejection below: the id errors were already spent at
   reserve, and definition validation was spent at spec construction
   (§9.3's eager rule), exactly as on the builder flavor — admission
-  validates nothing. Whether the builder and dynamic flavors
+  validates no definition data. Whether the builder and dynamic flavors
   are one generic slot type or two parallel families is implementation
   latitude; the operation inventory and this error split are not.
 - **Fused admission futures abort on drop; split ones detach.** Dropping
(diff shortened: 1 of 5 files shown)


ralexstokes:
Adversarial review of exact head `775118b781916b977b869724ac64e0b980db283c` against `0b4eba749cfaebb0208867bb1c8995911919b2be`:

**[P2] Update the normative dynamic-admission error contract for `NoRuntime`.**

This change adds public `ReserveError::NoRuntime` and deliberately returns it both from synchronous dynamic reservation and from the first `Admission` poll. However, `SPEC.md` Appendix B says its operation/variant shapes are content-normative, and B.8 currently says dynamic `add_*` fails with **exactly** the old union (`EmptyId`, `DuplicateId`, `RemovalInProgress`, `IdentityExhausted`, and `NotAdmitting`) while separately describing `BuildError::NoRuntime` as spawn-only. The implementation and normative contract therefore disagree at this exact head.

Please update the SPEC in this PR to include `ReserveError::NoRuntime` and pin the two relevant edges: (1) reservation requires an entered runtime, and (2) losing runtime before first poll returns `NoRuntime`, synchronously terminalizes/releases that reservation, and leaves the ID immediately reusable. The contract should also state error precedence (the code currently validates `EmptyId` before checking the runtime, then prefers `NoRuntime` over scope-state and resident-ID errors). Without that, two conforming consumers cannot infer the same observable result for compound failure cases.

I otherwise found no actionable defect in the enum transitions, pin/fuse behavior, fused-cancel race, split detachment boundary, response fallback, or same-membership cancellation logic.

Evidence from the exact-head worktree:

- complete diff inspection for `driver.rs`, `plan.rs`, `tree.rs`, and `tests/dynamic.rs`
- `dynamic::admission_runtime_guards_leave_reservation_ids_reusable` — passed
- `dynamic::select_and_timeout_preserve_fused_and_split_admission_ownership` — passed
- `dynamic::fused_drop_withdraws_or_removes_while_split_drop_detaches` — passed
- `dynamic::actor_and_subtree_slots_preserve_fused_and_split_drop_ownership` — passed
- GitHub CI at this exact head — passed

ralexstokes:
Adversarial re-review of corrected exact head `a6207362651cabd24d3133882419fb557bf8f080` against exact base `b4c8c2c407908040318dc09a544272632d6fcd15`: **no actionable findings**.

The prior P2 is resolved. `SPEC.md` now makes `ReserveError::NoRuntime` part of the content-normative dynamic-admission contract and matches the implementation on all requested edges:

- dynamic reservation validates `EmptyId` first, then checks runtime availability before scope-state, removal, duplicate-ID, or identity-minting work;
- a first poll outside a runtime returns `NoRuntime` before submission, synchronously releases and terminalizes the reservation as `NeverStarted`, leaves the ID reusable before returning `Ready`, and does not cross the split-detach boundary;
- completed admissions are specified as fuse-like and remain pending on later polls;
- B.8 now includes `NoRuntime` in the exact public error union and distinguishes dynamic define from builder define behavior.

I re-inspected the complete restacked diff and found the Admission enum transitions, pin/fuse behavior, fused cancellation races, split ownership boundary, response fallback, same-membership cancellation, and actor/task/subtree funnels consistent with that contract.

Exact-head evidence:

- `git diff --check b4c8c2c407908040318dc09a544272632d6fcd15..a6207362651cabd24d3133882419fb557bf8f080` — clean
- all 25 `dynamic::` integration tests — passed
- focused runtime-guard reuse, select/timeout ownership, fused/split task drop, and actor/subtree drop tests — passed individually
- GitHub `ci` for exact head `a6207362651cabd24d3133882419fb557bf8f080` — passed

The corrected PR is clean from this review.

ralexstokes:
Final exact-head verification for `aaa745923e0e73cad531efc013264470bce31912` against base `da960bddb7e7a456a08e53085b8ba895a01a0015`: **no actionable findings**.

`git range-diff` reports the patch exactly equal to corrected/reviewed commit `a6207362651cabd24d3133882419fb557bf8f080`, `git diff --check` is clean, and GitHub `ci` passed on this exact final head. The prior SPEC finding remains resolved.

PR #75, opened 2026-08-08
Validate public policy literals at admission

ralexstokes:
Part of #56. Stacked on #74.

## Summary
- validate public readiness, intensity, backoff, defaults, and inherited policy literals before ownership transfer
- reject invalid static trees, dynamic admissions, nested one-shot trees, and runtime-produced subtrees with structured provenance
- revalidate resolved values so literal enum fields cannot bypass constructor checks
- preserve the shared static/dynamic ChildPlan construction funnel and cleanup ordering

## Verification
- table and integration coverage for every invalid literal/admission surface
- ./scripts/dev just ci (350/350 tests)
- ./scripts/dev just ci-nix
- git diff --check

PR #76, opened 2026-08-08
Close standalone supervision races

ralexstokes:
Part of #56. Stacked on #75.

## Summary
- assert the lifecycle event ring capacity is a power of two at compile time
- publish only the first root StartupFailed outcome under concurrent startup failures
- route construction disposal completion through a runtime-independent, nonblocking completion lane
- release dynamic-state locking before entering the observation transition gate
- exercise trusted-ID validation directly at the driver boundary

## Verification
- concurrent startup-failure, disposal fallback, lock-order, and EmptyId regressions
- ./scripts/dev just ci (353/353 tests)
- ./scripts/dev just ci-nix
- git diff --check

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -406,9 +406,12 @@ pub(crate) fn remove_dynamic(
     // registration is reclaimed before the removal response completes.
     let member = Arc::clone(&entry.slot.member);
     // Dynamic-state protects admission/removal bookkeeping; the observation
-    // gate protects the public projection. Never nest the former around the
-    // latter: shutdown takes the same resources in the opposite order while
-    // clearing residents and closing dynamic state.
+    // gate protects the public projection. Release the former before entering
+    // the latter. No path takes the two in the opposite order, so this is not
+    // breaking an existing cycle; it keeps an unbounded wait out of the
+    // bookkeeping mutex. Any thread may hold the gate across arbitrary
+    // observation work, and blocking there while holding dynamic state would
+    // stall every concurrent reservation, removal, and driver admission.
     drop(state);
     scope.transition_child(&member, |record| record.removing = true, None);
     member.removal.fire();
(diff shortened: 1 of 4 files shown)


ralexstokes:
Exact-head adversarial review of `dbe57b46a744a7914112fc27bf9620524cdc4edb..6c80c6ed8daf0cd0472e62b1f598503bbaebdcf2`: **no blocking findings**.

I traced the standalone changes through their surrounding ownership and wake paths:

- Startup failure is first-wins in the authoritative `ScopeLifecycle`; later failures still terminalize their children but cannot replace the cause or republish the `StartupFailed` scope edge.
- Construction disposal reports over a dedicated nonblocking lane. Completion is enqueued before the retained scope signal is pulsed, the driver drains that lane before every sleep, and driver cancellation safely falls back to `ScopeRuntime` terminality discharge. I found no lost-completion or lost-wake path. The unbounded queue is limited in practice by outstanding terminal disposals and removes the bounded-queue worker wedge this change targets.
- Dynamic removal releases `DynamicState` before entering the observation transition, removing the lock inversion while preserving the entry/removal obligation until driver handling or teardown.
- The lifecycle-capacity assertion correctly locks the public constant to Tokio broadcast's effective power-of-two capacity assumption.
- Empty IDs are rejected by `checked_id` at the trusted `reserve_dynamic` boundary before runtime or scope-state precedence checks.
- New channel aliases/helpers remain `pub(crate)`; I found no public API or layering regression.

Verification on the exact head:

- GitHub Actions run 31281540532 passed at head `6c80c6ed8daf0cd0472e62b1f598503bbaebdcf2`.
- Focused tests passed: `dynamic_removal_releases_state_before_waiting_for_the_observation_gate`, `incarnation_exhaustion_uses_post_disposal_retention_routing`, `concurrent_initial_failures_publish_one_startup_failed_scope_edge`, and `reserve_dynamic_rejects_an_empty_id_at_the_driver_boundary`.
- The exact base/head diff passes `git diff --check`.

ralexstokes:
Final exact-head revalidation for `6a688fe46ef23bb3626397c29c188262a78e2d43..1fa3ac823324087449d4f17df747340a6bed7d66`: **no blocking findings**.

`git range-diff` reports the restacked commit as patch-identical to the previously reviewed clean patch (`6c80c6e = 1fa3ac8`), and the exact range passes `git diff --check`. GitHub Actions run 31282446523 passed with `headSha` exactly `1fa3ac823324087449d4f17df747340a6bed7d66`.

PR #77, opened 2026-08-08
Trim mailbox and raw runtime machinery

ralexstokes:
Part of #56. Stacked on #76.

## Summary
- remove the dead frozen-mailbox branch from next_ready and correct stop/receive plus batch-fairness documentation
- trim dead mailbox diagnostics and collapse disposal, queued-event, one-shot timer, and slot-claim wrappers
- centralize hostile-safe panic-containment vocabulary in the runtime layer
- preserve mailbox ordering assertions and document Drop completion, lock order, spawn-latch/report topology, and member-watch wake-bus invariants
- directly test RestartPolicy::is_never and the consolidated mailbox/slot behavior

## Verification
- integrated with the later engine, admission, validation, and standalone-race changes
- ./scripts/dev just ci (356/356 tests; 3/3 doctests)
- ./scripts/dev just ci-nix
- git diff --check

### DIFF
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -30,22 +30,12 @@ use crate::{
 
 /// Isolated payload returned after mailbox termination has synchronously
 /// published all waiter outcomes.
-pub(crate) trait MailboxDisposal: Send {}
+pub(crate) type MailboxDisposal = Box<dyn Send>;
 
 /// Prepared terminal mailbox transition. Finishing it wakes terminal waiters
 /// before returning unread payload ownership for detached disposal.
 pub(crate) trait MailboxTermination: Send {
-    fn finish(self: Box<Self>) -> Option<Box<dyn MailboxDisposal>>;
-}
-
-#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
-pub(crate) struct MailboxStats {
-    pub(crate) accepted: u64,
-    pub(crate) delivered: u64,
-    pub(crate) conflated: u64,
-    pub(crate) sends_rejected: u64,
-    pub(crate) depth: usize,
-    pub(crate) capacity: usize,
+    fn finish(self: Box<Self>) -> Option<MailboxDisposal>;
 }
 
 /// Type-erased mailbox lifecycle surface owned by a member cell.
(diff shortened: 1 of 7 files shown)


ralexstokes:
Adversarial review at exact base `6c80c6ed8daf0cd0472e62b1f598503bbaebdcf2` and head `aee079e8301d5348e6745c596d809d3d62810296`: no findings.

I reviewed the full patch and its range-diff from the original isolated trim, with special attention to the integrations over the engine, admission, policy-validation, and standalone-tail changes.

- `next_ready` had only live-mailbox callers; removing the permanently-false frozen branch preserves fairness and stop behavior, while stopping still drains only through the direct `try_recv` path.
- Removing `MailboxStats` deletes internal diagnostic counters only; the accepted-sequence watermark used by timer/mailbox ordering remains intact.
- The centralized panic helpers preserve first-panic precedence, run every cleanup, contain recursive panic-payload destructors, and avoid double panic during an existing unwind.
- `MailboxDisposal`, `QueuedEvent`, and one-shot `TimerMessage` are ownership-equivalent simplifications.
- The consolidated fallible `SlotCell` claim preserves both old modes: strict static lowering explicitly expects success, while disposal, never-started, and dynamic race paths treat an already-claimed definition as absent.
- Mailbox configure/close-before-bind assertions, report/latch topology, lock-order integration, and the member-watch wake-bus documentation are consistent with the stacked runtime changes.

Independent verification at that exact head:
- `./scripts/dev cargo test -p shelterwood --lib`: 117 passed
- `./scripts/dev cargo test -p shelterwood --test integration`: 236 passed
- `git diff --check`: clean
- GitHub `ci`: passed in 1m33s

ralexstokes:
Final exact-head adversarial verdict for base `1fa3ac823324087449d4f17df747340a6bed7d66` and head `01257142752f3b946f45c7147bf53f901b7167cd`: no findings.

`git range-diff` reports the final commit as exactly patch-equivalent to the previously reviewed clean commit (`aee079e = 0125714`), so the earlier behavioral and conflict-adaptation review applies unchanged. The GitHub CI rerun, run 31282456851, passed in 1m28s. The prior exact-patch independent verification also remains applicable: 117 library tests and 236 integration tests passed, with a clean diff check.

PR #93, opened 2026-08-09
Extract shared cancellation token

ralexstokes:
## Summary

- move `CancellationToken` from the task implementation into a dedicated shared cancellation module
- import the token directly from that module in task, raw-actor, and callback-actor code
- move the cancellation-specific unit tests with the implementation
- keep the layering scanner and its fixture tree comprehensive for the new module

## Why

`CancellationToken` is a read-only cancellation capability shared across supervised tasks, raw actors, callback actors, and actor-owned blocking work. Defining it in `task.rs` implied task ownership that the runtime architecture does not have. The original location was historical: tasks introduced the type before actors reused it.

The public API remains `shelterwood::CancellationToken`; this is an internal source-layout reorganization with no behavioral change.

## Validation

- `./scripts/dev just ci`

### DIFF
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -3,10 +3,10 @@
 use std::{fmt, future::Future, hash::Hash, marker::PhantomData, sync::Arc, time::Duration};
 
 use crate::{
-    ActorRef, Blocking, CancellationToken, ChildId, DeadlineElapsed, ExitError, ExitResult, Guard,
-    Incarnation, Mailbox, MailboxShutdown, RawActor, RawContext, RawDef, RawOnceDef, Readiness,
+    ActorRef, Blocking, ChildId, DeadlineElapsed, ExitError, ExitResult, Guard, Incarnation,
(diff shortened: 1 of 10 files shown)


ralexstokes:
Adversarial review complete. I found one actionable coverage gap: although `cancellation.rs` was added to the hand-maintained below-driver scan list, no negative fixture exercised that specific path, so a future accidental omission would still pass the enforcement suite. I added a cancellation-specific upward-driver violation fixture and wired it into the harness in `787fdb8`.

The implementation move, crate-root `shelterwood::CancellationToken` API, relocated tests, rustdoc reachability, and packaged-crate behavior otherwise look sound.

Validation: `./scripts/dev just enforcement-fixture-check` and full `./scripts/dev just ci` pass. The CI mirror explicitly ran `cargo +nightly fmt --all --check` and `cargo +nightly clippy --locked --workspace --all-targets --all-features -- -D warnings`; all 418 tests, doctests, docs/API checks, packaging checks, and enforcement checks passed.

PR #95, opened 2026-08-09
Move child IDs into identity

ralexstokes:
## Summary

- move `ChildId` and its internal validation error from policy into the identity module
- update lowering, policy, and exit internals to import the type from its domain owner
- keep the public crate-root path `shelterwood::ChildId` unchanged

## Why

`ChildId` is the stable textual identity used by child handles, mailboxes, observations, exits, and membership generation. Its previous location in `policy.rs` grouped an identity primitive with configuration data even though policy is only one consumer.

This is a source-layout reorganization with no behavioral or public API change.

## Validation

- `./scripts/dev just ci`

### DIFF
--- a/crates/shelterwood/src/exit.rs
+++ b/crates/shelterwood/src/exit.rs
@@ -7,7 +7,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use crate::{identity::Membership, policy::ChildId};
+use crate::identity::{ChildId, Membership};
 
 /// The terminal reason for a scope incarnation or root system.
 #[derive(Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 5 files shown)


ralexstokes:

I traced the full diff and all direct `ChildId` / `IdError` references through identity, policy, lowering, exits, and the crate-root API. The move preserves `shelterwood::ChildId`, its derives/conversions/display/accessor behavior, and the empty-ID validation/error mapping. The layering direction is also sound: policy now consumes the identity primitive, with no reverse dependency.

Validation: `./scripts/dev just ci` passed, including explicit nightly formatting and linting (`cargo +nightly fmt --all --check` and `cargo +nightly clippy ...`), 418 tests, doctests, rustdoc/API reachability, layering checks, packaged-doc sync, packaged-crate verification, and Nix formatting. Worktree is clean; no follow-up commit was needed. Reviewed head: `4e167da73e3d0dbcb96be7cf270dd5bf9b88af7c`.

Issue #91, opened 2026-08-09
Audit: Arc/Mutex/dyn usage — hot-path wins, dead weight, and spec-conformance findings


ralexstokes:
A full audit of `Arc` / `Mutex` / `dyn Trait` usage across `crates/shelterwood/src` (≈350 uses, ~20k lines). Headline: the architecture is sound and most usage is earned — the driver's decision state is lock-free single-owner `&mut self` (driver.rs `run_scope_incarnation`), no lock is held across `.await`, lock ordering is documented and acyclic, and the `dyn` surface is essentially one fundamental erasure (a supervision tree holds children of heterogeneous `M`/`R`) expressed at four seams (`Arc<dyn MailboxControl>`, `Box<dyn ErasedRawInstance>`, `RawFuture`, `RawFactory`). No re-architecting is warranted.

What remains is a targeted punch list, in priority order.

## Hot-path wins

1. **Lazy `SendOperation` allocation.** `SendFuture::new` (mailbox.rs:1348) eagerly allocates the `Arc<Mutex<SendOperation>>` even for sends accepted immediately. A fast-path `send().await` on a bound, non-full mailbox costs 1 allocation + 4 mutex acquisitions where `try_send` costs 1 acquisition + 0 allocations. Hold an enum `{ Immediate(Option<M>), Parked(Arc<SendOperation<M>>) }` and only materialize the `Arc` when `submit` parks. Biggest single win.
2. **Shrink `SendFuture`'s handle.** It clones a full `ActorRef<M>` (mailbox.rs:1335) but only uses the mailbox `Arc` and the actor id — which `MailboxCell` already stores (mailbox.rs:760). Holding `Arc<MailboxCell<M>>` alone halves per-send refcount atomics. (`CallOperation` genuinely needs the full `ActorRef`.)
3. **Remove the per-event lock in the observation hubs.** `SnapshotHub::state` (observe.rs:337) and `LifecycleHub::channels` (observe.rs:481) take a mutex per published event, per ancestor (via cells.rs:914-916, 945-950) — but the mutex only guards an `Option`; the inner `watch`/`broadcast` channels are already internally synchronized. Restructure to an atomic `closed` flag + always-present sender (the `receiver_count() == 0` check at observe.rs:371 already provides laziness). Also resolves a spec finding, see below.
[...]

PR #98, opened 2026-08-09
Remove redundant type erasure and allocations

ralexstokes:
Stacked on #96; review only the top commit for this PR.

Part of #91.

## Summary

- replace the `Arc<dyn Any>` dynamic-control escape hatch with a named `DynamicRoute` capability implemented by `DynamicControl`, while preserving stale-route identity rejection
- store subtree factories in their final erased `Arc<dyn Fn() -> BuilderCore>` form instead of adding a second closure/vtable at lowering
- remove the unexercised inner `Arc` from application `ExitError`s
- use `Box<dyn Any + Send>` directly for heterogeneous timer keys
- represent interval cloning as `(M, fn(&M) -> M)` instead of a boxed closure
- use a `Box` for the move-once `RawDef` factory
- freeze through `MailboxReceiver`, removing `RawContext`'s third handle to the same mailbox cell

## Validation

- `./scripts/dev just ci`
- 419/419 nextest cases passed, plus doctests, rustdoc, API reachability, runtime/exit/layering enforcement, packaged-crate verification, and formatting/lint checks

### DIFF
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -3,10 +3,10 @@
 //! These cells are the neutral synchronization and observation projection
 //! layer shared by public handles, mailboxes, actor/task contexts, and the
 //! mutable supervision driver. In particular, this module does not depend on
-//! driver state or declaration-plan slots.
+//! mutable driver state. Its dynamic-route interface names declaration slots
+//! only as opaque capability payloads; their state remains owned elsewhere.
 
 use std::{
-    any::Any,
     fmt,
     sync::{
         Arc, Mutex, OnceLock, RwLock, Weak,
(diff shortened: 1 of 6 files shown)


ralexstokes:
Adversarial review confirmed the erasure removals preserve semantics: `ExitError` outer-`Arc` `ptr_eq` identity, timer-key `TypeId` dispatch, interval tick sequence (stored message never sent, cloned per tick as before), `RawDef` factory re-invocation across restarts, and `MailboxReceiver::freeze` incarnation guards are all byte-identical to the previous design.

One regression found and fixed (**`8d5ae75`**): the defensive stale-route branch in `handle_admission` disposed the reservation through the *current* control — where the stale entry can never live, making cleanup a guaranteed no-op — and the route-gone branch skipped disposal entirely with a blanket `Terminal` cause. Unreachable by construction today, but it contradicted the invariant documented at the rejection site. `AdmissionRequest` again holds the concrete `Weak<DynamicControl>` so rejection cleans up through the control that reserved (also sidestepping `dyn`-fat-pointer `ptr_eq` semantics).

PR #100, opened 2026-08-09
Align construction spec and harden invariants

### LINKED ISSUE #91: Audit: Arc/Mutex/dyn usage — hot-path wins, dead weight, and spec-conformance findings

ralexstokes:
Stacked on #98; review only the top commit for this PR.

Closes #91.

## Summary

- apply #91's latest take-once decision by scoping the `Mutex<Option<_>>` prohibition to construction payloads, keeping `DisposalJob`'s non-panicking claim protocol
- update every normative restartable-constructor signature and B.10 to the post-#37 `Send + Sync` reality, and document why `ActorDef::cloned` requires `Args: Clone + Sync`
- replace `ReportReceiver`'s potentially blocking receive with `try_recv().expect(...)`, encoding the post-join invariant directly
- document the `current_children` residency-drop rule that prevents watch-guard self-deadlock
- assert that observation-gate adoption is parent-to-child and uses the parent's installed gate
- explain why offload cancellation deliberately performs both synchronous state cancellation and runtime-task abort

## Validation

- `./scripts/dev just ci`
- 422/422 nextest cases passed, plus doctests, rustdoc, API reachability, runtime/exit/layering enforcement, packaged-crate verification, and formatting/lint checks

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -84,10 +84,13 @@ these, in order.
    escalation, exit classification, child identity: each exists exactly once,
    as a named primitive, and every subsystem consumes that primitive.
    [#361, #363, #364, #370]
-3. **Invariants live in ownership and types, not comments.** A rule that two
-   modules must both remember is a design defect. Exactly-once effects are
-   expressed by consuming owned values (drop = the fallback effect), not by
-   `Mutex<Option<_>>` take-once flags. [#360, #363]
+3. **Construction invariants live in ownership and types, not comments.** A
+   rule that two modules must both remember is a design defect. Exactly-once
+   construction effects are expressed by consuming owned values (drop = the
+   fallback effect), not by `Mutex<Option<_>>` take-once flags. Internal
+   synchronization may still use an optional payload as a non-panicking claim
+   protocol when it does not re-assert a construction capability. [#360,
+   #363]
 4. **Capabilities are never erased and re-asserted.** If the caller proved a
    capability statically (a dynamic scope, a restartable actor), every value
    derived from it carries that proof in its type. Runtime downgrades
@@ -565,8 +568,9 @@ established in the construction-path types, before erasure:
   what they can accept (owned, non-re-mintable args) and return
   (`OneShotTaskRef<T>`), not in how long the child runs.
 - **Restartable** (`add_*`-family): the caller supplies an args *source* —
-  `Args: Clone`, or `Fn() -> Args` re-minted at restart time (so a restart
-  can observe the current world: re-resolve an address, fresh timestamp).
+  `Args: Clone + Sync`, or `Fn() -> Args + Send + Sync` re-minted at restart
+  time (so a restart can observe the current world: re-resolve an address,
+  fresh timestamp).
   This is OTP's `{M, F, A}` translated into ownership: the args value is the
   per-spawn argument, and whether you can clone or re-mint it *is* your
   restart capability. For subtree children the args source *is* the
@@ -576,11 +580,13 @@ established in the construction-path types, before erasure:
 - The erased internal representation of one-shot construction is
   `FnOnce`-shaped all the way down: the closure owns the resource, the
   runner owns the closure, and init-panic / startup-failure /
-  shutdown-before-start all reduce to dropping the owner. `Mutex<Option<_>>`
-  take-once tricks MUST NOT appear, publicly or privately — and the same
-  owned-token shape MUST be reused for the other take-once sites the origin
-  accumulated (cancel guards, readiness signals, exit-report claims): build
-  the consuming primitive once (§1 principle 3, §7).
+  shutdown-before-start all reduce to dropping the owner. Within construction
+  payloads, `Mutex<Option<_>>` take-once tricks MUST NOT appear, publicly or
+  privately — and the same owned-token shape MUST be reused for construction
+  claims throughout the lowering and incarnation path. This prohibition does
+  not cover internal, non-panicking synchronization claims such as disposal
+  completion, where losing the claim race is an ordinary no-op rather than a
+  re-asserted construction capability (§1 principle 3, §7).
 - **Every user-supplied construction source executes inside the single
   incarnation runner** (§7): the restartable forms' `Fn() -> Args` factory
   and `Args::clone`, task body factories, raw-actor factories, and subtree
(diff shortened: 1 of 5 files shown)


ralexstokes:
Adversarial review: the `try_recv().expect` conversion is sound — the report obligation discharges on every return, panic, and abort edge before `runtime::join` resolves, so no path reaches the receiver without a report. Gate-adoption assertions are debug-only (no new release panic path), SPEC constructor signatures/B.10/take-once carve-out are internally consistent and match the code, and the wording commit weakens no normative MUST.

One fix (**`46e60fe`**): the report-ordering comment credited "Rust" for what is a tokio `JoinHandle` guarantee (the future is destroyed before the handle resolves); it now names the runtime invariant a replacement executor must preserve.

Coverage cross-check against #91: with the `ChildId` → `Arc<str>` fix on #96, every audit item is now addressed by the stack, so `Closes #91` is accurate.

Issue #101, opened 2026-08-09
Newtype and enum opportunities from the full-codebase review at PR #100


ralexstokes:
Results of a full-`src/` review (at PR #100's head, a967211) looking for places where a newtype — or a small enum replacing a positional bool — would turn a discipline-enforced invariant into a compiler-enforced one. Line numbers are at that commit. Companion to #97, which parked three cells.rs candidates behind the #91 audit; PR #100 closes #91, so those three unblock when it merges (re-verified below).

Findings are grouped by priority. Each item states the concrete mix-up that compiles today.

### A1. Named latch carriers in scope/task plumbing

- `run_nested_tree(..., ready: Latch, cancel: Latch, abort: Latch, abort_ack: Latch)` (driver.rs:2609) and the matching `SpawnBody::ScopeRestartable`/`ScopeOnce` fields.
- `run_scope_incarnation(..., parent_ready: Option<Latch>, incarnation_cancel: Option<Latch>, incarnation_abort: Option<Latch>, incarnation_abort_ack: Option<Latch>, ...)` (driver.rs:2686) — call sites read `(plan, false, Some(ready), Some(cancel), Some(abort), Some(abort_ack), epoch)`.
- `TaskContext::new(id, incarnation, shutdown: Latch, abort: Latch, ready: Latch)` (task.rs:36).
- `CancellationToken::from_latches(primary: Latch, secondary: Latch)` (cancellation.rs:20) — argument order decides which latch `child()` derives from.

Transposing `abort`/`abort_ack` compiles and makes ancestor abort un-acknowledgeable (caught only by the tidy-beat backstop — a silent latency bug); swapping `shutdown`/`abort` in `TaskContext::new` inverts graceful-stop vs hard-abort for the task body. Fix: named-field carrier structs (e.g. `NestedScopeLatches`, `AncestorCommandLatches` — the latter maps 1:1 onto the existing `ScopeRuntime.ancestor_*` fields), built once from `SpawnLatches`. Fold `is_root: bool` (driver.rs:2675, 2686; call site `run_scope(plan, true, None)` at driver.rs:730) into a `ScopeRole { Root, Nested }` enum in the same pass.

### A2. `DisposalOutcome` replacing `Option<Option<String>>`

[...]

PR #104, opened 2026-08-09
Make disposal and admission outcomes explicit

ralexstokes:
Part of #101 (A2 and A5).

## Summary

- replace disposal completion's ambiguous `Option<Option<String>>` with `Option<DisposalPanic>`, whose `message: Option<String>` field cannot be flattened into a clean result
- thread the typed panic through the driver event and terminal-verdict override path while accurately representing no construction or an intentionally unobserved detached disposal as no recorded panic
- pass `AdmissionOwnership` directly into `Admission::new` and derive fused cancellation state there instead of erasing the enum to a positional bool
- cover a non-string factory-destructor panic after a real incarnation exit, proving it remains a panic verdict with no message

## Why

The nested option encoded two independent facts—whether destruction panicked and whether its payload was a string—without naming either state. A routine `flatten` or wrapping mistake could silently turn a destructor panic into clean disposal. `Option<DisposalPanic>` preserves the same semantics without nesting and remains truthful on hard-force paths where destruction is detached and its result is intentionally not observed.

The admission path similarly discarded an existing enum into a bool that controlled drop semantics, then reconstructed the distinction later.

## Impact

There is no public API or intended behavior change. Destructor-panic verdict precedence and fused-versus-split admission ownership are preserved while invalid state mix-ups become compile errors.

## Validation

- `./scripts/dev just ci`
  - Clippy with warnings denied
  - 427/427 tests

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -781,7 +781,7 @@ enum ChildEvent {
     },
     ConstructionDisposed {
         child: ChildKey,
-        outcome: runtime::DisposalOutcome,
+        panic: Option<runtime::DisposalPanic>,
     },
 }
 
@@ -1929,7 +1929,7 @@ impl ScopeRuntime {
             // The incarnation has already exited; only its retained factory
             // remains. Hard escalation detaches that cleanup, but must not
             // rewrite the actor's recorded verdict.
-            self.handle_construction_disposed(key, runtime::DisposalOutcome::Clean);
+            self.handle_construction_disposed(key, None);
         }
     }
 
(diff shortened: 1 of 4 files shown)


ralexstokes (commented):
Reviewed the full diff.

**What it does:** Replaces the disposal completion's `Option<Option<String>>` with `Option<DisposalPanic>` (threaded through `DisposalJob`, `dispose_then`, the `ConstructionDisposed` driver event, and the terminal-verdict override), and passes `AdmissionOwnership` directly into `Admission::new` instead of collapsing it to a positional bool.

**Assessment:**
- Semantics are preserved exactly: the driver's `Some(DisposalPanic { message })` destructure binds the same `Option<String>` the old inner option carried, and `fused_cancel` derivation via the `match` on ownership is equivalent to the old `fused.then(Latch::default)`.
- The hard-force/detached-disposal case (`None` = no recorded panic) remains truthful.
- Removing `AdmissionOwnership::is_fused` in favor of matching at the single consumption site is the right call — no dead helper left behind.
- New test covers the previously untested non-string panic payload (`panic_any(17_u8)`) after a real incarnation exit, asserting `ExitKind::Panicked { message: None }`. Good edge-case coverage.

No findings. CI green. Merging.

PR #105, opened 2026-08-09
Name scope and task latch roles

ralexstokes:
Part of #101 (A1).

## Summary

- replace the positional task shutdown/abort/readiness arguments with `TaskContextLatches`
- build one `NestedScopeLatches` value from `SpawnLatches` and carry its parent-readiness plus ancestor shutdown/abort/ack roles through both restartable and one-shot subtree paths
- replace the independent root flag and optional nested-only latches with `ScopeRole::{Root, Nested}`
- remove `CancellationToken::from_latches(primary, secondary)` and construct derived cancellation sources by named fields

## Why

The old call chains repeatedly lined up interchangeable `Latch` values. Swapping abort with abort acknowledgement compiled but could leave recursive abort unacknowledged until the tidy-beat backstop; swapping task shutdown and abort inverted graceful versus hard-stop behavior. Root/nested state was likewise spread across a bool and four independently optional latches.

The new carriers are constructed once at the ownership boundary and preserve those roles through every downstream hop. `ScopeRole` also makes root-with-parent-latches and nested-with-missing-command-latches unrepresentable.

## Impact

There is no public API or intended runtime behavior change. This is compiler-enforced wiring for task and nested-scope control signals.

## Validation

- `./scripts/dev just ci`
  - Clippy with warnings denied
  - 426/426 tests

### DIFF
--- a/crates/shelterwood/src/cancellation.rs
+++ b/crates/shelterwood/src/cancellation.rs
@@ -17,16 +17,12 @@ impl CancellationToken {
         }
     }
 
-    pub(crate) fn from_latches(primary: Latch, secondary: Latch) -> Self {
-        Self {
-            primary,
-            secondary: Some(secondary),
-        }
-    }
-
     pub(crate) fn child(&self, cancellation: Latch) -> Self {
         debug_assert!(self.secondary.is_none());
-        Self::from_latches(self.primary.clone(), cancellation)
+        Self {
+            primary: self.primary.clone(),
+            secondary: Some(cancellation),
+        }
     }
 
     /// Reports whether cancellation has been requested.
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -32,7 +32,7 @@ use crate::{
     policy::{DefaultsInheritance, ResolvedDefaults, ScopeFlavor},
     raw::{CatchUnwindFuture, RawRunContext, RawSpawn},
     runtime::{self, Latch},
-    task::{TaskContext, TaskFactory},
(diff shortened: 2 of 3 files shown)


ralexstokes (commented):
Reviewed the full diff, tracing each latch role from the old positional arguments to the new named carriers.

**What it does:** Introduces `TaskContextLatches`, `NestedScopeLatches`/`AncestorCommandLatches`, and `ScopeRole::{Root, Nested}`; removes `CancellationToken::from_latches`, replacing the root bool + four independent `Option<Latch>` fields on `ScopeRuntime`.

**Correctness check (all verified against the pre-change call sites):**
- `SpawnLatches::nested_scope()` maps `ready → parent_ready`, `shutdown → ancestor.shutdown`, `framework_abort → ancestor.abort`, `framework_abort_ack → ancestor.abort_ack` — exactly the tuple the old `run_nested_tree` callers lined up positionally.
- The fire sites are preserved: `Pending::Shutdown` fires `ancestor.shutdown` (old `cancel`), `Pending::AncestorAbort` fires `ancestor.abort_ack` (old `ack`), startup completion fires `parent_ready`.
- The `pending.is_empty()` select-arm derivation via `role.ancestor().filter(...).map(...)` is equivalent to the old `then(...).flatten()` on the option fields.
- The two driver tests that previously passed `None` for some nested latches now pass fresh unfired `Latch::default()` values — behaviorally identical (unfired latch ≡ absent latch in every consumer: `is_fired()` false, select arm pends forever), and it's the price of making partial nested wiring unrepresentable, which is the point of the PR.

No findings. CI green (426/426). Merging.

PR #106, opened 2026-08-09
Bind identity counters to their domains

ralexstokes:
Part of #101 (A3).

## Summary

- add a private `Lineage` newtype and store `FenceCounter.current` as `Generation`
- reconstruct adopted counters through `FenceCounter::from_fence`, so lineage and current generation are never lined up as raw `u64`s
- give non-identity sequence counters an explicit `FenceCounter::sequence()` constructor
- replace the forgeable `(membership, FenceCounter)` incarnation API with `IncarnationCounter { membership, counter }` and a no-argument `mint()`
- update driver, mailbox, task, and identity tests to carry the bound counter rather than repeatedly supplying a membership

## Why

A lineage and generation were adjacent `u64` fields, including one production construction that unwrapped a `Generation`; swapping them compiled and silently broke fence comparability. Incarnation minting separately accepted any membership/counter pairing, while the driver captured the counter at child setup but re-read the member's potentially rebased membership on every mint.

The new types preserve both pairings at construction time. A counter can mint only the membership it was created for, even if the surrounding member later rebases.

## Impact

There is no public API or intended behavior change. Fence exhaustion, poison handling, and identity ordering are preserved while cross-domain swaps become compile errors.

## Validation

- `./scripts/dev just ci`
  - Clippy with warnings denied
  - 426/426 tests

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -23,7 +23,7 @@ use crate::{
         JoinVerdict, RecordedOutcome, StartupError, StopReason, classify_exit,
         reconcile_recorded_outcomes,
     },
-    identity::{FenceCounter, ScopeIdentity},
+    identity::IncarnationCounter,
     observe::LifecycleEventKind,
     plan::{
         BuilderCore, ChildConstruction, ChildPlan, LowerError, ScopeFactory, ScopePlan, SlotCell,
@@ -852,7 +852,7 @@ struct ChildRuntime {
     construction: runtime::Isolated<ChildConstruction>,
     pending_terminal: Option<PendingTerminal>,
     options: crate::policy::ResolvedCommonOptions,
-    incarnations: FenceCounter,
+    incarnations: IncarnationCounter,
     restarts: RestartState,
     restart_deadline: Option<DeadlineHandle>,
     active: Option<ActiveChild>,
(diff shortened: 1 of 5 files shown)


ralexstokes (commented):
Reviewed the full diff, plus the driver call sites for the one real semantic risk.

**What it does:** Private `Lineage` newtype, `FenceCounter.current` stored as `Generation`, `FenceCounter::from_fence`/`sequence()` constructors, and `IncarnationCounter { membership, counter }` with a no-argument `mint()` replacing the forgeable `(membership, &mut FenceCounter)` pairing.

**Correctness check:**
- The one behavioral subtlety: the old `mint_child_incarnation` re-read `slot.member.membership()` on every mint, while `IncarnationCounter` captures membership at creation. This is safe because the counter is created in `ChildRuntime::from_plan` — after lowering, which is where `rebase_membership` happens (`plan.rs`) — and a subtree restart re-lowers into fresh `ChildRuntime`s. The extended `rebased_declared_handles_and_incarnations_keep_identity` test pins exactly this: the post-rebase `Started` event's incarnation carries the rebased membership.
- `mint()` poison/exhaustion logic is unchanged (`checked_add` → `Generation::new` → `POISON` on failure), and `from_fence` matches the old inline adopted-counter construction.
- `incarnation_counter` now actually binds the membership it previously ignored (`_membership`), with the lineage-0 sequence counter reserved for the non-identity domain via `sequence()`.

No findings. Merge conflict with #105 (both edited the driver test-import block) was resolved as the union of both sides; full local CI mirror passed on the merge, and GitHub CI is green. Merging.

PR #107, opened 2026-08-09
Type mailbox and timer sequence domains

ralexstokes:
Part of #101 (A4).

## Summary

- add private `ArmingOrder` and `KeyHash` newtypes throughout `TimerStore`, its deadline index, and fired-timer batches
- add a crate-private ordered `AcceptedSequence` type at the mailbox boundary and carry it through accepted envelopes, receive watermarks, and raw-actor timer fairness snapshots
- store `RawResources.next_timer_order` as `ArmingOrder`
- store `WaiterQueue.next_id` as the existing `WaiterId`, including its zero/poison values and successor logic

## Why

Timer armings, erased key hashes, and mailbox accepted-sequence watermarks were all raw `u64`s in adjacent storage and control flow. In particular, the arming index was `HashMap<u64, u64>` and removal held both values in one scope, so reversing key and value compiled. The waiter queue similarly left its counter in the raw domain even though its map key already had a dedicated type.

The newtypes preserve each sequence domain end to end. Only the atomic accepted counter remains raw at the atomic boundary; its values are wrapped immediately when minted or loaded.

## Impact

There is no public API or intended behavior change. Timer ordering, hash collision handling, mailbox fairness watermarks, saturation, and waiter poison semantics are preserved.

## Validation

- `./scripts/dev just ci`
  - Clippy with warnings denied
  - 426/426 tests

### DIFF
--- a/crates/shelterwood/src/mailbox.rs
+++ b/crates/shelterwood/src/mailbox.rs
@@ -425,8 +425,11 @@ enum MailboxKind {
 
 struct Envelope<M> {
     message: M,
-    accepted_sequence: u64,
+    accepted_sequence: AcceptedSequence,
 }
+
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub(crate) struct AcceptedSequence(u64);
 
 enum OperationOutcome<M> {
     Waiting {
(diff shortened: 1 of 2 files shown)


ralexstokes (commented):
Reviewed the full diff.

**What it does:** Newtypes for the remaining raw-`u64` sequence domains: `ArmingOrder` and `KeyHash` through `TimerStore` (including the `armings: HashMap<ArmingOrder, KeyHash>` index whose key/value used to be swappable), `AcceptedSequence` at the mailbox boundary, and `WaiterQueue.next_id` promoted to the existing `WaiterId`.

**Correctness check:**
- `WaiterId::checked_next` + the poison assignment in `push_back` reproduce the old two-branch exhaustion logic exactly: overflow and poison-value both leave `next_id = POISON` and panic, which the existing `waiter_identity_exhaustion_poison_is_never_minted` test still pins (now via `WaiterId::POISON`).
- `take_due`'s `..=(now, ArmingOrder::MAX)` range bound is the old `(now, u64::MAX)`; derived `Ord` on the newtype keeps the `BTreeSet<(Instant, ArmingOrder)>` ordering identical.
- `saturating_next` mirrors the old `saturating_add(1)` on `next_timer_order` (pre-existing saturation behavior unchanged).
- `mint_accepted_sequence` and `accepted_sequence` wrap at the atomic boundary only, as described — no change to ordering semantics (`Release`/`Acquire` preserved).

No findings. Will merge once PR 106 is in (both touch `mailbox.rs`, so mergeability needs re-checking after that lands).

PR #108, opened 2026-08-09
Name cell state and lifecycle domains

### LINKED ISSUE #97: Newtype candidates in cells.rs deferred behind the #91 audit
Follow-up to #92, which introduced the `Epoch` newtype. The same review pass identified three more places in `cells.rs` where a dedicated type would carry an invariant currently enforced only by discipline. All three contend with open work claimed by the #91 audit punch list, so they are parked here rather than done now — each should be folded into (or scheduled after) the corresponding #91 item.

## 1. `ObservationGate` wrapper for the tree gate — after #91 item 9

The resident-tree gate is passed around as a bare `Arc<Mutex<()>>` with two behaviors re-implemented at each site:

- pointer-identity comparison via `Arc::ptr_eq` ("one resident tree shares one observation gate"),
- poison-*tolerant* locking (`unwrap_or_else(PoisonError::into_inner)` in `adopt_observation_gate` and `with_observation_gate`) — a deliberate semantic choice (an observer panicking mid-observation must not wedge gate handoff) that exists only as a repeated idiom, while every other mutex in the file uses `expect`.

A newtype with `same_gate(&other)` and a `lock()` encoding the poison policy would centralize both and give the adopt/handoff protocol a home for its docs (and for the adoption-directionality debug assertion #91 asks for).

**Contention:** #91 item 9 replaces the `Mutex<Arc<Mutex<()>>>` handoff with a swap-tolerant fast path (RwLock outer / arc-swap). The wrapper is a natural substrate for that change — do them together.

## 2. `MemberMailbox` struct → enum — with the #91 `Mutex<Option<_>>` decision

`MemberMailbox { control: Option<Arc<dyn MailboxControl>>, terminal: Option<Exit>, teardown: Option<Box<dyn MailboxTermination>> }` has coupled invariants enforced by scattered asserts: `control` set at most once; `teardown` only `Some` while `terminal` is `Some`; a terminal staged before attach has no teardown yet. An enum along the lines of `Unattached` / `Attached(control)` / `Terminal { control: Option<_>, exit, teardown: Option<_> }` makes the illegal combinations unrepresentable and turns "terminality wins the race permanently" into a plain state transition.

**Contention:** `teardown` is exactly the take-once `Mutex<Option<_>>` shape whose SPEC scoping #91's spec-conformance finding is deciding; the enum restructure should implement that decision, not preempt it.

## 3. Public `LifecycleSeq` newtype — after #91 items 3 and 5

Lifecycle sequence numbers are bare `u64` on public API (`LifecycleEvent.seq`, `ScopeSnapshot.lifecycle_seq`, `ChildSnapshot.scope_seq`, `ScopeSnapshot::watermark`) with `u64::MAX` as an implicit "sequence space exhausted / permanently lagged" sentinel, set in `ScopeCell::emit_locked` when `FenceCounter::mint_sequence` saturates. Consumers must know the sentinel by convention (SPEC B.4/B.6). A newtype (Ord, `get()`, documented `EXHAUSTED`) would document the sentinel where consumers see it. Public-API change, so also needs a SPEC appendix touch (`scope_seq: Option<u64>` in the B.6 sketch) and doc updates.

**Contention:** threads through `emit_locked`/`publish_snapshot_chain_locked` and the observation hubs, which #91 item 3 (remove per-event hub locks) and item 5 (`lifecycle_sequence` mutex) rewrite.

## Explicitly considered and rejected

- Scope tag on `Epoch` (reviewed on #92): epochs never travel without their `ScopeCell`, and guarded ordering would make the stale-driver `<=` comparisons silently false both ways. Revisit with the `Incarnation` shape (identity + `supersedes`) only if epochs ever escape their cell.
- Newtypes for `restart_count`/`total_restarts`, the `bool` returns on `terminalize_child`/`prune_child`/`take_*_request`, and `ScopeRecord.startup`'s `Option<Result<_>>` — names and privacy already carry the meaning.

ralexstokes:
Closes #97.
Part of #101 (section B).

## Scope

This PR implements the three domain carriers deferred by #97:

- replaces the coupled `MemberMailbox` option fields with explicit `Unattached`, `Attached`, and `Terminal` states;
- introduces `ObservationGate` as the single home for resident-tree gate identity and poison-tolerant locking;
- introduces public `LifecycleSeq` across lifecycle events and snapshot watermarks, including `get()` and the never-emitted `EXHAUSTED` sentinel.

It also updates the SPEC appendix and observation documentation for the typed lifecycle sequence surface.

## Behavioral intent

The changes preserve mailbox terminality ordering, subtree gate handoff behavior, lifecycle ordering, and exhaustion semantics while making their invariants explicit in the types.

A regression test deliberately poisons an observation gate and verifies later observation remains live.

## Validation

- `./scripts/dev just ci`
- 427/427 tests passed
- doctests, rustdoc warnings, public API reachability, enforcement checks, packaged-doc sync, and packaged-crate verification passed

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -3076,7 +3076,7 @@ LifecycleEvent { scope_path, scope, seq, kind }
               cell through the same §3.2 machinery; its token and
               sequence behave exactly like a nested scope's — one
               uniform identity for every emitting scope, root included
-  seq:        u64 from the emitting scope's single monotone sequence,
+  seq:        LifecycleSeq from the emitting scope's single monotone sequence,
               owned by the scope's membership cell — continuous across
               subtree restart (a rebuilt incarnation continues it; only
               a replacement membership starts fresh, and `scope`
@@ -3164,7 +3164,8 @@ Ordering and delivery contract:
   it. The same
   protocol is the post-`Lagged` resync. The documentation MUST teach
   it in this form.
-- `seq`/`lifecycle_seq` mint through §3.1's one primitive: `u64`,
+- `LifecycleSeq` exposes `get()` plus the documented `EXHAUSTED` sentinel;
+  `seq`/`lifecycle_seq` mint through §3.1's one primitive: `u64`,
   saturating advance, the saturated value poisoned and never minted.
   Exhaustion — unreachable at `u64` scale, pinned per §3.1's
   decide-once rule — fails closed for observation: the scope mints no
@@ -3276,7 +3277,7 @@ ChildSnapshot   { id, membership,                       // §3 identity types
                                                          //   ScopeSnapshot — Stopped { NeverStarted }
                                                          //   included (§3.2) — so traversal reaches
                                                          //   the terminal scope state
-                  scope_seq: Option<u64> }               // scope children only (None otherwise): the
+                  scope_seq: Option<LifecycleSeq> }      // scope children only (None otherwise): the
                                                          //   nested scope's lifecycle_seq watermark,
                                                          //   sampled at the same publication point —
                                                          //   equal to nested.lifecycle_seq whenever
@@ -3305,7 +3306,7 @@ ScopeSnapshot   { state: Unstarted                              // membership ex
                                                                 //   scope-state twin of §7's exit
                   kind: Ordered | Dynamic, strategy (ordered only), intensity,
                   total_restarts,                        // charges per §9.2 — group respawns count
-                  lifecycle_seq,                         // aligns events with snapshots (§12)
+                  lifecycle_seq: LifecycleSeq,           // aligns events with snapshots (§12)
                   children: Vec<ChildSnapshot> }         // declaration order (ordered scopes);
                                                          //   admission order (dynamic scopes) —
                                                          //   pre-admission reserved cells are
--- a/crates/shelterwood/doctests/docs/observation.md
+++ b/crates/shelterwood/doctests/docs/observation.md
@@ -25,6 +25,10 @@ temporarily absent. For an event emitted by the subscribed scope itself, compare
 `snapshot.watermark(event.scope)` finds the matching watermark. An event is
 already reflected when `event.seq <= watermark`.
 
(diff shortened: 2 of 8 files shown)


ralexstokes (commented):
Reviewed the full diff.

**What it does:** The three carriers deferred from #97: `MemberMailbox` becomes an explicit `Unattached | Attached | Terminal` state machine, `ObservationGate` extracts the poison-tolerant resident-tree gate, and public `LifecycleSeq` types the lifecycle sequence surface (events, snapshot watermarks, `watermark()`), with SPEC/docs updated.

**Correctness check:**
- The `MemberMailbox` transitions reproduce the old option-struct behavior exactly: `attach_mailbox` keeps the one-mailbox panic (now also covering `Terminal{control: Some}`), `terminalize` stays idempotent (existing `Terminal` returns the recorded exit unchanged) and preserves the attached control into `Terminal`, and teardown preparation still happens inside the same lock hold.
- The new `unreachable!` in the teardown-take path is sound: `publish_terminal` has exactly two callers (`attach_mailbox`'s terminal branch and `terminalize`), and both leave the state `Terminal` inside the same lock before publishing. This upgrades a silent no-op into an invariant check — good.
- `ObservationGate` is a faithful extraction: `same_gate` replaces `Arc::ptr_eq`, `lock()` keeps the `PoisonError::into_inner` tolerance, and the adopt/retry handoff loop logic is untouched. The new poison-injection regression test pins the tolerance.
- `LifecycleSeq` minting is unchanged under the wrapper (saturating advance, `u64::MAX` poisoned and stored as the permanent watermark, never emitted on an event — now nameable as `EXHAUSTED`). The public API change to `LifecycleEvent.seq`/`ScopeSnapshot.lifecycle_seq`/`watermark()` is the sanctioned #97 surface, and SPEC + both observation docs are updated consistently.

No findings. Branch updated with a merge of main (PRs 104–107) and CI re-ran green on the combined tree. Merging.

PR #109, opened 2026-08-09
Type exit cancellation and grace phases

ralexstokes:
Part of #101 (section C: exit cancellation and grace-phase domains).

## Scope

This PR removes two boolean domains from the exit and stop-ladder pipeline:

- `Exit::new` and the exit classifier now take `Cancellation::{Observed, NotObserved}`;
- the report token, child-exit event, and driver handler carry `Cancellation` from the latch observation through final classification, without a positional-bool conversion boundary;
- `Exit::cancellation()` returns that typed value; the old boolean accessor is removed;
- `ExitKind::Aborted`, recorded/join verdicts, stop actions, and the driver's hard-abort provenance use `GracePhase::{WithinGrace, AfterGrace}`;
- public exports, trait-contract tests, SPEC text, and packaged shutdown documentation are updated.

## Why

Previously, readers had to decode calls such as `Exit::new(ExitKind::Aborted { after_grace: false }, true)`. Both boolean positions were semantically independent and transposable. The new API makes cancellation observation and grace provenance distinct compiler-checked domains throughout the reporting and stop-ladder paths.

## Validation

- `./scripts/dev just ci`
- 426/426 tests passed
- doctests, rustdoc warnings, public API reachability, enforcement checks, packaged-doc sync, and packaged-crate verification passed

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1229,23 +1229,25 @@ One classification, produced at one point, used by every consumer.
   never input to restart or intensity accounting (both consume
   incarnation exits), and counts as a failure for `is_failure()` —
   awaiting a child that never ran is not success.
-- **`cancelled` is one state-machine fact**, not a narrative judgment:
-  the flag is true iff the incarnation's own shutdown token (B.1/B.2)
-  had fired before its outcome was recorded (the record phase below).
+- **Cancellation observation is one state-machine fact**, not a narrative
+  judgment: the value is `Cancellation::Observed` iff the incarnation's own
+  shutdown token (B.1/B.2) had fired before its outcome was recorded (the
+  record phase below), and `Cancellation::NotObserved` otherwise.
   The token fires for every engine-initiated stop — scope teardown,
   dynamic removal, readiness-timeout teardown, `shutdown(0)`'s immediate
   escalation — and for local `ctx.stop()`, which arms the same ladder
   (§10), so a self-stopped actor's exit reads `Cancellation::Observed`. A stop
   racing natural completion needs no third rule: whichever of token-fire
-  and outcome-record happened first decides the flag. `NeverStarted`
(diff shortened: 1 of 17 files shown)


ralexstokes (commented):
Reviewed the full diff.

**What it does:** Replaces the two transposable exit booleans with `Cancellation::{Observed, NotObserved}` and `GracePhase::{WithinGrace, AfterGrace}` end to end: `Exit::new`/`cancellation()`, the report token, `ChildEvent::Exited`, `classify_exit`, `RecordedOutcome`/`JoinVerdict`, `StopLadder`/`StopAction`, the driver's hard-abort provenance, public exports, SPEC, and the shutdown docs.

**Correctness check (every bool→enum site traced):**
- `cancelled: true → Observed` / `false → NotObserved` is consistent everywhere: the report-token derivation (`shutdown.is_fired() || local_stop`), driver `JoinOutcome::Cancelled` kill verdicts (`Observed`), driver-panic exits (`NotObserved`), the root `Completed` exit deriving from `reason == ShutdownRequested`, and `Exit::never_started()` (`NotObserved`).
- `after_grace: true → AfterGrace` / `false → WithinGrace` likewise: `StopLadder` starts `WithinGrace` and flips to `AfterGrace` at exactly the two sites the old code set `after_grace = true`; `classify_exit`'s fail-closed `(None, None)` default stays `WithinGrace`; the hard-abort override in `handle_exit` carries the phase unchanged.
- Verdict-precedence and reconciliation test tables are translated one-for-one, and the conformance test now pins `Copy + Eq + Hash + Send + Sync` for both new enums.
- SPEC §7/§10/appendix and both copies of `shutdown-and-resources.md` stay in sync with the new field names.

No findings. Branch updated with a merge of main (through #108) — clean merge, no conflicts — and CI is green on the combined tree. Merging.

PR #110, opened 2026-08-09
Replace supervision mode booleans with domain types

ralexstokes:
Implements the supervision-mode portion of #101.

- replaces `startup_aborted` parameters with `StartupDisposition`
- passes `ScopeFlavor` and a named `ChildCompletionState` into scope finish policy
- passes `ChildMode` into common policy resolution for restartable vs one-shot children
- updates tests to construct each domain explicitly

Validation:
- `./scripts/dev just ci` (426/426 tests, doctests, docs, API/enforcement checks, package verification, and Nix formatting)

### DIFF
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -80,6 +80,13 @@ pub(crate) enum MemberStage {
     Terminal(Exit),
 }
 
+/// Whether a terminal child incarnation failed during aggregate startup.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum StartupDisposition {
+    NotAborted,
+    Aborted,
+}
+
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub(crate) struct MemberRecord {
     pub(crate) stage: MemberStage,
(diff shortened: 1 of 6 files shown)


ralexstokes (commented):
Reviewed the full diff.

**What it does:** Three positional-bool domains become types: `startup_aborted: bool` → `StartupDisposition::{NotAborted, Aborted}` through `PendingTerminal`, `terminalize_child`, and terminal disposal; `finish_if_ready(ordered, has_children, all_terminal)` → `(ScopeFlavor, ChildCompletionState)`; and `one_shot: bool` → `ChildMode::{Restartable, OneShot}` through policy resolution.

**Correctness check:**
- The `pre_ready` derivations (`child.initial && !startup_complete && !initial_ready`) map to `Aborted` exactly where the old code computed `true`, at both the exhaustion path and the exit-dispatch path; `handle_construction_disposed` keeps the incarnation-gating (`exited_incarnation.is_some()` else `NotAborted`) and the drain check still reads the ungated `terminal.startup`, matching the old `terminal.startup_aborted` read.
- `finish_if_ready` now takes the flavor directly instead of a pre-computed `flavor == Ordered` bool; the engine test's old `finish_if_ready(false, ...)` case correctly becomes `ScopeFlavor::Dynamic`.
- `ChildMode` mapping is faithful: `Task → Restartable`, `TaskOnce → OneShot`, raw/scope constructions via `mode()` derived from `source.is_one_shot()`; both `resolve_common` consumption sites (`restart` forcing and `retention` default) compare against `OneShot`.

No findings. The merge with main hit four import-block conflicts (vs #105/#106/#107/#109's edits) — all resolved as straight unions, dropping `mint_child_incarnation` which #106 removed — and CI is green on the combined tree. Merging.

PR #111, opened 2026-08-09
Replace runtime control booleans with domain enums

ralexstokes:
Implements the remaining private positional-bool cluster from #101.

- names live-only versus frozen-prefix mailbox receives with `ReceiveMode`
- carries `DeadlinePhase` through the shared mailbox deadline scaffold and its operations
- distinguishes scoped and unscoped offloads with `OffloadScope`
- distinguishes pending and finished offload polls with `OffloadPoll`
- replaces handler-context `draining` construction booleans with `DeliveryStage`

Validation:
- `./scripts/dev just ci` (426/426 tests, doctests, docs, API/enforcement checks, package verification, and Nix formatting)

PR #112, opened 2026-08-09
Type the three restart counter domains

ralexstokes:
Implements issue #101 section D.

- adds public opaque `RestartAttempt`, `RestartCount`, and `TotalRestarts` domains
- limits each counter to `ZERO`, saturating `bump()`, and `get()` with no arithmetic traits
- carries the domains through backoff derivation, restart state/decisions, intensity charges, observation records, lifecycle events, and public snapshots
- replaces raw `IntensityTrip` field construction in the driver with one derived constructor from the intensity policy and authoritative charge
- updates API/integration tests and the normative specification

Validation:
- `./scripts/dev just ci` (427/427 tests, doctests, docs, API/enforcement checks, package verification, and Nix formatting)

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -142,7 +142,7 @@ any single design rule:
   invariant makes bit-equality correct: §9.2's backoff factor),
   serializable where Part II §21 needs it, and carrying **no
   behavior** beyond small pure derivation functions of the
-  `should_restart(is_failure)` / `next_delay(attempt, jitter_sample)` /
+  `should_restart(is_failure)` / `next_delay(restart_attempt, jitter_sample)` /
   `validate()` shape. Runtime behavior is *derived from* the data by local
   functions; it is never encoded as trait objects, callbacks, or builder
   side effects.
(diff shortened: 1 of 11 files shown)


ralexstokes (commented):
Reviewed the full diff.

**What it does:** Public opaque `RestartAttempt`, `RestartCount`, and `TotalRestarts` — each exposing only `ZERO`, saturating `bump()`, and `get()`, with no arithmetic traits — carried through backoff derivation (`Backoff::next_delay`), `RestartState`/`RestartDecision`, intensity charges, member/scope records, lifecycle events, and public snapshots; plus `IntensityTrip::new(policy, charge)` replacing raw field construction in the driver; SPEC updated.

**Correctness check:**
- All three counters keep the old raw-u64 semantics exactly: `bump()` is the old `saturating_add(1)`, `settled()` resets only `attempt`, and `next_delay`'s `attempt.get().max(1)` preserves the one-origin clamp. The new saturation unit test pins `u64::MAX.bump() == u64::MAX` for all three.
- `IntensityTrip::new`'s `debug_assert_eq!(charge.tripped, charge.in_window > policy.max_restarts)` matches the trip condition in `IntensityState::charge`, so the constructor can't silently disagree with the charge that produced it.
- The conformance test pins `Copy + Eq + Hash + Send + Sync` plus `ZERO.bump().get() == 1` for the public surface, and integration tests were translated one-for-one (`restart_count == 1` → `RestartCount::ZERO.bump()` etc.).

No findings. The merge with main hit seven conflicts (import-list unions and side-by-side additions vs #108–#111) — all resolved as unions with no logic changes; one follow-up commit restored rustfmt normalization after the hand-resolved imports. CI is green on the combined tree. Merging.

PR #113, opened 2026-08-09
Harden restart derivation inputs

ralexstokes:
Implements the restart-related items from issue #101 section E.

- carries stable-run endpoints in named `IncarnationRun { started_at, stopped_at }`
- adds public `JitterSample`, whose constructors own the `[0, 1)` clamp and integer-ratio normalization
- makes `RestartPolicy` evaluate `&Exit` directly instead of accepting a failure boolean
- derives every collected driver's arbitration class from `Pending::class()`
- updates invariant tests and the normative specification

Validation:
- `./scripts/dev just ci` (427/427 tests, doctests, docs, API/enforcement checks, package verification, and Nix formatting)

PR #114, opened 2026-08-09
Name runtime boundary carriers

ralexstokes:
Implements the remaining runtime-boundary items from issue #101.

- carries primary and cleanup unwind payloads in named `UnwindPanics`
- carries same-domain scope wake futures in named `ScopeWait { signal, parent_shutdown }`
- renames both public `wait_for_child` duration parameters from `deadline` to `timeout`
- adds regression coverage for the biased scope-control priority
- updates the normative specification without changing runtime behavior

Validation:
- `./scripts/dev just ci` (427/427 tests, doctests, docs, API/enforcement checks, package verification, and Nix formatting)

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -3521,7 +3521,7 @@ fn wait_for_child(
     &self,
     id: impl Into<ChildId>,
     pred: impl FnMut(&ChildSnapshot) -> bool + Send,
-    deadline: Duration,                       // trailing, per Appendix B
+    timeout: Duration,                        // trailing, per Appendix B
 ) -> impl Future<Output = Result<ChildSnapshot, WaitError>> + Send;
 // WaitError { TimedOut, ScopeTerminated { state } } — non-exhaustive;
 // ScopeTerminated carries the scope's terminal B.6 state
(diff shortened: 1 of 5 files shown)


ralexstokes (commented):
Reviewed the full diff.

**What it does:** The final runtime-boundary carriers from #101: `UnwindPanics { primary, cleanup }` through both `resume_preferred_panic` variants and `PanicAccumulator::drop`, `ScopeWait { signal, parent_shutdown }` into `wait_scope`, and the public `wait_for_child` duration parameter renamed `deadline` → `timeout` (a span, not an instant — the new name matches the semantics); SPEC updated.

**Correctness check:**
- Every `UnwindPanics` construction maps the old positional arguments correctly: first → `primary`, second → `cleanup`, including the easy-to-miss `PanicAccumulator::drop` case (`primary: None, cleanup: first`). The precedence logic itself is untouched — the functions destructure and proceed identically.
- `ScopeWait` maps `signal.changed() → signal` and the ancestor-command future → `parent_shutdown`, and `wait_scope` pins in the same order, so the biased select priority is unchanged — now pinned by the new `scope_wait_prefers_signal_when_both_control_futures_are_ready` regression test, which is exactly the kind of coverage this refactor family needed.
- The `wait_for_child` rename is parameter-name-only on both `ScopeRef` and `DynamicScopeRef` (no call-convention change), with the SPEC signature updated to match.

No findings. The merge with main (through #113) was conflict-free; `cargo check` and fmt verified locally, CI green on the combined tree. Merging.

Issue #116, opened 2026-08-09
Review round: confirmed bugs, spec tightenings, and simplification backlog


ralexstokes:
Consolidated findings from a full-codebase review (eight parallel file/architecture/test reviews plus adversarial verification of an external review's claims). Every bug listed as **confirmed** was verified by an independent trace through the code; refuted claims are recorded at the bottom so they don't get re-reported.

### 1.1 Manual readiness is lost when a child exits immediately after `mark_ready()` — spurious root `StartupFailed` (HIGH)
A task running `mark_ready(); Ok(())` without yielding races its `Ready` and `Exited` events into the same driver batch; arbitration sorts `ChildExit` before `ReadinessSignal` (engine.rs:18-28), `handle_exit` takes the active incarnation and classifies the child as pre-ready startup abort without consulting the retained readiness latch (driver.rs:2065-2131), and the trailing `Ready` is dropped by the staleness guard (driver.rs:2037-2043). The readiness-*deadline* path already consults the latch precisely to avoid this (`signal_seen`, driver.rs:2364-2367, engine.rs:427-435, per SPEC "ready-at-deadline beats timeout") — the exit path lacks the analogous consult, which is demonstrably unintentional asymmetry.
**Fix:** in `handle_exit`, if `active.ready_signal.is_fired()` while the gate is `Waiting`, latch `initial_ready` before dispatch (e.g. `ReadinessEvent::Exit { signal_seen }` symmetric to `Deadline`).
**Decided semantics for `mark_ready(); Err(..)`:** ready-then-failed is a *post-ready* failure — restart policy applies and scope startup keeps waiting, exactly as if the events had arrived in the natural order (the same linearization argument that justifies the fix). This is the expected outcome of that regression test and goes into the §4.1 spec sentence.
**Tests:** `mark_ready(); Ok(())`, `mark_ready(); Err(..)`, `mark_ready(); stop()` with no intervening awaits, as initial children during startup.

### 1.2 Restart-deadline respawn of an `Immediate`-readiness child during startup hangs `wait_started()` (HIGH)
[...]

PR #118, opened 2026-08-09
Preserve mailbox capacity across kind defaults

ralexstokes:
Part of #116.

## Summary

- retain the nearest queue capacity while scope defaults switch through other mailbox kinds
- resolve `queue_inherit()` through an intervening `latest()` to the outer queue capacity
- preserve `DefaultsInheritance::Reset` as a hard boundary back to the library queue capacity
- add unit and three-level end-to-end coverage plus the normative §9.3 worked example

## Why

`ResolvedDefaults` previously retained only the currently selected mailbox value. Choosing `latest()` therefore erased an outer queue capacity even though the specification requires kind-mismatched defaults to be passed over during the outward capacity walk.

## Validation

- focused policy, defaults, and subtree test suites
- `./tools/dev just ci` (432 tests, doctests, rustdoc/API reachability, runtime/exit/layering enforcement, packaged-crate verification, and nixfmt)

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1831,7 +1831,13 @@ Resolution and mechanics:
   a child declaring `queue` with deferred capacity under a scope
   default of `latest()` gets `queue` at the library default capacity —
   the scope default neither converts the declared kind nor supplies a
-  capacity across kinds.
+  capacity across kinds. For the full outward walk, suppose the root
+  defaults to `queue(10)`, an inheriting child scope defaults to
+  `latest()`, and an inheriting grandchild scope contains an actor declared
+  with `queue_inherit()`: that actor resolves to `queue(10)`, because the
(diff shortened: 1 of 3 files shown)


ralexstokes:

The review traced the new `queue_capacity` invariant through scope overlay, direct actor/raw resolution, static lowering, restartable and one-shot subtree inheritance, reset boundaries, and dynamic admission. Counterexamples with repeated kind switches resolve as specified: the nearest explicit queue capacity survives intervening `latest()` defaults, a nearer explicit queue replaces it, and any `Reset` restarts the walk at the library capacity. Omitted mailbox declarations still inherit the active kind unchanged.

The §9.3 worked example matches the implementation, focused policy/integration tests pass, `git diff --check` is clean, and hosted CI is green. No follow-up commit was needed. Reviewed head: `cccefee`.

PR #119, opened 2026-08-09
Fix timer-batch offload fairness

ralexstokes:
Part of #116.

## Summary

- clear stale steady-state offload credit when a fired timer batch takes over scheduling
- pin the ordering so completions created during that batch cannot jump mailbox input accepted after the batch snapshot
- define timer “fire” in the specification as the loop’s batch-take linearization point

## Why

`offloads_lead` captured a completion watermark at a prior mailbox delivery. A later timer batch superseded that snapshot but retained the credit, allowing completions pushed during the batch to run ahead of post-snapshot mailbox input and violate the documented fairness boundary.

## Validation

- deterministic regression that reverses order without the fix
- `./tools/dev just ci` (432 tests, doctests, rustdoc/API reachability, runtime/exit/layering enforcement, packaged-crate verification, and nixfmt)

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -835,6 +835,11 @@ batch in arming order. Within class 3, ordering between mailbox and
 offload deliveries stays deliberately unspecified (§5.1's ordering
 contract promises per-sender FIFO and nothing more).
 
+For this ordering rule, a timer **fires** when the event loop observes due
+armings and takes their timer batch. "Queued at the fire instant" therefore
+means present at that batch-take linearization point, not merely present when
+the wall clock passed the timer's deadline.
+
 Already-queued messages get one bounded turn to retract an elapsed timer:
 when a timer fires, the messages queued *at that instant* are delivered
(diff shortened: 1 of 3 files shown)


ralexstokes:
Adversarial review found and fixed one specification overclaim.

- **Fixed in `46dcb04`**: the initial wording described timer fire as one global cross-source “batch-take linearization point.” In reality, batch formation snapshots continuations, mailbox acceptance, and offload completion through independent source-specific cutoffs. The spec now defines fire operationally as beginning batch formation, describes the bounded prefix recorded for each source, explicitly declines a global cross-source linearization claim, and phrases timer precedence in terms of those prefixes.

The code change itself was confirmed sound: clearing stale `offloads_lead` is covered by the timer batch’s own offload watermark, including cancellation holes, and does not starve completions across cancellation, re-arming, or intervals. The deterministic regression precisely establishes the stale-credit ordering failure. Full `./tools/dev just ci` passes at the reviewed follow-up head.

PR #120, opened 2026-08-09
Fix framework message disposal venues

ralexstokes:
Part of #116.

## Summary

- let handler `MailboxShutdown::Discard` return without draining so the framework disposes the frozen prefix detached and panic-contained
- route messages produced after a call-constructor budget expires through detached disposal
- document the deliberate live-`latest()` exception: displaced payloads drop inline on the displacing task after replacement acceptance is visible
- specify disposal venues and pin hostile-destructor regressions

## Why

Dropping a discarded message inline on the actor task allowed a hostile payload destructor to reclassify an otherwise clean stop and skip `on_stop`. The overdue call-constructor path similarly dropped user payloads on the caller task even though it is framework timeout cleanup.

## Validation

- focused drain and disposal test suites
- `./tools/dev just ci` (433 tests, doctests, rustdoc/API reachability, runtime/exit/layering enforcement, packaged-crate verification, and nixfmt)

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -800,10 +800,13 @@ the survivor is the newest accepted value.
 payload inline on the displacing task, after acceptance of its replacement
 is visible. This is the deliberate hot-path exception: a panicking foreign
 payload destructor surfaces on that task even though the replacement remains
-accepted. Every framework-initiated discard — including teardown,
-timeout/withdrawal cleanup, and batch disposal — runs detached from the
-initiating task with per-element panic containment. No single disposal-thread
-identity is promised.
+accepted. Framework-initiated disposal of externally submitted mailbox or
+reply-bearing payloads — including mailbox teardown, timeout/withdrawal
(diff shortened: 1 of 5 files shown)


ralexstokes:
Fresh adversarial review found three documentation/test overclaims; all are addressed in `e48004c`:

- Narrowed the detached-disposal guarantee to externally submitted mailbox and reply-bearing payloads. Incarnation-owned continuations, timers, and offload state remain governed by the incarnation teardown/verdict rules.
- Clarified `Discard` ordering: freezing makes the accepted prefix undeliverable before `on_stop`, while detached physical destruction is scheduled after the actor run and is not ordered before `on_stop` or exit publication.
- Removed the obsolete assertion that all fallback disposal uses one thread, while retaining off-submitter and post-panic-progress coverage.

Validation: `./tools/dev just ci` passed in full (including 433 tests), as did the focused hostile-disposal and drain/discard regressions.

PR #121, opened 2026-08-09
Fix terminal wakeup after mailbox panic

ralexstokes:
## Summary

- preserve the terminal membership wakeup when mailbox teardown invokes a panicking waker
- pulse terminal waiters before resuming the teardown panic while retaining panic precedence
- specify terminal publication as store, mailbox discharge, then pulse
- add focused unit and end-to-end regressions for panic surfacing, waiter wakeup, and driver completion

Part of #116 (items 1.3 and 4.7).

## Validation

- `./tools/dev just ci`
- focused mailbox-waker regressions
- library tests: 153 passed

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -385,6 +385,14 @@ Consequences (normative):
   `ScopeState` event (B.6, B.4) — before its streams close, so
   `wait_stopped()` (B.9) and snapshot/lifecycle subscribers resolve
   structurally, never merely by stream closure.
+- Terminal publication has one precise internal order: store the terminal
+  cell record, synchronously discharge parked mailbox operations, then pulse
+  the cell's single change signal. A direct or reentrant borrow MAY therefore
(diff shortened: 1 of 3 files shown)


ralexstokes:
Fresh adversarial review found three issues; all are addressed in `aef05e0`:

- Mailbox teardown panics are now carried through the parent `Exited`/snapshot/nested-close transaction and resumed only after the child terminality obligation is completed. The end-to-end regression asserts `Exited → Removed → lifecycle closure`.
- The spec now distinguishes an attached mailbox’s `store → discharge → pulse` order from terminality that wins before late mailbox attachment.
- A deterministic dual-panic regression proves mailbox teardown remains the primary panic while a membership-pulse panic is contained.

Validation: focused mailbox regressions and the full `./tools/dev just ci` pass (434 tests plus docs, API/layering, packaging, and formatting checks).

PR #122, opened 2026-08-09
Fix fused admission restart cancellation race

ralexstokes:
## Summary

- synchronously publish fused-admission cancellation as membership removal state before bounded driver forwarding
- recheck level-triggered restart suppression when an already-due restart deadline executes
- retain a weak owning-scope link for coherent snapshot publication, with a teardown-safe member fallback
- add a deterministic saturated-driver-lane regression using a real fused `Admission` drop

Part of #116 (item 1.5).

## Validation

- `./tools/dev just ci`
- deterministic saturated-lane fused-cancellation regression

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1601,7 +1601,13 @@ Rules (normative):
   **level-triggered cancellation latch** owned by the fused future and
   registered on the cell: dropping the future flips the latch
   synchronously (a token edge that always succeeds, like §10's shutdown
-  latch), and the driver converts the edge into an engine event that
+  latch). For an admitted cell that same synchronous edge publishes its
+  membership as `Removing` before attempting the bounded command channel;
+  both exit-time restart scheduling and execution of an already-due restart
+  deadline MUST recheck that level-triggered suppression. Queue saturation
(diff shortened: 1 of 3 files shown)


ralexstokes:
Fresh adversarial review found three issues; all are addressed in `087558b`:

- Exit dispatch now consults the fused cancellation latch before any restart schedule, counter, or intensity charge; the saturated-lane regression drops the real admission before handling exit and asserts zero restart accounting.
- Fused drop no longer performs public observation publication. The infallible latch/queued edge is authoritative, and reservation cancellation is completed through panic-contained cleanup. A reentrant hostile membership waker regression drops a second unpolled admission and proves both ids are released despite the surfaced panic.
- Dynamic restart suppression now uses the existing `ChildId` map key plus membership equality, avoiding an O(n) entry scan and stale-id confusion.

The due-deadline suppression remains as defense in depth; the spec now explicitly distinguishes the latch synchronization point from the later public `Removing` projection.

Validation: focused cancellation/reentrancy regressions and full `./tools/dev just ci` pass (433 tests plus Clippy, docs, API/layering, packaging, and formatting).

PR #123, opened 2026-08-09
Clarify supervision and lifecycle contracts

ralexstokes:
## Summary

- describe `Handler` as the public decorator composition point that encapsulates callback-loop teardown discipline
- document the once-per-incarnation-root `RawActor::run` contract and retained offload-panic behavior
- add the actor self-post/scope-shutdown deadlock warnings required by the capability contract
- define the practical large-attempt backoff domain without promising impossible full-`u64` exponentiation
- correct `Removed` to describe the later residency-pruning edge
- document the local-stop handshake and completed-output destruction behavior

Part of #116 (§4.6, §4.9, and the independent §5 accuracy fixes). Deadline and disposal changes are intentionally handled in separate code/spec PRs.

## Validation

- `./tools/dev just ci` (431 tests plus docs, API/layering, packaging, and formatting checks)

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -812,13 +812,6 @@ decided semantics are: `call` is allowed, conflation-away surfaces as
 `ReplyDropped`, and the documentation teaches this next to the §3.3 retry
 discipline.
 
-Where this specification calls for **isolated** or **detached** disposal, the
-guarantee is that user destruction runs off the submitting caller's thread and
-that each value has its own panic-containment boundary. The worker's identity,
-the number of disposal workers, and whether distinct disposal jobs share a
-thread are implementation details, not observable ordering or affinity
-contracts.
-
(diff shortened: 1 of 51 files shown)


ralexstokes:
Fresh adversarial review found three documentation overclaims; all are addressed in `1fa6aff`:

- `RawActor::run` is now specified as **at most once** per incarnation root, since shutdown may destroy a constructed root before `run` begins.
- Handler summaries now consistently say it encapsulates the callback error-path freeze-and-join discipline; clean/panic epilogues remain owned by the outer raw runner.
- Receive docs distinguish retained offload-work panics from continuation-materialization panics, which surface directly on that receive call.

Validation: full `./tools/dev just ci` passed (431 tests plus docs, API/layering, packaging, and formatting checks).

PR #126, opened 2026-08-09
Discover layering-check modules automatically

ralexstokes:
## Summary

- derive the below-driver module set from every top-level Rust source file instead of a hand-maintained list
- explicitly exempt only the driver, tree, and crate-root orchestration layers
- add an enforcement fixture proving a newly introduced module cannot silently reference upward

Part of #116 (§5 layering-check accuracy).

## Validation

- `./tools/dev just ci` on the audited source branch (436 tests plus Clippy, docs, API/layering, packaging, and formatting)
- enforcement fixture suite, including the new-module counterexample

### DIFF
--- a/tools/check-enforcement-fixtures.sh
+++ b/tools/check-enforcement-fixtures.sh
@@ -81,6 +81,8 @@ expect_case exit-downcast-mut check-exit-paths.sh \
   "runtime type recovery found on the exit path:"
 expect_case layering-driver-below check-layering-paths.sh \
   "upward driver or tree references found below the driver layer:"
+expect_case layering-new-module-driver-below check-layering-paths.sh \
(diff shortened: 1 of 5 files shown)


ralexstokes:
Fresh adversarial review found one high-priority blind spot: the first discovery pass only covered flat `src/*.rs` modules, so conventional `src/module/mod.rs` and future driver/tree submodules could escape enforcement.

Addressed in `c00d05d`:

- recursively discover every Rust source below `src/`
- classify driver and tree files plus their submodules into their specialized layers
- treat every other recursive source as below-driver
- add a directory-module enforcement fixture proving the counterexample fails

Validation: Bash syntax, the complete enforcement fixture suite, and full `./tools/dev just ci` pass (431 tests plus Clippy, docs, API/layering, packaging, and formatting).

PR #129, opened 2026-08-09
Defer observation wakes beyond the tree gate

ralexstokes:
## Summary

- mutate watch, snapshot, lifecycle, and terminal-mailbox state silently while holding the resident-tree observation gate
- flush every wake-producing action only after releasing that gate
- contain hostile waker panics per notification so later committed notifications still run, then resume the first panic
- add bounded custom-waker regressions for snapshots, lifecycle events, and `wait_stopped`, each synchronously reentering `ScopeRef::snapshot()`

Part of #116 (§2 observation-gate reentrancy robustness).

This PR is intentionally stacked on #121 because terminal-mailbox discharge is one of the synchronous wake sources it generalizes. The combined branch preserves teardown-before-member-pulse ordering and #121’s panic precedence/terminal observation guarantees.

## Validation

- focused snapshot/lifecycle/terminal/mailbox reentry regressions
- `./tools/dev just ci` on the resolved stack (437 tests plus Clippy, docs, API/layering, packaging, and formatting)

### DIFF
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -230,8 +230,9 @@ impl MemberCell {
         self.record.send_modify(update);
     }
 
-    fn update_locked(&self, update: impl FnOnce(&mut MemberRecord)) {
-        self.record.send_modify(update);
+    fn update_locked(&self, wakes: &mut ObservationWakes, update: impl FnOnce(&mut MemberRecord)) {
+        self.record.modify_silently(update);
+        wakes.pulse(&self.record);
     }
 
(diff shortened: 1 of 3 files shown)


ralexstokes:
Addressed the adversarial terminal-ordering finding in ef217b7. Supervised terminalization now folds startup_aborted into the terminal record mutation and queues mailbox teardown before any terminal record pulse, preserving #121's discharge-before-pulse and mailbox-panic precedence while keeping every wake outside the observation gate. Added supervised regressions for parked-sender ordering and dual hostile-waker precedence. Full ./tools/dev just ci passes (439 tests plus all other checks).

PR #133, opened 2026-08-09
Unify raw event arbitration batches

ralexstokes:
Part of #116; stacked on #119 because it replaces the fairness state touched by that fix.\n\nThis removes the parallel steady-state and fired-timer selectors. One ReadyBatch now owns both: steady state has a one-mailbox budget, a due timer promotes and refreshes the same batch cutoffs, and later due timers cannot overwrite already-fired armings. The duplicate offloads_lead state and try_recv_live path disappear.\n\nDeterministic traces cover steady priority, fired promotion, mid-batch timer arming, and #119's stale-credit regression on the combined branch.\n\nValidation: ./tools/dev just ci (435 tests plus all lint, docs, API/layering, packaging, and Nix checks).\n\nDraft until fresh adversarial review completes.

### DIFF
--- a/crates/shelterwood/src/mailbox.rs
+++ b/crates/shelterwood/src/mailbox.rs
@@ -1854,11 +1854,6 @@ impl<M: Send + 'static> MailboxReceiver<M> {
             .receive(self.incarnation, ReceiveMode::IncludeFrozen, None)
     }
 
-    pub(crate) fn try_recv_live(&self) -> Option<M> {
-        self.mailbox
-            .receive(self.incarnation, ReceiveMode::LiveOnly, None)
-    }
-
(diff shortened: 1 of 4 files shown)


ralexstokes:
Addressed all adversarial findings in 4c04634. Steady-state continuations remain live across handler calls, so a continuation queued by a mailbox/offload handler retains next-message priority. Before a second continuation runs, external work beyond the retained steady cutoff starts a fresh bounded turn and receives its mandatory fairness opportunity. Fired batches keep immutable source cutoffs. Added both deterministic traces and aligned offload docs with bounded arbitration. Full ./tools/dev just ci passes (437 tests plus all other checks).

PR #134, opened 2026-08-09
Remove dead internal bookkeeping

ralexstokes:
Part of #116.\n\nThis removes three behavior-neutral internal leftovers: Deadlined's write-only done bit, BlockingWork's redundant Option around an owned JoinHandle, and the test-only FenceCounter mint_sequence wrapper with its inaccurate observation claim. Broader lookalikes that still encode distinct lifecycle boundaries or test observability were left intact.\n\nValidation: focused identity coverage and full ./tools/dev just ci (431 tests plus all lint, docs, API/layering, packaging, and Nix checks).\n\nDraft until fresh adversarial review completes.

### DIFF
--- a/crates/shelterwood/src/identity.rs
+++ b/crates/shelterwood/src/identity.rs
@@ -201,15 +201,6 @@ impl FenceCounter {
             generation,
         })
     }
-
-    /// Mints a generation for a non-identity monotone sequence.
-    ///
-    /// Observation uses the same saturating, poison-never-minted primitive as
-    /// membership and incarnation fencing.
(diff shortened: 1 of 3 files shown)


ralexstokes:
Fresh adversarial review found no actionable issues. It confirmed Deadlined::done was write-only, BlockingWork still detaches its owned JoinHandle correctly before and after polling, double join remains type-impossible, and the sole mint_sequence call now executes its exact former body. Focused deadline, identity, and run_blocking verification passed; hosted CI is green.

PR #136, opened 2026-08-09
Replace incarnation report channel with a once cell

ralexstokes:
Part of #116.\n\nThis replaces the per-incarnation std::sync::mpsc report channel with an owned Arc<OnceLock<RecordedReport>>. The child-owned token still fills synchronously on explicit record or Drop fallback; post-join consumption asserts sole ownership and initialized state, preserving exact cancellation sampling without channel machinery.\n\nNew regressions cover unwind fallback and runtime-task abort/join ordering.\n\nValidation: ./tools/dev just ci (433 tests plus all lint, docs, API/layering, packaging, and Nix checks).\n\nDraft until fresh adversarial review completes.

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -3,7 +3,7 @@
 use std::{
     collections::{BTreeMap, HashMap},
     ops::{Bound, Index, IndexMut},
-    sync::{Arc, Mutex, Weak, mpsc},
+    sync::{Arc, Mutex, OnceLock, Weak},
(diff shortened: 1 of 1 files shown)


ralexstokes:
Fresh adversarial review found no actionable issues. It verified the two-owner Arc/OnceLock lifecycle, synchronous fallback across normal return, construction/run/destructor panic, cancellation and abort, identical cancellation sampling linearization, sufficient publication ordering, and disconnected-claim equivalence. Focused report and full-path panic tests passed; hosted CI is green.

PR #141, opened 2026-08-09
Park closed watch change waits

ralexstokes:
Part of #116.\n\nThis fixes the unit-return WatchReceiver::changed wrapper becoming permanently always-ready after publisher closure. Closure-agnostic waits now park forever; consumers that intentionally observe closure continue to use changed_or_closed. The call-site audit confirmed all unit-return waits structurally retain their sender.\n\nTests pin first-poll and repeated-poll pending behavior after closure plus the explicit closure-aware path.\n\nValidation: ./tools/dev just ci (433 tests plus all lint, docs, API/layering, packaging, and Nix checks).\n\nDraft until fresh adversarial review completes.

### DIFF
--- a/crates/shelterwood/src/runtime.rs
+++ b/crates/shelterwood/src/runtime.rs
@@ -876,8 +876,15 @@ impl<T> Clone for WatchReceiver<T> {
 }
 
 impl<T> WatchReceiver<T> {
+    /// Waits for a new value without treating publisher closure as a change.
+    ///
+    /// Callers that need to observe publisher closure must use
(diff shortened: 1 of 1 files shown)


ralexstokes:
Fresh adversarial review found no actionable issues. Every unit-return changed() path structurally retains its sender; snapshot/lifecycle closure consumers already use changed_or_closed. Tokio cancellation/final-update ordering remains intact, parking retains no wake loop, and the regression test distinguishes the prior first-poll Ready behavior. Focused closed-watch, lib, integration, diff, and hosted CI checks passed.

PR #142, opened 2026-08-09
Consolidate stopped scope publication

ralexstokes:
Part of #116; stacked on #129 so the helper preserves its reviewed no-wake-under-gate and #121 terminal ordering guarantees.\n\nThis introduces one ScopeCell::publish_stopped_locked transition for live-epoch stops, no-live root terminal stops, and never-started nested/root publication. It carries epoch ownership through retained mutations, queues mailbox teardown before terminal pulses, releases control before snapshot/lifecycle emission, and leaves observation closure at callers where parent/nested ordering differs.\n\nupdate_locked remains because its deferred-pulse semantics are not equivalent to immediate update. Tests cover all three paths plus hostile mailbox/member waker precedence.\n\nValidation: ./tools/dev just ci (441 tests plus all other lanes).\n\nDraft until fresh adversarial review completes.

### DIFF
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -996,7 +996,7 @@ impl ScopeCell {
             if record.last_incarnation.is_none()
                 && let Some(scope) = &nested
             {
-                scope.publish_never_started_locked(wakes);
+                scope.publish_stopped_locked(wakes, StopReason::NeverStarted, None, None);
             }
             if let Some(incarnation) = exited_incarnation {
                 self.emit_locked(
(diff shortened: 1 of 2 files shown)


ralexstokes:
Fresh adversarial review found no actionable issues. It verified live/stale epoch ownership, no-live root safety, never-started/nested distinctions, teardown-to-pulse-to-observation ordering, nested-parent closure order, startup fill-once, terminal idempotence, and hostile-waker panic precedence. New and existing focused publication, observation, shutdown, and abort tests all passed; hosted CI is green.

PR #143, opened 2026-08-09
Fail closed on lost removal responses

ralexstokes:
Part of #116.\n\nThis aligns Removal with Admission when an internal owned-completion invariant regresses. A closed removal response still trips a debug assertion, but release builds fail closed as RemoveOutcome::Removed because the already-latched removal route becoming terminal satisfies the removal goal.\n\nPublic rustdocs and the normative spec now state the policy. A deterministic closed-response regression covers both build modes.\n\nValidation: ./tools/dev just ci (432 tests plus all lint, docs, API/layering, packaging, and Nix checks).\n\nDraft until fresh adversarial review completes.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -3376,7 +3376,7 @@ path. If that invariant regresses, release builds nevertheless fail
 closed: admission observes `NotAdmitting(Terminal)` and a latched removal
 observes `Removed`, since its route becoming terminal satisfies the
 removal goal. Debug builds MAY instead assert and panic at that boundary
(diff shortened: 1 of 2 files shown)


ralexstokes:
Addressed both low-severity review gaps in 0ad12de. The spec now distinguishes release fallback from debug assertion/panic, and the Removed choice lives in a profile-independent helper covered by ordinary CI while the debug/release execution-path test remains. Focused debug and release tests pass; full ./tools/dev just ci passes with 433 tests and all other checks.

PR #145, opened 2026-08-09
Simplify reply channel ownership

ralexstokes:
Part of #116.\n\nThis removes redundant Option wrappers around the owned, non-cloneable Reply and ReplyReceiver channel halves plus the stored answered bit. Consuming methods now move the halves directly; the underlying one-shot/disposal wrappers retain exactly-once close and detached late-value disposal. CallOperation's phase Option remains because it encodes a real not-yet-constructed state.\n\nThe public Debug shape remains Reply { answered: false, .. }, derived from the consuming capability. Tests cover success, sender drop, receiver cancellation, and Debug output.\n\nValidation: ./tools/dev just ci (433 tests plus all docs, lint, API/layering, package, and formatting checks).\n\nDraft until fresh adversarial review completes.

### DIFF
--- a/crates/shelterwood/src/mailbox.rs
+++ b/crates/shelterwood/src/mailbox.rs
@@ -293,15 +293,17 @@ impl<F: DeadlineOperation + Unpin> Future for Deadlined<F> {
 /// the channel, so a late [`Reply::send`] safely discards its value through
 /// isolated disposal.
 pub struct Reply<T> {
-    sender: Option<OneShotSender<T>>,
-    answered: bool,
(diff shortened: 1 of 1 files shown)


ralexstokes:
Fresh adversarial review found no actionable issues. It verified exactly-once consuming sender/receiver ownership, drop closure, detached late/stored value disposal and panic containment, unchanged CallOperation phase state, identical Debug output, unchanged pin/drop cancellation chain, and no hidden answered/Option uses. Focused reply, delivery, disposal, call, and API checks passed.

PR #146, opened 2026-08-09
Break the plan and cell route cycle

ralexstokes:
Part of #116; stacked on #126 so the new dependency boundary is recursively enforced.\n\nDynamicRoute now has an associated slot type, and the restart-stable cells layer stores an object-safe erased route instead of importing plan::SlotCell. Slot erasure/recovery stays in plan.rs; driver-owned dynamic state remains concrete, preserving canonical Arc identity, definition ownership, and disposal behavior.\n\nA pointer-identity test pins the erased round trip, and layering enforcement forbids plan references in cells with a known-bad fixture. Downcasts remain outside exit paths.\n\nValidation on the stacked branch: ./tools/dev just ci (432 tests plus all docs, API/path/enforcement, packaging, and formatting checks).\n\nDraft until fresh adversarial review completes.

### DIFF
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -7,6 +7,7 @@
 //! only as opaque capability payloads; their state remains owned elsewhere.
 
 use std::{
+    any::Any,
     fmt,
     sync::{
         Arc, Mutex, MutexGuard, OnceLock, RwLock, Weak,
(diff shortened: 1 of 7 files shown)


ralexstokes:
Fresh adversarial review found one MEDIUM enforcement gap: the new cells-to-plan layering guard covered only cells.rs, so a nested cells module could silently restore the cycle. Addressed in f506310 by recursively discovering cells.rs plus cells/** and adding a directory-module known-bad fixture. Focused checks and full local CI pass; hosted CI is rerunning.

PR #147, opened 2026-08-09
Replace receiverless scope watches with plain state

ralexstokes:
Part of #116.\n\nThis replaces ScopeCell's receiverless config, current_children, and parent watch channels with poison-tolerant mutex state. The tree observation gate still owns compound consistency; recursive projections clone before descent, and removed residents always move outside the state guard before Drop/Removed publication. Member/scope record watches and terminal publication are unchanged.\n\nDead WatchSender read/replace/take helpers disappear. Tests cover atomic concurrent config snapshots, recursive removal after unlock, and nested parent snapshot propagation.\n\nValidation: ./tools/dev just ci (434 tests plus all docs, layering, package, and formatting checks).\n\nDraft until fresh adversarial review completes.


ralexstokes:
Fresh adversarial review found no actionable issues. It traced every converted config/current_children/parent access; verified guards are released before recursive publication, dropping, observation-gate handoff, and parent traversal; confirmed consistent poison recovery and no API regression; and passed all 154 library tests plus hosted CI.

PR #148, opened 2026-08-09
Wait for live scope epilogues

ralexstokes:
Part of #116.

Fixes the confirmed `wait_stopped` ordering race during parent teardown. A parent can publish terminal membership before Tokio destroys an aborted nested driver and that driver publishes its final stopped scope record; waiters now reserve `NeverStarted` for truly unstarted scopes and continue waiting for the live epoch epilogue otherwise.

Adds a deterministic regression for the parent-terminalizes-before-nested-epilogue ordering. Full `./tools/dev just ci` passes locally (432 tests plus docs, layering, packaging, and formatting checks).

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -2189,7 +2189,14 @@ cooperative cancel → grace expiry → tidy-abort beat → hard abort
   at its parent as an ordinary child failure; the parent's restart
   policy applies as usual — each retry re-invokes the factory, so a
   stateful source can heal, while a deterministic factory bug churns
-  to the parent's intensity trip (§9.2), the designed containment. In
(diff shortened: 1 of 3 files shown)


ralexstokes:
Fresh adversarial review found a HIGH pre-driver race: a restartable nested factory could execute before reserving its scope epoch, allowing membership terminality and wait_stopped completion before that same poll later began an incarnation. Fixed in 4f17e47 by acquiring ScopeEpochGuard on the task first poll before invoking the factory and holding it through lowering/driver handoff. Added deterministic initial-factory, restarted stale-Stopped factory, and factory-panic regressions. Focused tests and full local CI pass (435 tests); hosted CI is rerunning.

Issue #151, opened 2026-08-10
Post-review backlog: simplification remainder, robustness residuals, tooling


ralexstokes:
Successor to #116, which is closed with its review round complete: all 33 PRs (#117–#149, plus the #150 build repair) merged 2026-08-10 — every §1 confirmed bug fixed and regression-pinned, all §4 spec tightenings landed with their pinning tests, §5 doc accuracy done, §2 robustness bullets acted on. This issue carries only what remains open, consolidated from #116's residuals and the findings of the per-PR adversarial reviews.

## 1. Simplification backlog (open remainder of #116 §3)

- **Collapse the remaining parallel state encodings** (#116 §3.3): `engine::ScopePhase` ≈ `observe::ScopeState` ≈ `cells::ScopeRecord.state` are still paired by hand at three driver sites, and the `MemberStage` transition graph is still scattered across driver mutation closures. (The `MembershipMode`/`MembershipStatus` unification landed via #140.) Related: removal now has **three** hand-projected encodings — `MemberRecord.removing`, public `MembershipStatus`, and control-plane `DynamicMembershipState` (#149) — extract e.g. `MembershipStatus::from_record` and consolidate.
- **Finish exit-classification unification** (#116 §3.4; #131 delivered the precedence/encoding slice): the driver still edits verdicts around the classifier (hard-abort join rewrite; disposal-panic exit rewrite), `StopReason → Exit` is folded two ways, and `JoinVerdict` remains a lossy copy of `runtime::JoinOutcome`.
- **Collapse `readiness`/`readiness_override` to one field** resolved where the child kind is known (#116 §3.8 clause; #130 delivered the rest of that item).
- **tree.rs remainder** (#116 §3.9; #135 delivered the forwarding surface): `impl_common_builder_surface!` doc-string double expansion; `System::armed` → `Drop for SystemRun`; dead defensive arms in `AdmissionState`; `AdmissionOwnership` threaded through static paths that ignore it; reconsider public `AdmissionReceipt`.
- **Driver split remainder** (#116 §3.1/§3.6): extract `driver/tests.rs`; make readiness a definition-level property to delete the `Constructed` round-trip; merge or unbound the spillover lane to delete the admission request forwarder task.

## 2. Robustness residuals (found by the review round)

[...]

PR #153, opened 2026-08-10
Polish startup and reply diagnostics

ralexstokes:
## Summary

- include the child exit-kind summary in `StartupFailure` display text, making panic payloads visible through erased error rendering
- preserve recursive nested failure detail while retaining the existing typed source chain
- relax the `Reply` debug test to pin only the meaningful derived state instead of the complete non-contractual format

## Impact

Startup failures now carry actionable child detail in ordinary error output, including panic messages that cannot be represented as an `Error::source`. Debug formatting remains free to evolve while its useful state stays covered.

Part of #151.

## Validation

- `./tools/dev cargo test -p shelterwood` (201 unit, 294 integration, 3 doctests)
- `./tools/dev just ci` (499 nextest tests plus formatting, Clippy, docs, API/layering checks, packaging, and Nix formatting)


ralexstokes:
Fresh adversarial review found one low-severity coverage gap: the new `StartupFailure` display branches were only exercised for application failure and panic-with-message.

Addressed in `0730f32` with a table-driven unit test that pins the complete rendering for every `ExitKind`, including message-less panic, readiness timeout, both abort phases, completion, and never-started.

Validation after the review fix: focused test pass and full `./tools/dev just ci` pass (500 nextest tests plus formatting, Clippy, docs, API/layering checks, packaging, and Nix formatting). The reviewer reported no other concrete findings.

PR #154, opened 2026-08-10
Preserve restart-window stop reasons

ralexstokes:
## Summary

- distinguish a truly never-started nested scope from the no-live-incarnation gap between restarts
- preserve the last real nested-scope stop reason when parent-driver destruction terminalizes the restarting membership
- add a driver-level regression that exercises the armed destruction fallback and verifies stop reason, abort exit, final snapshot, observation closure, and `wait_stopped()` completion

## Root cause

`discharge_child_terminality` used `record.incarnation.is_none()` to decide that a nested scope needed a synthesized `NeverStarted` stop. That condition also holds during restart backoff, after an incarnation has already published a real `Stopped { reason }` state. Parent-driver destruction therefore overwrote the real reason with `NeverStarted`.

The fallback now keys the synthesis on `last_incarnation.is_none()`, which is the actual never-ran predicate. Restart-window terminalization retains the prior scope epilogue while still terminalizing and closing the membership.

The regression directly discharges the same armed `Obligation<ChildTerminality>` used by `ScopeRuntime::drop`. Reverting the predicate to the pre-fix `exited_incarnation.is_none()` makes it fail with `NeverStarted` instead of the prior `Finished` reason.

Part of #151.

## Validation

- `./tools/dev cargo test -p shelterwood terminality_fallback_preserves_restart_window_scope_reason`
- pre-fix predicate mutation: focused test fails with `left: NeverStarted, right: Finished`
- `./tools/dev just ci` (500 nextest tests plus formatting, Clippy, docs, API/layering checks, packaging, and Nix formatting)

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -385,7 +385,10 @@ fn discharge_child_terminality(completion: ChildTerminality) {
     if matches!(record.stage, MemberStage::Terminal(_)) {
         return;
     }
-    let (exit, exited_incarnation) = if record.last_incarnation.is_some() {
+    let never_started = record.last_incarnation.is_none();
+    let (exit, exited_incarnation) = if never_started {
+        (Exit::never_started(), None)
+    } else {
         (
             Exit::new(
                 ExitKind::Aborted {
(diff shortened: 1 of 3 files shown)


ralexstokes:
Fresh adversarial review found that the initial integration assertion exercised graceful drain, which disarmed the terminality obligation before the changed destruction fallback ran.

Addressed in `65cdafe`:

- removed the misleading integration assertion and restored that test's original scope;
- added a driver-level regression that directly discharges the still-armed `ChildTerminality` obligation in the restart window;
- asserted preservation of the prior `Finished` stop reason, the parent-destruction abort exit, the retained final stopped snapshot, observation closure, and `wait_stopped()` completion.

Mutation check: changing the condition back to the pre-fix `exited_incarnation.is_none()` makes the new test fail with `left: NeverStarted, right: Finished`; restoring the fix makes it pass.

Validation after the review fix: focused regression pass and full `./tools/dev just ci` pass (500 nextest tests plus formatting, Clippy, docs, API/layering checks, packaging, and Nix formatting). The reviewer found no predicate correctness defect.

PR #156, opened 2026-08-10
Pin dynamic admission and removal interleavings

ralexstokes:
Part of #151.

## Summary

- deterministically fire fused cancellation after its admission is queued but before the driver handles it, proving the pre-conversion latch check rejects before child conversion
- pin the under-lock latch re-check separately: park child conversion on the held child-identity mutex, fire fused cancellation while the driver is parked past the pre-conversion check, then prove the re-check under the control-plane lock rejects with `ReservationEnded` and admits no resident
- pin the reachable reserved-membership behavior: cancellation cannot emit a key-addressed `RemovalRequest` until admission has promoted the entry
- exercise duplicate fused and explicit removal delivery both while a child is active and while retained construction disposal is pending
- add a fused-only removal variant (fused drop, no explicit `remove`) pinning the driver-side `publish_dynamic_removal` call: the fused source marks the control-plane entry but never touches the member record, so `handle_removal` is the sole writer of the public `Removing` projection on that path
- assert idempotent stop-ladder/deadline state, delayed removal completion, complete cleanup, and exactly one Added/Removed lifecycle edge

## Validation

- `./tools/dev cargo test -p shelterwood fused_cancellation` (pre-conversion and parked-conversion interleavings)
- `./tools/dev cargo test -p shelterwood double_removal_is_idempotent`
- `./tools/dev cargo test -p shelterwood fused_only_removal`
- mutation checks, each verified against the exact test that kills it:
  - removing the pre-conversion latch check in `handle_admission` fails the pre-conversion test
  - removing the fused-latch disjunct of the under-lock re-check fails only the parked-conversion test; the pre-conversion test cannot reach that check
  - deleting the `publish_dynamic_removal` call in `handle_removal` fails only the fused-only test; both explicit-remove duplicate variants shadow it because `remove_dynamic` publishes the projection before its request reaches the driver
  - removing the live stop-ladder guard in `begin_stop_child` fails the while-active duplicate test
  - the disposal-time duplicate is protected jointly by `begin_stop_child`'s terminal/disposing guard and `begin_terminal_disposal`'s pending-terminal guard; either check alone suffices, so that test kills only their joint mutation (stated in the test comment)
  - the record-`removing` guard inside `publish_dynamic_removal` has no distinguishing public observable (skipping it only avoids re-publishing an identical record); it is documented in code as idempotency hygiene rather than claimed as pinned
- `./tools/dev just ci`

## Review follow-up

The first draft injected an artificial removal after dequeuing admission, which did not prove a production-reachable ordering. The follow-up removes that event, explicitly pins why a reserved entry cannot emit Removal, and reframes the test around the real latch-vs-admission race. It also replaces a fixed-yield startup poll with a bounded latch wait.

A second adversarial pass found two of the four originally claimed guard pins did not hold: the under-lock latch re-check was shadowed by the pre-conversion check, and the driver-side `Removing` publication was shadowed by the explicit-remove path in both duplicate variants. This revision adds the parked-conversion and fused-only tests, re-verifies every mutation kill individually, and downgrades the unpinnable record guard to a documented hygiene check.

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -959,6 +959,14 @@ impl ScopeRuntime {
         else {
             return;
         };
+        // Idempotency hygiene rather than a load-bearing guard: an explicit
+        // removal publishes this projection before its request reaches the
+        // driver, and duplicate deliveries re-enter here after the first
+        // publication. Skipping the transition only avoids re-publishing an
+        // identical record under the observation gate; no public observable
+        // distinguishes that redundant publication, so tests pin the call
+        // sites (the fused-only removal path, where this write is the sole
+        // Removing-projection writer) rather than this check.
         if !member.record().removing {
             self.root
                 .transition_child(&member, |record| record.removing = true, None);
--- a/crates/shelterwood/src/driver/tests.rs
+++ b/crates/shelterwood/src/driver/tests.rs
@@ -12,8 +12,8 @@ use std::{
 use crate::{
     ActorRef, Backoff, Cancellation, ChildId, ChildState, DynamicTree, Exit, ExitError, ExitKind,
     GracePhase, Intensity, LifecycleEventKind, LifecycleItem, LifecycleTryRecvError, Mailbox,
-    RawOnceDef, Readiness, ReadinessDeadline, RemoveOutcome, ReserveError, RestartCondition,
-    RestartPolicy, Retention, ScopeRef, ScopeState, SendErrorKind, StartupError,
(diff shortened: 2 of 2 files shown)


ralexstokes:
Addressed the adversarial review in `f9d08a5`:

- removed the artificial arbitrary-key Removal event and reframed the regression around the production-reachable race: fused cancellation fires after Admission is queued but while the entry is still Reserved
- added an assertion that this state emits no key-addressed Removal, documenting the control-plane invariant behind the latch re-check
- renamed the test and updated the PR title/body so they no longer claim an unreachable Removal-over-Admission ordering
- replaced the fixed 64-yield startup poll with a latch awaited under a two-second timeout

Both focused test groups pass locally; full CI revalidation is running.

PR #171, opened 2026-08-10
Resolve lifecycle and admission review residuals

### LINKED ISSUE #170: Residuals from the #151 review round
Successor to #151, which closed with every enumerated item delivered (PRs #153–#169, all merged 2026-08-10). This carries only the small adjacent gaps surfaced by that round's per-PR adversarial reviews — none are regressions; all are pre-existing or accepted-by-design edges recorded in https://github.com/ralexstokes/shelterwood/issues/151#issuecomment-5243531604.

## Code cleanups

- **Last scope-state hand-pairing at incarnation start**: the driver constructs `ScopeLifecycle::starting()` while `cells::begin_incarnation` independently writes `record.state = ScopeState::Starting` (#163 closed the three driver transition sites; this construction-time pairing remains). Pass the lifecycle's initial state into `begin_incarnation`, or fold construction into the machine. Related: `ScopeCell::set_state`'s `Starting` arm (`total_restarts` reset) is dead in production — production Starting goes through `begin_incarnation`, which duplicates the reset.
- **Route `discharge_child_terminality` through `classify_exit`**: it still hand-builds its `Aborted{WithinGrace}`/`Observed` exit (#162 routed `join_driver`/`spawn_system`; this site was deferred to avoid conflicting with #154's then-open edits to the same lines).

## Test gaps

- **Fused-cancel window projection skew**: the window where the public snapshot reads `Active` while exit dispatch classifies `Removing` (`dispatch_membership_status`) is documented in SPEC (control-plane wins for restart decisions) but not pinned by a test.
- **Dynamic-add readiness-panic unwind**: a panicking `R::readiness()` during a dynamic add unwinds the *adding* actor (#165 made evaluation definition-time); only the static builder unwind is pinned.

## Recorded, no action planned

- The member `Terminal` edge intentionally stays a raw multi-field publication in `publish_terminal_with_wakes` (idempotency latch + mailbox-teardown transaction); the rest of the transition graph is centralized behind `MemberTransition` with per-arm legal-source `debug_assert!`s.
- `publish_dynamic_removal`'s record guard is documented as idempotency hygiene — proved unpinnable by any public observable; the call itself is pinned by the fused-only test.

ralexstokes:
## Summary

- make the pre-driver epoch guard own the matching initial `ScopeLifecycle` and pass that machine state into scope-incarnation publication
- remove the dead `ScopeCell::set_state(Starting)` restart reset
- route child-terminal fallback aborts through the canonical `classify_exit` path
- pin the fused-cancel control-plane/public-snapshot projection window
- pin dynamic one-shot raw readiness panics, including single resource drop and reservation reuse

## Why

Issue #170 recorded the remaining adjacent gaps from the #151 review round. Scope incarnation startup still hand-paired two copies of `Starting`, one fallback exit path bypassed canonical classification, and two already-documented edge behaviors lacked regression coverage.

## Impact

Scope startup now projects the exact state owned by its lifecycle machine, fallback abort classification shares the same precedence rules as other join paths, and the fused-cancel plus dynamic-definition unwind contracts are protected by deterministic tests.

## Validation

- `./tools/dev just ci`
  - nightly rustfmt and clippy with warnings denied
  - 513 nextest tests
  - doctests and rustdoc/API reachability
  - runtime/exit/layering enforcement checks

Closes #170

### DIFF
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -1026,9 +1026,6 @@ impl ScopeCell {
     pub(crate) fn set_state(&self, state: ScopeState) {
         self.with_observation_gate(|wakes| {
             self.record.modify_silently(|record| {
-                if state == ScopeState::Starting {
-                    record.total_restarts = TotalRestarts::ZERO;
-                }
                 record.state = state.clone();
             });
             wakes.pulse(&self.record);
@@ -1383,11 +1380,10 @@ impl ScopeCell {
         }
     }
 
-    pub(crate) fn begin_incarnation(&self) -> Option<Epoch> {
+    pub(crate) fn begin_incarnation(&self, state: ScopeState) -> Option<Epoch> {
         self.with_observation_gate(|wakes| {
             let mut control = self.control.lock().expect("scope control mutex poisoned");
             let epoch = control.epochs.begin()?;
-            let state = ScopeState::Starting;
             self.record.modify_silently(|record| {
                 record.total_restarts = TotalRestarts::ZERO;
                 record.state = state.clone();
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -1381,11 +1381,16 @@ impl ScopeCell {
     }
 
     pub(crate) fn begin_incarnation(&self, state: ScopeState) -> Option<Epoch> {
+        debug_assert!(
+            matches!(state, ScopeState::Starting),
+            "a fresh incarnation publishes its lifecycle machine's initial state"
+        );
         self.with_observation_gate(|wakes| {
             let mut control = self.control.lock().expect("scope control mutex poisoned");
             let epoch = control.epochs.begin()?;
             self.record.modify_silently(|record| {
                 record.total_restarts = TotalRestarts::ZERO;
+                record.startup = None;
                 record.state = state.clone();
             });
             // Hold epoch ownership through its observation projection. A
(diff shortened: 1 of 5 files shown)


ralexstokes:
## Review

Ran a multi-agent review of this PR (8 finder angles, adversarial verification of each candidate). 11 candidate findings; 8 refuted, 3 survived and are addressed in 51e771b:

**1. `begin_incarnation` never cleared the write-once startup verdict** (`cells.rs`). It reset `total_restarts` and `state` but not `record.startup`, while `set_startup` only writes when the field is `None` — so incarnation 2+ of a restartable nested scope could never publish its own verdict; every later `set_startup` silently no-opped. Latent today (the only non-test reader is `wait_started`, reachable only for the single-incarnation root), but a genuine inconsistency with the sibling fields reset in the same block. Fixed by resetting `startup` alongside them, with a cell-level regression test.

**2. No guard against a non-`Starting` state reaching `begin_incarnation`.** The new parameter is correctly sourced from the lifecycle machine, but a future caller passing a terminal state would publish it against a live epoch and still zero restarts. Added a `debug_assert`.

**3. `ScopeEpochGuard`'s `Option<ScopeLifecycle>` + `take`/`expect` plumbing.** Unlike the epoch, the lifecycle carries no drop obligation, so the `Option` bought nothing but a runtime panic path. Replaced with a plain field and a cloning `lifecycle()` accessor, preserving the transfer-last ordering.

Notable refuted candidates (all verified against code/tests): the fallback `Aborted{WithinGrace}` over a recorded `last_exit` is the documented, test-pinned kill-path contract; the "missing" `Exited` event for `incarnation: None` children is already emitted atomically by `publish_child_restart` (the proposed fix would duplicate it); routing through `classify_exit` was explicitly requested by #170 and is behavior-identical; and the four test preambles that resemble `running_dynamic_fixture` differ in load-bearing channel wiring (merged vs. separate dynamic lanes), so they aren't collapsible without changing what the tests assert.

Issue #172, opened 2026-08-10
Full-codebase review 2026-08-10: merged findings (correctness, tests, simplification, tooling)


ralexstokes:
Merged findings from two independent full-codebase reviews conducted 2026-08-10: a 13-agent module-by-module + test-suite review, and a second independent review. Duplicates are merged; where the reviews disagree, both positions are recorded.

Each finding carries a status:

- **verified** — mechanics traced through the code by at least one reviewer.
- **corroborated** — found independently by both reviews.
- **reported** — from the second review; code shape spot-checked at the cited lines, but the claimed interleaving/consequence not independently re-derived. Verify before (or by) writing the regression test.
- **disputed** — the reviews reached opposite conclusions; adjudicate with a targeted test first.

Headline: neither review found a reachable production bug in the happy paths — the correctness core is strong. The findings cluster around (a) one spec-vs-code contract divergence, (b) races and latent hazards on the dynamic-scope and terminalization edges, (c) invariants held by convention at distant call sites rather than by construction, (d) test gaps, and (e) ~600+ lines of mechanical simplification.

---

## Part 0 — Structural moves (decided direction: fix classes, not instances)

Breaking changes are explicitly in scope. The race-shaped findings below are not independent bugs: they share root causes, and each move here dissolves a cluster of items by construction. Fix rounds should implement these moves and check off the subsumed items, rather than patching windows individually. The crate's own pattern is the proof this works: everywhere semantics already live in the pure core as a total function (exit classification's `prefer_earlier`, `arbitrate`'s class order), both reviews found the code airtight; every race lives where that principle stops.

[...]

PR #180, opened 2026-08-10
Implement Part 0.3 drain verdict lattice

ralexstokes:
## Summary

- replace the scope drain's first-writer-wins reason with the Part 0.3 total precedence order: `ShutdownRequested > StartupFailed > IntensityTripped > Finished`
- upgrade in-progress drain verdicts monotonically without repeating teardown side effects
- make `StopLadder::force` retain the earliest deadline
- add coverage for the cross-pass `max_restarts = 0` shutdown/intensity race, every drain-reason upgrade, downgrade rejection, and already-due forced deadlines

## Why

Part 0.3 of #172 identifies a shared first-writer-wins root cause. If shutdown latched immediately after a driver pass sampled the control plane, a restartable exit could trip intensity first. The next pass observed shutdown, but `ScopeLifecycle::begin_drain` returned without changing the already-latched intensity verdict.

The same monotonicity requirement applies to forced stop deadlines: forcing an already-due ladder must not move its deadline later.

## Impact

Explicit shutdown now deterministically supersedes lower-precedence causes even when they begin teardown one pass earlier. Lower-precedence causes cannot overwrite a stronger verdict, and forcing a stop ladder can only preserve or shorten its deadline.

This implements move 0.3 from #172, including item 5 and the `StopLadder::force` bullet from item 10.

## Validation

- `./tools/dev just ci`
  - 520 nextest tests
  - doctests and rustdoc
  - packaged-crate verification
  - Nix formatting checks

### DIFF
--- a/crates/shelterwood/src/driver/tests/shutdown.rs
+++ b/crates/shelterwood/src/driver/tests/shutdown.rs
@@ -74,7 +74,7 @@ async fn latched_shutdown_upgrades_an_intensity_drain() {
     );
 }
 
-#[crate::runtime::test]
+#[crate::runtime::test(start_paused = true)]
 async fn force_uses_the_stop_funnel_for_every_ordered_child() {
     let mut tree = Tree::new();
     let first = tree
--- a/crates/shelterwood/src/driver/tests/shutdown.rs
+++ b/crates/shelterwood/src/driver/tests/shutdown.rs
@@ -1,6 +1,80 @@
 use super::support::*;
 
 #[crate::runtime::test]
+async fn latched_shutdown_upgrades_an_intensity_drain() {
+    let mut tree = Tree::new();
+    tree.intensity(Intensity::new(0, Duration::from_secs(10)).expect("valid intensity"));
+    tree.add_task(
+        "worker",
+        TaskDef::new(|_| future::pending::<crate::ExitResult>()),
+    )
+    .expect("valid task");
+    let mut plan = tree.lower_for_test();
+    let root = Arc::clone(&plan.root);
+    let epoch = root
+        .begin_incarnation(ScopeState::Starting)
+        .expect("test scope epoch is available");
+    root.set_admitted_children(
+        plan.children
+            .iter()
+            .map(|child| resident_projection(&child.slot))
+            .collect(),
+    );
+    root.set_state(ScopeState::Running);
+    root.set_startup(Ok(()));
+    let (events, _event_receiver) = crate::runtime::unbounded_mpsc();
+    let (disposal_events, _disposal_event_receiver) = crate::runtime::unbounded_mpsc();
+    let child = ChildRuntime::from_plan(plan.children.pop().expect("one child plan"), &root);
+    let mut children = ChildArena::default();
+    let key = children
+        .insert(child)
+        .unwrap_or_else(|_| panic!("the test fixture fits in the child-key domain"));
+    let mut scope = ScopeRuntimeBuilder::new(Arc::clone(&root), epoch, events, disposal_events)
+        .with_defaults(plan.defaults.clone())
+        .with_intensity_policy(plan.config.intensity)
+        .with_children(children)
+        .with_lifecycle(ScopeLifecycle::running())
+        .build();
+    plan.armed = false;
+    drop(plan);
+
+    scope.spawn_child(key);
+    let active = scope.children[key]
+        .active
+        .as_ref()
+        .expect("worker is active");
+    let incarnation = active.incarnation;
+    active.abort_handle.abort();
+
+    // Model a shutdown request that latches after this pass sampled the
+    // control plane but before the collected child exit is dispatched.
+    assert!(root.request_shutdown().is_some());
+    scope.handle_exit(
+        key,
+        incarnation,
+        Some(RecordedOutcome::returned(Err(ExitError::message(
+            "trip intensity",
+        )))),
+        crate::runtime::JoinOutcome::Ok { value: () },
+        Cancellation::NotObserved,
+        false,
+    );
+    assert!(matches!(
+        scope.lifecycle.draining_reason(),
+        Some(StopReason::IntensityTripped(_))
+    ));
+
+    // The request's next-pass follow-up owns the stronger verdict even
+    // though teardown was already started by the intensity trip.
+    assert!(root.take_shutdown_request(scope.epoch));
+    scope.begin_drain(StopReason::ShutdownRequested);
+    assert_eq!(
+        scope.lifecycle.draining_reason(),
+        Some(&StopReason::ShutdownRequested)
+    );
+}
+
+#[crate::runtime::test]
 async fn force_uses_the_stop_funnel_for_every_ordered_child() {
     let mut tree = Tree::new();
     let first = tree
--- a/crates/shelterwood/src/driver/tests/shutdown.rs
+++ b/crates/shelterwood/src/driver/tests/shutdown.rs
@@ -74,6 +74,81 @@ async fn latched_shutdown_upgrades_an_intensity_drain() {
     );
 }
 
+#[crate::runtime::test]
+async fn force_upgrades_an_intensity_drain_to_shutdown_requested() {
+    let mut tree = Tree::new();
+    tree.intensity(Intensity::new(0, Duration::from_secs(10)).expect("valid intensity"));
+    tree.add_task(
+        "worker",
+        TaskDef::new(|_| future::pending::<crate::ExitResult>()),
+    )
+    .expect("valid task");
+    let mut plan = tree.lower_for_test();
+    let root = Arc::clone(&plan.root);
+    let epoch = root
+        .begin_incarnation(ScopeState::Starting)
+        .expect("test scope epoch is available");
+    root.set_admitted_children(
+        plan.children
+            .iter()
+            .map(|child| resident_projection(&child.slot))
+            .collect(),
+    );
+    root.set_state(ScopeState::Running);
+    root.set_startup(Ok(()));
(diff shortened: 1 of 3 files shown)


ralexstokes:

I checked the full drain-reason order, equal/lower-rank retention, upgrade-without-replaying teardown effects, final-reason lookup through finish_if_ready, NeverStarted remaining outside the live-incarnation lattice, the cross-pass max_restarts = 0 shutdown/intensity race, and StopLadder::force for already-due and repeated deadlines. The implementation matches the settled SPEC §10 contract.

Validation: ./tools/dev just ci passed, including formatting, Clippy with warnings denied, all 520 nextest tests, doctests/rustdoc, API reachability, runtime/exit/layering enforcement, packaged-crate verification, and Nix formatting.

crates/shelterwood/src/engine.rs
```diff
@@ -788,12 +810,27 @@ impl ScopeLifecycle {
...
             ScopeState::Starting => StartupPhase::Pending,
             ScopeState::Running => StartupPhase::Complete,
             ScopeState::StartupFailed => StartupPhase::Failed,
-            ScopeState::Draining => return None,
+            ScopeState::Draining => {
+                let drain = self
+                    .drain
+                    .as_mut()
+                    .expect("the drain invariant supplies a reason while draining");
+                if incoming_precedence > DrainReasonPrecedence::of(&drain.reason) {
```
ralexstokes, line 828: **[review] The `StartupFailed > IntensityTripped` edge of this lattice is unreachable in production.** The only `begin_drain(StartupFailed)` site (driver/startup.rs:170) sits behind `!self.lifecycle.is_draining()` (driver/child.rs:1050) and `ScopeLifecycle::fail_startup`'s `state == Starting` gate; and once `ScopeMode::Draining`, `dispatch_exit` returns `Terminal` for every exit, so `begin_drain(IntensityTripped)` cannot arrive mid-drain either. In production only `ShutdownRequested` can reach this upgrade arm — the other rungs are exercised only by the unit test. Not a wrong verdict (SPEC ~1185-1202 pins the trip's arbitration win as owning the startup transition, so no `StartupFailed` cause exists once draining), but SPEC §10's "upgrades ... when a higher-precedence cause arrives" overstates what can arrive. Worth an adjudication note recording the dead edges as defensive generality.

crates/shelterwood/src/engine.rs
```diff
@@ -666,6 +666,28 @@ enum StartupPhase {
     Failed,
 }
 
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
+enum DrainReasonPrecedence {
```
ralexstokes, line 670: **[review] Consider placing this ranking next to `StopReason` in exit.rs, beside `OutcomeRank`.** The payload-erased rank enum follows the established `OutcomeRank` idiom, but `OutcomeRank` lives next to `ExitKind` at its definition site while this ranking of `StopReason` lives in engine.rs, away from the enum it orders. A future `StopReason` variant (Part II §19 group restarts is the named extension point) gets ranked — or missed — far from the definition. A `StopReason::drain_precedence()` (or this enum moved to exit.rs) would make every variant addition meet the SPEC §10 lattice where the variants are defined.

crates/shelterwood/src/engine.rs
```diff
@@ -666,6 +666,28 @@ enum StartupPhase {
...
+
+impl DrainReasonPrecedence {
+    fn of(reason: &StopReason) -> Self {
+        match reason {
+            StopReason::Finished => Self::Finished,
+            StopReason::IntensityTripped(_) => Self::IntensityTripped,
+            StopReason::StartupFailed(_) => Self::StartupFailed,
+            StopReason::ShutdownRequested => Self::ShutdownRequested,
+            StopReason::NeverStarted => {
+                unreachable!("NeverStarted is not a live-incarnation drain reason")
```
ralexstokes, line 685: **[review] The `NeverStarted` exclusion is enforced only by this runtime panic.** `begin_drain` is `pub(crate)` and cells.rs already mints `NeverStarted` verdicts through a separate terminalization path; a future refactor routing a no-incarnation stop through `begin_drain` would panic mid-teardown where a compile error was available. Verified unreachable today (all production call sites pass literal live-reason variants), so this is fragility, not a bug — but given the codebase's newtype discipline, a `DrainReason` type that structurally excludes `NeverStarted` would make SPEC §10's "sits outside the order" clause a type-level fact.

crates/shelterwood/src/engine.rs
```diff
@@ -788,12 +810,27 @@ impl ScopeLifecycle {
         Some(self.state.clone())
     }
 
+    /// Begins draining or monotonically upgrades an in-progress drain.
+    ///
+    /// The returned effect exists only for the initial transition: upgrades
+    /// change the eventual verdict without repeating teardown side effects.
     pub(crate) fn begin_drain(&mut self, reason: StopReason) -> Option<DrainEffect> {
+        let incoming_precedence = DrainReasonPrecedence::of(&reason);
```
ralexstokes, line 818: **[review] Minor:** `incoming_precedence` is computed on every call but consumed only in the `Draining` arm — the common Starting/Running/StartupFailed paths discard it. It can be computed inline in that arm (`if DrainReasonPrecedence::of(&reason) > DrainReasonPrecedence::of(&drain.reason)`; no borrow conflict), keeping the upgrade logic in one place.

crates/shelterwood/src/driver/tests/shutdown.rs
```diff
@@ -1,6 +1,80 @@
 use super::support::*;
 
 #[crate::runtime::test]
+async fn latched_shutdown_upgrades_an_intensity_drain() {
+    let mut tree = Tree::new();
+    tree.intensity(Intensity::new(0, Duration::from_secs(10)).expect("valid intensity"));
+    tree.add_task(
```
ralexstokes, line 7: **[review] Minor:** this hand-builds the single-pending-task tree that `pending_tree()` (driver/tests/support.rs) already encapsulates — `let mut tree = pending_tree(); tree.intensity(...)` works, since the child name is not load-bearing here.

ralexstokes:
**Review — two findings on lines outside this diff** (so they can't be inline comments), both consequences of the new `begin_drain` upgrade contract:

### 1. `force_all` still guards `begin_drain` and can lose a `ShutdownRequested` upgrade (confirmed race)

`crates/shelterwood/src/driver/shutdown.rs:148`:

```rust
if !self.lifecycle.is_draining() {
    self.begin_drain(StopReason::ShutdownRequested);
}
```

This guard was written when a mid-drain `begin_drain` was a pure no-op; under this PR it actively suppresses the monotone upgrade on the force path. There is a reachable loss today via `Pending::AncestorAbort`: the parent always fires the child's shutdown latch before its abort latch, but the child driver samples the two latches non-atomically in one pass (driver.rs:681-688 vs 690-698). On a multi-threaded runtime the parent can fire both between the two reads (a forced ladder emits Cancel+Escalate in one `advance_ladder` loop), so the child collects only `AncestorAbort` and runs `force_all` on a scope already draining for `IntensityTripped`. If all children are already exited with only retained-factory disposals pending, `force_all`'s tail detaches those synchronously and `finish_if_ready` terminalizes in the same pass with the stale `IntensityTripped` reason — the fired shutdown latch is never observed, violating SPEC §10's "an explicit shutdown request cannot lose its verdict to a lower-precedence cause that latched one wake earlier."

Fix: make the call unconditional — the new wrapper semantics make it upgrade-safe (`ScopeRuntime::begin_drain` early-returns on `None`, and the lifecycle performs the upgrade before returning `None`, so nothing is repeated).

### 2. Verdicts minted directly against the epoch bypass the lattice (plausible; adjudication call)

`crates/shelterwood/src/driver.rs:543` — the lowering-failure path calls `epoch.finish(StartupFailed)` without consulting the shutdown latch, and `finish_incarnation_with_terminal` (cells.rs:1438-1443) unconditionally clears a latched request and publishes the caller's reason verbatim. A `ShutdownRequested` latched during lowering (the epoch is begun at driver.rs:479/493 before `tree.lower` at :517, and `disposal.fired().await` at :540 is a guaranteed async window on the failure path itself) is silently dropped and the scope records the lattice-lower `StartupFailed`. The identical race one instant later — inside the driver loop — yields `ShutdownRequested`. SPEC §10 scopes the mechanism to "a drain already in progress," so this may be intentional, but the sentence "an explicit shutdown request cannot lose its verdict to a lower-precedence cause" reads broadly enough to cover it. Suggest adjudicating: either the pre-loop finish sites take `max(reason, latched request)`, or the SPEC bullet explicitly excludes pre-drain verdicts.

---

Inline comments cover the remaining findings: the production-dead `StartupFailed`/`IntensityTripped` upgrade edges, `DrainReasonPrecedence` placement, the `NeverStarted` `unreachable!`, a hoisted precedence computation, and a test-fixture reuse nit. Verified non-findings worth recording: the unbounded upgrade window and the startup-latch/terminal-verdict divergence are both exactly the adjudicated SPEC §10 contract (checked against #172's adjudications and a4d3fd6); the deadline `min` clamp direction and grace-phase provenance check out in every phase/policy combination traced; the new test's "next-pass follow-up" premise is sound (a restartable child's retained construction forces async disposal, guaranteeing pass N+1).

crates/shelterwood/src/engine.rs
ralexstokes: Fixed in d78ed61: begin_drain is now unconditional in force_all (the stale `!is_draining()` guard was suppressing the ShutdownRequested upgrade on the force path); regression test driver::tests::shutdown::force_upgrades_an_intensity_drain_to_shutdown_requested pins the upgrade.

PR #182, opened 2026-08-10
Part 0.4: carry control subjects into the driver

ralexstokes:
## Why

Part 0.4 of #172 requires driver wakes to carry their subject instead of rediscovering rare control-plane facts through full-arena scans. Item 11 also requires shutdown completion to stop rescanning every child on every pass.

Stacked on #181.

## What

- publish pending-incarnation restart shutdowns as transactional control events carrying exact `Membership`
- resolve event subjects through an incrementally maintained membership-to-`ChildKey` index, so stale subjects safely miss
- republish a pre-admission pending request when parent adoption commits
- replace `pending_restart_shutdowns()` and its per-child record clone/lock scan
- maintain an incremental incomplete-child count and use it for O(1) shutdown completion checks
- cover subject publication, pre-admission adoption, same-batch arbitration, and exact terminal-count completion

This completes Part 0.4 and subsumes item 11 from #172.

## Test plan

- `./tools/dev cargo test -p shelterwood driver::tests::`
- pending restart-shutdown integration tests
- `./tools/dev just ci`
  - clippy with warnings denied
  - 523 nextest tests
  - doctests and rustdoc
  - packaged-crate verification
  - Nix formatting check

### DIFF
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -8,6 +8,7 @@
 
 use std::{
     any::Any,
+    collections::VecDeque,
     fmt,
     sync::{
         Arc, Mutex, MutexGuard, OnceLock, RwLock, Weak,
@@ -797,12 +798,23 @@ struct ScopeControl {
     epochs: ScopeEpochs,
     shutdown: Option<ScopeRequest>,
     force: Option<ScopeRequest>,
+    events: VecDeque<ScopeControlEvent>,
 }
 
 #[derive(Clone, Copy, Debug)]
 struct ScopeRequest {
     epoch: Epoch,
     consumed: bool,
+}
+
+/// One fact published by a scope control-plane transaction for its driver.
+///
+/// The payload is restart-stable identity rather than a mutable driver key.
+/// The driver resolves it through its incrementally maintained membership
+/// index, so stale events miss instead of addressing a replacement child.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum ScopeControlEvent {
+    RestartShutdown { membership: Membership },
 }
 
 /// Shared scope state follows two distinct synchronization regimes.
@@ -948,11 +960,24 @@ impl ScopeCell {
             .and_then(Weak::upgrade)
     }
 
-    fn set_parent(&self, parent: &Arc<ScopeCell>, _txn: &mut ObservationTxn<'_>) {
+    fn set_parent(&self, parent: &Arc<ScopeCell>, txn: &mut ObservationTxn<'_>) {
         *self
             .parent
             .lock()
             .unwrap_or_else(std::sync::PoisonError::into_inner) = Some(Arc::downgrade(parent));
+        let control = self.control.lock().expect("scope control mutex poisoned");
+        let pending_shutdown = control.shutdown.is_some_and(|request| {
+            !request.consumed && control.epochs.request_is_pending(request.epoch)
+        });
+        drop(control);
+        if pending_shutdown {
+            parent.publish_control_event_locked(
+                ScopeControlEvent::RestartShutdown {
+                    membership: self.member.membership(),
+                },
+                txn,
+            );
+        }
     }
 
     #[cfg(test)]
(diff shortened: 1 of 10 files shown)


ralexstokes (commented):
Adversarial review found one lost-event window, addressed in `78b5839`:

A `RestartShutdown` control event could arrive in a driver batch just before the corresponding child `Exited` event. The subject lookup succeeded, but `spawn_child` rejected the still-active old incarnation; because the new queue is edge-triggered, the expedite fact was then consumed permanently and the replacement waited for its ordinary backoff. The child runtime now retains that fact while the old incarnation is active and retries it immediately after exit, while still clearing it when drain/removal/terminality suppresses restart. A deterministic regression pins the cross-batch ordering with a 60-second ordinary backoff.

The branch was rebased onto the reviewed #181 head. Validation: `./tools/dev just ci` passes on the combined stack (524 nextest tests, clippy with warnings denied, doctests/rustdoc, API/runtime/exit/layering checks, packaged-crate verification, and Nix formatting). No further findings after re-review.

ralexstokes (commented):
Multi-agent review complete (8 finder angles → adversarial verification). Two confirmed correctness bugs — both fixable within this PR's design — plus residuals.

## Correctness — fixing on this branch

1. **`expedite_restart_shutdown` lost the old `MemberStage::Restarting` gate; a never-started child can be spawned out of order** (`events.rs:204`) — *confirmed independently by three finder angles.* The deleted `pending_restart_shutdowns()` scan required `matches!(stage, MemberStage::Restarting)`, which excluded never-started (Admitted-stage) children and left them to `progress_startup`'s strict in-order gating. The new path checks only suppression / pending target / terminal / disposing / active — all of which pass for a never-started child. Scenario: Ordered scope `[A, nested B]`; `set_admitted_children` runs before ordered startup spawns anything, so a user holding B's scope handle can call shutdown while the cursor is still on A; `request_shutdown` publishes a `RestartShutdown` event for B's *first* epoch (`pending_incarnation=true`), and the parent driver spawns B before A is ready — B's `Started` precedes A's readiness, and a pre-ready stop routes `StartupDisposition::Aborted` into `fail_startup`, deciding the scope's startup verdict out of turn. The existing test `pre_admission_restart_shutdown_does_not_expedite_the_following_incarnation` masks this by spawning the child before processing the event. **Fix:** restore the started-before gate in `expedite_restart_shutdown` (skip never-started children; the pending request stays latched on the scope cell for the child's ordered turn) + regression test.

2. **`handle_exit`'s retained-retry runs mid-batch, before later exits in the same batch can trip intensity** (`child.rs:977-988`) — *confirmed, with full trace.* The tail retry calls `expedite_restart_shutdown` synchronously while the rest of the arbitration batch is unprocessed, so with `[A's exit, B's exit]` in one wake where B's exit trips intensity, A's retry spawns a fresh incarnation (user factory runs, `Started` published) that the same batch immediately drains — exactly what `restart_shutdown_work`'s own comment forbids ("a child exit collected in the same wake must first get the chance to trip intensity or fail startup"). The pre-PR wake-start scan structurally deferred this case by one wake. Neither existing test covers the combination (`same_batch_intensity_exit_suppresses_real_expedited_factory` exercises only the arbitration path; the retry-path test has no second exit). **Fix:** defer the retry through arbitration — retain `(key, target)` on the runtime, drain into the pending batch as `restart_shutdown_work` at the next collection phase (classified after `ChildExit`, so the execution-time suppression re-check observes any drain the batch latched) + regression test.

## Hardening — folding into the fix commit

3. The incompleteness predicate `!is_terminal() || is_disposing()` is hand-copied at five sites across four files, and `(child_keys, incomplete_children)` is derived in three places (`driver.rs:292/664`, `tests/support.rs:111`, negation in `child.rs:550`, `shutdown.rs:135/153`) — the counter that `finish_if_ready` now depends on is maintained by convention. Adding `ChildRuntime::is_incomplete()` plus one shared index/count helper closes the drift surface.
4. `reclaim_child` (`removal.rs:128`) neither decrements nor asserts `incomplete_children` completeness; correct today only because every reclaim path runs post-terminalization. A `debug_assert!` of completeness at this choke point turns a future silent shutdown-hang into a loud failure.

## Residuals (tracked on #172)

- `take_control_events` acquires the tree-wide gate on every driver wake to observe an almost-always-empty queue; peek under the control mutex (race-safe: publication pulses `member.record` inside its txn) or fold into the existing per-wake gate transaction (`cells.rs:1710`).
- `control_event_work` makes "early" indistinguishable from "stale" (silent drop on `child_keys` miss); the arena-insert-before-publication ordering that prevents early events is unstated at the insertion sites (`events.rs:169`, `driver.rs:415-424`).
- `spawn_child` declines silently; `expedite_restart_shutdown` hand-mirrors its decline set before clearing the one-shot retained fact (`events.rs:195`).
- Pending-shutdown predicate duplicated between `set_parent` (`cells.rs:963`) and `has_pending_incarnation_shutdown` (`cells.rs:1721`); publish-to-parent event construction duplicated (`cells.rs:968` vs `1688`); `restart_shutdown_pending` cleared on three exit paths instead of set-once/take-once; `expedite_restart_shutdown` does up to four arena lookups for one key; `ScopeControl.events` VecDeque could be a `Vec` + `mem::take`; two driver tests hand-roll fixtures that `running_dynamic_fixture`/`start_admission` already provide.

Fixes for items 1–4 will be pushed to this branch; residuals will be rolled up on #172.

PR #184, opened 2026-08-10
Part 0.5/0.6: make disposal and mailbox intake structural

ralexstokes:
## Why

Part 0.5/0.6 of #172 moves two cleanup and intake invariants from repeated call-site convention into types:

- every raw-incarnation-owned user value must be destroyed through one panic-contained cleanup route
- bound mailbox waiters may exist only when a queue is full, and acceptance must have one implementation with displaced payloads destroyed after unlocking

Stacked on #182.

### Part 0.5

- add one contained raw-payload owner backed by a shared disposal/panic sink
- route continuations, queued offload events, timer keys/messages, and offload work through it
- contain the zero-budget offload work destructor and classify its panic as cleanup
- make bulk freeze independently contain every discarded payload
- split mailbox freeze and raw-resource freeze into separate epilogue panic boundaries

### Part 0.6

- replace independent mailbox status/waiter fields with typed binding states
- make `BoundState::Full` the only bound state that can own waiters
- centralize `submit` and `try_send` acceptance in `accept_locked`
- return latest-value displacement in a must-finish bundle whose completion happens after the mutex is released
- preserve promotion, withdrawal, freeze/rebind, and terminal teardown semantics

This completes Part 0.5/0.6 and subsumes all item 8 bullets plus item 32's acceptance/invariant bullets from #172.

## Test plan

- focused raw disposal and zero-budget offload tests
- focused mailbox state/acceptance tests
- `./tools/dev just ci`
  - clippy with warnings denied
  - 526 nextest tests
  - doctests and rustdoc
  - packaged-crate verification
  - Nix formatting check

### DIFF
--- a/crates/shelterwood/src/mailbox/cell.rs
+++ b/crates/shelterwood/src/mailbox/cell.rs
@@ -282,7 +282,9 @@ impl WaiterId {
 /// FIFO registrations with direct removal by a send operation.
 ///
 /// Monotonic keys are insertion order, so the first map entry is the oldest
-/// waiter. Keys are never reused, making stale cancellation ids harmless.
+/// waiter. Keys are never reused within one queue instance; a queue is only
+/// replaced when it is empty or by the absorbing Terminal state, so no
+/// registration outlives its queue and stale cancellation ids remain harmless.
 /// `u64::MAX` is a poison key and is never minted; exhaustion remains poisoned
 /// instead of wrapping back into the live id domain.
 pub(super) struct WaiterQueue<M> {
--- a/crates/shelterwood/src/mailbox/cell.rs
+++ b/crates/shelterwood/src/mailbox/cell.rs
@@ -25,6 +25,35 @@ enum BindingStatus {
     Terminal(Option<Incarnation>),
 }
 
+enum MailboxBinding<M> {
+    Unbound(WaiterQueue<M>),
+    Bound(BoundState<M>),
+    Frozen {
+        incarnation: Incarnation,
+        waiters: WaiterQueue<M>,
+    },
+    Terminal(Option<Incarnation>),
+}
+
+/// A bound mailbox either has no parked senders or is explicitly blocked.
+/// Only the blocked variant can own waiters, and it is constructed only when
+/// a queue mailbox has reached capacity.
+enum BoundState<M> {
+    Available(Incarnation),
+    Full {
+        incarnation: Incarnation,
+        waiters: WaiterQueue<M>,
+    },
+}
+
+impl<M> BoundState<M> {
+    fn incarnation(&self) -> Incarnation {
+        match self {
+            Self::Available(incarnation) | Self::Full { incarnation, .. } => *incarnation,
+        }
+    }
+}
+
 /// Which mailbox binding states may yield accepted messages.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 enum ReceiveMode {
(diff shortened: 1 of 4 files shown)


ralexstokes (commented):
Adversarial review completed at `37358017bf8d5d7eb0d81a6f3d1aa27e58853933`: no actionable findings.

I traced the raw ownership funnel through continuations, queued offload events/results, timer keys/messages (including replacement, clearing, one-shot delivery, and interval cloning), zero-budget and ordinary offload work, freeze, join, and the split epilogue panic boundaries. Stored ownership remains contained until an explicit delivery/execution transfer, and bulk cleanup independently contains each destructor.

I also traced the mailbox state machine through unbound bind promotion, capacity overflow, cancellation/withdrawal, live promotion, freeze/close/rebind, latest displacement, and terminal waiter detachment. Acceptance is centralized, user payload destruction occurs after unlocking, and every reachable bound waiter state is the full-queue variant.

Validation:
- `./tools/dev just ci` — passed (527 tests plus clippy, docs, API/layering checks, packaging, nixfmt)
- release-mode mailbox cell tests — 8 passed
- release-mode multi-panicking-payload freeze regression — passed
- GitHub CI is green on this head

ralexstokes (commented):
Multi-agent review complete (8 finder angles → adversarial verification). Verified findings, split by disposition.

## Correctness — fixing on this branch

1. **`bind()` mutates before validating; release-mode contract violations now poison-cascade** (`mailbox/cell.rs:788-802`) — *confirmed-latent, found independently by three finder angles.* The `mem::replace` to `Bound(Available)` runs before (a) the non-Unbound check — a rebind without close now hits `unreachable!` in release, where base degraded gracefully — and (b) `state.kind.expect("a bound mailbox is configured")` — an unconfigured bind now panics in release where base parked senders. Both panics fire with the mailbox mutex held and state half-mutated: displaced waiters drop un-terminated during unwind and the poisoned mutex converts every later `send`/`try_send`/`withdraw` into a panic across all sender threads. Contract-gated today (driver configures before bind, closes before rebind), and the PR's own contract tests are `#[cfg(debug_assertions)]`, so the release escalation is an unintended byproduct of moving waiters into `MailboxBinding`. **Fix:** validate before mutating, and drop the guard before panicking so a violation stays a driver-thread panic instead of a poison cascade.

2. **Offload completion arm bypasses the PR's own `RawDisposal::record` funnel** (`raw.rs:1418`) — *confirmed.* The arm still does raw `panic.record(payload); events.signal.pulse();` while `RawDisposal::record` (introduced here) guards the pulse with `catch_panic` and retains a pulse panic. Verified same `Arc<PanicSlot>` and same `Signal`, so `disposal.record(payload)` is drop-in compatible. A panicking pulse currently unwinds through the spawned offload task instead of being retained as cleanup evidence.

3. **`WaiterQueue` doc rot** (`mailbox/cell.rs:283-287`) — "Keys are never reused, making stale cancellation ids harmless" is now per-queue-instance only: fresh `WaiterQueue::default()` instances (park, freeze/close/prepare_termination from Available) restart `next_id` at zero. No aliasing is reachable — every replacement happens with the predecessor provably empty or absorbing-Terminal — but harmlessness now rests on that emptiness invariant, not key monotonicity. **Fix:** amend the doc to state the actual invariant.

4. **`EventQueue::default`/`TimerStore::default` mint orphan disposal funnels** (`raw.rs:342/480`) — fresh `PanicSlot` + watcherless `Signal` that nothing can ever drain; used only from `#[cfg(test)]` today, but they are ordinary impls on production types and the pre-PR idiom constructed exactly `EventQueue::default()` at the site now using `::new(disposal)` — a one-line future regression that silently disconnects containment from any observer. **Fix:** gate both `Default` impls `#[cfg(test)]` (and drop the dead `signal` local clone in `EventQueue::default`).

## Residuals (tracked on #172)

- `freeze`/`close`/`prepare_termination` each hand-roll the same waiter-extraction `mem::replace` with a transient placeholder `Terminal` binding and bespoke `unreachable!` arms; a single transition helper on `MailboxBinding`/`MailboxState` removes the triplication and the placeholder state (`cell.rs:829-905`).
- `accept_locked` collapses four rejection causes into a bare `Err(M)`; `try_send` pre-derives one of them and `submit` assumes "park" for the rest — a reason-carrying rejection enum would make the mapping explicit (`cell.rs:935`, `598`).
- `BindingStatus` is a derived shadow of `MailboxBinding`: mutation paths gate on `status()` then re-match `binding` with `unreachable!` arms asserting the gate held (`cell.rs:190`).
- Aliased fields: `RawResources.panic` ≡ `disposal.panic`, `EventQueue.signal` ≡ `disposal.signal` — two names for one slot, an invariant held only by the constructors (`raw.rs:338/881`).
- `Contained` stores a `RawDisposal` clone per element (continuations, queued events, timers — 3 clones per timer arm via the wrap/unwrap/rewrap in `TimerStore::replace`); wrap-at-drain or a `Contained::map` would keep the containment at zero per-message cost (`raw.rs:363/526/1153`).
- Zero-period `set_interval` re-implements the containment wrapping inline instead of routing through the `TimerStore` funnel; the explicit `drop(key); drop(message);` reads as load-bearing ordering but isn't (`raw.rs:1188-1194`).
- `SharedOffloadState::new` decomposes the caller's `RawDisposal` only to reassemble it; `accept_locked`/`promote_waiter_queue` duplicate the mint+insert envelope block; `Acceptance::finish` drops a displaced Latest envelope raw in the sender's context (pre-existing, but now the one displaced-payload path outside the disposal discipline).
- Behavior note (intended per the new tests, recording for the changelog): a zero-deadline `offload()` destructor panic is now contained and resurfaces later via `resume_pending_panic` reclassified as cleanup, where base propagated it synchronously out of `offload()` — callers treating that as an atomicity boundary see partial work committed before the panic is observed.

Fixes for items 1–4 will be pushed to this branch; residuals will be rolled up on #172.

PR #185, opened 2026-08-10
Part 0.7/0.8: make ownership and counters structural

ralexstokes:
Implements the remaining Part 0 work from #172.

Stacked on #184.

## Part 0.7 — structural ownership

- make membership minting carry the sole incarnation counter and evict exact terminal identities
- select fused versus split dynamic admission ownership when constructing the slot
- transfer scope plans into the runtime through a consuming handoff
- represent internal scope lifecycle state with only live, reachable variants
- reserve parent-capable cancellation tokens for internal construction
- resolve mailbox and readiness policies into types that cannot retain inherited or incomplete states

Terminalized remove/re-add identities are deliberately incomparable, matching the Part 0 adjudication's fail-closed rule.

## Part 0.8 — poisoned monotonic counters

- share local and atomic poisoned-counter primitives across identity fencing, mailbox acceptance and waiters, timers, lifecycle sequencing, arena/deadline handles, and snapshot generations
- permanently poison counters at exhaustion so the maximum value is never minted
- document the remaining diagnostic-only counters that intentionally saturate

## Validation

- `./tools/dev just ci`
- 530 tests passed

### DIFF
--- a/crates/shelterwood/src/cancellation.rs
+++ b/crates/shelterwood/src/cancellation.rs
@@ -3,7 +3,7 @@
 use crate::runtime::{self, Latch};
 
 /// A library-owned cancellation token.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct CancellationToken {
     primary: Latch,
     secondary: Option<Latch>,
@@ -14,14 +14,6 @@ impl CancellationToken {
         Self {
             primary: latch,
             secondary: None,
-        }
-    }
-
-    pub(crate) fn child(&self, cancellation: Latch) -> Self {
-        debug_assert!(self.secondary.is_none());
-        Self {
-            primary: self.primary.clone(),
-            secondary: Some(cancellation),
         }
     }
 
@@ -41,19 +33,50 @@ impl CancellationToken {
     }
 }
 
+/// Internal capability that alone may derive a locally cancellable token.
+#[derive(Clone, Debug)]
+pub(crate) struct ParentCancellationToken {
+    primary: Latch,
+}
+
+impl ParentCancellationToken {
+    pub(crate) fn from_latch(primary: Latch) -> Self {
+        Self { primary }
+    }
+
+    pub(crate) fn token(&self) -> CancellationToken {
+        CancellationToken::from_latch(self.primary.clone())
+    }
+
+    pub(crate) fn child(&self, cancellation: Latch) -> CancellationToken {
+        CancellationToken {
+            primary: self.primary.clone(),
+            secondary: Some(cancellation),
+        }
+    }
+
+    pub(crate) fn is_cancelled(&self) -> bool {
+        self.primary.is_fired()
+    }
+
+    pub(crate) async fn cancelled(&self) {
+        self.primary.fired().await;
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use std::time::Duration;
 
     use crate::runtime::{self, JoinOutcome, Latch, Timeout};
 
-    use super::CancellationToken;
+    use super::ParentCancellationToken;
 
     #[crate::runtime::test]
     async fn local_cancellation_cancels_only_the_derived_token() {
         let primary = Latch::default();
         let local = Latch::default();
-        let supervisor = CancellationToken::from_latch(primary.clone());
+        let supervisor = ParentCancellationToken::from_latch(primary.clone());
         let operation = supervisor.child(local.clone());
 
         assert!(local.fire());
@@ -70,7 +93,7 @@ mod tests {
     async fn supervisor_cancellation_cancels_the_derived_token() {
         let primary = Latch::default();
         let local = Latch::default();
-        let supervisor = CancellationToken::from_latch(primary.clone());
+        let supervisor = ParentCancellationToken::from_latch(primary.clone());
         let operation = supervisor.child(local.clone());
 
         assert!(primary.fire());
@@ -88,7 +111,8 @@ mod tests {
         for _ in 0..128 {
             let primary = Latch::default();
             let local = Latch::default();
-            let operation = CancellationToken::from_latch(primary.clone()).child(local.clone());
+            let operation =
+                ParentCancellationToken::from_latch(primary.clone()).child(local.clone());
             let waiter = runtime::spawn(async move {
                 operation.cancelled().await;
             });
--- a/crates/shelterwood/src/cells.rs
+++ b/crates/shelterwood/src/cells.rs
@@ -12,23 +12,23 @@ use std::{
     fmt,
     sync::{
         Arc, Mutex, MutexGuard, OnceLock, RwLock, Weak,
-        atomic::{AtomicBool, AtomicU64, Ordering},
+        atomic::{AtomicBool, Ordering},
     },
     time::Instant,
 };
 
 use crate::{
-    ChildId, Exit, Incarnation, Intensity, Mailbox, Membership, Readiness, RestartCount, Strategy,
+    ChildId, Exit, Incarnation, Intensity, Membership, Readiness, RestartCount, Strategy,
     TotalRestarts,
     admission::{RemoveOutcome, ReserveError},
     engine::{Epoch, MembershipStatus, RequestTarget, ScopeEpochs, ScopeState},
     exit::{StartupError, StopReason},
-    identity::ScopeIdentity,
+    identity::{AtomicPoisonedCounter, IncarnationCounter, MintedMembership, ScopeIdentity},
     observe::{
         ChildSnapshot, ChildState, LifecycleEvent, LifecycleEventKind, LifecycleEvents,
         LifecycleHub, LifecycleSeq, ScopeKind, ScopeSnapshot, SnapshotHub, SnapshotReceiver,
     },
-    policy::{ResolvedCommonOptions, ScopeFlavor},
+    policy::{ResolvedCommonOptions, ResolvedMailbox, ScopeFlavor},
     runtime::{self, Latch},
 };
 
@@ -52,7 +52,7 @@ pub(crate) trait MailboxTermination: Send {
 pub(crate) trait MailboxControl: fmt::Debug + Send + Sync {
     /// Installs the declaration-time mailbox policy before the first bind.
     /// Reconfiguration may only repeat the same resolved policy.
-    fn configure(&self, mailbox: Mailbox);
+    fn configure(&self, mailbox: ResolvedMailbox);
     /// Makes one incarnation live after configuration and prior-close cleanup.
     /// A bind after terminal preparation is deliberately ignored because
     /// terminality wins that race permanently.
@@ -188,6 +188,7 @@ pub(crate) struct MemberCell {
     id: ChildId,
     membership: Membership,
     rebased_membership: OnceLock<Membership>,
+    incarnations: Mutex<Option<IncarnationCounter>>,
     record: runtime::WatchSender<MemberRecord>,
     // Guards only a gate-pointer swap, so no torn state is possible; every
     // access deliberately tolerates poisoning (mirroring
@@ -232,7 +233,8 @@ impl fmt::Debug for MemberMailbox {
 }
 
 impl MemberCell {
-    pub(crate) fn new(id: ChildId, membership: Membership) -> Arc<Self> {
+    pub(crate) fn new(id: ChildId, identity: MintedMembership) -> Arc<Self> {
+        let (membership, incarnations) = identity.into_pair();
         let (record, _) = runtime::watch(MemberRecord {
             stage: MemberStage::Reserved,
             incarnation: None,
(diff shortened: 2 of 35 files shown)


ralexstokes:
Adversarial review completed. I found and fixed three substantive issues in the rebased PR head `acc7782`:

1. The normative spec, guides, packaged docs, and several test/comment names still described remove/re-add memberships as superseding even though the adjudicated implementation deliberately makes terminalized lineages incomparable. Those contracts now consistently document fail-closed incomparability.
2. Scope epochs still carried a bespoke exhaustion rule instead of the shared poisoned-counter policy. `Epoch::successor`, the local counter, and the atomic counter now use one `PoisonedCounter::minted_after` decision, preserving the rule that `u64::MAX` is poison and is never minted.
3. GitHub's stacked synthetic merge exposed a newer startup regression test that still mutated the removed `ScopePlan::armed` field. I rebased PR 184 onto the updated Part 0.4 head, rebased PR 185 onto that rewritten Part 0.5/0.6 head, and converted both the integrated and existing startup tests to the consuming `take_for_runtime().finish_transfer()` handoff.

I also re-audited structural ownership transfer, dynamic admission ownership, identity eviction, resolved policy types, and atomic mint races; I found no further actionable defects.

Validation: `./tools/dev just ci` passes on both rewritten PR heads (528 tests for PR 184 and 531 tests for PR 185), including docs, public-API reachability, packaged-crate verification, and formatting/lint checks. GitHub Actions passes on both PR 184 and PR 185.

crates/shelterwood/src/driver/child.rs
ralexstokes, line 1: Line 141: **discharge_child_terminality's never-started non-resident branch terminalizes the slot without evicting its child identity — the only terminalization path in this PR that skips evict_child_identity, so the terminal lineage stays retained and a rebuilt same-id child mints a comparable (superseding) membership.**

Nested scope with declared children [a, b]: converting b in run_scope_incarnation's from_plan loop panics (e.g. poisoned incarnation-counter mutex, as in scope_plan_conversion_panic_terminalizes_every_child). Converted child a's Obligation fallback runs this branch (never started, not yet resident), calls slot.terminalize_never_started() and returns without root.evict_child_identity(). On the supervisor restart, BuilderCore::lower's adopt_or_mint_membership hits the leaked Occupied entry and mints an ordered successor: replacement.membership().supersedes(terminalized_predecessor) == true, violating the PR's own SPEC contract (terminalization evicts the lineage; re-adds are incomparable in both directions). Every sibling path pairs the two calls (cells.rs:1337, plan.rs:387/428, admission_control.rs take_terminal_reservation, driver.rs:414/431); the deeper fix is performing eviction inside the terminal transition itself rather than at each caller.

crates/shelterwood/src/plan.rs
```diff
@@ -384,6 +384,7 @@ impl BuilderCore {
     fn terminalize(&self) {
         for slot in &self.slots {
             slot.terminalize_never_started();
+            self.root.evict_child_identity(&slot.member);
```
ralexstokes: **BuilderCore::terminalize evicts identities from self.root, but lower(root_override) adopts/mints the slots' lineages into the override (stable) scope's identity map — failed nested lowering evicts from the wrong ScopeIdentity and leaves stale lineages in the stable scope.**

run_nested_tree_with_epoch lowers with Some(stable_scope). The adopt loop (plan.rs:317-335) inserts each slot's lineage into stable_scope.child_identity; if a later slot fails with LowerError::IdentityExhausted (or a panic unwinds after partial adoption), BuilderCore drops armed and terminalize() calls self.root.evict_child_identity(&slot.member) — the builder's throwaway root, not the override. For rebased members the builder-map evict is also a guaranteed no-op (issued() no longer matches the rebased membership). The stable scope retains lineages for terminalized never-started members, so the next restart's rebuild mints comparable, superseding memberships for terminal predecessors, and repeated failed lowerings keep advancing the stale counters toward per-id exhaustion.

crates/shelterwood/src/plan.rs
```diff
@@ -403,27 +404,81 @@ pub(crate) struct ScopePlan {
...
 
-impl Drop for ScopePlan {
-    fn drop(&mut self) {
-        if !self.armed {
-            return;
-        }
-        for child in &self.children {
+struct ScopePlanTerminality;
+
+pub(crate) struct RuntimeScopePlan {
```
ralexstokes, line 438: **RuntimeScopePlan is a field-for-field copy of ScopePlan with a duplicated Drop impl, forcing take_for_runtime to clone config/defaults and Arc::clone the root on every scope incarnation.**

Both structs have identical fields and identical Drop bodies delegating to terminalize_plan, and any new plan field must be added in three places (both structs plus take_for_runtime). take_for_runtime pays ScopeConfig/ResolvedDefaults clones per scope start (including every nested-scope restart) because Drop blocks destructuring; a future heap field silently becomes a per-incarnation allocation. Simpler: one shared inner struct (or a newtype over destructured fields) — the Option<ScopePlanTerminality> token already makes the handoff exactly-once.

crates/shelterwood/src/identity.rs
```diff
@@ -267,62 +382,133 @@ impl ScopeIdentity {
...
-    /// pre-spawn handles retain their identity. Later declarations of that id
-    /// (including declarations produced after a scope restart) mint from the
-    /// retained counter and therefore supersede their predecessors.
+    /// The first declaration of an untracked id donates its already-minted
+    /// lineage so pre-spawn handles retain their identity. If the scope still
+    /// tracks that lineage, a later provisional declaration mints its ordered
+    /// successor. Terminalization evicts the lineage, so an ordinary later
+    /// remove-and-re-add or post-restart rebuild donates a fresh, incomparable
+    /// identity instead.
     pub(crate) fn adopt_or_mint_membership(
```
ralexstokes, line 412: **adopt_or_mint_membership returns Option<Option<MintedMembership>>, where the two Option layers carry unrelated meanings (outer None = id exhausted, inner None = provisional adopted unchanged).**

A caller collapsing the layers with ?/flatten compiles fine and silently converts 'adopted' into 'exhausted' or vice versa; the call site in plan.rs binds the value as `rebased` and the PR's own tests had to downgrade from assert_eq to matches!(.., Some(None)). A dedicated enum (e.g. AdoptOutcome::{Adopted, Minted(MintedMembership)}) inside the Option names both cases and cannot be layer-swapped silently.

crates/shelterwood/src/identity.rs
```diff
@@ -185,22 +271,30 @@ impl FenceCounter {
...
-        };
-        self.current = generation;
+        let generation = Generation::new(self.generations.mint()?)?;
         Some(Fence {
             lineage: self.lineage,
             generation,
         })
     }
+
+    fn issued(&self) -> Fence {
```
ralexstokes, line 286: **FenceCounter::issued() reconstructs the last-issued generation from counter state with a hand-rolled poison clamp (current == u64::MAX => u64::MAX - 1), re-encoding PoisonedCounter's domain rule outside the new shared primitive and reaching into its private field.**

The clamp encodes an implicit invariant (poison is only reached after minting MAX-1) that from_fence/near_exhaustion or a future counter change can break, after which evict() silently mismatches and retains a terminal lineage — re-enabling the supersedes-after-terminal bug this PR removes. Recording the issued Fence alongside the counter (or a PoisonedCounter::last_issued() next to minted_after) removes the arithmetic reconstruction; even keeping the logic, Generation::new(self.generations.current().min(u64::MAX - 1)) is one expression with no field reach-in.

crates/shelterwood/src/cells.rs
```diff
@@ -274,6 +278,27 @@ impl MemberCell {
         self.rebased_membership
             .set(membership)
             .expect("a reservation can be rebased at most once");
+        *self
+            .incarnations
+            .lock()
+            .expect("incarnation counter mutex poisoned") = Some(incarnations);
+    }
+
+    pub(crate) fn take_incarnation_counter(&self) -> IncarnationCounter {
```
ralexstokes, line 291: **The 'sole incarnation counter' invariant is enforced by Mutex<Option<IncarnationCounter>> plus take().expect() — a runtime panic check in a PR whose stated goal is structural ownership.**

take_incarnation_counter panics on a second take, and rebase_membership silently re-fills the Option, so an ordering bug (take before rebase, or two runtimes racing) surfaces as a driver-path panic rather than a compile error; the MintedMembership pairing is split by into_pair and the counter parked back in shared mutable state. Moving the IncarnationCounter by value through ChildPlan/ChildConstruction into ChildRuntime::from_plan would make exactly-once issuance ownership instead of an expect.

crates/shelterwood/src/identity.rs
```diff
@@ -267,62 +382,133 @@ impl ScopeIdentity {
...
         }
     }
 
-    pub(crate) fn incarnation_counter(&self, membership: Membership) -> IncarnationCounter {
-        // Membership already supplies the complete incarnation lineage. The
-        // per-membership counter therefore needs only an ordered generation.
-        IncarnationCounter {
-            membership,
-            counter: FenceCounter::sequence(),
+    pub(crate) fn evict(&mut self, id: &ChildId, membership: Membership) {
```
ralexstokes, line 436: **ScopeIdentity::evict clones the ChildId (Arc refcount RMW) via the entry API for a read-mostly lookup, and every terminalization loop acquires the scope-wide child_identity mutex once per child.**

evict runs on every child terminalization and every dynamic-reservation cancel/remove/close; entry(id.clone()) pays a refcount round-trip even when the entry is absent or the issued fence mismatches. terminalize_plan, BuilderCore::terminalize, and DynamicControl::close additionally lock child_identity once per child inside a single teardown/observation transaction. Cheaper: get(id).is_some_and(..) then remove(id) (borrowing the id), and one guard held across each loop.

crates/shelterwood/src/mailbox/cell.rs
```diff
@@ -756,14 +750,10 @@ impl<M> MailboxCell<M> {
 }
 
 impl<M: Send + 'static> MailboxControl for MailboxCell<M> {
-    fn configure(&self, mailbox: Mailbox) {
+    fn configure(&self, mailbox: ResolvedMailbox) {
```
ralexstokes, line 755: **MailboxKind {Queue(NonZeroUsize), Latest} is now a variant-for-variant duplicate of the new policy::ResolvedMailbox, and configure() is a pure identity mapping between them.**

Any new mailbox policy variant must be added and mapped in two enums; storing ResolvedMailbox directly in MailboxState deletes the shadow enum and the mapping.

crates/shelterwood/src/identity.rs
```diff
@@ -267,62 +382,133 @@ impl ScopeIdentity {
...
             Entry::Occupied(mut entry) => entry.get_mut().mint().map(Membership),
             Entry::Vacant(entry) => {
                 let mut counter = Self::fresh_counter()?;
                 let membership = counter.mint().map(Membership)?;
                 entry.insert(counter);
                 Some(membership)
             }
-        }
+        }?;
+        Some(MintedMembership {
```
ralexstokes, line 395: **The MintedMembership { membership, incarnation_counter: IncarnationCounter { membership, counter: FenceCounter::sequence() } } literal is constructed inline three times (mint_membership, adopt_or_mint_membership, IncarnationCounter::fixture).**

If the counter pairing rule changes (e.g. seeding the generation), three sites must change together or silently diverge. A MintedMembership::new(membership) constructor used by all three keeps the pairing invariant in one place.

crates/shelterwood/src/cancellation.rs
```diff
@@ -41,19 +33,50 @@ impl CancellationToken {
     }
 }
 
+/// Internal capability that alone may derive a locally cancellable token.
+#[derive(Clone, Debug)]
+pub(crate) struct ParentCancellationToken {
```
ralexstokes, line 38: **ParentCancellationToken reimplements is_cancelled/cancelled instead of delegating to the CancellationToken it can already mint; the only genuinely new capability is child().**

Two parallel implementations of the primary-latch firing semantics must stay in sync. Holding a primary-only CancellationToken internally (token() clones it; is_cancelled/cancelled delegate) leaves child() as the sole new code.

crates/shelterwood/src/driver/child.rs
ralexstokes: Fixed in 88d3797: eviction now happens inside `SlotCell::terminalize_never_started{,_locked}` (which take the owning scope), so the never-started non-resident fallback in `discharge_child_terminality` evicts via `completion.root` like every other path; regression test `conversion_unwind_evicts_never_started_child_identities` drives the conversion panic, restarts the rebuild against the same scope, and asserts the re-minted memberships are incomparable in both directions.

crates/shelterwood/src/plan.rs
ralexstokes: Fixed in 3dd8a71: `BuilderCore` records the adopting (override) root when `lower` begins adoption, and `terminalize` evicts each slot from that scope's identity map instead of the throwaway builder root (not-yet-adopted slots no-op there via the fence-mismatch check); regression test `failed_override_lowering_evicts_adopted_lineages_from_the_override` fails a nested lowering with `IdentityExhausted` after partial adoption and asserts a rebuild against the stable scope mints an incomparable membership.

Issue #206, opened 2026-08-11
Double Stopped publication with divergent reason on root-driver panic mid-drain


ralexstokes:
**Severity: low — edge-reachable only via root-driver panic; possibly adjudicate-by-design.** (Internal review round finding, medium confidence.)

`finish_live_root_incarnation`'s no-live-epoch fallback (`cells.rs:1688-1701`, `2007-2031`) calls `publish_stopped_locked` without the already-`Stopped` guard that `ScopeRuntime::drop` has (`driver.rs:269`).

Scenario: the root driver panics mid-drain after its scope began draining with reason `Finished`. Unwind runs `ScopeRuntime::drop`, which publishes `Stopped{Finished}` via `finish_incarnation` (member not terminal, streams stay open, epochs go `Idle`). The join monitor (`driver.rs:100-103/141-148`) then calls `finish_live_root_incarnation(ShutdownRequested, panic_exit)`, hits the `else` branch, and publishes a **second** `Stopped{ShutdownRequested}` — overwriting the record's reason and emitting a second `ScopeState::Stopped` lifecycle event with a different reason into still-open streams before closing them.

The final verdict (panic exit, `ShutdownRequested`) is arguably the intended one, and this is reachable only on a root-driver panic/cancel — so this may be adjudicated as acceptable. If not: cheaply mirror the drop path's "already `Stopped`" guard for the record/event (keeping the member terminalization and stream closure), so the lifecycle stream ends with exactly one `Stopped`.

Producers of the double-publish are only `ScopeRuntime::drop` and the two join monitors; a cells.rs-side guard needs no driver change.

ralexstokes:
Fixed by PR #218 (merged). Adjudicated beyond the original first-publication-wins approach: stopped publication is now monotone in a single precedence lattice (Finished < IntensityTripped < StartupFailed < ShutdownRequested < NeverStarted, StopPrecedence in exit.rs, shared with begin_drain). Root-driver panic mid-drain upgrades Finished to the join monitor's ShutdownRequested (the verdict this issue identified as intended), and the B.6 never-started twin invariant holds in both arrival orders. SPEC B.4/B.6 amended accordingly; all five repo-wide collision sites verified to resolve to the lattice verdict.

Issue #214, opened 2026-08-11
Correctness-first pass: reducer core, crate split, spec recast (tracking)


ralexstokes:
Tracking issue for the correctness-first pass decided 2026-08-11. Standing constraints for everything below: **no users — public-API breaks are acceptable; correctness and verifiability outrank compatibility; SPEC.md may be hardened or rewritten wherever that makes semantics cleaner** (extends the #116 §4 adjudication protocol). Findings inventory: #200–#213 plus the #191/#192 comment addenda.

The destination: a pure synchronous supervision reducer (`step(&mut SupervisorState, Event, &mut EffectSink)`) in a tokio-free crate, with the spec recast as numbered invariants over that transition system, a hard "no user behavior under locks" boundary in the concurrent substrate, and loom covering what the reducer cannot.

Two design rules to write into the spec as part of this pass (both resolve tensions found during review):

- **Latches are sampled inputs.** Removal/shutdown latches stay synchronously writable from the caller's observation gate (that synchronous latching is what defends §11's "suppressed from that instant" — a correctness property). The reducer reads them as inputs sampled at step entry; nothing mutates reducer state behind its back. This is how the engine already models shutdown requests — name it and enforce it.
- **Structural over temporal arbitration.** Outcome rules are decided by which transition won, never by clock comparison (the existing `TimedOut` rule stays; see #213's item-6 comment). New rules follow the same shape — it is the more verifiable discipline.

---

## Stage 0 — land the confirmed bugs as point fixes (no architecture gate) — **done**

All small, all with verified fix sketches; the repro/pinning tests carry into any later shape.

- [x] #200 startup hang: `progress_startup()` after removal (interim: at the `finalize_removal` site; stage 1 moves it to the loop epilogue) + both repro tests + SPEC sentence on aggregate shrink during `Starting`
- [x] #201 `shutdown_and_wait` fence: `scope_settled` predicate in cells, used by all three `shutdown.rs` checks; record the two follow-up adjudications in the issue
- [x] #202 waker drops under locks: move-out-then-drop-after-unlock at both sites + hostile-waker-**Drop** test
- [x] #203 `wait_for_child` Send: eager-convert wrapper (breaking-friendly shape fine) + `RcId` conformance probe
- [x] #205 disposal shut-down-pool window: `strong_count` fallthrough to the fallback thread + the #192 doc sentence
- [x] #206 double `Stopped`: idempotence guard inside `publish_stopped_locked` (publisher-side, not call-site)
[...]

PR #218, opened 2026-08-11
Make stopped publication idempotent

ralexstokes:
## Summary

- resolve competing stopped publications through a single precedence lattice instead of arrival order: `Finished < IntensityTripped < StartupFailed < ShutdownRequested < NeverStarted` (`StopPrecedence` in `exit.rs`, the one source of truth)
- `publish_stopped_locked` commits only when the incoming reason strictly outranks the recorded one; an upgrade republishes the record, snapshot, and a corrected `ScopeState` edge, while equal or weaker verdicts are idempotent repeats that mutate nothing
- `ScopeLifecycle::begin_drain` joins through the shared lattice (its private `DrainReasonPrecedence` copy is deleted), so a drain verdict and a published verdict can never resolve the same pair in opposite directions
- drop `ScopeRuntime::drop`'s already-`Stopped` guard: the publisher is the idempotence point, but epoch retirement is not idempotent and has no other owner
- regression tests pin the no-live-epoch branch and all four collision shapes (upgrade in both directions, suppression, exact duplicate)

## Root cause

Two owners can each reach a stop verdict for one incarnation — a driver's drain epilogue, the join monitor's fallback after that driver panicked or was cancelled, a never-started terminalization — and the publisher resolved them by arrival order, picking a winner for reasons unrelated to which verdict is authoritative.

An earlier revision of this PR replaced that with "first publication wins". Adversarial review showed first-wins gets two cases backwards: a root driver that dies mid-drain reported the abandoned drain's `Finished` instead of the join monitor's `ShutdownRequested` (the verdict #206 identified as intended, and the opposite of `begin_drain`'s own monotone upgrade rule), and a never-started terminalization only won when it arrived first, breaking B.6's twin invariant in the other arrival order.

## Adjudication (recorded per the spec-adjustment protocol)

- Stopped publication is **monotone in the precedence lattice**, not first-wins and not last-wins. `NeverStarted` ranks top because B.6 makes it the mandated scope-state twin of a never-started membership terminalization — it must win in both arrival orders. Below it, the order is exactly the pre-existing drain lattice.
- **SPEC B.6**: records the lattice; the never-started twin is restated as an invariant the lattice preserves.
- **SPEC B.4**: amended — a non-final `Stopped` is followed either by the next incarnation's `Starting` or by a strictly-higher-precedence `Stopped` for the same incarnation; the final `Stopped` is the one no restart and no precedence upgrade follows. Rationale: a lifecycle stream must not end on an event that disagrees with the final record; a corrected final event beats a single-but-wrong one.
- Residual (documented in the publisher's doc comment): a subscriber attaching between a publication and a later suppressed no-op repeat can still see closure without an event — the same pre-existing class as the stale-epoch branch; its initial snapshot is already terminal.

Instrumented whole-suite sweep after the change: exactly five collision sites exist repo-wide, and each resolves to the lattice verdict (`NeverStarted` in both arrival orders for the twin cases, `ShutdownRequested` over `Finished` for the mid-drain panic, pure suppression for the exact duplicate). Removing the `ScopeRuntime::drop` guard introduced no new collision sites.

## Validation

- `./tools/dev cargo nextest run -p shelterwood terminalization`
- `./tools/dev cargo test -p shelterwood root_driver_panic_mid_drain_upgrades_to_the_join_monitor_verdict`
- `./tools/dev just ci` (580 tests plus formatting, clippy, docs, layering/API enforcement, and packaging) — exit 0

Refs #206

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -3301,7 +3301,11 @@ Ordering and delivery contract:
   descendant memberships). `Stopped` is an incarnation edge, not a
   closure signal: only membership terminality ends the stream (closure
   rule below), so every non-final `Stopped` is followed on the same
-  sequence by the next incarnation's `Starting`, and the positive
+  sequence either by the next incarnation's `Starting` or by a
+  strictly-higher-precedence `Stopped` for the *same* incarnation
+  (B.6's stop-reason lattice — a bounded, monotone correction, never a
+  repeat of an equal verdict), and the final `Stopped` is the one
+  followed by neither. The positive
   terminality signals are the membership edges — the parent's `Removed`
   for this scope child, or this stream's own closure, always preceded
   by the final event. Snapshot receivers hold the last published snapshot through the
@@ -3350,7 +3354,8 @@ Ordering and delivery contract:
   scope's final event — closure is always preceded by one (under
   sequence exhaustion, by the final `Lagged`), and per the restart
   rule above a `Stopped` alone is not closure: the final `Stopped` is
-  the one no restart follows. For a scope membership that never spawns
+  the one no restart and no precedence upgrade follows. For a scope
+  membership that never spawns
   (a declaring tree dropped unspawned, a withdrawn or rejected
   insertion, §3.2), that terminal event is
   `ScopeState { Stopped { reason: NeverStarted } }` (B.6), published at
@@ -3480,7 +3485,9 @@ ScopeSnapshot   { state: Unstarted                              // membership ex
                                                                 //   tree, rejected/withdrawn insertion,
                                                                 //   removal before first spawn, startup-
                                                                 //   aborted ordered sibling (§6) — the
-                                                                //   scope-state twin of §7's exit
+                                                                //   scope-state twin of §7's exit, an
+                                                                //   invariant the stop-reason lattice
+                                                                //   below preserves in either order
                   kind: Ordered | Dynamic, strategy (ordered only), intensity,
                   total_restarts: TotalRestarts,         // charges per §9.2 — group respawns count
                   lifecycle_seq: LifecycleSeq,           // aligns events with snapshots (§12)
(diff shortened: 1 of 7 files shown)


ralexstokes:
## Adversarial review — issue #206

Reviewed PR head `ed8a039d487569c9e74be1245930842993be9522` against #206, the Stage 0 requirement in #214, the full `origin/main...HEAD` diff, and both commits.

### Areas attacked

- publisher-side guard placement versus a call-site workaround
- record, snapshot, and lifecycle-event idempotence
- first `Stopped` reason authority
- continued member terminalization and stream closure on the fallback
- live-, stale-, and no-live-epoch behavior
- observation-gate locking and deferred watcher notification ordering
- fidelity to the root-driver panic-mid-drain sequence

### Confirmed finding and fix

The original regression constructed the two cell calls directly, so it did not prove that a panicking root driver actually runs `ScopeRuntime`'s synchronous unwind epilogue before the join monitor's no-live fallback.

Commit `ed8a039` replaces that setup with an actual panicking driver task carrying `Draining(Finished)`. The test now verifies that:

1. `ScopeRuntime::drop` retires the epoch and publishes the first `Stopped(Finished)`;
2. join classification supplies the panicked terminal exit;
3. `finish_live_root_incarnation(ShutdownRequested, ...)` still terminalizes the member and closes both observation streams;
4. the retained reason remains `Finished`, with exactly one stopped lifecycle event and no duplicate snapshot edge.

### Review verdict

No remaining correctness findings. The guard is inside `publish_stopped_locked`, so all publisher paths share the rule. It suppresses only an already-published scope projection/event; terminal membership publication remains active and caller-owned closure still runs. Epoch ownership is released before deferred pulses, stale epochs cannot rewrite a newer incarnation, and a fresh incarnation resets the projection away from `Stopped`.

### Checks

- `./tools/dev cargo test -p shelterwood root_driver_panic_mid_drain_preserves_the_first_stopped_publication -- --nocapture`
- `./tools/dev cargo test -p shelterwood driver::tests::terminalization` — 14 passed
- `./tools/dev cargo test -p shelterwood driver::tests::observation` — 25 passed
- `./tools/dev cargo test -p shelterwood scope_plan_conversion_panic_terminalizes_every_child`
- `./tools/dev just ci` — 576/576 tests plus fmt, clippy, doctests/docs, enforcement, and packaging
- `./tools/dev just ci-nix` — all flake checks passed

PR intentionally remains draft for the parent orchestration gate.

Issue #205, opened 2026-08-11
Runtime-teardown disposal runs user destructors inline on the submitting thread (shut-down blocking pool)


ralexstokes:
**Severity: low — real defect, teardown-window only; cheap fix.** (Internal review round finding; independently corroborated by the external review's runtime-lifetime claim.)

## Mechanism

When the tokio runtime's blocking pool has already shut down but `Handle::try_current()` still succeeds, `task::spawn_blocking` does **not** panic — it returns `Ok(handle)` after synchronously dropping the closure on the calling thread. In `dispatch_disposal` (`runtime/disposal.rs:195-204`), the closure's `Arc` clone drops there, so when the function returns and drops its own `job` Arc (now the last reference), `DisposalJob::drop → finish()` runs the **user destructor inline on the submitting thread** — exactly what the module header (lines 209-213: "still never runs on (or unwinds into) the submitting thread") promises to prevent. The fallback disposal thread (line 214), which was available and would have isolated the destructor, is never attempted because the `Ok(handle)` branch returns unconditionally.

Concrete scenario: user drops (or `shutdown_timeout`s) the runtime while a shelterwood tree is alive; task drop glue drops an `Isolated<T>`/`DisposingReceiver<T>` → `dispose_detached` → shut-down-pool path → a blocking user destructor blocks the runtime-teardown thread. Panics remain contained; the exposure is blocking-only. `tests/disposal.rs` covers the no-runtime and live-runtime paths but not the shutting-down-pool window.

## Fix

After the `Ok(handle)` branch, detect that the closure was destroyed without running (`Arc::strong_count(&job) == 1`) and fall through to the fallback-thread enqueue instead of returning.

## Relationship to #192

#192 item 42 carries "runtime/ in-runtime disposal silently loses thread isolation during runtime shutdown (intentional; say so)". This issue proposes the loss is **fixable**, not just documentable — if fixed, drop that doc bullet. Independent of the fix, one adjacent doc sentence is still warranted (verified separately): `dispose_detached` is detached from the *caller*, not from the runtime's lifetime — on a healthy runtime, disposal rides `spawn_blocking`, so a permanently blocked destructor makes default `Runtime::drop` wait indefinitely (mitigation: `shutdown_timeout`/`shutdown_background`; an embedding-host decision). docs/shutdown-and-resources.md / embedding.md don't state this today. That sentence belongs with #192's doc work.

## Related test gap

[...]

Issue #192, opened 2026-08-11
Invariant and public-doc documentation (items 38, 39, 42 + Part 0 doc follow-ups)


ralexstokes:
Split out of #172 (P4 items 38–39, P5 item 42, plus doc follow-ups from the Part 0 review round). Anchors predate the item-37 splits and Part 0 rewrites — relocate by symbol; several invariants below were made structural by Part 0, so first verify each still holds by convention before documenting it (if it's now by construction, drop it from the list).

## 38. Write down the load-bearing invariants, at the site that depends on them

- [ ] `select_two`'s left-bias tie rule
- [ ] `LifecycleChannels` field-drop order
- [ ] observe's zero-watcher `send_modify` safety — *may be structurally moot post-0.1 (wakes flush at txn commit); verify, then document or drop*
- [ ] `run_blocking`'s deliberate works-while-stopping asymmetry
- [ ] `handle_removal`'s two distant preconditions — *site reworked by #181; re-verify*
- [ ] `CompletionGatedLatch`'s wake-lags-state ordering
- [ ] the reserved-lineage-0 subtlety in `IncarnationCounter`

(The `armed` handshake bullet is done — replaced by consuming `take_for_runtime()` in 0.7.)

## 39. Public-doc fixes

- [ ] `LifecycleItem::Lagged` doc contradicts SPEC B.4's marker-leads-episode semantics
- [x] ~~the `Structured*` wrappers in `exit.rs` silently block `downcast_ref` on authentic framework errors — comment whether non-forgeability is intentional and pin it with a test~~ — **done** in #214 stages 2/4: pinned by `exit::application_errors_cannot_forge_structured_provenance` and `exit::framework_errors_expose_structured_provenance_and_erased_views`, intent stated on `IntensityTrip`/`StartupFailure`, SPEC B.9 reworded to *authenticate*. See the verification comment below.
- [ ] `StopReason::NeverStarted`'s nested projection is a stringly error among structured siblings
- [ ] `Handler` docs should state post-error context capabilities for resuming decorators
- [ ] document `next_delay`'s ZERO-attempt coercion and `from_u64_ratio`'s non-finite mapping

## 42. Doc/comment polish

- [ ] `runtime/` in-runtime disposal silently loses thread isolation during runtime shutdown (intentional; say so)
[...]

PR #219, opened 2026-08-11
Isolate disposal during blocking-pool shutdown

ralexstokes:
## Summary

- Detect Tokio's accepted-but-synchronously-dropped `spawn_blocking` closure during blocking-pool shutdown and fall through to the shared fallback disposal thread.
- Distinguish a synchronously rejected closure from a healthy blocking task that already completed, avoiding empty fallback jobs behind a blocked destructor.
- Add two deterministic regressions: one occupies the sole blocking worker and proves a hostile blocking destructor neither runs on nor blocks the *submitting* thread, and one tears a runtime down from the embedding host's own thread and proves the same for the host.
- Document that detached disposal is detached from the caller, not from a healthy runtime's blocking-pool lifetime, plus the embedding-host `Runtime::shutdown_timeout` / `Runtime::shutdown_background` mitigation and what that mitigation costs.

## Root cause and impact

Tokio preserves compatibility by returning a `JoinHandle` when `spawn_blocking` is called after its blocking pool starts shutting down, while synchronously destroying the submitted closure. The previous unconditional return then dropped the disposal job's last `Arc` on the submitting thread, so a blocking user destructor could stall runtime teardown despite Shelterwood's isolation guarantee.

The fix detects that the caller's job is the sole remaining strong reference and still contains pending work, then routes it through the existing non-runtime fallback queue. The pending-work check prevents a fast, successfully completed blocking task from spuriously enqueueing an empty job.

## Scope of the guarantee

The isolation guarantee this restores is about the **submitting** thread. Two adjacent behaviors are unchanged and intentional:

- **Plain `drop(runtime)` still blocks indefinitely on a hostile destructor.** `Runtime::drop` shuts the scheduler down *before* the blocking pool, so a disposal submitted from task drop glue is accepted and queued rather than rejected. A blocking-pool worker later drains and destroys it, and `BlockingPool::shutdown(None)` joins that worker. Verified by repro. This is the case the new doc paragraph describes, and the fix is what makes its stated mitigation actually work: before this change `shutdown_timeout` / `shutdown_background` hit the *rejected* path and deadlocked the host thread too, so the documented escape hatch was not one.
- **Disposal is deferred past the return of `shutdown_timeout` / `shutdown_background`.** Where the old inline path destroyed the value before returning, it now runs on a Shelterwood-owned thread that nothing joins, so a host that exits the process immediately afterwards loses those destructors. Now stated in `docs/shutdown-and-resources.md`.

## Item / commit mapping

1. **Shut-down blocking-pool fallback and regression** — `8c710c4 fix disposal during blocking-pool shutdown`
2. **Runtime-lifetime documentation from #192** — `28e6d85 document disposal runtime lifetime`
3. **Adversarial review hardening of closure-outcome detection** — `e08e8a7 distinguish completed disposal submissions`
4. **Second adversarial review round: embedder-thread regression, non-vacuous assertions, doc and comment gaps** — `3dee73d Pin embedder-thread disposal isolation`

## Review notes addressed in commit 4

- The original regression's "teardown thread" is a `tokio-rt-worker` draining its queue, not the embedding host's thread. The host-thread shape — the one #205 actually names — was untested, and it is strictly worse pre-fix: `shutdown_background()` never returns. `embedder_runtime_teardown_isolates_a_task_held_value` covers it.
- `assert_ne!(destructor_thread, shutdown_thread)` was vacuous; that variable held the test's own thread, which never submits the disposal on either path. Replaced with an assertion that the destructor lands on the named shared fallback thread, which pins where isolated destruction goes rather than only where it does not.
- Both hostile destructors get a bounded escape hatch, unreachable on a correct run, so a regression reports a thread-identity assertion instead of hanging until nextest's `terminate-after`.
- The classifier's dependence on Tokio destroying a rejected closure before `spawn_blocking` returns is now recorded, along with the fact that a future Tokio deferring that drop degrades fail-safe to the old inline behavior rather than misrouting a live closure.

## Not in scope

- The driver-task cancellation test gap noted in the second half of #205 (`classify_root_driver_join` / the `spawn_system` lifecycle monitor, `driver.rs:107-154`). **#205 stays open after this merges.**
- `spawn_blocking_work` reaches the same Tokio rejection path with no handling, and `join_resuming` turns the resulting cancellation into an invariant panic. Filed separately as #222.

## Validation

- `./tools/dev cargo nextest run -p shelterwood --lib runtime::disposal` — 4 passed; both regressions confirmed to fail with clean thread-identity assertions when the fall-through is disabled.
- Flake checks on the disposal tests: 200/200 serial, 384/384 at 48-way process concurrency.
- Race stress (not committed): 40 rounds × 64 concurrent `dispose_then` submissions staggered against `shutdown_background`, plus `dispose_all` across the same edge — every value destroyed exactly once and every completion fired exactly once.
- `./tools/dev just ci` — passed, including clippy, 578 nextest tests, doctests, rustdoc, path/API checks, packaged docs/crate, and Nix formatting.
- `./tools/dev just ci-nix` — all authoritative clean Nix checks passed (run on commit 3).

Refs #205
Related to #192, #222

### DIFF
--- a/crates/shelterwood/doctests/docs/shutdown-and-resources.md
+++ b/crates/shelterwood/doctests/docs/shutdown-and-resources.md
@@ -113,7 +113,10 @@ runtime's lifetime. On a healthy runtime it uses Tokio's blocking pool, so a
 permanently blocked destructor can make the default `Runtime::drop` wait
 indefinitely. Embedding hosts that cannot accept that wait must choose Tokio's
 `Runtime::shutdown_timeout` or `Runtime::shutdown_background` as a host-level
-mitigation, accepting that blocking work may then outlive the runtime.
+mitigation, accepting that blocking work may then outlive the runtime. Once
+either returns, disposals submitted during teardown may still be in flight on
+a Shelterwood-owned thread that nothing joins, so a host that exits the process
+immediately afterwards loses those destructors.
 
 A shutdown request through a pre-spawn scope handle is retained as a pending
 stop. The first incarnation starts and immediately enters teardown; the timeout
--- a/crates/shelterwood/doctests/docs/shutdown-and-resources.md
+++ b/crates/shelterwood/doctests/docs/shutdown-and-resources.md
@@ -108,6 +108,13 @@ requires `panic = "unwind"`. The standard double-panic exclusion remains: a
 destructor that panics while its user future is already unwinding aborts the
 process before the supervisor can publish an exit.
 
+Shelterwood's detached disposal is detached from the caller, not from the
+runtime's lifetime. On a healthy runtime it uses Tokio's blocking pool, so a
+permanently blocked destructor can make the default `Runtime::drop` wait
+indefinitely. Embedding hosts that cannot accept that wait must choose Tokio's
+`Runtime::shutdown_timeout` or `Runtime::shutdown_background` as a host-level
+mitigation, accepting that blocking work may then outlive the runtime.
+
 A shutdown request through a pre-spawn scope handle is retained as a pending
 stop. The first incarnation starts and immediately enters teardown; the timeout
 for `shutdown_and_wait` begins when that incarnation starts, because there is
--- a/crates/shelterwood/src/runtime/disposal.rs
+++ b/crates/shelterwood/src/runtime/disposal.rs
@@ -200,6 +200,12 @@ fn run_fallback_disposals() {
 /// The pending check distinguishes rejection from a task that ran to
 /// completion before the submitter observed its reference count; requeueing
 /// the latter would leave empty jobs behind a blocked fallback disposal.
+///
+/// This pins Tokio's `spawn_task` shutdown path: a rejected closure is
+/// destroyed synchronously, before `spawn_blocking` returns. A future Tokio
+/// that deferred that drop would leave the count at two and degrade
+/// fail-safe to the old inline behavior rather than misroute a live closure.
+/// The end-to-end regression below pins the behavior we rely on.
 fn blocking_spawn_needs_fallback<T, C>(job: &Arc<DisposalJob<T, C>>) -> bool
 where
     T: Send + 'static,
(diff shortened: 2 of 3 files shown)


ralexstokes:
## Adversarial review: pass after one hardening fix

I reviewed the full `origin/main...e08e8a7` diff and both original item commits against #205, the #192 relationship, and the #214 stage-0 bullet. I traced the regression through Tokio 1.53.1's blocking-pool implementation rather than relying only on the test outcome.

### Confirmed finding, fixed directly

- **Closure-run vs. closure-drop ambiguity:** `Arc::strong_count(&job) == 1` is true both when Tokio synchronously rejects and destroys the submitted closure and when a healthy blocking task runs to completion before the submitter checks the count. The latter path could enqueue an already-finished job behind a legitimately blocked fallback destructor; repeated fast disposals could then accumulate empty jobs in the unbounded queue. Commit `e08e8a7` now requires both sole ownership and still-pending work. Its unit test pins the accepted, synchronously rejected, and already-completed outcomes.

### Adversarial checks

- The sole-reference test is race-safe for this ownership graph: the submitted closure is the only possible extra `Arc`, and pending state is inspected only after sole ownership is observed. A running/queued closure keeps count > 1; a completed closure has empty state; a synchronously destroyed closure leaves pending state for fallback.
- The fallback queue's `worker_live` transition and push/pop decisions remain under one mutex. Successful enqueue has a live worker destined to drain it; thread-spawn failure removes the just-added job and reaches the existing contained inline last resort rather than stranding completion.
- User destruction and completion callbacks remain outside the fallback queue lock and within unwind boundaries. The shutdown regression proves a blocking destructor runs on neither the runtime-shutdown caller nor the blocking-pool teardown worker, and proves submission returns while that destructor is still blocked.
- The regression is deterministic for Tokio 1.53.1: the occupied worker cannot run the outer task until after `shutdown_background()` has set pool shutdown; Tokio drains that already-queued outer task, whose nested disposal submission is then synchronously rejected. I corrected the test comment to describe that exact path.
- Healthy-runtime behavior is preserved: accepted work stays on Tokio's blocking pool. The docs accurately qualify the lifetime statement to a healthy runtime and accurately describe `Runtime::drop` as potentially indefinite and `shutdown_timeout` / `shutdown_background` as host-level mitigations that may let blocking work outlive the runtime.
- Scope remains limited to #205's shut-down blocking-pool window and the #192 documentation sentence. The separate driver-task cancellation gap is explicitly excluded.
- Commit structure remains separated: `8c710c4` is code + regression, `28e6d85` is documentation only, and `e08e8a7` is the additive adversarial code/test hardening.

### Validation

- `./tools/dev cargo test -p shelterwood --lib runtime::disposal::tests` — 3 passed
- `./tools/dev cargo test -p shelterwood --test integration disposal::` — 36 passed
- `./tools/dev just ci` — passed; 577/577 nextest tests plus all lint, doctest, rustdoc, API/path/layering, packaged-doc/crate, and formatting gates
- `./tools/dev just ci-nix` — passed; authoritative clean Nix lane

**Verdict:** no remaining blocking findings in the requested review scope. PR intentionally remains draft for the parent orchestration step.

PR #221, opened 2026-08-11
Avoid dropping mailbox wakers under core locks

### LINKED ISSUE #202: Mailbox drops caller-supplied wakers under core locks (poison/deadlock/abort via hostile waker Drop)
**Severity: high — confirmed; reachable from the public API; inconsistent with the module's otherwise-systematic hostile-waker containment.** (External review claim, adversarially verified with a full enumeration of waker-touching sites.)

## The two defective sites

Every place the mailbox touches a stored `Waker` was enumerated; exactly two drop caller-supplied wakers while holding core locks:

1. `mailbox/futures.rs:308` (`SendFuture::poll`, `Waiting` arm, operation mutex held since :280-283): `operation_state.waker = Some(context.waker().clone());` — runs the new waker's `clone` **and the drop of the previously stored waker** under the operation mutex.
2. `mailbox/cell.rs:763` (`withdraw`, `Waiting` arm): `state.waker = None;` — runs the stored waker's drop while holding **both** the mailbox mutex (:736) and the operation mutex (:744-747). Reached from `SendFuture::drop` (cancellation) and from `withdraw_send` (`send_timeout`/`call` timeout arbitration).

All other sites already follow the module's own discipline (move the waker out, wake/drop after unlock, `PanicAccumulator` containment): `SendOperation::accept`, `SendOperation::terminate`, the retired-op disposal path (`cell.rs:461-488`).

## Reachability

`ActorRef::send/send_timeout/call` return public `SendFuture`/`SendTimeout`/`CallFuture` futures; nothing restricts who polls them or with what `Context`. A user hand-polling with a custom `RawWaker` whose drop/clone vtable fns are arbitrary **safe** code hits site 1 by re-polling with a different waker, and site 2 by dropping the future or letting the deadline expire.

## Consequences

- Panic at site 2 unwinds through both guards, poisoning the mailbox mutex → every subsequent `send`/`try_send`/`receive`/`bind` panics on `.expect("mailbox mutex poisoned")` (the code explicitly treats mailbox-mutex poisoning as unacceptable elsewhere, `cell.rs:585-591`, `:819-821`). Reached from `SendFuture::drop` during an unwind, it's a double panic → **process abort**.
- Panic at site 1 poisons the op mutex; the next `operation.accept`/`observe` under the mailbox mutex in `promote_waiters`/`bind` poisons the mailbox mutex transitively.
- Blocking drop at site 2 stalls the whole mailbox; at site 1 it stalls `promote_waiters` transitively.
- Re-entrancy: site 2 re-entering any mailbox entry point self-deadlocks (std Mutex, non-reentrant); site 1 re-entering inverts the documented lock order (`cell.rs:100-104`: mailbox → operation, never reverse) → ABBA deadlock against a concurrent `promote_waiters`.

The pinned hostile-waker tests (`promotion_wakes_every_sender_when_multiple_wakers_panic` `cell.rs:1313`, `termination_discharge_...` `cell.rs:1360`) cover panics in `wake()` only — no test covers a panicking waker **Drop**, which is exactly this vector.

## Fix sketch

- `futures.rs` poll: clone the caller waker before taking the operation lock; `mem::replace` it in, release the guard, then drop the stale waker. Consider `will_wake` to skip churn.
- `cell.rs` `withdraw`: `state.waker.take()` into a local carried out of both lock scopes, dropped after the mailbox guard is released (matches the existing convention for rejected payloads; route through `dispose_detached` if stronger containment is wanted).
- Add a hostile-waker-**Drop** test alongside the existing hostile-wake tests.

ralexstokes:
## Summary

Keep every caller-supplied waker vtable call out of the mailbox's core mutexes, and keep a hostile waker destructor from escaping cancellation drop glue.

## Issue items and commits

1. **`SendFuture::poll` waker clone/replacement/drop** — `62d9541`
   - Clone the incoming context waker before acquiring the send-operation mutex.
   - Replace the registered waker while locked, then release the mutex before dropping the stale waker.
   - Add a hostile raw-waker regression whose clone and drop callbacks attempt to reacquire the operation mutex.

2. **Mailbox withdrawal waker removal/drop** — `a61e49d`
   - Take the registered waker while holding the operation mutex.
   - Finish direct waiter unlinking, release both the operation and mailbox mutexes, then drop the waker.
   - Add a hostile waker-drop regression that reacquires the operation mutex, re-enters `try_send`, panics, and verifies neither core mutex is poisoned.

3. **Cancellation containment and clone elision** — `15c354d`
   - `MailboxCell::withdraw` returns the released waker instead of destroying it, handing the recovered message over first, so the caller chooses where a `RawWaker` destructor runs.
   - `SendFuture::drop` routes that waker through isolated disposal beside the message. A panicking destructor can no longer escape drop glue, so cancellation reached during an unwind is no longer a double panic and a process abort, and it can no longer divert the message from its own isolated disposal.
   - `SendFuture::poll` stops cloning the context waker speculatively: it inspects the outcome locked, takes a `will_wake` fast path, and only otherwise unlocks to clone and reacquires to install. Reacquiring re-reads the outcome, and the outcome transition and waker take stay atomic under the operation lock, so the unlocked window cannot lose a wakeup.
   - `withdraw`'s accepted and terminated arms `debug_assert` that acceptance and termination already took the waker.
   - Regression coverage moves onto the public `SendFuture::drop` -> `withdraw` path, including the unwind case that aborted before.

## Semantics

No caller vtable call happens under a core mutex, and none happens at all on a completing poll or on a repoll with an equivalent waker. Waker destruction is exactly-once on every path.

Where a waker destructor runs is now explicit per path:

- **Cancellation (`SendFuture::drop`)** — isolated disposal. There is no caller left to receive a panic and this is drop glue, so nothing user-controlled runs inline.
- **Timeout arbitration (`send_timeout`/`call`)** — inline on the caller's own task, as before. This is a normal return, not drop glue, so a hostile destructor's panic stays visible to the caller who supplied the waker. No lock is held and the recovered message already belongs to the returned result.

## Validation

- `./tools/dev cargo nextest run -p shelterwood --lib mailbox::` — 24 passed
- `./tools/dev just ci` — passed (582 nextest tests plus lint, docs, API/layering, and package checks)
- `./tools/dev just ci-nix` — passed

Each new regression was also run against the pre-fix code and fails there: the unwind case reproduces the `SIGABRT`, the containment case surfaces `injected waker drop panic` out of `SendFuture::drop`, the completing poll records 1 speculative clone instead of 0, and four repolls record 4 clones instead of 1.

Fixes #202

### DIFF
--- a/crates/shelterwood/src/mailbox/cell.rs
+++ b/crates/shelterwood/src/mailbox/cell.rs
@@ -1203,6 +1203,12 @@ pub(super) mod tests {
                 .mailbox
                 .upgrade()
                 .expect("the cancelling send retains its mailbox");
+            drop(
+                mailbox
+                    .state
+                    .try_lock()
+                    .expect("waker drop runs after the mailbox lock is released"),
+            );
             let error = mailbox
                 .try_send(2)
                 .expect_err("a reentrant try-send observes the unbound mailbox");
--- a/crates/shelterwood/src/mailbox/cell.rs
+++ b/crates/shelterwood/src/mailbox/cell.rs
@@ -732,7 +732,18 @@ impl<M: Send + 'static> MailboxCell<M> {
 }
 
 impl<M> MailboxCell<M> {
-    pub(super) fn withdraw(&self, operation: &Arc<SendOperation<M>>) -> Withdrawal<M> {
+    /// Withdraws a send operation and releases the waker it had registered.
+    ///
+    /// The waker is returned rather than destroyed here: a `RawWaker` vtable is
+    /// caller code, so only the caller knows whether its destructor may run
+    /// inline. Withdrawal guarantees only that neither core lock is held once
+    /// the waker is handed back, and that the recovered message is handed over
+    /// first, so a hostile waker destructor can neither poison a core mutex nor
+    /// divert the message from the caller's chosen disposal route.
+    pub(super) fn withdraw(
+        &self,
+        operation: &Arc<SendOperation<M>>,
+    ) -> (Withdrawal<M>, Option<Waker>) {
         let mut mailbox = self.state.lock().expect("mailbox mutex poisoned");
         let current_observation = match mailbox.status() {
             BindingStatus::Bound(incarnation) | BindingStatus::Frozen(incarnation) => {
@@ -766,24 +777,34 @@ impl<M> MailboxCell<M> {
                         state.waker.take(),
                     )
                 }
-                OperationOutcome::Accepted(incarnation) => (
-                    Withdrawal::Accepted(*incarnation),
-                    state.registration.take(),
-                    None,
-                ),
+                OperationOutcome::Accepted(incarnation) => {
+                    let incarnation = *incarnation;
+                    // Acceptance took the waker in the same critical section
+                    // that published this outcome, so no registration survives
+                    // it for withdrawal to release.
+                    debug_assert!(state.waker.is_none());
+                    (
+                        Withdrawal::Accepted(incarnation),
+                        state.registration.take(),
+                        None,
+                    )
+                }
                 OperationOutcome::Terminated {
                     message,
                     final_incarnation,
-                } => (
-                    Withdrawal::Terminated {
-                        message: message
-                            .take()
-                            .expect("a terminal operation must retain its message"),
-                        observed: *final_incarnation,
-                    },
-                    state.registration.take(),
-                    None,
-                ),
+                } => {
+                    let message = message
+                        .take()
+                        .expect("a terminal operation must retain its message");
+                    let observed = *final_incarnation;
+                    // Termination likewise took the waker before publishing.
+                    debug_assert!(state.waker.is_none());
+                    (
+                        Withdrawal::Terminated { message, observed },
+                        state.registration.take(),
+                        None,
+                    )
+                }
                 OperationOutcome::Withdrawn => {
                     panic!("a send operation was withdrawn more than once")
                 }
@@ -797,8 +818,7 @@ impl<M> MailboxCell<M> {
             }
         }
         drop(mailbox);
-        drop(waker);
-        result
+        (result, waker)
     }
 }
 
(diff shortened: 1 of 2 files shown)


ralexstokes:
Adversarial review completed against issue #202 and the Stage 0 scope in #214 at `803da487`.

Confirmed finding and direct fix:
- The withdrawal regression's hostile `Drop` callback called blocking `try_send` to prove mailbox re-entry. A regression that released the operation mutex but retained the mailbox mutex would therefore hang the test instead of failing diagnostically.
- Added `803da48 Make waker unlock regression fail fast`: the callback now first probes the mailbox mutex with `try_lock`, then exercises reentrant `try_send`. Both lock-boundary regressions now fail promptly.

Adversarial audit:
- `SendFuture::poll`: incoming waker clone runs before the operation lock; replacement is move-only while locked; stale waker destruction runs after unlocking. Acceptance/termination races leave the cloned waker to drop outside the lock. No mailbox acquisition starts from the operation lock, so the documented mailbox -> operation order remains acyclic.
- `MailboxCell::withdraw`: the stored waker is taken under mailbox -> operation locking; the operation guard is released, the waiter registration is unlinked while the mailbox guard remains, and only then are the mailbox guard and waker dropped in that order. Reentrant clone/drop behavior cannot poison either core mutex or create the prior ABBA/self-deadlock.
- Panic behavior remains intentionally caller-visible outside the locks, consistent with this point-fix PR and its stated semantics. In `SendFuture::drop`, a hostile waker-drop panic can still participate in Rust's ordinary double-panic abort behavior during an existing unwind; changing that to detached/contained disposal would be a separate semantic expansion, not the #202/#214 Stage 0 move-out-after-unlock fix.
- Repeated polling still replaces the registered waker exactly as before; `will_wake` could reduce clone/drop churn but is not required for correctness and was not added.
- No broader Stage B effects refactor is present.

Commit/scope audit:
- `62d9541`: poll-time clone/replacement/drop site.
- `a61e49d`: withdrawal take/unlink/drop site.
- `803da48`: adversarial-review test hardening only.
- Diff remains limited to the two mailbox source/test files.

Validation at `803da487`:
- `./tools/dev cargo test -p shelterwood waker`: passed.
- `./tools/dev cargo test -p shelterwood mailbox::`: 25 passed.
- `./tools/dev just ci`: passed, including 577/577 nextest tests plus lint/docs/API/layering/package checks.
- `./tools/dev just ci-nix`: passed.
- GitHub CI run 31516118041: passed.

Verdict: no remaining correctness findings within #202's Stage 0 scope. The PR is suitable to be marked ready after the parent orchestration gate; it remains draft in this review step.

PR #227, opened 2026-08-11
Consolidate supervision in the core reducer

### LINKED ISSUE #210: Driver cleanup: restart_is_suppressed double-naming, handle_removal collapse, lookup helper, small dead weight
Driver-layer cleanups from the 2026-08-11 review round. The rename in item 1 is the load-bearing one; the rest are line-count and false-degree-of-freedom reductions (~50-60 lines total).

## 1. Rename one of the two `restart_is_suppressed` predicates

`ScopeRuntime::restart_is_suppressed` (`driver/events.rs:212`) is the *broad* consult (drain, force, latched stop requests, ancestor latches, plus removal sources) and must never be used at exit dispatch; `DynamicEntry::restart_is_suppressed` (`driver/admission_control.rs:158`) is the *narrow* removal-sources-only predicate exit dispatch depends on via `dispatch_membership_status` (`events.rs:185`). The doc comments at `events.rs:176-221` exist precisely to keep these apart; the identical name invites exactly the misclassification they warn about. Rename the entry-level one (e.g. `removal_latched(key)` — it's the keyed sibling of `DynamicEntry::removal_requested`).

## 2. `handle_removal` re-implements `dynamic_membership_is_removing` inline (~15-20 lines)

`driver/removal.rs:25-47` spells out exactly the predicate defined 15 lines above at `removal.rs:10`. The extra `member.membership() == removal.membership` filter is provably always-true (`ChildKey`s are never reused — `storage.rs` comment, pinned at `storage.rs:215` — and a slot's `Membership` is immutable), so the body collapses to `if !self.dynamic_membership_is_removing(key) { return; }` plus the existing branch — after which `RemovalRequest.membership` has no reader and can be dropped from the struct and its two construction sites in `admission_control.rs`. (The synchronous-disposal re-entry at `removal.rs:52-58` was verified safe — it uses `get(...).is_some_and(...)`, not indexing.) Note #192 item 38 lists "handle_removal's two distant preconditions — re-verify"; this collapse would make that bullet largely structural.

## 3. Shared dynamic-entry lookup helper (~15 more lines)

The "lock `dynamic.state`, `entry(child_id)`, filter entry membership against the child's membership, apply a predicate" pattern is spelled out five times (`events.rs:189-197`; `removal.rs:14-22, 38-44, 71-75, 99-102`). A `ScopeRuntime` helper taking `(ChildKey, impl FnOnce(&DynamicEntry) -> bool)` centralizes the membership-match invariant.

## 4. Smaller items

- `SpawnDispatch.declared_readiness` (`driver/child.rs:289`) is set identically in all four dispatch arms and read once where `child.options.readiness` is equally in hand — a false degree of freedom, ~8 lines.
- `SpawnLatches` allocates `framework_abort`/`framework_abort_ack` for every incarnation (`child.rs:609-616`) then discards them for the three non-scope construction kinds (`:685-686`) — two wasted latch allocations per non-scope spawn including every restart; make the pair scope-child-only (optional; current shape buys uniformity).
- `ChildArena::storage_len` duplicates `len` (`storage.rs:127-143`, `driver.rs:309-314`) — vestigial slab-era telemetry shape; collapse opportunistically.

## 5. Optional: derive `incomplete_children` instead of maintaining it

The counter (`driver.rs:201`) is maintained at three sites with one consumer (`all_terminal` in `finish_if_ready`, `shutdown.rs:210`, once per batch). Deriving it (`children.values().all(|c| !c.is_incomplete())`) is feasible and safe — one short-circuiting scan per batch, with a real but usually small cost for `Retention::Retain` scopes. Verified NOT implicated in the startup-hang bug (that was a missed recomputation *trigger*, not counter drift); the higher-value structural change is running `progress_startup()` in the per-batch epilogue next to `finish_if_ready` (see the startup-hang issue).

ralexstokes:
## Summary

- add a tokio-free `SupervisorState` and total synchronous `step(&mut SupervisorState, Event, &mut EffectSink)` reducer in `shelterwood-core`
- move child-key/index ownership, authoritative child/removal state, startup aggregation, ordered start/stop cursors, drain/finish state, and derived completion into the reducer
- leave the façade driver as the runtime-resource/effect shell, sampling synchronous removal latches before reducer transitions and again at the single construction funnel
- expose the membership-terminal / incarnation-complete / joined predicates behind the single `settled(epoch)` shutdown surface
- absorb #210: distinct suppression naming, one dynamic-entry lookup helper, collapsed removal dispatch, no duplicate arena telemetry, no false readiness field, and scope-only framework latches
- add table-driven transitions, derived-state properties, key exhaustion/totality checks, legal-phase guards, ordered-stop coverage, and exhaustive 2–3 child event-order enumeration in both scope flavors

## Why

Stage 3 of #214 requires one structural authority that can be tested without Tokio, clocks, randomness, user code, or I/O. Previously the driver maintained arena indexes, lifecycle flags, startup bookkeeping, ordered cursors, removal state, and an incremental incomplete-child counter across separate structures and update sites.

The reducer now owns those decisions. Runtime latches remain synchronously writable, but are sampled into reducer events; removal is represented by `ChildState::Removing(...)` rather than a parallel driver flag.

A fresh adversarial pass after the draft was opened found and directly fixed a queued-`StartChild` removal race, illegal stale phase transitions, sampled-removal startup scheduling, and missing ordered-flavor/ordered-stop coverage.

## Validation

- `./tools/dev just ci`
  - formatting and `-D warnings` clippy
  - 614 tests, doctests, rustdoc/API reachability, and packaged-crate validation
- `./tools/dev just ci-nix`
- reducer suite enumerates all 8! two-child and all 9! three-child schedules for both ordered and dynamic scopes

Stacked on #226 (`agent/issue-214-stage-2`).

Part of #214.
Closes #210.

### DIFF
--- a/crates/shelterwood-core/src/lib.rs
+++ b/crates/shelterwood-core/src/lib.rs
@@ -13,6 +13,7 @@ pub mod exit;
 pub mod identity;
 pub mod panic;
 pub mod policy;
+pub mod supervisor;
 
 pub use deadline::*;
 pub use engine::{MembershipStatus, ScopeState};
--- a/crates/shelterwood-core/src/supervisor.rs
+++ b/crates/shelterwood-core/src/supervisor.rs
@@ -356,13 +356,21 @@ impl SupervisorState {
             .map(|(key, _)| *key)
     }
 
-    fn transition_incarnation(&mut self, child: ChildKey, next: IncarnationState) {
+    fn transition_incarnation(
+        &mut self,
+        child: ChildKey,
+        expected: &[IncarnationState],
+        next: IncarnationState,
+    ) -> bool {
         if let Some(record) = self.children.get_mut(&child) {
-            if record.state.membership_terminal() {
-                return;
+            if record.state.membership_terminal() || !expected.contains(&record.state.incarnation())
+            {
+                return false;
             }
             record.state = record.state.with_incarnation(next);
+            return true;
         }
+        false
     }
 
     fn mark_removing(&mut self, child: ChildKey) -> Option<IncarnationState> {
@@ -425,6 +433,13 @@ impl SupervisorState {
                         self.next_ordered_start = next;
                         continue;
                     };
+                    if record.state.membership_status() == MembershipStatus::Removing {
+                        // The queued removal transition owns stop and commit.
+                        // Until that commit shrinks the initial set, this
+                        // membership still gates ordered startup but can no
+                        // longer schedule construction.
+                        return;
+                    }
                     if !record.spawned_once {
                         effects.push(Effect::StartChild { child });
                         return;
@@ -440,6 +455,7 @@ impl SupervisorState {
                 for (&child, record) in &self.children {
                     if matches!(record.startup, StartupMembership::Initial { .. })
                         && !record.spawned_once
+                        && record.state.membership_status() == MembershipStatus::Active
                     {
                         effects.push(Effect::StartChild { child });
                     }
@@ -529,7 +545,11 @@ impl SupervisorState {
             } => self.admit(membership, initial, start_immediately, effects),
             Event::Spawned { child } => {
                 if let Some(record) = self.children.get_mut(&child)
-                    && !record.state.membership_terminal()
+                    && matches!(record.state, ChildState::Resident(_))
+                    && matches!(
+                        record.state.incarnation(),
+                        IncarnationState::Unstarted | IncarnationState::RestartPending
+                    )
                 {
                     record.spawned_once = true;
                     record.state = record.state.with_incarnation(IncarnationState::Active);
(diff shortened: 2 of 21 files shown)


ralexstokes:
Reviewed at `59ab093`; `just ci` was green there (614 tests). Findings and what I did with each — three fix commits pushed to this branch (`0672c13`, `ac2f3c2`, `456920b`), `just ci` green after them (616 tests).

## Fixed: `settle_supervisor()` spins forever

`settle_startup` decided to emit `StartChild` from `spawned_once` alone, while `spawn_child` declines a child that is already disposing or terminal and so never sends `Spawned` back. Since `settle_supervisor` loops until a `Settle` emits nothing, the reducer re-derived the same start from unchanged state forever.

Reachable via incarnation exhaustion: `spawn_child` → `mint()` returns `None` → `begin_terminal_disposal` hands the disposal to the blocking pool, leaving the child `Disposing` with `spawned_once` false while `lifecycle.is_starting()` still holds. Confirmed on the branch — a driver fixture calling `settle_supervisor()` was still running at 60s and had to be SIGTERM'd. In production this fires at the pre-loop `scope.settle_supervisor()` in `run_scope_incarnation`, so the scope hangs before entering the event loop. Pre-consolidation, `progress_startup` returned at its `initial_ready` check and ran once per batch, so the same state was a startup stall rather than a spinning core.

Exhaustion needs u64 restarts so this is not a realistic production trigger, but it is a route the repo treats as real and fail-closed, and turning a graceful terminal outcome into an unpreemptable loop is strictly worse.

`0672c13` gates both flavor arms on `ChildRecord::startable`, spelled as `Event::Spawned`'s own acceptance set so emission and acceptance cannot drift. The structural point is the general one: the loop's termination argument assumes the shell honours every effect, and the reducer has no feedback channel for one it declines. Pinned three ways — the emission/acceptance agreement over every child state in both flavors and both membership statuses, the same agreement as a property inside the exhaustive enumeration, and the production shape driven through the real `settle_supervisor` loop. Verified all three fail (the last by timing out) with the gate removed.

## Fixed: `RestartPending` narrowed the readiness reset

`Event::RestartPending` clears initial readiness under `is_starting`, where the driver it replaced used `!startup_complete`. Both readers of the flag gate on `startup_complete`, and the predicates differ exactly where the readers still care: in a root scope that has published `StartupFailed` but is not yet draining, `dispatch_exit` can still return `ScheduleRestart`, and the membership then kept `ready` set — so a later pre-ready exit published `NotAborted` where it previously published `Aborted`. `ac2f3c2` restores the readers' predicate; a deliberate change there belongs to stage 4's recast, not to an otherwise behaviour-preserving consolidation.

## Fixed: silent transition drops, and a tautological invariant

`transition_incarnation` returns whether it applied and nearly every caller discards it. A dropped `DisposalStarted` would make the following `Terminalized` unreachable too, so the membership never reaches `Joined` and the scope simply never finishes — a completion stall is the worst presentation of that mistake. `456920b` asserts at both publication sites. I could not reach either assertion from the current call graph; they exist so a future caller cannot reintroduce the stall quietly.

Same commit replaces the invariant asserting `membership_terminal() == joined()`. Those are the same expression, so the check could never fail. They coincide *by construction* — this model routes every terminal through `Disposing` and the shell publishes the terminal stage only once disposal joins, so the state the old `!is_terminal() || is_disposing()` test existed to catch is unrepresentable. That is now documented on the type, and the check spends itself on something that constrains instead.

## Filed, not fixed

- **#228** — stage 3's last checkbox (re-establish the intra-façade layering) is not addressed, and I don't think it can be here. Both retired rules verify clean on this branch — no module outside `driver`/`tree` names `driver::` or `tree::`, `cells` never names `plan`, and the old exit path has no downcast — so nothing is violated; what is missing is enforcement. The checkbox's premise was that moving decision state into core would make the layering structural, but Rust has no intra-crate layering enforcement, so a crate boundary is the only mechanism and the reducer move does not create one. #228 covers extracting `cells`/`observe`/`admission`/`cancellation` into `shelterwood-cells` (no cycle: `shelterwood-mailbox` defines the traits `cells` implements), and notes that `plan` cannot move because it depends on the definition surface. Either way SPEC item 13's final paragraph is now stale — it says the lapse persists "until the reducer consolidation moves the driver's decision state into core," and that has happened. Worth unchecking or re-scoping the #214 bullet rather than leaving it ticked.
- **#229** — `exhaustive_small_scope_interleavings_preserve_reducer_invariants` is 21.4s of a 21.7s suite. Not a blocker, and revisiting belongs after stage 4 gives it a real invariant list to assert; filed so the cost is recorded. Note it did not catch the spin above despite the bug living squarely in its state space — the schedules were the expensive part, the missing property was the cheap part.

## Not addressed

The `--offline` drop in `check-packaged-crate.sh` (`59ab093`) — superseded, that lane is being removed separately.

#210's items are all genuinely absorbed: the `restart_is_suppressed` double-naming, the `handle_removal` collapse with `RemovalRequest.membership` dropped, the `with_dynamic_entry` helper, `declared_readiness`, `storage_len`, the scope-child-only `SpawnLatches` pair, and `incomplete_children` derived. `shelterwood-core` still has only `thiserror`, and `step` reads no clock and samples no randomness.

PR #233, opened 2026-08-11
Extract state projections into shelterwood-cells

### LINKED ISSUE #228: Extract the state/projection layer into `shelterwood-cells` to make intra-façade layering structural
Extract the state/projection layer into a `shelterwood-cells` crate so the intra-façade layering the stage-2 retirement left unenforced becomes a compiler-checked dependency direction rather than a convention. Follow-up to #214 stage 3 (#227); do this after #227 merges.

## Why this is still open after #227

Stage 3's last checkbox assumed the reducer consolidation would make this fall out of the dependency direction: *"Moving the decision state into core makes both fall out of the dependency direction rather than out of a grep."*

Half of that happened. Arena keys, ordered cursors, the startup aggregate and authoritative child state did move into `shelterwood-core`. But `driver/**` still owns exit dispatch, removal, readiness gating and the stop ladder, and `cells`/`observe`/`cancellation` are still **in the same crate as** `driver`. Rust has no intra-crate layering enforcement — a crate boundary is the only mechanism that makes "below the driver" structural. The reducer landing in core was necessary but not sufficient.

## Current state: nothing is violated

This is an enforcement gap, not a regression. Verified on `agent/issue-214-stage-3`:

- No façade module outside `driver`/`driver/**` and `tree`/`tree/**` names `driver::` or `tree::`.
- `cells` never names `plan`.
- The old exit path (`driver.rs`, `driver/**`, `engine.rs`, `exit.rs`) contains no `downcast{,_ref,_mut}`. The façade's `engine.rs` and `exit.rs` are now one/two-line re-export shims over core; the only downcasts in the crate are `plan.rs` (the `DynamicRoute` slot type) and `raw.rs` (panic payloads), neither of which was ever on the retired `check-exit-paths.sh` list.

So the work here is to make the invariant unbreakable going forward, not to fix a break.

## The move

Extract into `shelterwood-cells`, depending on `shelterwood-core`, `shelterwood-mailbox`, `shelterwood-runtime`:

| module | lines | notes |
|---|---|---|
| `cells.rs` | 2094 | the bulk |
| `observe.rs` | 1018 | contributes public façade types |
| `admission.rs` | 90 | imports `identity` only |
| `cancellation.rs` | 143 | imports `runtime` only |

`driver`, `tree`, `scope`, `plan`, `definition`, `raw`, `task`, `actor` stay in the façade.

**No cycle, no new inversion seam.** `shelterwood-mailbox` *defines* `MailboxControl` / `ActorIdentity` / `MailboxTermination` (`shelterwood-mailbox/src/lib.rs:47-82`) and `cells` implements them, so the new crate depends on mailbox and not the reverse. `cells`'s import set is entirely core shims (`engine`, `exit`, `identity`, `policy` + crate-root re-exports), the mailbox crate, the runtime crate, and `observe`/`admission` — which move with it. It names nothing above itself today.

## `plan` cannot move — scope this deliberately

The #214 checkbox lists `cells`/`plan`/`observe`/`cancellation`, but `plan.rs` imports `definition::DefinitionSource`, `raw::RawConstruction` and `task::{OnceTask, TaskDef}` — the user-facing definition surface, which stays in the façade. So `plan` stays co-resident with `driver`, along with `raw`/`task`/`actor`/`definition`.

That leaves an honest residue: the *state/projection* layer becomes compiler-enforced, the *definition* layer stays convention-only. Decide explicitly whether that residue is acceptable (I think it is — it is a much smaller and more legible claim than the original blanket rule) or whether a second `shelterwood-plan` boundary is wanted, and record the answer.

## The no-downcast rule is a separate decision

SPEC item 13 covers both rules in one paragraph, so whoever does this has to resolve both. Recommendation for the downcast half: **retire it.** Its content now holds by construction where it matters — exit classification lives in core with typed internal verdicts, and the `Any`/`dyn Error` surface exists only at the façade's boundary. The residual surface is `driver/**` alone, and the property that actually protects the design (a blanket user conversion cannot mint an authenticated `Structured*` payload) is covered by a probe rather than a grep.

## Tasks

- [ ] Create `crates/shelterwood-cells`; move `cells`, `observe`, `admission`, `cancellation`; leave shims in the façade the way `engine.rs`/`exit.rs`/`mailbox.rs` do today
- [ ] Accept and document the visibility trade (more `pub` across the boundary), and do the within-crate tightening as part of the move, per the stage-2 precedent
- [ ] Add a `runtime-api-check` lane for the new crate — `justfile:32-36` runs the rustdoc-JSON walk once per crate contributing public façade items, because cross-crate re-exports are not inlined into the façade's document. `observe`'s types make `shelterwood-cells` such a crate. Mirror the change in `flake.nix`
- [ ] Decide the `plan`/definition-layer residue and record it
- [ ] Rewrite SPEC item 13's final paragraph (`SPEC.md:2718-2723`). It currently says the lapse persists *"until the reducer consolidation moves the driver's decision state into core"* — that has now happened, so the sentence is stale regardless of outcome. Replace with what is actually true for each half
- [ ] Re-scope the corresponding #214 stage 3 checkbox to point here rather than leaving it ticked

ralexstokes:
## Summary

- extract `cells`, `observe`, `admission`, and `cancellation` into a new `shelterwood-cells` crate beneath the mutable façade driver
- leave narrow façade re-export shims and preserve test-only cell probes behind a `test-util` feature
- add the new crate to both local and clean-Nix public API reachability enforcement
- update SPEC item 13 to record the accepted `plan`/definition-layer residue and retire the former driver-path downcast source rule

## Why

The reducer consolidation moved authoritative supervision state into `shelterwood-core`, but Rust still could not enforce the remaining state/projection-to-driver direction while both layers lived in the same crate. This crate boundary makes that dependency direction compiler-checked without introducing an inversion seam.

## Impact

The supported `shelterwood` façade remains the public entrypoint. Cross-crate implementation seams become `pub` inside `shelterwood-cells` and are hidden from generated API documentation where they are not supported façade surface. Existing façade unit-test probes remain available only through the crate's `test-util` feature.

## Validation

- `./tools/dev just ci` — 616/616 tests passed, plus fmt, clippy, doctests, strict docs, API reachability, packaged docs, and nixfmt
- `./tools/dev just ci-nix` — all authoritative clean Nix checks passed

Closes #228


ralexstokes:
## Review

Verified the extraction is faithful and the enforcement claim holds, with three fixes pushed (`d229786`, `e5785a3`).

**Move fidelity.** Diffed each moved module against its `main` original: `admission.rs` is byte-identical; `cells.rs`, `observe.rs`, and `cancellation.rs` differ only in visibility, import paths, and two doc-text edits (the `[crate::ScopeRef::wait_for_child]` link becomes prose, correctly, since the target no longer resolves from this crate). No logic moved or changed. Test attribute count is identical on both sides (396), and the façade's `pub use observe::{…}` list in `lib.rs:40-44` matches the shim's export list exactly, so the supported public surface is unchanged.

**The `#[doc(hidden)]` on `mod cells` is load-bearing, and that was worth recording.** Removing it makes the new reachability lane fail with 24 leaks (`ScopeCell`, `MemberCell`, `ResidentProjection`, `DynamicRoute`, `ObservationTxn` → `shelterwood_runtime::sync::{Latch, OneShotReceiver, WatchReceiver, WatchSender}`) — rustdoc omits hidden items from its JSON, so the walk over `shelterwood_cells` covers `admission`, `cancellation`, and `observe` only. That is the right outcome, but it is an exemption rather than coverage, and nothing in the PR said so. What actually keeps the seam out of the public API is the façade's `pub(crate) use shelterwood_cells::*` shim: I checked that `pub use cells::ScopeCell;` in the façade is E0365, so the accident this walk would otherwise have caught is a compile error instead. `e5785a3` records that in SPEC item 13 and above the attribute, and re-wraps the ragged line the SPEC edit left at `2715-2716`.

**Two items were widened past what crosses the boundary** (the issue's tightening task). Audited mechanically by narrowing every `pub` in `cells.rs` back to `pub(crate)` and reading off what the façade actually demands:

- `ObservationTxn::pulse` — no caller outside this crate; the façade reaches its own senders through `defer`. (`.pulse()` hits in `driver/child.rs` and `raw.rs` are `WatchSender::pulse`, not this method.)
- `ScopeRecord::total_restarts` — read only by snapshot publication inside the crate; the driver takes its totals from the decision that produced them.

Both are `pub(crate)` in `d229786`. Everything else that became `pub` is genuinely reached by `driver`/`plan`/`tree` or by the façade's own tests, including `ObservationGate::{lock, same_gate}` and `MemberCell::terminal_disposal_pending`, which I checked individually.

**Scope decisions.** The `plan`/definition residue and the retired downcast rule read as recorded in SPEC item 13; the #214 stage-3 checkbox was already re-scoped to this issue in the prior round, so all six issue tasks are covered. `tracing` and `thiserror` are both used in the new crate, and both remain used in the façade after the move; no stale path references to the old module locations remain in docs, `justfile`, or `flake.nix`.

**Validation.** `./tools/dev just ci` green on the pushed head: 616/616 tests, clean on all three reachability walks, plus fmt, clippy, doctests, strict docs, packaged docs, and nixfmt.

Issue #211, opened 2026-08-11
Small simplifications batch: identity lineage, raw/cells/observe items, Option<ReadinessDeadline>, visibility tightening


ralexstokes:
Batch of small, independent simplifications from the 2026-08-11 review round. Total ~100-130 lines plus several removed reader hazards. Source is otherwise near its complexity floor — the round's verdict on every major module was "no reachable bugs, no simplification worth its churn" beyond what's listed here and in the sibling driver/mailbox issues.

## identity.rs — dead lineage in `IncarnationCounter` (~20-25 lines)

Every incarnation counter is built from `FenceCounter::sequence()` (the constant `Lineage(0)`), and `IncarnationCounter::mint` immediately discards `fence.lineage` (`identity.rs:264-270, 252-254, 312-318`) — incarnation ordering is gated by membership equality instead. Beyond dead weight, it's a reader hazard: all incarnation counters *appear* to share one global lineage domain. Replace the embedded `FenceCounter` with a bare `PoisonedCounter`; also dedupe the verbatim `MintedMembership { … FenceCounter::sequence() }` block (`identity.rs:381-387` vs `406-412`) behind a private constructor. Note: this would moot #192 item 38's "reserved-lineage-0 subtlety in `IncarnationCounter`" doc bullet by removing the subtlety.

## raw.rs (~30 lines)

- `EventQueue.signal` is always `disposal.signal.clone()` and `RawResources.panic` always aliases `RawResources.disposal.panic` (`raw.rs:352-355, 987-1013`); route through `disposal` and have `SharedOffloadState::new` accept a `RawDisposal` — removes two fields and an implicit "these are the same object" invariant.
- `OffloadScope` (`raw.rs:830-834, 1340-1373, 1466-1492`) exists only to avoid building latch clones `offload` would discard; always returning a `Guard` with `offload` calling `.map(|g| g.detach())` deletes the enum and an `expect`.
- `TimerStore::replace` (`raw.rs:583-612`) invokes the user `Hash` impl twice per arming; hoist one `hash_key`.

## cells.rs (~20 lines)

- `ResidentChild::drop` (`cells.rs:785-804`) re-implements the `Removed`-event body `publish_removal` (`cells.rs:766-777`) already owns — reduce to takeing the completion and delegating.
- `adopt_descendant_observation_gates_locked` (`cells.rs:1143-1154`): both arms call `install_observation_gate_locked` identically; hoist above the branch.
[...]

Issue #208, opened 2026-08-11
SPEC text amendments: admission-channel bounded wording, zero-duration observation waits, Graceful{0} provenance


ralexstokes:
Three places where SPEC text and the deliberately-pinned implementation have drifted apart. Per the issue #116 §4 adjudication protocol, these are spec amendments, not code bugs — the observable semantics are pinned by tests and were reviewed as intentional.

## 1. Admission command channel: "bounded, caller is the backpressure" — no longer true, and never really was

SPEC Appendix A (~3064: "Control/command channel | 64, floored to 1 | Internal; awaited insertion commands…"), §10 (~2067: "bounded command channel … callers hold futures and absorb backpressure"), §11 (~1680: "that awaiting caller is the channel's backpressure") describe a bounded lane. Commit `9516ca7` ("Unify the driver event lane", the #116 round) deliberately converted both lanes to unbounded with a per-wake batch cap replacing the capacity bound (`driver.rs:663-679`); SPEC was never updated. Caller-absorbed backpressure never actually existed — the pre-9516ca7 architecture had an unbounded per-scope request queue plus a forwarder feeding the bounded lane (needed because split-`define` detaches at first poll and removal can be signalled off-runtime), and that forwarder caused the confirmed #116 race. The queued `AdmissionRequest` is small; the user payload lives in the reservation the producer already retains, so bounding the channel would bound ingest rate, not memory. **Amend**: update Appendix A's row and the §10/§11 wording to describe the unbounded lane + per-wake batch cap. Do **not** reintroduce a bounded insertion channel (it would either break split-detach semantics or resurrect the forwarder pattern).

## 2. Zero-duration budgets: observation waits are predicate-first, not short-circuit

[...]

PR #234, opened 2026-08-12
Complete Stage 4 semantic contract

ralexstokes:
## Summary

- recast SPEC §§6, 7, 10, 11, and 14.2 as numbered reducer invariants and transition rules, with named checked evidence or explicit unverified host preconditions
- introduce `DeadlineBudget` as the public deadline type and pin each zero-width API to no-attempt, poll-once, or immediate-escalation behavior
- make graceful child shutdown structurally non-zero, simplify readiness-default inheritance, and make dynamic-scope capability access explicit through `as_scope()`
- reconcile the Stage 4 public surface: correct the unbounded admission-lane contract, apply the remaining small simplifications, and retain `#[non_exhaustive]` only for four documented additive axes

## Stack

This work was branched from #233. That PR merged while this change was in progress, so this branch was rebased onto its merge commit and the PR now targets `main`; the diff contains only Stage 4.

## Verification

- `./tools/dev just ci`
- `./tools/dev just ci-nix`

Addresses Stage 4 of #214 and the semantic drift tracked in #208 and #211.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ A task-first application needs no actor merely to obtain supervision:
 ```rust
 use std::time::Duration;
 
-use shelterwood::{DeadlineBudget, ExitError, TaskOnceDef, Tree};
+use shelterwood::{ExitError, TaskOnceDef, Tree};
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -45,9 +45,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     system.wait_started().await?;
 
     // The owner drives bounded teardown and waits for every child to join.
-    system
-        .shutdown(DeadlineBudget::new(Duration::from_secs(5)))
-        .await?;
+    system.shutdown(Duration::from_secs(5)).await?;
     Ok(())
 }
 ```
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ A task-first application needs no actor merely to obtain supervision:
 ```rust
 use std::time::Duration;
 
-use shelterwood::{ExitError, TaskOnceDef, Tree};
+use shelterwood::{DeadlineBudget, ExitError, TaskOnceDef, Tree};
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -45,7 +45,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     system.wait_started().await?;
 
     // The owner drives bounded teardown and waits for every child to join.
-    system.shutdown(Duration::from_secs(5)).await?;
+    system
+        .shutdown(DeadlineBudget::new(Duration::from_secs(5)))
+        .await?;
     Ok(())
 }
 ```
--- a/SPEC.md
+++ b/SPEC.md
@@ -777,7 +777,7 @@ Delivery is at-most-once (§1 principle 6). Send flavors on `ActorRef`
   fn call<T: Send + 'static>(
       &self,
       make_msg: impl FnOnce(Reply<T>) -> M,
-      deadline: DeadlineBudget,              // trailing, per Appendix B
+      deadline: impl Into<DeadlineBudget>,   // trailing, per Appendix B
   ) -> impl Future<Output = Result<Replied<T>, CallError>>;
   // Replied<T>: the reply value plus the accepting Incarnation (§3.3)
   ```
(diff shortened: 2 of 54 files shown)


ralexstokes:
## Review

Reviewed the code and the spec recast separately. No reachable bugs in the diff; the behavior changes are all deliberate and pinned. Four things landed as fixes in a86ef26, all in the zero-width deadline area.

### The zero-behavior seam did not carry its weight

`select_zero_budget_behavior(budget, behavior)` was `budget.is_zero().then_some(behavior)`, and all five call sites discarded the returned value to test `.is_some()`/`.is_none()`. So `ZeroBudgetBehavior` named which contract a site implemented without anything checking that it did — a comment in type clothing, plus a `doc(hidden) pub` cross-crate export and a unit test asserting the identity function. Collapsed to `is_zero()` with the behavior named in a comment pointing at Appendix B's table, which is the actual inventory.

`wait_for_child`'s `poll_once` flag went with it: `Deadline::after(now, ZERO)` returns `Some(now)` via the explicit zero carve-out, so `expires.is_due(now')` three lines below was already true on the first pass. That rule is what the whole poll-once contract rests on, so it now has a direct unit test in `deadline.rs` instead of being an unstated coincidence.

### `impl Into<DeadlineBudget>` at the public boundary

The budget type permits zero and has `From<Duration>`, so it enforces no invariant — its value is being the one name for the semantics and the anchor for Appendix B's table. Demanding the exact type at every call site bought nothing for that and cost 330 `.into()`s, including the README's headline example:

```rust
system.shutdown(DeadlineBudget::new(Duration::from_secs(5))).await?;
```

Public surfaces now take `impl Into<DeadlineBudget>` (the house idiom, cf. `id: impl Into<ChildId>` on the same method). `Duration` reads naturally again, `DeadlineBudget::ZERO` still works where you want to be explicit, and the conformance test pins both forms.

### `BoundedReadinessDeadline` was already `NonZeroDuration`

Same shape: private-field newtype over a validated non-zero `Duration`, constructed through a fallible associated fn, one accessor. It now carries a `NonZeroDuration`, so the policy surface has one validated-non-zero type and one validation site rather than two idioms for the same property. Public shape unchanged (`duration()` is still `const fn -> Duration`).

`Shutdown::Graceful { grace: NonZeroDuration }` stays as landed — it deletes the `AfterGrace`/`WithinGrace` provenance corner instead of documenting it, which is the point of the stage. Note `Backoff` and `Intensity` still validate non-zero at their constructors with sealed payload *fields*; that is a defensible difference (no public field to carry the invariant) but worth a deliberate look in the release-tagging pass.

### Spec

I checked every qualified test name the recast cites as evidence — all 75 resolve to a real `fn`, and every module prefix matches where that function actually lives. Three corrections:

1. **The immediate-escalation row cited tests that do not discriminate it.** `zero_timeout_reports_recursive_straggler_paths_and_joins_them` and `start_or_shutdown_rollback_timeout_preserves_the_startup_cause_and_stragglers` both use a `pending()` child, so both stay green under the old `runtime::timeout(ZERO, ..)` spelling that granted the operation its one poll. I reverted `shutdown_scope` to that spelling and ran the suite: the only failure is `disposal::zero_shutdown_reports_the_live_child_but_detaches_blocking_factory_disposal`, which this PR rewrote for exactly that purpose. The row now cites it first and says why.

2. **B.9's forgery sentence became false as written.** "only the library can construct the payloads" was true while `IntensityTrip` and `StartupFailureCause` were `#[non_exhaustive]`; with the attribute gone and the fields `pub` (and `Exit::new` public), an application can build the payload values. The security property is unaffected — authentication lives in `ExitError`'s private provenance, as the type docs already say correctly — so the sentence now says only the library can *authenticate* a payload into an `ExitError`, and records that constructibility is deliberate.

3. Appendix B and the two signature sketches record the accepted parameter shape.

The rest of the recast holds up. Worth noting what the behavior change actually is, since it is easy to miss under the recast: `shutdown(ZERO)` on a scope that would settle on the very next poll now returns `Err(ShutdownTimeout)` and forces, where it previously could return `Ok`. Deliberate, spec'd, and pinned — just a real observable change.

### Smaller

- `LifecycleEventKind`'s non-exhaustive rationale sat between the `derive` and the item, so it rendered as a second doc paragraph out of order; moved to match the three in `policy.rs`.
- `reply_receiver_reports_drop_and_is_safe_to_abandon` now carries the zero-budget `recv` pin, which the name no longer advertises. Left alone, but a rename would help the next reader.

Verification: `just ci` (618/618) and `just ci-nix`.

PR #242, opened 2026-08-13
Pin product behavior edge cases

### LINKED ISSUE #213: Product test gaps: nine one-test-each pins (readiness arm, exhausted-id recovery, poison shutdown, cancellation path, ...)
Small, one-test-each gaps in product behavior coverage, from the 2026-08-11 review round. Each pins a behavior that currently has no direct test; a regression would surface only as a rare production anomaly.

1. **Engine `ReadinessGate` un-signaled exit arm** (`engine.rs:499-509`): `(Waiting, Exit { signal_seen: false }) -> Disarmed` — the common production case — shares a match arm with `Shutdown` and is exercised by no engine unit test (`engine.rs:1421` covers `signal_seen: true` and `Shutdown` only). Two assertions in the existing test.

2. **identity.rs exhausted-id recovery** (`identity.rs:272-283, 422-429`): the poisoned branch of `issued()` exists for one scenario — generation counter exhausts, the still-live membership at `MAX-1` later terminalizes, `evict` must match so the id becomes mintable again under a fresh incomparable lineage. Exhaustion and eviction are tested separately, never combined; a regression silently makes an exhausted id permanently stuck. ~10-line test.

3. **`request_shutdown_ignoring_poison`** (`cells.rs:1715-1723`): exists solely so `SystemRun::drop` doesn't abort on an already-unwinding thread with a poisoned control mutex — no test poisons the mutex and exercises the entry point.

4. **Driver-task cancellation path** (`driver.rs:107-154`, `classify_root_driver_join` + the `spawn_system` monitor): the cancelled-root-driver branch (runtime torn down mid-run) publishing `finish_live_root_incarnation` so `wait_stopped` observers unblock has no test. A test dropping a dedicated runtime mid-run and asserting the root reaches `Stopped` pins this and the disposal-teardown window (see the runtime-teardown disposal issue).

5. **Multi-waiter FIFO promotion through the mailbox receive path** (`mailbox/cell.rs:703-714, 1007-1034`): promotion at bind with multiple waiters and single-waiter promotion after receive are tested; no test drives `receive` freeing one slot at a time with 2+ parked senders (capacity 1, three parked senders, assert delivery order), including the Full→Full refill intermediate state. (Loom exploration tracked separately on #152.)

6. **Timeout-flavored terminality race** (`mailbox/cell.rs:1391` pins only the cancellation flavor): mirror `cancellation_can_race_detached_terminal_teardown` with a retained teardown plus an expired `send_timeout`, asserting `TimedOut` — freezing the adjudicated arbitration (SPEC defines `TimedOut` structurally by successful withdrawal; verified spec-consistent).

7. **`fire_silently`/`notify` deferred-wake contract** (`runtime/sync.rs:96-120`): load-bearing for the admission-control observation gate; only the combined `fire()` is tested. Assert: after `fire_silently()` alone, `is_fired()` is true, a parked waiter stays parked, and a subsequent `notify()` releases it.

8. **`CompletionGatedLatch::completed()`** (`sync.rs:220-222`): no test awaits it (neither parked-then-woken nor already-completed fast path).

9. **`JitterSample::from_u64_ratio(n, 0)`** (`policy.rs:138-140`): the zero-denominator edge of this public constructor (NaN/inf → 0.0 totality) has no assertion; `policy.rs:965` covers `0/MAX` and `MAX/MAX` only.

Also worth one comment each while in the area (load-bearing invariants a future simplification pass could break — candidates for #192's invariant list): `raw.rs:1416` (recv's drain contract relies on every driver token-firing site freezing the mailbox first — `driver/child.rs:714-716`, `driver.rs:247-253`); `tree/admission.rs:157-181` (`Admission::poll` replaces state only after the fallible operation so an unwind still reaches Drop's annul path); `runtime/sync.rs:328-331` (comment misattributes which poll armed the waker).

ralexstokes:
## Summary

- add direct regression coverage for all nine product-behavior gaps from #213
- exercise the cancelled root-driver monitor with a dedicated-runtime teardown fixture
- clarify the accepted timeout-versus-terminality arbitration in public docs and code comments
- record the three load-bearing invariants called out in the issue

## Why

These paths were implemented but only indirectly covered, so regressions could surface as rare production anomalies. The timeout race was also easy to misread as a lost-result bug even though the existing arbitration is intentional and spec-consistent.

Closes #213.

## User and developer impact

Runtime behavior is unchanged. The new tests pin readiness disarming, exhausted-id recovery, poison-tolerant shutdown, driver cancellation, FIFO waiter promotion, timeout arbitration, deferred latch wakes, completion waits, and zero-denominator jitter normalization. Public retry guidance now explains why an expired timeout can win before deferred terminal waiter discharge and what a retry observes.

## Validation

- `./tools/dev just ci`
- `./tools/dev just ci-nix`

### DIFF
--- a/crates/shelterwood/src/driver/tests/shutdown.rs
+++ b/crates/shelterwood/src/driver/tests/shutdown.rs
@@ -4,29 +4,7 @@ use super::support::*;
 async fn runtime_teardown_finishes_the_cancelled_root_driver() {
     let plan = DynamicTree::new().lower_for_test();
     let root = Arc::clone(&plan.root);
-    let (driver_sender, driver_receiver) = std::sync::mpsc::sync_channel(1);
-    let (teardown_sender, teardown_receiver) = std::sync::mpsc::sync_channel(0);
-    let runtime_thread = std::thread::spawn(move || {
-        let runtime = tokio::runtime::Builder::new_multi_thread()
-            .worker_threads(2)
-            .enable_all()
-            .build()
-            .expect("dedicated driver runtime");
-        let guard = runtime.enter();
-        let driver = crate::runtime::spawn(async move { run_scope(plan, ScopeRole::Root).await });
-        drop(guard);
-        driver_sender
-            .send(driver)
-            .expect("the monitor receives the root driver");
-        teardown_receiver
-            .recv()
-            .expect("the test requests runtime teardown");
-        drop(runtime);
-    });
-
-    let driver = driver_receiver
-        .recv()
-        .expect("dedicated runtime publishes the driver handle");
+    let (hosted, driver) = DedicatedRuntime::spawn(run_scope(plan, ScopeRole::Root));
     let monitor = monitor_root_driver(Arc::clone(&root), driver);
     root.wait_started().await.expect("root starts");
     let wakes = Arc::new(AtomicUsize::new(0));
@@ -40,10 +18,7 @@ async fn runtime_teardown_finishes_the_cancelled_root_driver() {
         "the live root observer parks before runtime teardown"
     );
 
-    teardown_sender
-        .send(())
-        .expect("dedicated runtime remains live");
-    runtime_thread.join().expect("runtime teardown completes");
+    hosted.shutdown().await;
 
     assert!(matches!(
         crate::runtime::timeout(DRIVER_PROGRESS_WAIT, crate::runtime::join(monitor)).await,
--- a/crates/shelterwood/src/driver/tests/shutdown.rs
+++ b/crates/shelterwood/src/driver/tests/shutdown.rs
@@ -1,5 +1,52 @@
 use super::support::*;
 
+#[crate::runtime::test(flavor = "multi_thread", worker_threads = 2)]
+async fn runtime_teardown_finishes_the_cancelled_root_driver() {
+    let plan = DynamicTree::new().lower_for_test();
+    let root = Arc::clone(&plan.root);
+    let (driver_sender, driver_receiver) = std::sync::mpsc::sync_channel(1);
+    let (teardown_sender, teardown_receiver) = std::sync::mpsc::sync_channel(0);
+    let runtime_thread = std::thread::spawn(move || {
+        let runtime = tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(2)
+            .enable_all()
+            .build()
+            .expect("dedicated driver runtime");
+        let guard = runtime.enter();
+        let driver = crate::runtime::spawn(async move { run_scope(plan, ScopeRole::Root).await });
+        drop(guard);
+        driver_sender
+            .send(driver)
+            .expect("the monitor receives the root driver");
+        teardown_receiver
+            .recv()
+            .expect("the test requests runtime teardown");
+        drop(runtime);
+    });
+
+    let driver = driver_receiver
+        .recv()
+        .expect("dedicated runtime publishes the driver handle");
+    let monitor = monitor_root_driver(Arc::clone(&root), driver);
+    root.wait_started().await.expect("root starts");
+
+    teardown_sender
+        .send(())
+        .expect("dedicated runtime remains live");
+    runtime_thread.join().expect("runtime teardown completes");
+
+    assert!(matches!(
+        crate::runtime::timeout(Duration::from_secs(1), root.wait_stopped()).await,
+        crate::runtime::Timeout::Completed(StopReason::ShutdownRequested)
+    ));
+    assert!(matches!(
+        crate::runtime::join(monitor).await,
+        crate::runtime::JoinOutcome::Ok {
+            value: StopReason::ShutdownRequested
+        }
+    ));
+}
+
 #[crate::runtime::test]
 async fn latched_shutdown_upgrades_an_intensity_drain() {
     let mut tree = Tree::new();
--- a/crates/shelterwood/src/driver/tests/shutdown.rs
+++ b/crates/shelterwood/src/driver/tests/shutdown.rs
@@ -29,6 +29,16 @@ async fn runtime_teardown_finishes_the_cancelled_root_driver() {
         .expect("dedicated runtime publishes the driver handle");
     let monitor = monitor_root_driver(Arc::clone(&root), driver);
     root.wait_started().await.expect("root starts");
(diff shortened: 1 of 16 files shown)


ralexstokes (commented):
Reviewed the full diff and verified the pins by mutating the code they claim to pin, rather than by reading the test names.

## Verification

`./tools/dev just ci` green (622 tests, includes the packaged-docs sync check, so both copies of `retry-and-ordering.md` match). `runtime_teardown_finishes_the_cancelled_root_driver` ran 30/30 clean.

Each pin was checked against a deliberate regression; all six failed as they should:

| Mutation | Test that caught it |
| --- | --- |
| `WaiterQueue::pop_front` → `pop_last` (LIFO promotion) | `receive_promotes_multiple_parked_senders_in_fifo_order` |
| `request_shutdown_ignoring_poison` → `.expect(...)` on the control lock | `destructor_shutdown_tolerates_a_poisoned_control_mutex` |
| `FenceCounter::issued` drops its poison branch | `terminal_eviction_recovers_an_exhausted_id_with_a_fresh_lineage` |
| `fire_silently` also notifies | `silent_fire_defers_a_parked_waiters_wake_until_notify` |
| `(Waiting, Exit { signal_seen: false })` moved to the `None` arm | `readiness_configuration_and_signal_deadline_race_are_engine_owned` |
| monitor drops `finish_live_root_incarnation` | `runtime_teardown_finishes_the_cancelled_root_driver` |

The FIFO test also pins the Full→Full refill intermediate the issue asked for (`waiters.len() == remaining` between deliveries), and the promotion *wake* stays covered by `promotion_wakes_every_sender_when_multiple_wakers_panic`, so the noop waker there is fine.

The three invariant comments hold against the code:

- `raw.rs`: both token-firing sites do freeze first — `begin_stop_child` freezes before the ladder reaches `StopAction::Cancel` (`driver/child.rs`), and `ScopeRuntime::drop` freezes and closes before `active.shutdown.fire()` (`driver.rs`). The scope-level `latches.*.shutdown.fire()` sites are nested-scope roles with no mailbox, so they don't weaken the claim.
- `tree/admission.rs`: `pending.start()` does run with `Unpolled` still installed, so an unwind reaches Drop's annul arm.
- `sync.rs`: the corrected attribution is right — in the `SENDING | SENT` branch it is that branch's own `poll` that registers the waker; the pre-CAS poll never ran.

`SendErrorKind::TimedOut` / `CallErrorKind::AcceptanceTimedOut` doc claim checks out: the `BindingStatus::Terminal` arm of `try_send` carries `final_incarnation`, which is `None` exactly when the mailbox was never bound — so "when one existed" is the correct qualification, and the arbitration matches SPEC's rule that the tie is decided by the withdrawal race, never by clock comparison. The `monitor_root_driver` extraction is behavior-preserving (same clone, same spawn order).

Also confirmed the poison test isn't half a pin: `ObservationGate::lock` is already poison-tolerant, so the drop path has no second poison-fatal lock behind it.

## Findings

**1. The new driver test is the only direct `tokio::` reference outside `shelterwood-runtime`.** `crates/shelterwood/src/driver/tests/shutdown.rs:10` builds a `tokio::runtime::Builder` directly; every other line in every crate's `src/` goes through `crate::runtime`. tokio is a dev-dependency of `shelterwood`, so this compiles, but the point of the runtime crate is that it is the single place that names tokio — a dedicated-runtime fixture in `shelterwood-runtime` behind `test-util` would keep that boundary intact and give the "spawn on a runtime, then tear it down" shape one home. Same fixture would also encapsulate the blocking `std::sync::mpsc::recv` / `thread::join` inside the async test body — benign here (the other side is a real OS thread and `worker_threads = 2`), but it is a landmine for anyone who later changes the flavor.

**2. The PR head is not on current main.** `d4e3284` is based on `349a00b`; `origin/main` is now `f1000f7` (#243). Per the combined-tree rule this needs updating before merge — my local copy of the branch is already rebased onto `f1000f7` and `just ci` is green there, so this is mechanical.

Nothing else — the nine gaps from #213 and all three comment requests are covered, and behavior is unchanged.

crates/shelterwood/src/driver/tests/shutdown.rs
```diff
@@ -1,5 +1,72 @@
 use super::support::*;
 
+#[crate::runtime::test(flavor = "multi_thread", worker_threads = 2)]
+async fn runtime_teardown_finishes_the_cancelled_root_driver() {
+    let plan = DynamicTree::new().lower_for_test();
+    let root = Arc::clone(&plan.root);
+    let (driver_sender, driver_receiver) = std::sync::mpsc::sync_channel(1);
+    let (teardown_sender, teardown_receiver) = std::sync::mpsc::sync_channel(0);
+    let runtime_thread = std::thread::spawn(move || {
+        let runtime = tokio::runtime::Builder::new_multi_thread()
```
ralexstokes: This is the only direct `tokio::` reference outside `shelterwood-runtime` — every other line under any crate's `src/` reaches the runtime through `crate::runtime`. It compiles (tokio is a dev-dependency here), but it pierces the crate boundary that makes `shelterwood-runtime` the single place that names tokio. A `test-util` fixture in `shelterwood-runtime` for "spawn onto a dedicated runtime, hand back the handle, tear it down on demand" would keep the boundary intact and would also encapsulate the blocking `recv`/`join` in this async test body — harmless at `worker_threads = 2`, but a trap for whoever changes the flavor later.

ralexstokes: Addressed in `70f3edb`: the shape now lives in `shelterwood-runtime` as `DedicatedRuntime`, so no crate outside the adapter names tokio again (`grep -rn "tokio::" crates/*/src` is clean outside `shelterwood-runtime/`). `test-util` now pulls `tokio/rt-multi-thread` itself rather than relying on feature unification from a sibling's dev-dependencies — verified by dropping it and watching `cargo check -p shelterwood-runtime --features test-util` fail on `new_multi_thread`.

The fixture also removes the blocking handshake: the hosted task is spawned before its runtime moves to the teardown thread (so no handle channel, no blocking `recv`), and `shutdown().await` offloads the thread join, so nothing blocks a worker in the async body. The test is 24 lines shorter and still fails when the monitor drops `finish_live_root_incarnation`; 25/25 clean on repeat, and `just ci` is green.

PR #248, opened 2026-08-13
Simplify remaining identity and runtime plumbing

### LINKED ISSUE #211: Small simplifications batch: identity lineage, raw/cells/observe items, Option<ReadinessDeadline>, visibility tightening

ralexstokes:
## Summary

- collapse incarnation generation onto the shared poison-aware counter and centralize membership grants
- make `ScopeCell` identity operations explicit while simplifying removal, gate handoff, and snapshot assembly
- share raw disposal/signal state, use one RAII offload path, and avoid hashing replacement timer keys twice

## Why

This completes the remaining low-risk simplifications from #211 while preserving identity exhaustion, observation, panic-containment, and disposal behavior.

## Validation

- `./tools/dev just ci`
- `./tools/dev just ci-nix`

Closes #211

### DIFF
--- a/crates/shelterwood-cells/src/cells.rs
+++ b/crates/shelterwood-cells/src/cells.rs
@@ -762,16 +762,7 @@ impl Drop for ResidentChild {
         let Some(parent) = completion.parent.upgrade() else {
             return;
         };
-        parent.with_observation_gate(|wakes| {
-            parent.emit_locked(
-                wakes,
-                LifecycleEventKind::Removed {
-                    id: completion.projection.member.id().clone(),
-                    membership: completion.projection.member.membership(),
-                    last_incarnation: completion.projection.member.record().last_incarnation,
-                },
-            );
-        });
+        parent.with_observation_gate(|txn| Self::publish_removal(completion, txn));
     }
 }
 
@@ -840,7 +831,7 @@ struct ScopeObservation {
 pub struct ScopeCell {
     pub member: Arc<MemberCell>,
     pub flavor: ScopeFlavor,
-    pub child_identity: Mutex<ScopeIdentity>,
+    child_identity: Mutex<ScopeIdentity>,
     control: Mutex<ScopeControl>,
     dynamic_route: Mutex<Option<Arc<ErasedDynamicRoute>>>,
     observation: ScopeObservation,
@@ -878,6 +869,24 @@ pub struct RuntimeStorage {
 }
 
 impl ScopeCell {
+    pub fn mint_membership(&self, id: &ChildId) -> Option<MintedMembership> {
+        self.child_identity
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner)
+            .mint_membership(id)
+    }
+
+    pub fn adopt_or_mint_membership(
+        &self,
+        id: &ChildId,
+        provisional: Membership,
+    ) -> Option<Option<MintedMembership>> {
+        self.child_identity
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner)
+            .adopt_or_mint_membership(id, provisional)
+    }
+
     pub fn evict_child_identity(&self, member: &MemberCell) {
         self.child_identity
             .lock()
@@ -1113,15 +1122,11 @@ impl ScopeCell {
             .map(|resident| resident.projection.clone())
             .collect::<Vec<_>>();
         for descendant in descendants {
+            descendant
+                .member
+                .install_observation_gate_locked(previous, gate);
             if let Some(scope) = descendant.scope {
-                descendant
-                    .member
-                    .install_observation_gate_locked(previous, gate);
                 scope.adopt_descendant_observation_gates_locked(previous, gate, _txn);
-            } else {
-                descendant
-                    .member
-                    .install_observation_gate_locked(previous, gate);
             }
         }
     }
@@ -1389,14 +1394,10 @@ impl ScopeCell {
             .config
             .lock()
             .unwrap_or_else(std::sync::PoisonError::into_inner);
-        let children = self
-            .current_children()
-            .iter()
-            .map(|resident| resident.projection.clone())
-            .collect::<Vec<_>>();
+        let children = self.current_children();
         let children = children
             .iter()
-            .map(|child| self.child_snapshot_locked(child))
+            .map(|resident| self.child_snapshot_locked(&resident.projection))
             .collect::<Vec<_>>();
         Arc::new(ScopeSnapshot {
             state: record.state,
--- a/crates/shelterwood-cells/src/cells.rs
+++ b/crates/shelterwood-cells/src/cells.rs
@@ -836,8 +836,6 @@ pub struct ScopeCell {
     dynamic_route: Mutex<Option<Arc<ErasedDynamicRoute>>>,
     observation: ScopeObservation,
     #[cfg(any(test, feature = "test-util"))]
-    child_identity_reconciliations: AtomicUsize,
-    #[cfg(any(test, feature = "test-util"))]
     ancestor_parent_reads: AtomicUsize,
     #[cfg(any(test, feature = "test-util"))]
     runtime_storage: Mutex<RuntimeStorage>,
@@ -883,18 +881,10 @@ impl ScopeCell {
         id: &ChildId,
         provisional: Membership,
     ) -> Option<Option<MintedMembership>> {
-        #[cfg(any(test, feature = "test-util"))]
-        self.child_identity_reconciliations
-            .fetch_add(1, Ordering::Relaxed);
         self.child_identity
             .lock()
             .unwrap_or_else(std::sync::PoisonError::into_inner)
             .adopt_or_mint_membership(id, provisional)
-    }
-
-    #[cfg(any(test, feature = "test-util"))]
-    pub fn child_identity_reconciliations(&self) -> usize {
-        self.child_identity_reconciliations.load(Ordering::Relaxed)
     }
 
     pub fn evict_child_identity(&self, member: &MemberCell) {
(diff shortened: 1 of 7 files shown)


ralexstokes:
Reviewed the full diff and mutation-tested the claims that pin behavior. One real finding, addressed in 3df782b; everything else checks out.

## Finding: the rewritten pre-conversion proof was vacuous

`fused_cancellation_overtaking_admission_rejects_before_conversion` claimed the reconciliation count "preserves that proof". It doesn't: `child_identity_reconciliations` is only incremented by `adopt_or_mint_membership`, and nothing on the admission path calls it — `ChildRuntime::from_plan` touches `take_incarnation_counter` and the mailbox, never `child_identity`. Deleting `handle_admission`'s *entire* pre-conversion latch re-check leaves the test green.

In fairness the probe it replaced was equally vacuous — I ran the same mutation against `main` and that test passed too. Conversion never locks `child_identity`, and the eviction that does (`evict_child_identity`) is poison-tolerant, so the poisoned mutex was never reachable. Both variants pinned nothing; the commit message just made the claim explicit.

Fixed by probing the artifact conversion actually produces: `from_plan` issues the membership's incarnation counter to the child runtime and never gives it back, so `member.lock_incarnation_counter().is_some()` after `handle_admission` is real evidence that rejection preceded conversion. The same mutation now fails at that assertion. It's also exactly what the sibling under-lock test parks on, so the two tests now pin the same seam from opposite sides — I corrected that test's stale "parked on the child identity mutex" comment (it parks on the incarnation-counter mutex).

With its only caller gone, `child_identity_reconciliations` (field, ctor init, accessor) is deleted rather than left as an unexercised test hook.

Also added a doc line to `SnapshotHub::subscribe` recording why it still takes an `ObservationTxn` it no longer uses — otherwise the bare `_txn` is a reader hazard right where the pulse used to be.

## Verified, no action needed

- **`SnapshotHub::subscribe` pulse deletion** — sound. `receiver_count()` is tokio's live-receiver count and `watcher()` is `subscribe()`, so the refresh arm runs with zero receivers; the only receiver that could observe the deferred pulse is the one minted immediately after, whose `seen_generation` is read *after* the mint. `changed()` re-parks in every case, including multiple subscribes inside one gate hold.
- **`IncarnationCounter` on a bare `PoisonedCounter`** — `mint` is now a verbatim copy of `FenceCounter::mint` minus the lineage, and `PoisonedCounter::near_exhaustion()` matches `FenceCounter::near_exhaustion(0)`. `FenceCounter` keeps `from_fence`/`issued` for memberships.
- **`OffloadScope` deletion** — `Rejected` returns before the `Guard` is constructed, so no error path can fire a spurious cancellation, and `detach` only clears `armed`. Behavior-identical to the old `None`.
- **`TimerStore::replace` hoisted hash** — the ordering is unchanged in the way that matters (`remove`'s first action was always `hash_key`, so a hostile `Hash` unwinds at the same point relative to disposal). `timer_replacement_hashes_each_incoming_key_once` has teeth: reverting to `remove(key); hash_key(key)` fails it 4 vs 2.
- **`SlotCell::terminalize_never_started_locked` if/else** — `slot.member` and `slot.scope.member` are the same `Arc` (`mint_reserved_slot`), so the old shape really did terminalize twice. The one behavior delta is the `observation.closed` early return skipping member terminalization; every `close_observation_locked` call site is a stop epilogue where the member is already terminal, so `terminalize_locked` would be a no-op there anyway. Now mirrors the non-locked variant exactly, as #211 asked.
- **`snapshot_locked` single pass** — the `current_children` guard is now held across recursion, but `child_snapshot_locked` only ever descends, and `record()`/`options()` are leaf locks released before returning. Lock order stays strictly downward.
- **`BuilderCore::lower`** — dropping the batch-wide identity guard loses only cross-id atomicity, and `ScopeIdentity` has no cross-id invariant (it's a per-`ChildId` counter map); the builder's ids are distinct by `reserve`. Each operation is still individually atomic. Worth knowing, not worth holding.
- **Poison policy unification** — `mint`/`adopt_or_mint` moving from `expect` to `into_inner` matches `evict_child_identity` and is what #211 asked for. `ScopeIdentity` operations can't realistically unwind mid-mutation (`ChildId` is an `Arc<str>`), so the poisoned state is unreachable either way.

`./tools/dev just ci` clean on the pushed tip (629 tests).

PR #249, opened 2026-08-13
Consolidate test polling and repeated fixtures

### LINKED ISSUE #212: Test-suite hygiene: two flake risks, assert_eventually helper, ~800-1000 lines of duplication, quality items
From the 2026-08-11 test-suite review (362 tests examined; overall verdict: unusually good shape — consistent paused-clock discipline, no dead helpers, no ignored tests, clean unit/integration layering). Two flake risks, one big helper win, then duplication and quality items.

## Flake risks (the only two found)

1. `tests/readiness.rs:727` — `assert!(runtime_started.load(..))` immediately after `scope.add_task(...).await`; the `Admission` future resolves at driver admission, not first poll of the child body. Passes only via current_thread scheduling order. Fix: `poll_until` (as at line 701) or delete the assert.
2. `acceptance/assistant.rs:583-636` — the redelivery section (not `start_paused`) must fit call + injected crash + supervised restart with backoff + `wait_for_child` + second call inside a 2s real-clock budget with hard asserts on the remaining budget; loaded CI can blow it. shard_store.rs does the same dance with 5s; use the suite's `POLL_TIMEOUT` scale.

## The big one: `assert_eventually` helper (~150-250 lines + better diagnostics)

The 4-6-line `assert!(poll_until(POLL_TIMEOUT, 1ms, || ...)).await` block appears at ~116 sites, and the bare `assert!` loses the predicate text on failure. One `assert_eventually(predicate)` in `tests/common/timing.rs` with `#[track_caller]` collapses each site to 1-2 lines.

## Duplication clusters (~800-1000 removable lines total; refactors of working tests, not defects)

Driver unit tests (`src/driver/tests/`, ~400 lines): (a) the ~28-line plan→`ScopeRuntime` setup block repeated ~13× across exhaustion/shutdown/startup/events (one support.rs builder helper); (b) the 19-line `ChildEvent::Exited` recv-and-destructure into a 6-tuple, 7× (also removes unreadable `exit.0..exit.5` indexing); (c) the dynamic admission handshake ~8× in dynamic.rs; (d) dynamic.rs:1256-1443 nearly duplicates `exercise_coalesced_removal` (dynamic.rs:1099-1249) — a source-enum parameter folds them.

Integration side (~400-500 lines): disposal-thread-id recv/assert pattern ~20 sites in disposal.rs; readiness.rs (5), subtrees.rs (2), actor/lifecycle/deadlines (4) hand-roll what `common/waiting.rs` already exports; the subtrees inherit/reset two-subtree skeleton 5-6×; the events.rs trace-log harness ~10×; the gate-released manual-ready TaskDef shape 4×; the liveness-probe select body triplicated (common/waiting.rs:49, dynamic.rs:132/160); `InertActor` (actor.rs:495) duplicates `ManualActor` (actor.rs:165) byte-for-byte.

## Test-quality items (can pass without testing; cannot flaky-fail)

- `tests/subtrees.rs:483-516` `ordered_graces_sum_while_dynamic_graces_overlap` asserts only `>= 20s` — under `start_paused`, adding `< 25s` is free and makes "sum" distinguishable from worse serialization.
- `acceptance/assistant.rs:819-830` idle-eviction can't distinguish timer stacking from replacement — add a closing `try_recv().is_err()` ("exactly one eviction") plus `assert_quiet`.
- `src/driver/tests/reports.rs:103` `assert!(!readiness.fire(), ...)` is true by construction (the latch was already fired at :101); assert the closed state directly.
- `tests/offloads.rs:780` `run_blocking_returns_on_the_actor_task` never tests task-locality — use the file's own `tokio::task::id()` technique (offloads.rs:53-62) or rename.
- `src/driver/tests/events.rs:21` asserts the `saturated` flag but nothing about the driver loop actually yielding.
- Real-clock quiet windows that regressions could out-wait (all backstopped, none can flake): `tests/dynamic.rs:875,949,978,1089,1122`, `tests/offloads.rs:439`, `tests/subtrees.rs:674-689`; two stress tests can silently stop interleaving under load (`src/driver/tests/observation.rs:641`, `dynamic.rs:852`).
- ~11 sites bound positive events at 1s real time where the suite's own `POLL_TIMEOUT` (5s) costs nothing under nextest parallelism (`tests/disposal.rs:180,730,844,891,991,1066`; `tests/dynamic.rs:295,592,608,1608`; `tests/offloads.rs:968`; plus `observation.rs:23`, `lifecycle.rs:119-133`, `sidecar.rs:166-171,223-226`, `drain.rs:513`).

ralexstokes:
## Summary

- add a source-reporting `assert_eventually!` helper and migrate 101 repeated polling assertions
- consolidate driver exit/admission fixtures plus integration waiting, disposal, liveness, and inherit/reset setup
- strengthen timing, exact-once, task-locality, latch-state, and concurrency coverage while widening positive real-clock backstops to the suite-wide budget

The two original flake fixes and end-to-end driver-yield coverage had already landed on `main`; this completes the remaining #212 cleanup. The reviewed branch removes 555 net lines.

## Validation

- `./tools/dev cargo nextest run -p shelterwood --test integration` (321 passed)
- `./tools/dev just ci` (629 tests passed)
- `./tools/dev just ci-nix`

Closes #212

### DIFF
--- a/crates/shelterwood-runtime/src/sync.rs
+++ b/crates/shelterwood-runtime/src/sync.rs
@@ -191,6 +191,14 @@ impl CompletionGatedLatch {
         )
     }
 
+    #[cfg(any(test, feature = "test-util"))]
+    pub fn is_completed(&self) -> bool {
+        matches!(
+            self.state.load(Ordering::Acquire),
+            COMPLETION_GATE_CLOSED | COMPLETION_GATE_CLOSED_FIRED
+        )
+    }
+
     pub async fn fired(&self) {
         self.fired.fired().await;
     }
--- a/crates/shelterwood/src/driver/tests/dynamic.rs
+++ b/crates/shelterwood/src/driver/tests/dynamic.rs
@@ -657,15 +657,7 @@ async fn admission_conversion_panic_does_not_poison_dynamic_cleanup() {
     reservation
         .slot
         .define(ChildConstruction::Task(TaskDef::new(|_| future::pending())));
-    let response = super::super::start_admission(
-        Arc::clone(&reservation.control),
-        Arc::clone(&reservation.slot),
-        None,
-    )
-    .expect("admission starts inside the runtime");
-    let Some(DriverEvent::Admission(request)) = event_receiver.recv().await else {
-        panic!("admission enqueueing submits the request")
-    };
+    let (response, request) = begin_admission(&reservation, &mut event_receiver, None).await;
 
     assert!(
         catch_unwind(AssertUnwindSafe(|| {
@@ -723,15 +715,8 @@ async fn annulment_before_admission_owns_never_started_terminality() {
     reservation
         .slot
         .define(ChildConstruction::Task(TaskDef::new(|_| future::pending())));
-    let response = super::super::start_admission(
-        Arc::clone(&reservation.control),
-        Arc::clone(&reservation.slot),
-        None,
-    )
-    .expect("admission starts inside the runtime");
-    let Some(DriverEvent::Admission(request)) = dynamic_event_receiver.recv().await else {
-        panic!("admission enqueueing submits the request")
-    };
+    let (response, request) =
+        begin_admission(&reservation, &mut dynamic_event_receiver, None).await;
 
     cancel_dynamic_reservation(
         &reservation.scope,
@@ -809,15 +794,8 @@ async fn annulment_after_promotion_is_inert_and_supervision_owns_the_exit() {
                 }
             }
         })));
-    let mut response = super::super::start_admission(
-        Arc::clone(&reservation.control),
-        Arc::clone(&reservation.slot),
-        None,
-    )
-    .expect("admission starts inside the runtime");
-    let Some(DriverEvent::Admission(request)) = dynamic_event_receiver.recv().await else {
-        panic!("admission enqueueing submits the request")
-    };
+    let (mut response, request) =
+        begin_admission(&reservation, &mut dynamic_event_receiver, None).await;
     scope.handle_admission(request);
     assert!(matches!(response.try_receive(), Some(Ok(()))));
     assert!(matches!(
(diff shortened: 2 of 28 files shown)


ralexstokes:
Reviewed and pushed fixes in ebfde7a. The consolidation itself is sound — I checked the mechanical rewrites are token-identical to what they replaced, and the fixture extractions (`recv_child_exit`/`ObservedExit`, `begin_admission`, `gate_released_manual_ready_task`, `liveness_probe`, `add_inherit_reset_subtrees`, `ManualActor` for `InertActor`) all preserve behaviour. `assert_eventually!` reports well: predicate source, caller location, and lazily-evaluated context all land in the failure message.

Findings, in order of weight:

**1. The new task-locality assertion could not fail the test.** `offloads.rs` `run_blocking_returns_on_the_actor_task` gained `assert_eq!(tokio::task::id(), actor_task)` inside `BlockingActor::handle`. Inverting it to `assert_ne!` still passes: supervision contains the handler panic, `observed` was already stored, and the one-shot actor terminalizes with `StopReason::Finished` either way. So #212's "never tests task-locality" item was not actually closed.

Fixed by publishing the evidence instead of judging it in the handler — both task ids plus the blocking token's cancellation flag go into an `Arc<Mutex<Option<BlockingObservation>>>`, and the test body compares them. Re-ran the same inversion against the fix; it now fails as it should.

(Pre-existing, not touched: `ZeroDeadlineActor` at `offloads.rs:53-62` has the same shape — its in-continuation asserts are only detectable as a hang, since a panic there means `ZeroMessage::Done` is never delivered and `context.stop()` never runs.)

**2. All 100 migrated sites were left with stale indentation.** The rewrite kept the body indented for the `poll_until(...)` nesting it replaced and crammed `}).await;` onto one line. This is invisible to CI: `cargo fmt` cannot see `crates/shelterwood/tests/**` at all, because its modules are declared inside `register_test_modules!` rather than as real `mod` items — I confirmed by dropping deliberately malformed code into `drain.rs` and watching `cargo +nightly fmt --all -- --check` still exit 0. (`src/driver/tests/` is fine; it's declared with a plain `mod tests;`.)

Re-laid-out every site the way rustfmt would, imported the macro per file instead of repeating `crate::common::` 100 times, and verified the result is token-identical to the pushed version modulo whitespace and the removed braces on collapsed single-expression predicates. Two of the PR's over-100-column lines went away; no new ones.

The rustfmt blind spot itself predates this PR and is worth its own issue — happy to file it.

**3. One hand-rolled poll left behind.** `dynamic.rs:1539` still spelled out `poll_until(Duration::from_secs(5), Duration::from_millis(1), ...)` — the same budget and interval as the helper, just hardcoding the value rather than naming `POLL_TIMEOUT`. Converted. The other three remaining `poll_until` calls are genuinely different (9s/1s drain window, 50ms and 5s bounded probes) and stay as they are.

**4. Over-claiming `expect` messages in the annulment race.** `driver/tests/dynamic.rs` — the rendezvous is a real improvement over the barrier (holding the dynamic-state guard until both contenders are running is what forces the overlap), but the two receives share one channel, so neither names a contender, and each send happens before the lock attempt rather than "at the mutex boundary". Reworded, with a comment stating what the guard vs. the rendezvous each buy.

Verified: `./tools/dev just ci` green (629 tests), `cargo +nightly fmt --all -- --check` clean, 485 shelterwood tests pass.

**Scope note, no change made:** #212's "events.rs trace-log harness ~10x" item is untouched — `tests/events.rs` isn't in the diff. Each of those actors has a genuinely different handler body, so the shared part is thin and I'd call skipping it defensible; but with `Closes #212` on the PR, either mention it in the body or leave a residual comment on the issue before merging.

PR #250, opened 2026-08-13
Handle blocking work rejection during runtime shutdown

### LINKED ISSUE #205: Runtime-teardown disposal runs user destructors inline on the submitting thread (shut-down blocking pool)

### LINKED ISSUE #222: spawn_blocking_work has no rejection handling: a shut-down blocking pool drops the user closure on the submitting thread and join_resuming panics
**Severity: low — teardown-window only; same class as #205, different call site.** (Found while adversarially reviewing #219.)

## Mechanism

`spawn_blocking_work` (`crates/shelterwood/src/runtime/spawn.rs:63-68` → `spawn_blocking`, `spawn.rs:126-133`) calls `tokio::task::spawn_blocking` with no handling for a synchronously rejected submission. That is the exact Tokio behavior #205 documented and #219 fixed for the disposal path:

```rust
// tokio-1.53.1 runtime/blocking/pool.rs:393-404
fn spawn_task(&self, task: Task, rt: &Handle) -> Result<(), SpawnError> {
    let mut shared = self.inner.shared.lock();
    if shared.shutdown {
        task.task.shutdown();          // destroys the closure, synchronously, here
        return Err(SpawnError::ShuttingDown);
    }
    ...
}
```

`Spawner::spawn_blocking` maps `Err(SpawnError::ShuttingDown)` back to `Ok`-shaped output — it still returns the `JoinHandle`. So when the blocking pool is already shutting down while `Handle::try_current()` still succeeds:

1. **User destructors run on the submitting thread.** The only caller today is `RawContext::run_blocking` (`crates/shelterwood/src/raw.rs:1386`), whose closure is `F: FnOnce(CancellationToken) -> T + Send + 'static` — user-authored and holding whatever the user captured. Tokio drops that closure inline, so every captured user destructor runs on the submitting thread. Unlike the disposal path, there is no `catch_panic` boundary here either, so such a destructor can also unwind into the submitting task rather than merely block it.

2. **The awaiting future panics.** The rejected task's `JoinHandle` resolves to a cancelled `JoinError`, and `join_resuming` (`crates/shelterwood/src/runtime/spawn.rs:150-160`) turns that into `panic!("library-owned operation task was unexpectedly cancelled")`. That message asserts an invariant that a shutting-down blocking pool legitimately violates, so the diagnostic is misleading even where the panic itself is harmless.

## Evidence

The rejection behavior itself is pinned end-to-end by #219's `runtime::disposal::tests::shut_down_blocking_pool_falls_back_off_the_teardown_thread` and `embedder_runtime_teardown_isolates_a_task_held_value`. During that review I confirmed with a repro that before #219 a hostile destructor reached through this same Tokio path ran inline on the embedder thread and hung `Runtime::shutdown_background` indefinitely (the test process had to be SIGTERM'd at 119s). `spawn_blocking_work` reaches the same Tokio code with no equivalent handling.

Reachability is narrower than the disposal case: `run_blocking` is called from inside an actor poll, so the submitting thread is a runtime worker rather than the embedding host's teardown thread, and the runtime is on its way out. That is why this is low severity rather than a repeat of #205.

## Possible fix

Either detect the rejection (the closure's sole-ownership signal `#219` uses is not directly available here, since the closure is not `Arc`-wrapped) or, more simply, classify the cancellation instead of panicking: have `BlockingWork::join` return a `JoinOutcome`-style value so `run_blocking` can surface "the runtime went away" rather than an invariant panic. Decide whether `run_blocking`'s captured user state deserves the same isolation guarantee the disposal path gives, and state the answer in `docs/shutdown-and-resources.md` either way — the blocking-operations section there currently promises only that the *thread* may outlive the actor, not what happens to a closure that never starts.

Refs #205
Refs #219

ralexstokes:
Closes #222.
Closes #205.

## Summary

- retain each `run_blocking` closure behind a single-owner-detectable job so Tokio cannot destroy captured user state on the submitting thread when its blocking pool rejects a shutdown-time submission
- reroute synchronously rejected work to a detached `shelterwood-blocking` fallback thread and preserve ordinary return-value and panic propagation through an owned completion lane
- isolate pending closure destruction when accepted work is cancelled or the dedicated fallback cannot start, and replace the misleading internal-invariant panic with a runtime-teardown cancellation diagnostic
- contain panics raised while waking the receiver, plus unclaimed result/panic-payload destructors, at the detached worker boundary
- record the `run_blocking` shutdown contract in the public API docs, normative spec, and synchronized shutdown guide, including disposal's final synchronous no-loss fallback when the system cannot create any isolation worker

## Root cause and impact

Tokio returns a `JoinHandle` even when a shutting-down blocking pool synchronously destroys the submitted closure. The old `spawn_blocking_work` passed the user closure directly to Tokio and unconditionally awaited that handle. Rejection therefore destroyed user captures on the submitting runtime thread and later turned the legitimate cancelled handle into `library-owned operation task was unexpectedly cancelled`.

The job wrapper gives the submitter an ownership signal analogous to the disposal fix from #219. A sole, still-pending job means Tokio rejected it synchronously; Shelterwood moves that same operation to a detached native thread. The completion lane also makes a dropped future dispose an unclaimed result or panic payload away from its drop glue. If the dedicated fallback cannot start, the job transfers its closure to disposal and closes the lane; disposal retries isolation independently. Only total native-thread exhaustion reaches disposal's existing synchronous no-loss fallback.

This completes the remaining shutdown-time blocking-submission class paired with #205. The disposal fallback, runtime-lifetime documentation, and root-driver cancellation regression named by #205 are already present on `main`; this PR records the corresponding `run_blocking` decision and closes both issues together as planned.

## Validation

- `./tools/dev cargo test -p shelterwood-runtime spawn::tests -- --nocapture` — 9 passed
- end-to-end regression occupies Tokio's only blocking worker, begins shutdown, submits nested blocking work from the draining queue, and proves the captured hostile destructor runs on `shelterwood-blocking`, does not delay submission return, and the returned future still yields its value
- adversarial unit coverage pins sole-owner detection, accepted-then-cancelled destruction, injected native fallback failure, panicking receiver wakers, and unclaimed hostile result/panic-payload destructors
- mutation check: forcing rejection detection to return false makes the end-to-end shutdown regression fail with the runtime-teardown cancellation diagnostic; restoring it passes
- `./tools/dev just ci` — passed, including formatting, Clippy with warnings denied, 635 tests, doctests, rustdoc, API/runtime path checks, package verification, and Nix formatting

## Review round (d672c2b)

- fix an isolation hole this PR introduced: `receive_blocking` built its `DisposingReceiver` inside an `async fn` body, which runs only on first poll, so a `Blocking<T>` future dropped before it was ever polled reclaimed an already-stored return value or panic payload in its own drop glue — on the awaiting actor's thread. The wrapper is now built with the future; `dropping_an_unpolled_future_isolates_a_stored_outcome` fails on the previous ownership
- unify the Tokio rejection protocol: one `BlockingPoolJob` trait with `submit_blocking_job`/`blocking_pool_accepted` in `spawn.rs`, used by both `spawn_blocking_work` and `dispatch_disposal` (which also folds in the old `QueuedDisposal` erasure), so the `spawn_task` contract has one home and `spawn_blocking_work` picks up the `is_available` guard it lacked
- correct the teardown-cancellation docs in SPEC, both guide copies, and both `run_blocking` surfaces: Tokio discarding an already-accepted submission during teardown reaches the same panic, not only a fallback thread that cannot start

Rebased on `main` (through #245/#248/#249). `./tools/dev just ci` on the combined tree: 643 tests, all green.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1098,10 +1098,14 @@ handlers non-blocking. Contracts:
   discards a later panic along with the detached thread (documented). A
   submission synchronously rejected by Tokio during blocking-pool shutdown
   moves to a detached Shelterwood-owned thread; the operation and destruction
-  of its captured state never run on the submitting runtime thread. If native
-  thread creation also fails, the captured state still reaches isolated
-  disposal and awaiting the future panics with a runtime-teardown cancellation
-  diagnostic rather than an internal task-invariant claim.
+  of its captured state stay off the submitting runtime thread while an
+  isolation worker can be created. If the dedicated fallback thread cannot
+  start, the captured state transfers to detached disposal and awaiting the
+  future panics with a runtime-teardown cancellation diagnostic rather than an
+  internal task-invariant claim. If the system cannot create the disposal
+  worker either, disposal's final no-loss fallback destroys the state
+  synchronously; thread exhaustion is the sole exception to the isolation
+  guarantee.
 - On orderly return, error, or caught-panic teardown, offloads, lifetime
   tasks, and monitor leases are frozen, cancelled, and joined before actor
   state is dropped. Hard abort necessarily drops the handler future (and
--- a/SPEC.md
+++ b/SPEC.md
@@ -1095,14 +1095,18 @@ handlers non-blocking. Contracts:
   closure's return value; a panic in the closure is captured and resumes
   where the future is awaited — on the actor task, the ordinary
   actor-panic path (§7) — while a future dropped before completion
-  discards a later panic along with the detached thread (documented). A
+  discards a later panic along with the detached thread (documented); a
+  return value or panic payload that already arrived is discarded through
+  detached disposal rather than in the awaiting task's drop glue. A
   submission synchronously rejected by Tokio during blocking-pool shutdown
   moves to a detached Shelterwood-owned thread; the operation and destruction
   of its captured state stay off the submitting runtime thread while an
-  isolation worker can be created. If the dedicated fallback thread cannot
-  start, the captured state transfers to detached disposal and awaiting the
-  future panics with a runtime-teardown cancellation diagnostic rather than an
-  internal task-invariant claim. If the system cannot create the disposal
+  isolation worker can be created. An operation that never runs at all —
+  Tokio discarded an already-accepted submission during runtime teardown, or
+  the dedicated fallback thread could not start — transfers its captured
+  state to detached disposal, and awaiting the future panics with a
+  runtime-teardown cancellation diagnostic rather than an internal
+  task-invariant claim. If the system cannot create the disposal
   worker either, disposal's final no-loss fallback destroys the state
   synchronously; thread exhaustion is the sole exception to the isolation
   guarantee.
--- a/SPEC.md
+++ b/SPEC.md
@@ -1095,7 +1095,13 @@ handlers non-blocking. Contracts:
   closure's return value; a panic in the closure is captured and resumes
   where the future is awaited — on the actor task, the ordinary
   actor-panic path (§7) — while a future dropped before completion
-  discards a later panic along with the detached thread (documented).
+  discards a later panic along with the detached thread (documented). A
+  submission synchronously rejected by Tokio during blocking-pool shutdown
+  moves to a detached Shelterwood-owned thread; the operation and destruction
+  of its captured state never run on the submitting runtime thread. If native
+  thread creation also fails, the captured state still reaches isolated
+  disposal and awaiting the future panics with a runtime-teardown cancellation
+  diagnostic rather than an internal task-invariant claim.
 - On orderly return, error, or caught-panic teardown, offloads, lifetime
   tasks, and monitor leases are frozen, cancelled, and joined before actor
   state is dropped. Hard abort necessarily drops the handler future (and
--- a/crates/shelterwood-runtime/src/disposal.rs
+++ b/crates/shelterwood-runtime/src/disposal.rs
@@ -6,9 +6,8 @@ use std::{
     },
 };
 
-use tokio::task;
-
-use super::{Latch, catch_panic, contain_panic_payload, discard_panic, is_available};
+use super::{Latch, catch_panic, contain_panic_payload, discard_panic};
+use crate::spawn::{BlockingPoolJob, submit_blocking_job};
 
 /// Ownership wrapper for user values retained by framework state.
 ///
@@ -94,13 +93,6 @@ where
         // worker or double-panic while the job is being dropped.
         discard_panic(catch_panic(|| completion(panic)).err());
     }
-
-    fn is_pending(&self) -> bool {
-        self.state
-            .lock()
-            .expect("disposal job mutex poisoned")
-            .is_some()
-    }
 }
 
 impl<T, C> Drop for DisposalJob<T, C>
@@ -113,18 +105,20 @@ where
     }
 }
 
-/// Erased view of a queued disposal job for the shared fallback thread.
-trait QueuedDisposal: Send + Sync {
-    fn run(&self);
-}
-
-impl<T, C> QueuedDisposal for DisposalJob<T, C>
+impl<T, C> BlockingPoolJob for DisposalJob<T, C>
 where
     T: Send + 'static,
     C: FnOnce(Option<DisposalPanic>) + Send + 'static,
 {
     fn run(&self) {
         self.finish();
+    }
+
+    fn is_pending(&self) -> bool {
+        self.state
+            .lock()
+            .expect("disposal job mutex poisoned")
+            .is_some()
     }
 }
 
@@ -134,7 +128,7 @@ where
 /// under this lock, and submitters push and consult it under the same lock, so
 /// a queued job always has a live worker destined to drain it.
 struct FallbackDisposals {
-    queue: VecDeque<Arc<dyn QueuedDisposal>>,
+    queue: VecDeque<Arc<dyn BlockingPoolJob>>,
     worker_live: bool,
 }
 
@@ -146,7 +140,7 @@ static FALLBACK_DISPOSALS: Mutex<FallbackDisposals> = Mutex::new(FallbackDisposa
 /// Queues a disposal job for the shared fallback thread, lazily starting it.
 /// Returns `false` when no worker exists and none could be started; the
 /// caller must then finish the job itself.
-fn enqueue_fallback_disposal(job: Arc<dyn QueuedDisposal>) -> bool {
+fn enqueue_fallback_disposal(job: Arc<dyn BlockingPoolJob>) -> bool {
     let mut state = FALLBACK_DISPOSALS
         .lock()
         .expect("fallback disposal queue mutex poisoned");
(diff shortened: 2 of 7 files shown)


ralexstokes:
## Adversarial review of `3ffb6b5`

I reviewed this from the Tokio rejection edge outward, including the current `tokio 1.53.1` blocking-pool implementation, every `Arc<BlockingJob>` ownership transition, cancellation before execution, dropped-return handling, hostile panic payloads/destructors/wakers, fallback-spawn failure, the public `run_blocking` surface, and the #205 closeout claims.

### Actionable findings fixed

1. **A receiver waker panic could escape the detached worker.** `BlockingJob::run` contained closure panics and unclaimed-outcome destruction, but `completion.send(outcome)` itself was outside a panic boundary. Tokio stores the outcome before calling the registered receiver waker; a hostile `Wake` implementation could therefore unwind a Tokio blocking worker or the native fallback thread even though the join handle is deliberately detached. The send is now caught, the waker panic payload is safely discarded, and the already-published authoritative outcome remains receivable. `completion_contains_a_panicking_receiver_waker` pins this behavior.

2. **The documented thread-exhaustion guarantee was stronger than the code can provide.** The original text said captured state never reaches the submitting thread, including after native fallback creation fails. In reality, failure of the dedicated `shelterwood-blocking` thread transfers the closure to `dispose_detached`; if the system also cannot create the disposal worker, disposal deliberately uses its synchronous no-loss fallback. The spec, synchronized shutdown guide, public `Blocking` docs, both `run_blocking` method surfaces, and PR description now identify total native-thread exhaustion as the sole isolation exception while preserving the cancellation diagnostic.

3. **Important ownership branches were only argued, not pinned.** Added direct coverage for:
   - accepted ownership vs synchronous rejection vs fast completion;
   - an accepted-but-never-run job being cancelled, closing its waiter and disposing captures off the cancelling thread;
   - injected `std::thread::Builder::spawn` failure, including the consumed worker clone and subsequent disposal handoff;
   - unclaimed return-value and panic-payload destructors that themselves panic.

### Challenged and retained

- **Tokio rejection detection is sound for the pinned dependency.** `spawn_task` calls `task.task.shutdown()` before returning `ShuttingDown`; `spawn_blocking` still returns a handle. Thus its closure-held `Arc` is gone before `spawn_blocking` returns. `strong_count == 1` proves there is no executor-held owner able to become live again, and the mutex-protected pending check distinguishes rejection from a worker that already took/completed the job.
- **The count/pending split has no resurrection race.** Once the count reads one, only the submitter owns an `Arc`; no other owner exists that could clone it between the count and lock. If a worker still exists the count is at least two, and if it completed before sampling, `pending == None` suppresses the fallback. The new three-outcome unit test pins all cases.
- **Duplicate execution is excluded.** Tokio acceptance followed by fallback because of an unrelated `spawn_blocking` panic is still safe: every execution path takes the single `Option<(F, completion)>` under the mutex, so at most one path can run the operation.
- **Accepted-then-cancelled work preserves the intended semantics.** The last worker owner drops the pending job, closes completion under a panic boundary, and moves the never-run closure into disposal. Awaiting reports `blocking operation was cancelled during runtime teardown`; it does not revive an operation Tokio had already accepted and later cancelled.
- **Dropped futures remain detached.** A result/panic published before receiver drop is recovered by `DisposingReceiver` and isolated; receiver-first cancellation makes send return the outcome to the worker, where hostile destruction is contained. A caller/executor-provided waker destructor is not reclassified as captured operation state; the supported actor/offload path uses Tokio's executor waker. No job mutex is held while any waker callback or user destructor runs.
- **Ordinary `run_blocking` behavior is preserved.** Submission remains eager, the returned future remains `Send + 'static`, values are returned unchanged, operation panics resume at await, dropped futures cancel cooperatively and discard late outcomes, and rejected work executes on the named `shelterwood-blocking` thread.
- **The per-rejection native thread is intentional.** Reusing the serialized disposal queue would allow one legitimate long-running blocking operation to strand every later rejected operation during teardown.
- **#205 closeout is supported by current `main`.** The disposal rejection fix and pinning tests are present; the stale item-42 documentation claim is absent; the guide already states that detached disposal normally rides Tokio's runtime lifetime and names `shutdown_timeout`/`shutdown_background`; and `driver::tests::shutdown::runtime_teardown_finishes_the_cancelled_root_driver` covers the remaining root-driver cancellation path.

### Validation

- `./tools/dev cargo test -p shelterwood-runtime spawn::tests -- --nocapture`: 9 passed.
- `./tools/dev just ci`: passed formatting, warnings-denied Clippy, all 635 tests, doctests, rustdoc, API/runtime reachability, packaged-doc sync, and Nix formatting.
- Mutation discrimination: forcing `blocking_spawn_needs_fallback` to return `false` made `shut_down_blocking_pool_runs_rejected_work_off_the_submitting_thread` fail with the runtime-teardown cancellation panic. Restoring the ownership detector returns the test and full CI to green.

I found no remaining actionable issue in the reviewed scope after these fixes.

ralexstokes:
## Review

Reviewed the whole diff plus the surrounding blocking/disposal paths. The core mechanism is right: the `Arc`-held job is the correct ownership signal, the mutex — not the reference count — is what actually makes double-execution impossible, and the fallback/disposal/synchronous ladder is the right degradation order. Three findings, all fixed in d672c2b (`just ci` green, 636 tests).

### 1. An unclaimed outcome was destroyed on the awaiting actor's thread (correctness)

`receive_blocking` was an `async fn`, so `DisposingReceiver::new` ran on **first poll**. A `Blocking<T>` future that is built and then dropped without ever being polled — `let b = ctx.run_blocking(..); if !needed { drop(b) }` — dropped a bare `OneShotReceiver`, and Tokio's oneshot destroys a stored value in the dropping thread's drop glue. Since submission is eager, a fast operation genuinely lands a value in that window, so a user return value or panic payload could run its destructor on the actor task — exactly the inline destruction this path exists to prevent, and the one hole the PR's own `DisposingReceiver` usage was meant to close. Every other `DisposingReceiver` in the tree (`task.rs`, `reply.rs`, `futures.rs`) is built eagerly into a struct field; this was the only late one.

Fix: `receive_blocking` is now a plain `fn` that builds the wrapper before returning the future. New unit test `dropping_an_unpolled_future_isolates_a_stored_outcome` fails on the old ownership (`ThreadId(2) == ThreadId(2)`, destructor on the waiter) and passes after.

### 2. The rejection-detection protocol was copied, not shared

`blocking_spawn_needs_fallback` + the `catch_panic(spawn_blocking)` / `discard_panic` dance now existed verbatim in both `disposal.rs` and `spawn.rs`, so the Tokio `spawn_task` contract it encodes lived in two places — and only disposal's copy carried the explanation of *why* a future Tokio that deferred the drop degrades fail-safe. The copies had also drifted: disposal guards with `is_available()`, `spawn_blocking_work` did not, and so relied on catching a guaranteed panic when no runtime exists.

Fix: one `BlockingPoolJob` trait (`run` + `is_pending`) with `submit_blocking_job` and `blocking_pool_accepted` in `spawn.rs`. `dispatch_disposal` is now `if submit_blocking_job(&job) { return; }`, the old `QueuedDisposal` erasure trait folds into `BlockingPoolJob`, and both unit tests keep covering their own job type's `is_pending`. Mutation-checked: forcing `blocking_pool_accepted` to `true` fails 5 tests across both modules.

### 3. Docs and spec understated when the teardown-cancellation panic fires

All four sites (SPEC §, `docs/shutdown-and-resources.md` + its doctest copy, `raw.rs`, `actor.rs`) attributed the panic solely to a fallback thread that cannot start. But Tokio discarding an **already-accepted** submission during teardown reaches the same panic, and that is the ordinary outcome when a runtime is dropped with blocking work queued — not an exotic thread-exhaustion case. The PR already tests it (`accepted_then_cancelled_job_isolates_capture_and_reports_teardown`); the docs just didn't say it. Amended, plus one SPEC clause for the invariant fix #1 establishes: an already-arrived value or panic payload is discarded through detached disposal, never in the awaiting task's drop glue.

### Not changed

- **Timing-dependent split between "rerouted" and "cancelled".** Whether a teardown-time submission runs on `shelterwood-blocking` or reports cancellation depends on whether the submitter sampled ownership before or after Tokio dropped the accepted task. Both outcomes are safe and now both documented; making them uniform would mean holding the submitter across a race it has no reason to win.
- **No Tokio context on the fallback thread.** A user closure that calls `Handle::current()` panics there instead of running. Teardown-only, and the panic propagates through the normal lane.
- **Test-fixture duplication.** `BlockingDrop`, `describe_current_thread`, the thread-description type and `release` are now copied between the `spawn` and `disposal` test modules. Left alone here to keep this PR to its subject; it belongs with #212's duplication sweep.

PR #251, opened 2026-08-13
Make mailbox effects structural and runtime-neutral

### LINKED ISSUE #209: Mailbox cleanup: unreachable accept_locked rejection variants (~40-50 lines); configure kind-mismatch vs panic convention
Two related cleanups in `mailbox/cell.rs`, from the 2026-08-11 review round (no reachable bugs in the mailbox otherwise — the park/promotion/termination interleavings, waiter-id reuse, Full-invariant, and deadline-boundary races were all traced clean).

## 1. `accept_locked`'s `NotAccepting`/`Unconfigured` rejections are unreachable (~40-50 lines)

Both callers invoke `accept_locked` (`cell.rs:947-1005`) only from the `BindingStatus::Bound` arm under the same continuous mutex hold, so the binding cannot be Unbound/Frozen/Terminal inside it (`NotAccepting` dead); and `bind` panics unless `kind` is `Some`, so a Bound mailbox is always configured (`Unconfigured` dead). Passing the already-extracted incarnation in and reducing the rejection type to "full" deletes the `RejectionReason` enum, the `Rejected` struct, `submit`'s drop-lock-then-`unreachable!` block (`cell.rs:583-595`), and `try_send`'s two-way error mapping (`cell.rs:644-659`) — and makes the invariant ("a bound mailbox is configured and its binding is what `status()` said") structural instead of re-derived. If the team prefers keeping `accept_locked` total over all states, then at minimum `try_send`'s `Unconfigured => incarnation_observed: None` arm (`cell.rs:650`) is dead code implying a reachable behavior that doesn't exist.

## 2. `configure` silently ignores a kind mismatch in release (`cell.rs:801-812`)

Every other driver-contract violation on this surface deliberately panics after releasing the lock (`bind` at `cell.rs:822-829`, with a comment explaining the discipline), but a reconfigure with a *different* `MailboxKind` is only `debug_assert_eq!` — in release the new kind is silently discarded and the mailbox keeps behaving as the old kind. Unreachable today (the driver resolves the kind once per child), but if that regresses the failure mode is silent capacity/conflation misbehavior rather than the loud panic the module's convention promises. Align with `bind`'s drop-lock-then-panic pattern, or comment why the divergence is intended.

### LINKED ISSUE #236: Mailbox two-phase effects: make the under-lock waker/payload class unwritable, and fold in the runtime-neutrality inversion
Split out of #214 track B, which is being retired as an inline checklist. This is the largest item in that track and the one that unblocks a leftover from #214 stage 2.

## The problem

The mailbox is where the lock rule (#235) is followed most and structured least. Outside the `Promotion` path it is hand-rolled:

- ~10 explicit `drop(state)` / `drop(mailbox)` calls scattered through `shelterwood-mailbox/src/cell.rs` to end critical sections before user-owned values are destroyed.
- Tuple returns carrying values out by hand. `withdraw` (`cell.rs:754`) returns `(result, registration, waker)` specifically so the caller destroys the waker after both guards are gone; the doc comment explains why, and `debug_assert!`s at `:785` and `:801` pin that the other paths took the waker already.

That is correct code. It is also correctness re-derived at every site from a rule that lives in comments — which is exactly how #202 happened (two sites dropped caller-supplied wakers under one or both core mutexes; a hostile waker `Drop` reached from `SendFuture::drop` during an unwind was a process abort).

## The shape

`Promotion` (`cell.rs:402`) is already the right pattern, applied to one path: an effects struct that accumulates wakers and displaced payloads inside the critical section and flushes them from `Drop`, panic-contained, after the guard is gone.

Generalize it so the whole mailbox routes wakers, payloads, and completions out of the locked region **by construction** — a `WakerSlot`-style API where storing a waker hands back the displaced one rather than dropping it in place, and where there is no way to spell the #202 bug. The measure of success is that the fix is not "we found the sites", it is "the sites cannot be written".

## Fold in: `shelterwood-mailbox` is not runtime-neutral yet

#214 stage 2 recorded this explicitly and deferred it here rather than doing a standalone inversion:

> `shelterwood-mailbox` still reaches `shelterwood-runtime` for oneshot, signal, detached disposal, and the clock. Fold the rest into track B's two-phase mailbox effects work rather than doing a standalone inversion: the pure state machine is `cell.rs`'s transitions, while the futures are shell by construction, and that refactor produces the same seam. A type parameter is disqualified — it would infect the public API (`ActorRef<M, R>`); a `dyn` capability object installed by the façade is the shape.

Until it lands, the mailbox's contribution to the public API reachability walk is *checked* by its own lane rather than implied by its dependencies.

## Scope

- [ ] Effects struct generalized across the mailbox; the manual `drop(guard)` sites and hand-carried tuples retire into it
- [ ] The #202 class is unrepresentable, not merely absent — demonstrate by trying to write the old bug
- [ ] `dyn` capability object for the four runtime capabilities (oneshot, signal, detached disposal, clock), installed by the façade; no type parameter on `ActorRef`
- [ ] Existing hostile-waker tests (`cell.rs:1183`, `:1203`, `:1237`, `:1313`, `:1360`) keep passing unchanged — they are the regression guard for the class

## Related

- #235 — the lock rule this makes structural in the densest region
- #202 (closed) — the instance
- #209 — mailbox cleanup (dead `accept_locked` rejection variants, `configure` convention); natural companion, same files

ralexstokes:
Closes #236.
Closes #209.

## Summary

- replace mailbox transition handoffs with `MailboxTxn` + `MailboxEffects`: the transaction owns the mutex guard and drops it before pulses, waker vtables, payload destructors, or disposal capabilities can run
- replace the raw `Option<Waker>` surface with a private `WakerSlot` whose take/replace operations require an effects sink; cancellation now returns a withdrawal outcome plus deferred effects instead of the old `(outcome, waker)` tuple
- remove `RejectionReason::{NotAccepting, Unconfigured}` and make bound acceptance return only the reachable full case
- make a mailbox configuration-kind mismatch panic in release builds after unlocking, matching the bind-contract convention
- invert the remaining mailbox/runtime dependency through one façade-installed `dyn MailboxRuntime` capability covering one-shot delivery, signals, isolated disposal, and clock/timers; `ActorRef<M>` gains no runtime type parameter
- move split reply construction to `actor.reply_channel()` so standalone reply deadlines and disposal use the same injected capability as the actor

## Structural guarantee

The old #202 spelling no longer type-checks. `OperationState::waker` is a `WakerSlot`, not an `Option<Waker>`; its storage is private even from the parent module. Calling the old `state.waker.replace(next_waker)` or `state.waker.take()` form fails because every mutator requires an explicit `WakerEffects` sink (and take also requires a wake/drop/dispose action). Locked code cannot obtain a raw displaced waker to accidentally destroy beside either core guard.

All mutable mailbox transitions now enter through `MailboxTxn`. Its `Drop` empties the guard field before Rust drops the effects field, including on unwind. The remaining direct state locks are read-only snapshots or operation-state helpers whose waker actions flow into an effects sink declared outside the guard.

## Runtime inversion

`shelterwood-mailbox` no longer has a production dependency on `shelterwood-runtime`; its normal dependency tree is only `shelterwood-core` (plus core's dependencies). The Tokio implementation lives in `shelterwood-runtime::mailbox`, and the façade supplies it at every `MailboxCell` construction site. Reply and deadline state retain that same object so paused Tokio time and disposal behavior remain coherent.

This deliberately changes the no-user split-reply surface from `Reply::channel()` to `ActorRef::reply_channel()`: without an actor (or another façade-owned capability source), a runtime-neutral mailbox crate cannot select the correct clock/disposal adapter without a global ambient runtime.

## Validation

- `./tools/dev cargo tree -p shelterwood-mailbox -e normal` — only `shelterwood-core`; no runtime adapter or Tokio
- existing hostile-waker wake, replacement, cancellation, unwind, reentrancy, and termination tests all pass
- added `configuration_mismatch_panics_after_releasing_the_mailbox_lock`, which verifies the panic and proves the mutex remains unpoisoned
- added `bind_submits_displaced_payloads_for_disposal_before_waking_senders`, which pins the existing pulse → disposal-submission → promoted-sender wake order through an injected runtime capability
- `./tools/dev just ci` — passed: formatting, Clippy with warnings denied, 630 tests, doctests, rustdoc, all public API reachability lanes, packaged-doc sync, and Nix formatting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -135,6 +135,7 @@ name = "shelterwood-mailbox"
 version = "0.1.0"
 dependencies = [
  "shelterwood-core",
+ "shelterwood-runtime",
  "tokio",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -135,7 +135,6 @@ name = "shelterwood-mailbox"
 version = "0.1.0"
 dependencies = [
  "shelterwood-core",
- "shelterwood-runtime",
  "tokio",
 ]
 
@@ -145,6 +144,7 @@ version = "0.1.0"
 dependencies = [
  "fastrand",
  "shelterwood-core",
+ "shelterwood-mailbox",
  "tokio",
 ]
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -803,14 +803,34 @@ Delivery is at-most-once (§1 principle 6). Send flavors on `ActorRef`
   a pending acknowledgement; dropping it is observable to the caller as
   `ReplyDropped`. A successful `call` exposes the accepting incarnation
   alongside the reply value — §3.3's retry discipline needs it on success
-  as much as on failure. `Reply::channel()` exists as the split escape
+  as much as on failure. `ActorRef::reply_channel()` exists as the split escape
   hatch when the reply must be awaited elsewhere: it yields the `Reply`
   plus a `ReplyReceiver<T>`, whose `recv(deadline)` (trailing deadline
   covering only the response wait — acceptance evidence is the
   accompanying send's result) resolves per B.3. Awaiting `call` on
   `myself()` inside `handle` is a guaranteed deadlock — the reply needs
   the very handler being blocked; use `continue_with`, or
-  `Reply::channel()` with an offload (documented hazard).
+  `ActorRef::reply_channel()` with an offload (documented hazard).
+
+**Mailbox transition boundary.** Mailbox and send-operation mutexes protect
+only synchronous state transitions. A transition records signal pulses,
+waker wake/drop actions, displaced payloads, and isolated disposal requests in
+an effects sink paired with its guard; the guard is released before that sink
+can flush. No `RawWaker` vtable, message destructor, signal callback, or
+runtime-disposal capability runs under either mutex, including unwind paths.
+The registered-waker slot exposes no operation that returns or replaces a
+`Waker` without an effects sink, making the #202 failure shape structurally
+unrepresentable rather than a call-site convention. Cancellation returns one
+withdrawal outcome plus its post-unlock effects object, never a tuple carrying
+a raw waker across the boundary.
+
+The mailbox crate is runtime-neutral in its production dependency graph. Each
+mailbox receives one type-erased runtime capability object from the public
+façade; it supplies one-shot delivery, change signals, isolated disposal, and
+the clock/timer pair. The same object flows through reply channels and deadline
+futures, so virtual-time and disposal semantics cannot silently switch
+adapters. Runtime choice is deliberately absent from `ActorRef<M>`'s type
+parameters.
 
 **Binding.** The mailbox binding is membership-owned: created at insertion,
 outliving incarnations and actor destruction (§5.5). *Bound* — accepting
@@ -4011,7 +4031,8 @@ implied on all; error/outcome types are B.3 and B.8):
   graceful shutdown.
 - **`ActorRef<M>`**: cheap `Clone`, membership-addressed (§2); `send` /
   `try_send` / `send_timeout` (each resolving to the accepting
-  `Incarnation`) and `call` per §5.1, error and success shapes per B.3;
+  `Incarnation`), `call`, and `reply_channel` per §5.1, error and success
+  shapes per B.3;
   `contramap` *(II §17)*, `pinned` *(II §15)*.
 - **`ScopeRef`**: `snapshot()`, `subscribe_snapshots()` (conflating
   watch), `subscribe_lifecycle()` (B.4), the `wait_for_child` helper
--- a/SPEC.md
+++ b/SPEC.md
@@ -816,8 +816,13 @@ Delivery is at-most-once (§1 principle 6). Send flavors on `ActorRef`
 only synchronous state transitions. A transition records signal pulses,
 waker wake/drop actions, displaced payloads, and isolated disposal requests in
 an effects sink paired with its guard; the guard is released before that sink
-can flush. No `RawWaker` vtable, message destructor, signal callback, or
-runtime-disposal capability runs under either mutex, including unwind paths.
+can flush. On every path a transition can complete — acceptance, rejection,
+withdrawal, terminal teardown, and an unwind out of any of them — no `RawWaker`
+vtable, message destructor, signal callback, or runtime-disposal capability
+runs under either mutex. The one carve-out is a framework invariant break
+(identity-space exhaustion or an unreachable binding state): those unwind with
+the payload still under the guard, and they poison the mutex regardless, so the
+transition is abandoned rather than completed.
 The registered-waker slot exposes no operation that returns or replaces a
 `Waker` without an effects sink, making the #202 failure shape structurally
 unrepresentable rather than a call-site convention. Cancellation returns one
--- a/crates/shelterwood-mailbox/Cargo.toml
+++ b/crates/shelterwood-mailbox/Cargo.toml
@@ -16,4 +16,7 @@ shelterwood-core.workspace = true
 
 [dev-dependencies]
 shelterwood-core = { workspace = true, features = ["test-util"] }
-tokio = { workspace = true, features = ["test-util"] }
+shelterwood-runtime = { workspace = true, features = ["test-util"] }
+# The adapter's re-exported `#[test]` attribute expands to unqualified `tokio`
+# paths, so the macro's own crate must be linkable. No source here names it.
+tokio.workspace = true
--- a/crates/shelterwood-mailbox/Cargo.toml
+++ b/crates/shelterwood-mailbox/Cargo.toml
@@ -13,9 +13,7 @@ workspace = true
 
 [dependencies]
 shelterwood-core.workspace = true
-shelterwood-runtime.workspace = true
 
 [dev-dependencies]
 shelterwood-core = { workspace = true, features = ["test-util"] }
-shelterwood-runtime = { workspace = true, features = ["test-util"] }
 tokio = { workspace = true, features = ["test-util"] }
--- a/crates/shelterwood-mailbox/src/capability.rs
+++ b/crates/shelterwood-mailbox/src/capability.rs
@@ -148,7 +148,7 @@ impl<T> OneShotReceiver<T> {
     }
 }
 
-fn dispose_value<T: Send + 'static>(runtime: &dyn MailboxRuntime, value: T) {
+pub(crate) fn dispose_value<T: Send + 'static>(runtime: &dyn MailboxRuntime, value: T) {
     runtime.dispose(Box::new(value));
 }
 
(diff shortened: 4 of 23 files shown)


ralexstokes:
Adversarial post-publication review completed at `6f88738b5c9c2b5cd5a7d800b7b07a2b36bbc7fd`.

### Review method

I reviewed the full diff against both #236 and #209, then:

- traced every production mailbox-state and send-operation mutex acquisition through normal return and unwind;
- checked every waker clone/replace/take/wake/drop path, including cancellation during an existing unwind and panic precedence;
- compared signal, displaced-payload, promoted-sender, and terminal teardown ordering with `main`;
- followed all four injected runtime capabilities (one-shot, signal, disposal, and clock/timer) from façade construction through sends, calls, split replies, deadlines, and teardown;
- verified object safety and dependency direction, including the absence of a runtime type parameter on `ActorRef<M>`;
- checked every façade `MailboxCell::new` site and every migrated split-reply site;
- audited the #209 reduction of bound acceptance to the only reachable `Full` failure and exercised configuration mismatch in an optimized build.

### Finding and fix

One real ordering regression was present in the published draft. Before this PR, binding an unbound latest-value mailbox completed post-lock effects in this order:

1. pulse the mailbox change signal;
2. submit displaced payloads to isolated disposal;
3. wake the promoted senders.

The generalized `MailboxEffects::drop` initially did (1), (3), (2). A custom waker may execute synchronously, so this was observable: promoted sender code could run before ownership of the displaced payload had been transferred to disposal.

Commit `6f88738` restores the original ordering for isolated displacement while retaining the existing pulse → wake → inline-drop order used by receive promotion. It also adds `bind_submits_displaced_payloads_for_disposal_before_waking_senders`, using an injected recording runtime/signal and recording wakers to pin the complete pulse → disposal-submission → wake → wake sequence.

No other blocking correctness findings remained. In particular:

- `WakerSlot` exposes no raw waker extraction/replacement path; every mutation requires an effects sink.
- transaction unwind drops the mailbox guard before effects; operation-only paths declare their waker sink outside the operation guard.
- cancellation disposes both recovered payloads and registered wakers through the same injected adapter without running either core lock.
- reply receivers and deadline timers retain the originating actor’s adapter, including its clock, rather than selecting ambient runtime state.
- the release-mode configuration mismatch panic occurs after unlocking and leaves the mutex unpoisoned.
- `shelterwood-mailbox` has no normal dependency on Tokio or `shelterwood-runtime`.

### Validation

- `./tools/dev cargo test -p shelterwood-mailbox` — 28 passed.
- `./tools/dev cargo test --release -p shelterwood-mailbox cell::tests` — 16 passed, including the release-only contract check and ordering regression.
- `./tools/dev cargo tree -p shelterwood-mailbox -e normal` — only `shelterwood-core` (plus its dependencies).
- `./tools/dev just ci` — passed: formatting, Clippy with warnings denied, 630 tests, doctests, rustdoc, all public API reachability lanes, packaged-doc sync, and Nix formatting.
- `git diff --check origin/main...HEAD` — clean.

The PR body was updated to include the new regression test and corrected test count. The PR remains draft for the parent workflow to mark ready.

ralexstokes:
## Review

Read the whole diff against `origin/main`. The core of the change holds up: I traced every mailbox and send-operation critical section and found no reachable path where user code — a `RawWaker` vtable, a message destructor, a signal callback, or a disposal capability — runs under either mutex. Drop ordering is right in the places that matter (`MailboxTxn::drop` empties the guard field before Rust drops the effects field; `withdraw` declares `waker_effects` before the transaction so unwind releases both guards first; `SendOperation::poll` clones the waker outside the lock and drops an unused replacement after the guard). Effect ordering per operation matches main exactly — pulse-then-displaced on the accept path, isolated-disposal-then-wake on bind, wake-then-displaced on promotion — and the new `bind_submits_displaced_payloads_for_disposal_before_waking_senders` pins the one that was previously only implied.

Three findings, all fixed in 4cf745b (`just ci` green: 630 tests, exit 0).

### 1. The mailbox crate now names tokio in its own sources

Dropping the `shelterwood-runtime` dev-dependency left the tests reaching for tokio directly:

```
crates/shelterwood-mailbox/src/lib.rs:102:    pub(crate) use tokio::{test, time::advance};
crates/shelterwood-mailbox/src/capability.rs:212: use tokio::sync::{oneshot, watch};
crates/shelterwood-mailbox/src/capability.rs:386: tokio::time::Instant::now().into_std()
crates/shelterwood-mailbox/src/capability.rs:392: tokio::time::sleep_until(deadline.into()).await
```

That breaks the layering rule pinned by #213/#242 one day earlier: `shelterwood-runtime` is the only crate that may name tokio, tests included. The rule and the inversion aren't actually in tension — the inversion is a claim about the *production* graph, and a dev-dependency doesn't appear in `cargo tree -e normal`. Restored the dev-dependency (a dev-only cycle, which cargo resolves) and rebuilt the test capability object as delegation to the adapter's own `oneshot`/`Signal`/`now`/`sleep_until`/`dispose_detached`.

It can't be `mailbox_runtime()` directly — the adapter links the *non-test* build of this crate, so its `MailboxRuntime` is a distinct type from the `cfg(test)` one (`E0308: multiple different versions of crate shelterwood_mailbox`). The binding has to be rebuilt locally; what it must not do is rebuild the *semantics*.

Which is the second half of this finding: the removed `TestRuntime` had hand-rolled the adapter's five-state one-shot protocol (`OPEN`/`SENDING`/`SENT`/`SENDER_CLOSED`/`RECEIVER_CLOSED`), the same protocol `shelterwood-runtime/src/sync.rs` implements. `reply.rs`'s drop/cancellation lifecycle tests and `deadline.rs`'s expiry tests were validating against that restatement, so a divergence from the real adapter would have read as a passing test. Delegation removes ~130 lines and puts those tests back on the production primitives. `cargo tree -p shelterwood-mailbox -e normal` still shows `shelterwood-core` alone, and `grep -rn "tokio::" crates/*/src` outside the adapter is empty again.

### 2. `MailboxEffects` clones two `Arc`s per transition, including per-message receive

`MailboxEffects::new` cloned `cell.runtime` and `cell.changed` on every mailbox transition — four atomic refcount operations per `try_recv`, i.e. per message through the actor loop, plus per send. But the sink never outlives the `MailboxTxn` that constructed it, and `MailboxTxn<'a>` already holds `&'a MailboxCell<M>`. Changed it to borrow the cell. The owned `Arc` in `WakerAction::Dispose` stays, since that one genuinely escapes into the returned `Withdrawal`.

### 3. SPEC transition-boundary claim is stronger than the code

> No `RawWaker` vtable, message destructor, signal callback, or runtime-disposal capability runs under either mutex, including unwind paths.

Two paths in `accept_locked` unwind with `message: M` still owned by that frame while the caller's guard is live: the `mint_accepted_sequence` exhaustion `expect` (pre-existing) and the new `unreachable!`/`debug_assert_eq!` binding arms. The latter is where main deliberately did `drop(state); drop(message); unreachable!(...)` with prose explaining why — that convention is gone now that `accept_locked` borrows the state rather than owning the guard, and reinstating it would mean resurrecting the `RejectionReason` machinery this PR correctly deletes as unreachable.

Both carve-outs poison the mutex regardless, so the transition is abandoned rather than completed. Scoped the SPEC sentence to name that instead of implying a universal guarantee.

### Not changed

`ErasedOneShotSender::is_closed` is required of every runtime implementor but its only caller is a `#[cfg(test)]` helper (`reply.rs`'s cancellation assertion). It can't be `cfg`-gated across the crate boundary and the cost is four lines in one impl, so I left it — noting it in case you'd rather the test pinned closedness through an observable instead.

PR #256, opened 2026-08-13
Retire framework-held exits through RetainedExit

### LINKED ISSUE #254: Retire framework-held `Exit` payloads through a `RetainedExit` newtype, closing the destruction class by construction
Supersedes #244, #246 and #247, which are three symptoms of one problem and
were being decided site by site. This issue records the decided policy and the
whole remaining scope.

## The problem, stated once

`ExitError` is `Arc<ExitErrorInner>` (`shelterwood-core/src/exit.rs:118`), and
only the `Application` variant owns user code — a
`Box<dyn Error + Send + Sync>`. So every framework drop of an `Exit` is
refcount traffic *except the last one*, and the last one runs a user
destructor wherever it happens to land. A destructor that blocks stalls that
thread; one that panics unwinds from wherever the framework dropped it; one
that runs during an unwind aborts.

The framework retains `Exit` clones in three places, which is exactly the
three closed issues:

- the bounded lifecycle broadcast ring (#244) — under the observation gate
  *and* the signal watch's lock;
- the member record's `stage: MemberStage::Terminal(Exit)` and
  `last_exit: Option<Exit>`, reached through residency's `Arc<MemberCell>`
  (#247) — under the observation gate;
- every other inline framework retirement (#246): `ScopeRuntime::drop`'s
  `children.clear()`, `SnapshotHub`'s retired projection, the deferred queues
  #245 added, `LifecycleHub::publish`'s undelivered event.

## Why the site-by-site answer is not available

`LifecycleHub` publishes into a Tokio broadcast (`runtime::broadcast`,
`shelterwood-cells/src/observe.rs:620`). Tokio's ring owns the evicted slot
and exposes no API to take the value out. #246's option 1 — "isolate exactly
the superseded paths" — therefore **cannot** fix #244 at any number of sites.
Closing the class requires making the payload's destruction safe wherever it
dies, not fixing the places it dies today.

## Why #246's option 2 is not the answer either

"Hold the payload in `Isolated<ExitError>` inside `Exit`" fails on two counts
the original issue did not record:

1. **Layering.** `Isolated`/`dispose_detached` live in `shelterwood-runtime`,
   which *depends on* `shelterwood-core`. `ExitError` cannot reach them
   without inverting the graph, moving disposal into core (which drags Tokio
   and breaks the #242 pin that only `shelterwood-runtime` may name it), or
   adding a new leaf crate plus an install hook.
2. **Semantics.** It would detach destruction even when the *user* drops the
   last clone. `ExitError` is handed back through `wait_terminal` and
   snapshots; bouncing a user's own error destructor onto a shelterwood
   disposal thread changes drop timing for a value they own.

## Decided design: a `RetainedExit` newtype

A framework-internal, `Clone` newtype wrapping `Exit`, used by every
*framework-retained* copy. Its `Drop` is:

```rust
impl Drop for RetainedExit {
    fn drop(&mut self) {
        // Only `Failed` owns user code; every other kind is framework data.
        if matches!(self.0.kind(), ExitKind::Failed(_)) {
            dispose_detached(exit);
        }
    }
}
```

Unconditional on `Failed`, **deliberately**. A `strong_count == 1` probe to
skip the submission when another owner exists is a TOCTOU: two owners can each
read 2, one of them still ends up last, and that one runs the destructor
inline. The check cannot be made sound, so there is no check.

What this buys:

- **#244** closes: the ring stores `RetainedExit`, so eviction detaches
  instead of destroying under the gate and the watch lock.
- **#247** closes: the member record holds `RetainedExit`, so dropping
  `Arc<MemberCell>` under the gate is safe by construction. The unproved
  `Drop`-ordering argument in `CLAUDE.md`'s `Arc` exemption — that
  `ScopeRuntime`'s `Drop` clears residents before `self.children` — can then
  be **deleted**, not merely documented.
- **#246** closes: there is no per-site decision left to get wrong, so the
  class cannot be half-handled.
- Users' own `Exit` values keep ordinary drop semantics.
- No crate-layering change: `RetainedExit` lives in `shelterwood-cells`, which
  already depends on `shelterwood-runtime`.

Cost: one disposal-job submission per retired *failed* exit. Submission is an
`Arc` allocation and a queue push; the destructor itself has to run somewhere
regardless. Failed exits are not on a hot path, and the retirement is off the
critical section by construction.

## Scope

Wide but mechanical.

- Define `RetainedExit` in `shelterwood-cells` (`Clone`, `Debug`, `PartialEq`
  delegating to `Exit`'s shared-provenance equality).
- Convert the retained sites: `MemberStage::Terminal`, `MemberRecord::last_exit`,
  `LifecycleEventKind::Exited`'s payload, the resident projection.
- Convert back at each public read boundary — `ChildState::Stopped`,
  `ChildState::StartupAborted`, `ChildSnapshot::last_exit` and
  `LifecycleEvent` all hand `Exit` to users and must keep doing so.
- Delete the `Arc`-exemption ordering paragraph from `CLAUDE.md`'s lock rule
  and replace it with the by-construction statement.
- Pin with the `gate_probe_exit` / `ObservationGate::is_held` harness #245
  added: the eviction repro in #244 and the residency repro in #247 both
  become passing regression tests, plus one asserting a retired failed exit's
  destructor does not run on the retiring thread.

Blocked by nothing. Relates to #235, #245.

ralexstokes:
Closes #254.

## What changed

- add a `RetainedExit` cells-layer newtype that transfers failed-exit destruction to isolated disposal without a racy strong-count probe
- store or structurally guard retained exits in member and scope records, terminal mailbox state, lifecycle-ring events, snapshot projections, pending terminal disposal, scope completions, driver drain state, and root-driver completion outputs
- cover structured `StartupFailure` projections, which recursively own the triggering child's `Exit` through `ScopeState`, `StartupError`, and `StopReason`
- convert retained exits back to ordinary `Exit` values only at lifecycle, snapshot, and terminal-wait user boundaries
- remove the resident/member `Arc` drop-order exemption from the lock-rule documentation
- add regressions for direct and structured lifecycle-ring eviction, scope-record and snapshot retirement, last-resident member retirement, driver-completion retirement, and public conversion/drop timing

## Why

A framework-retained `ExitKind::Failed` can own the last reference to a user error. Ring eviction, residency teardown, scope-state retirement, or driver cleanup could therefore run a blocking or panicking user destructor on a framework thread or under framework synchronization. A retained newtype closes that destruction class by construction while preserving normal drop timing for exits handed to users.

## Validation

- `./tools/dev just ci`
  - formatting and Clippy with warnings denied
  - 653 nextest tests
  - doctests and rustdoc
  - public API reachability checks
  - packaged-doc and Nix formatting checks


crates/shelterwood-cells/src/cells.rs
```diff
@@ -77,7 +228,7 @@ pub struct MemberRecord {
     pub stage: MemberStage,
     pub incarnation: Option<Incarnation>,
     pub last_incarnation: Option<Incarnation>,
-    pub last_exit: Option<Exit>,
+    pub last_exit: Option<RetainedExit>,
```
ralexstokes: **A boolean record probe now submits two disposal jobs, several of them under the gate.**

`MemberRecord` is `Clone` and `MemberCell::record()` hands out a full clone. With `stage: Terminal(RetainedExit)` and `last_exit: Option<RetainedExit>` both retained per clone, every read of a terminal failed member's record submits two isolated-disposal jobs when the clone dies. Measured on this branch (temporary counter in `RetainedExit::drop`, reverted):

```
PROBE: 100 member.record() stage probes => 200 disposal submissions
PROBE: 10 root.snapshot() reads        => 0 disposal submissions
PROBE: 10 scope record() reads         => 0 disposal submissions
```

`ScopeRecord` costs nothing because its clones share one `Arc<Vec<RetainedExit>>` guard — exactly the reasoning in your own comment on that field. `MemberRecord` did not get that treatment, so the cost lands on the callers that only want a boolean:

- `cells.rs:1451` `set_child_removing_locked`
- `cells.rs:1521` terminalization's residency lookup, inside `with_observation_gate`
- `cells.rs:970`, `cells.rs:1953`, `cells.rs:2146`
- `driver/child.rs:220`, `:233`, `driver/shutdown.rs:6`, `driver/admission_control.rs:308`

The first four run with the observation gate held, so `matches!(member.record().stage, MemberStage::Terminal(_))` now performs a `spawn_blocking` — or, outside a runtime, a global-mutex enqueue that may `Builder::spawn` a native thread — while holding the outermost framework lock. That is framework work, not user code, so it is not a lock-rule violation; it is still a native thread creation under the gate that main did not have, and it is per lifecycle event rather than per poll only because nothing polls these probes today.

Suggested fix, mirroring `ScopeRecord`: keep `stage`/`last_exit` as raw `Exit` and add a trailing shared `Arc<Vec<RetainedExit>>` guard refreshed at mutation time (raw first, guard after, as `refresh_retained_exits` already does). Clone-and-drop of a record then becomes refcount-only, the last owner still disposes, and the `.map(RetainedExit::into_exit)` / `.into()` churn at the driver and test call sites goes away.

crates/shelterwood-runtime/src/disposal.rs
```diff
@@ -238,6 +269,20 @@ pub fn dispose_detached<T: Send + 'static>(value: T) {
     dispose_then(value, |_| {});
 }
 
+/// Detaches user destruction from a framework critical section.
+///
+/// Unlike [`dispose_detached`], exhausted task and native-thread creation
+/// never falls back to synchronous destruction. The pending job remains in a
+/// static queue until a later submission can start the shared disposal
+/// worker, preserving the lock rule even under resource exhaustion.
+pub fn dispose_critical<T: Send + 'static>(value: T) {
```
ralexstokes, line 335: **The "never synchronous" guarantee holds on the rejection path but not on the acceptance path.**

`submit_blocking_job` decides acceptance from an `Arc::strong_count` sample. If Tokio takes the closure, queues it, and then drops it unrun — a concurrent `shutdown_background`, the #205 embedder shape this crate already tests for — the count falls back to one *after* the sample. `dispose_critical` has already returned by then, and its own `drop(job)` is the last owner: `Drop for DisposalJob` calls `finish()`, which runs the user destructor inline on the submitting thread, i.e. inside the critical section `RetainedExit` exists to protect. The window is only the few instructions between the sample and the drop, but the doc claim above is absolute, and `Drop for RetainedExit` is exactly where an inline blocking destructor would sit under the observation gate.

The enqueue path has no such hole (the queue holds an `Arc`, so the local drop is never last with pending state), and `dispose_detached` is unaffected — inline fallback is its stated contract.

Two ways out, whichever you prefer: mark the job critical and have `Drop for DisposalJob` re-route a still-pending critical payload back through `enqueue_critical_disposal` instead of finishing it (a re-submission is guaranteed to transfer once the runtime is gone, since `is_available()` is then false), or narrow the doc to "a rejected submission never falls back to synchronous destruction" and name the residual.

crates/shelterwood/src/driver.rs
```diff
@@ -104,13 +104,14 @@ impl SystemRun {
...
         };
-        if let Err(exit) = classify_root_driver_join(runtime::join(driver).await) {
+        if let Err(exit) = classify_retained_root_driver_join(runtime::join(driver).await) {
             self.root
                 .finish_live_root_incarnation(StopReason::ShutdownRequested, exit);
         }
     }
 }
 
+#[cfg(test)]
```
ralexstokes: **The test now exercises a copy of the shipped classifier instead of the shipped one.**

`classify_root_driver_join` survives only under `#[cfg(test)]`, and its sole caller is `driver/tests/terminalization.rs:173`. That test is named for the root-driver panic reaching its join monitor, but after this change it classifies through a duplicate: `classify_retained_root_driver_join`'s `Panic`/`Cancelled` arms could be broken and the test would stay green.

The two bodies differ only in the `Ok` arm's payload type. Typing the test's injected driver as `RetainedStopReason` (its unreachable tail return becomes `RetainedStopReason::new(StopReason::Finished)`) lets the test call the production function and lets the duplicate be deleted.

crates/shelterwood-cells/src/observe.rs
```diff
@@ -362,17 +596,16 @@ impl SnapshotHub {
...
-    /// refresh does owe: retiring the projection it displaces, whose child
-    /// exits carry user payloads (§1's lock rule).
+    /// already seen. `RetainedScopeSnapshot` makes the displaced projection
+    /// safe to release before that gate is unlocked.
     pub(crate) fn subscribe(
         &self,
-        initial: Arc<ScopeSnapshot>,
-        txn: &mut crate::cells::ObservationTxn<'_>,
+        initial: RetainedScopeSnapshot,
+        _txn: &mut crate::cells::ObservationTxn<'_>,
```
ralexstokes: `_txn` is now unused: the doc comment above says the retained projection is what makes the displaced snapshot safe to release before unlock, and the body drops `retired` directly. If the parameter is being kept as a proof-of-gate token — which is a reasonable thing to keep — say so in the doc, since the paragraph currently explains why the transaction is *not* needed. Otherwise drop it from the signature; `subscribe_snapshots` already holds the gate around the call.

crates/shelterwood-runtime/src/disposal.rs
```diff
@@ -141,30 +144,58 @@ static FALLBACK_DISPOSALS: Mutex<FallbackDisposals> = Mutex::new(FallbackDisposa
...
+/// ownership and a later disposal submission retries the worker start. The
+/// fail-safe degradation is delayed reclamation, never user destruction in a
+/// framework critical section.
+fn enqueue_critical_disposal(job: Arc<dyn BlockingPoolJob>) {
+    let queued = enqueue_fallback_disposal_with(&FALLBACK_DISPOSALS, job, true, || {
+        std::thread::Builder::new()
+            .name("shelterwood-disposal".to_owned())
+            .spawn(run_fallback_disposals)
+    });
+    debug_assert!(queued, "critical disposal always transfers ownership");
```
ralexstokes: This assertion cannot fail: with `retain_on_spawn_failure = true`, every path in `enqueue_fallback_disposal_with` returns `true`. It reads like a check on the transfer but pins nothing, so a future edit to the retention branch would not trip it.

Since the two callers want different return contracts (`bool` vs. infallible), consider having the shared helper return an enum or splitting the critical wrapper so the type states "ownership transferred" instead of a `debug_assert!` restating it.

AGENTS.md
```diff
@@ -44,17 +44,21 @@ rests on:
...
-  audit confirmed was a drop that *looked* like refcount traffic. One
-  exemption still rests on such a proof: `ScopeCell::clear_residents_locked`
-  and `prune_child_locked` drop `Arc<MemberCell>`s under the gate, and a
-  member record owns an `Exit`. They are safe only because the driver's own
-  child map outlives residency (`ScopeRuntime`'s `Drop` clears residents
-  before `self.children`), which is an ordering nothing enforces.
+  audit confirmed was a drop that *looked* like refcount traffic. Resident
+  member records, scope records, lifecycle events, snapshot projections and
+  driver completions protect failed exits through `RetainedExit`; its drop
+  transfers destruction to a critical-safe disposal queue whose
```
ralexstokes: This paragraph now sanctions disposal submission from `RetainedExit::drop` under a framework lock, but the `dyn` seams bullet three lines below still reads:

> `MailboxRuntime` is nonetheless kept off every locked path: its disposal capability hands work to a blocking worker, so it belongs to the effects flush like the user code it carries.

Both can be true — submitting a job is bounded framework work, while routing a user value through a runtime capability that may block is not — but as written a reader gets opposite guidance on "may I hand a value to disposal while holding a lock?". Worth one clause reconciling them.

Separately, "the fail-safe degradation is delayed reclamation" is optimistic for the terminal case: if thread creation fails and no later disposal is ever submitted, the retained job is never reclaimed at all, and the user error takes whatever it owns (file descriptors, sockets) with it for the life of the process. That is still the right trade against destroying it in a critical section — it just isn't only a delay.

ralexstokes (commented):
Reviewed at `e71f0fb` (full PR, with attention on the new *Keep critical exit disposal queued* commit). `cargo nextest run --workspace` is green here: 654 passed.

The retention shape holds up. I mutation-checked the new regressions by replacing `RetainedExit::drop`'s `dispose_critical` with a plain `drop(exit)`: 9 of them fail, including all five new ones (ring eviction direct and structured, scope-record and snapshot retirement, last-resident member retirement), so the suite pins the behaviour rather than merely describing it. `dispose_critical`'s retain-on-spawn-failure path is also correctly reasoned: the `pop_back` reclamation the non-critical path still does stays correct now that "no live worker implies an empty queue" no longer holds, because the push and the pop happen under one continuous guard.

Six findings inline, none blocking correctness of the retention itself. The first is the one I would want resolved before merge; the rest are contract/coverage/doc issues.

One thing I checked and am satisfied with: `retain_exit`/`retain_owned`/`refresh_retained_exits` all lean on `Exit`'s equality, and for `ExitKind::Failed` that is `Arc::ptr_eq`, so every "an equal retained copy already exists, release this one inline" step is provably refcount-only. Structural equality only collapses payload-free variants.

ralexstokes:
All six review findings are addressed in four commits on top of `e71f0fb`. `./tools/dev just ci` is green (exit 0, 656 tests); the only warning it emits — `update_child_record` is never used without `test-util` — is present on `main` too.

**`6326e3e` Share one exit guard set per member record** — finding 1. `MemberRecord` now has the shape `ScopeRecord` already had: raw `Exit` in `stage`/`last_exit`, one shared `Arc<Vec<RetainedExit>>` declared after them, refreshed by `update_locked` and by terminalization's own mutation so every writer maintains the invariant. Field order keeps a clone's raw exit drop refcount-only, the shared allocation keeps a read refcount-only, and the last surviving clone still isolates. Re-measured with the same temporary counter: **100 record probes went from 200 disposal submissions to 0**. `RetainedExit::install` is factored out so both records share the guard-swap step, and the driver plus its tests go back to naming plain `Exit`. New regression `member_record_reads_share_one_guard_set_and_the_last_one_isolates` pins both halves — `Arc::ptr_eq` across two reads, nothing destroyed by those reads, and off-thread disposal once the cell dies.

**`82b3b21` Close the accepted-submission hole in critical disposal** — findings 2 and 5. Critical jobs are marked, and `Drop for DisposalJob` re-routes a still-pending critical payload onto the fallback queue instead of finishing it. `Drop` runs only when the last reference dies, so the re-routed copy cannot duplicate work, and the queued copy is dropped only after the worker has run it, so it cannot re-enter. That covers the teardown race a reference-count sample cannot see. The mode flag and its unfalsifiable `debug_assert!` are gone: `enqueue_fallback_disposal_with` reclaims and reports, `retain_fallback_disposal_with` returns `()` because every path transfers ownership, and both share `start_worker_locked` under one guard. Claiming a payload now tolerates a poisoned mutex, since the critical drop path can run inside an unwind. New regression `dropping_the_last_critical_job_reroutes_off_the_owning_thread`.

**`6a90fd4` Classify the root-driver join through one function** — finding 3. The injected driver's unreachable tail is typed as `RetainedStopReason`, so the panic regression calls the shipped classifier; the `#[cfg(test)]` duplicate is deleted.

**`5d619e0` Say why a gate token and a locked submission are kept** — findings 4 and 6. `subscribe`'s transaction is renamed `_gate` and documented as what it is: obtainable only under the gate, so requiring it makes the serialization static rather than conventional. In the notes, the `dyn` seams bullet now states the real distinction (a submission runs no user code and is legal under a lock, but can cost a thread start, so a caller that already owns an effects sink should still flush it; retained exits retire from drop glue and have no sink to reach), and the disposal convention records that the exhaustion fallback holds memory for the life of the process rather than merely delaying reclamation.

Mutation check on the final tree: replacing `dispose_critical` with an inline `drop` in `RetainedExit::drop` fails 12 tests, including the residency and ring-eviction regressions and the new member-record one — the restructure did not make any of them vacuous.

PR #257, opened 2026-08-13
Explore the reducer's reachable states instead of event schedules

### LINKED ISSUE #229: Revisit the exhaustive reducer enumeration's cost after stage 4's invariant list exists
`exhaustive_small_scope_interleavings_preserve_reducer_invariants` (`crates/shelterwood-core/src/supervisor.rs`) landed in #227 as stage 3's "exhaustive small-scope enumeration" checkbox. It is now the test suite's critical path and it is worth revisiting once stage 4 exists.

## The cost

Measured locally on `agent/issue-214-stage-3`, debug profile:

```
PASS [  21.390s] (614/614) shelterwood-core supervisor::tests::exhaustive_small_scope_interleavings_preserve_reducer_invariants
Summary [  21.709s] 614 tests run: 614 passed, 0 skipped
```

One test is 21.4s of a 21.7s suite. Everything else in the workspace runs in ~0.3s.

The shape: for each width in {2, 3} and each flavor in {Ordered, Dynamic}, every permutation of the schedule. Width 2 is 8 events (spawn/ready/remove/terminal per child) = 8! = 40,320; width 3 is 9 events (ready/remove/terminal per child) = 9! = 362,880. Roughly 806k full state-machine runs, each replaying admission, 8-9 events, a `Settle` per event, `check_invariants()`, and a per-child property sweep.

## Why it is worth revisiting rather than keeping as-is

- **The permutation count buys less than it costs.** Events for *distinct* children commute except through the shared startup aggregate and the ordered cursor. Most of the 9! schedules are re-exploring the same equivalence class. A canonical-form filter, or enumerating interleavings of per-child *sequences* rather than raw permutations, would likely cover the same reachable states at a fraction of the runs.
- **The properties it asserts are weak.** Today: removal monotonicity, joined non-regression, `all_children_joined` agreeing with a per-child fold, effects naming live keys, plus `check_invariants()`. Those are totality properties. The checkbox's actual wording is "asserting the **stage-4 invariant list**" — which does not exist yet, so the enumeration is currently paying full price for partial value.
- **It did not catch the bug it was best placed to catch.** #227's start-effect spin (a `Settle` emitting `StartChild` for a child that `Event::Spawned` would reject) lives squarely inside this state space. The enumeration missed it because it never asserts anything about the *relationship* between an emitted effect and the event that would acknowledge it — the schedules were the expensive part, the missing property was the cheap part.

## When to revisit

After stage 4's invariant recast, when there is a real list to assert. At that point decide together:

- which invariants belong in the enumeration versus as targeted table-driven cases;
- whether width 3 earns its 9! or whether a reduced event alphabet / canonical-form dedup gives the same coverage;
- whether this should move behind a slower lane (`--ignored`, or a nightly/`ci-nix`-only job) so the ordinary PR loop stays sub-second, with the ordinary lane keeping width 2 only.

Optional stage 5 in #214 (a stateright/TLA+ model) is the other end of this trade: the issue's own note says to hold the model checker "until the list exists and stage 3's exhaustive enumeration shows where it runs out of steam." This is the data point for that decision.

No action wanted before stage 4 — filing so the cost is recorded rather than rediscovered.

ralexstokes:
Closes #229 (the enumeration's cost/value question). Carries the stage-5 decision #214 parked here.

## What changed

The exhaustive enumeration permuted *schedules* — 8!/9! orderings of a four-event alphabet — and asserted totality. It is replaced by a walk of the reducer's *reachable state space*: expand every state against the whole 15-variant alphabet, visit each state once, assert SPEC's invariant list at every transition.

Exploring states rather than orderings is what pays for the wider alphabet. The old test could not reach a single drain, force, restart or reclaim rule, because those events were not in it.

| | old enumeration | new walk (width 2) |
|---|---|---|
| alphabet | 4 of 15 events | 15 of 15 |
| runs | ~806k schedules | 49k states / 1.3M transitions |
| debug cost | 21.4s | 3.5s |
| workspace suite | 21.7s | 3.8s |
| asserts | 4 totality properties | R1–R6, E4, S3–S5 |

## The mapping pass, which decided the rest

Of SPEC's invariant list, only part is stated over `SupervisorState` at all: **R1–R6, E4, S3–S5**. The stop ladder (S1/S2), sampled latches (S6), driver death (S7), the exit funnel (E1–E3/E5/E6) and tree lowering (T1–T7) are not reducer properties, and the walk does not claim them. The three rows that used to cite the enumeration — R5, E4, S5 — now cite a test that actually checks them.

## Mutation evidence

Nine mutations across R1–R6, E4 and S3–S5:

| | old enumeration | new walk |
|---|---|---|
| caught | 1 of 9 | 8 of 9 |

The one the old test caught (the #227 start-effect spin) is the one it was retrofitted for after the fact. The ninth (readiness ignoring the removal-latch sample, R3) survived both — a genuine hole in the first property set, now closed by `check_r3_removal_is_sampled_at_publication`, verified by re-running the mutation.

## Width 2, and that is not a budget compromise

Width 3 costs 1.4M states / 53M transitions against width 2's 49k / 1.3M — **40×** for the same properties — and no mutation covering R1–R6, E4 or S3–S5 survives width 2 but falls to width 3. The structural reason is the load-bearing one: every property here is per-child, a fold over children, or cursor-versus-one-child, so the reducer couples no three memberships and a third child adds combinations rather than cases. The one genuinely three-body distinction (`keys_after(child).next()` vs `.last()`, which needs a middle element to differ) is a progress rule no width detects — see below.

That reasoning lives on `CONFIGURATIONS`; widening is a one-line change, with `MAX_WIDTH` headroom already in the fingerprint. There is deliberately no `#[ignore]`d wide lane: an ignored test is not evidence of the claim it exists to defend.

## What the walk cannot see, and what that means for stage 5

Two mutations were designed to need width 3: an ordered *start* cursor that skips a middle member, and an ordered *stop* cursor that walks two keys back. **Neither is caught at any width**, because both are progress rules — a skipped member is a startup that never completes, not a state that is wrong, and no safety property over reachable states can see that.

- The stop one was already caught by `ordered_stop_releases_one_child_per_join_in_reverse_order`.
- The start one **survived the entire `shelterwood-core` suite**. That was a real coverage hole; `ordered_startup_advances_through_every_initial_member_in_order` closes it, verified by mutation.

So the division of labour is: the walk owns safety over all reachable states; targeted table-driven cases own progress rules and the named single-transition rows.

**Recommendation on optional stage 5 (stateright/TLA+): no, not now.** The walk *is* exhaustive state-space exploration, in the language the reducer is written in, with no model to drift and no CI input to add. What a model checker adds over it is temporal/liveness properties — exactly the gap above. But liveness is only meaningful against a faithful environment model, and this walk deliberately uses an adversarial one (any event in any state, because that is the reducer's actual contract — events arrive from concurrent runtime sources). Building a faithful environment means modelling the driver inside `shelterwood-core`, which is the drift risk the crate split exists to avoid. Two targeted tests cover the two progress rules that exist today. Revisit if the progress rules multiply.

## One spec/code divergence, pinned rather than changed

R6 states aggregate completion as an "iff": `StartupCompleted` iff no resident initial record remains unready. The walk's counterexample: an **ordered** scope whose last member is latched for removal but already ready parks the cursor, so every initial member is ready and startup still does not complete until the removal commits. Defensible as ordered sequencing, and it resolves (the removal commits, the cursor advances), so I pinned current behavior in `ordered_startup_waits_for_a_removing_member_to_commit` and narrowed R6's wording to say the predicate is sufficient for dynamic scopes and necessary for both. **Flagging for adjudication** rather than treating the wording change as settled.

## Notes

- `just ci` green; 647 tests in 3.8s.
- The `update_child_record` dead-code warning in `shelterwood-cells` is pre-existing on `main`, untouched here.

---

## Adversarial review round (4193c33)

An adversarial review reproduced the mutation table with an independent set of 29 mutations, restored the deleted enumeration onto this tree to re-measure it, instrumented all 24 assertion branches for vacuity, and checked the fingerprint for collisions by running a lossless projection beside the packed `u64`.

**Survived scrutiny:** fingerprint soundness (zero collisions at both widths, measured), the mutation ratio (independent set: old enumeration 2/20, walk 20/20), the width-3 claim (six mutations width 2 misses also survive width 3), the cursor-progress claim, and the R6 divergence. No checker is vacuous — every branch fires.

**Did not survive, all now fixed and mutation-verified:** R3 was enforced only on the latched `Ready` variant rather than over the membership, so a `Removing` record could still gain readiness — invisible to all 647 tests, and the one property the deleted enumeration asserted with no successor. R2 named the event but not the incarnation phase. S5 asserted only that a published finish was earned, not that an earned one is published. R6's exemption covered the whole roster where only the cursor can park. SPEC R6's word "resident" was contradicted by the test the row cites. R4's "per *unspawned* initial member" was pinned by nothing — now pinned, since no state-space property can see it.

Both automated Codex comments were confirmed and fixed, each with one correction to its reasoning; details in the [reply](https://github.com/ralexstokes/shelterwood/pull/257#issuecomment-5287485763).


ralexstokes (commented):
## Verdict

Adversarial review: I tried to falsify each load-bearing claim by reproducing the mutation table with an independent set, restoring the deleted enumeration to re-measure it, instrumenting every checker for vacuity, and checking the fingerprint for collisions at both widths.

**Survived scrutiny:** claim 2 (fingerprint soundness — *no collisions at either width, measured*), claim 3 (mutation ratio — my independent set reproduces it), claim 4 (width 3 earns nothing — *confirmed against six mutations width 2 misses*), claim 5 (both cursor-progress mutations are invisible at every width; the two new tests close them), claim 7 (the R6 divergence is real, with an explicit counterexample).

**Did not survive:** claim 1 is over-stated in three specific places (R2's phase constraint, R3's state-level half, S5's "iff" are all named in the module doc but not enforced), and claim 6 has one inaccurate row (R6's word "resident" is contradicted by the test the row now cites).

**No checker is vacuous.** I instrumented all 24 assertion branches; every one fires. Counts below.

Both Codex comments are **CONFIRMED**, with one correction to each — details and one-line fixes at the bottom. Neither is a live bug; both are gaps in the walk.

---

## Method

- Instrumented every branch of every `check_*` with an atomic counter (24 sites) and printed the counts.
- Added a second, non-lossy `String` projection of `SupervisorState` alongside the packed `u64` and asserted the two set cardinalities are equal, at width 2 and width 3.
- 29 mutations of `supervisor.rs` / `engine.rs`, each applied to a clean tree, run against `cargo test -p shelterwood-core --lib`; survivors re-run against `cargo nextest run --workspace` and against the `#[ignore]`d width-3 walk in release.
- Restored the deleted `exhaustive_small_scope_interleavings_preserve_reducer_invariants` onto this tree and ran 20 of those mutations against it alone.

Tree restored; `git status --short` is clean; nothing was committed or pushed.

---

## Confirmed claims, with the evidence

### Claim 2 — the fingerprint is sound. CONFIRMED, measured, not argued.

I replaced the pruning test with a comparison: for every visited state, insert both `fingerprint(...)` and a `format!` of `(flavor, lifecycle, children, child_keys.len(), keys.current(), next_ordered_start, ordered_stop_cursor, ordered_stop_waiting, hard_forced, finish_emitted)` into separate sets, and assert equal cardinality. Any collision shows up as `structural > fingerprints`.

```
COLLISION-CHECK: fingerprints=35293 structural=35293    Ordered [true, true]
COLLISION-CHECK: fingerprints=7725  structural=7725     Dynamic [true, true]
COLLISION-CHECK: fingerprints=5651  structural=5651     Dynamic [true, false]
COLLISION-CHECK: fingerprints=1190378 structural=1190378  Ordered [true, true, true]
COLLISION-CHECK: fingerprints=206039  structural=206039   Dynamic [true, true, true]
```

Bit accounting also checks out: 8 bits/slot × `MAX_WIDTH` 4 = bits 0–31, lifecycle 32–37, flavor/force/finish 38–40, three 3-bit cursors 41–49. `IncarnationState` max 6 fits the 3-bit field at offset 1; `ScopeLifecycleState` projects to 0–5 and `StopPrecedence` to 0–4, both inside 3 bits; the `None` cursor sentinel 7 cannot alias a slot while `MAX_WIDTH ≤ 4`.

One near-miss worth pinning (inline comment below): nothing *enforces* the 3-bit budget on `ScopeLifecycle::fingerprint`'s two `u8`s. The exhaustive `match` in `engine.rs` forces a new variant to *choose* a projection, but it does not stop it choosing `8`, which would silently alias into the flavor bit and prune the space. Two `debug_assert!`s at the pack site cost nothing.

### Claim 3 — old caught 1 of 9, new caught 8 of 9. CORROBORATED independently.

I restored the deleted enumeration onto this tree (baseline: passes, 21.25s debug — the PR's 21.4s is accurate) and ran 20 of my mutations against it alone:

| | old enumeration | new walk |
|---|---|---|
| killed | **2 of 20** | **20 of 20** |

The old test's two kills are `R5_unstartable_start_effect` (the #227 retrofit) and `S5e_all_joined_drift`. Notably the old test *asserted* `"removal is a monotone state transition"` and still let a removal-monotonicity break through (`E4c2`) — its 4-event alphabet has no `StopStarted`, so the mutation was unreachable. That is the strongest single argument in the PR and it holds up.

### Claim 4 — width 3 earns nothing. CONFIRMED against mutations the author did not use.

Six mutations that width 2 misses were run against the `#[ignore]`d width-3 walk in release. All six **survive width 3 as well**:

```
R3b_unlatched_ready_on_removing   :: WIDTH3-SURVIVED
R5b_dynamic_restart_start_effect  :: WIDTH3-SURVIVED
CODEX1a_dynamic_never_finishes    :: WIDTH3-SURVIVED
CODEX1b_dynamic_finishes_undrained:: WIDTH3-SURVIVED
CODEX2b_ready_in_complete         :: WIDTH3-SURVIVED
CODEX2c_ready_in_restart_pending  :: WIDTH3-SURVIVED
```

So the `#[ignore]` is not merely unrefuted, it is confirmed: none of these are width problems, they are *property* problems. My measured cost ratio is 53,063,846 / 1,314,063 = **40.4× transitions**, matching the stated 40×.

### Claim 5 — the two cursor-progress mutations. CONFIRMED.

`R4b_start_cursor_skips` (`keys_after(child).next()` → `.nth(1)`) is invisible to the walk at width 2 *and* width 3, and is killed only by `ordered_startup_advances_through_every_initial_member_in_order` and `ordered_startup_waits_for_a_removing_member_to_commit` — both new in this PR, so the hole was real. `S3d_stop_cursor_skips` is killed only by the pre-existing `ordered_stop_releases_one_child_per_join_in_reverse_order`. Both exactly as described.

### Claim 7 — the R6 divergence. CONFIRMED with a counterexample.

Deleting `&& !ordered_cursor_may_wait` from the checker makes the walk fail with a concrete witness:

```
a settle that finds no unready initial member completes startup: Ordered {
  ChildKey(1): state: Removing(Active),  startup: Initial { ready: true }, spawned_once: true,
  ChildKey(2): state: Resident(Active),  startup: Initial { ready: true }, spawned_once: true }
  left: 0  right: 1
```

The exemption is load-bearing in **147** of the 265 states it covers (instrumented). And the pin discriminates: mutating the ordered `Removing` early-return into a cursor advance fails `ordered_startup_waits_for_a_removing_member_to_commit` **and nothing else** in the crate — a proper single-rule pin.

### Vacuity — no decorative checker. CONFIRMED.

Every branch fires at width 2 (counts over 1,314,063 transitions):

```
e4 key-leaves-roster 23242 | e4 key-absent 162270 | e4 phase actually moved 100035
e4 before-Removing 1222633 | e4 before-spawned_once 2069042
r5 StartChild 135 | r5 Stop/Force 165404 | r5 FinalizeRemoval 31897
r5 fixed-point entered 48669 | ...with non-empty starts 134
r3 asserts reached 91328 | ...with before-ready true 36232
r6 before startup_complete 86103 | ready bit set 5900 | ready bit cleared 5900
StartupCompleted emitted 563 | r6 progress assert 445 | r6 exemption taken 265
r4 starts non-empty 134 | r4 starts >1 1
s3/s4 Force 48669 | fresh BeginDrain 15276 | ordered settle with a stop 14736
s5 Finished emitted 2988
```

Two are *thin* rather than vacuous and worth knowing about: `StartChild` appears on only **135** of 1.3M transitions (settlement emits starts only while `is_starting` and `!spawned_once`), and `starts.len() > 1` happens exactly **once** in the whole walk. R5's start-effect assertion still killed its mutation, so it is live — but that surface is one code change away from going quiet, and the walk would not tell you.

---

## Findings, ranked

### 1. HIGH — R3's rule is enforced only on the latched event variant, not on the state. Regression from the deleted test's intent.

`check_r3_removal_is_sampled_at_publication` returns early unless the event is `Ready { removal_latched: true }`. But SPEC R3 states the rule over the *membership*: a `Removing` membership rejects readiness. A child that became `Removing` earlier (via `RemovalSampled` or `RemovalLatched`) and then receives `Ready { removal_latched: false }` — a pair the alphabet offers in every such state — is asserted about by nothing.

Mutation `R3b`, narrowing the reducer's guard to `(status == Removing && removal_latched)`:

```
R3b_unlatched_ready_on_removing :: SURVIVED               (shelterwood-core)
R3b_unlatched_ready_on_removing :: WORKSPACE-SURVIVED     (all 647 tests)
R3b_unlatched_ready_on_removing :: WIDTH3-SURVIVED
```

**A `Removing` membership gaining readiness is invisible to the entire repository.** Fix, verified to pass clean and to kill `R3b`:

```rust
for &child in transition.keys {
    if transition.before.contains(child)
        && transition.before.membership_status(child) == MembershipStatus::Removing
    {
        assert!(
            !transition.after.initial_ready(child) || transition.before.initial_ready(child),
            "a Removing membership never gains readiness, got {:?}", transition.event
        );
    }
}
```

### 2. HIGH — R2's phase constraint is claimed but not checked (this is Codex's second comment, and it is right).

See the adjudication below. Two narrow mutations survive all 647 tests at both widths.

### 3. MEDIUM — SPEC R6's word "resident" is contradicted by the test the row now cites.

The rewritten row reads *"never while a **resident** initial record remains unready"*. The reducer's aggregate does not filter on residency:

```rust
.all(|record| !matches!(record.startup, StartupMembership::Initial { ready: false }))
```

and neither does the walk's `unready`. I made the code match the SPEC wording literally (skip `Removing` records in the aggregate) and **the walk fails**:

```
supervisor::tests::exploration::…preserve_the_reducer_invariants  FAILED  (exploration.rs:503)
supervisor::tests::sampled_removal_suppresses_start_effects_until_commit  FAILED
```

So a `Removing`-but-unready initial record *does* withhold completion, exactly as R3's "shrinks the initial set only at `Reclaim`" implies. The word "resident" is wrong here — and note R5 in the same section uses "resident" in the `ChildState::Resident` sense, so it reads as a real qualifier. Drop it. This is pre-existing wording, but the PR rewrote this row and added the citation, which makes it in scope for claim 6.

### 4. MEDIUM — S5's "iff" is asserted in one direction only (Codex's first comment, also right).

See adjudication. Both directions survive the core suite; the workspace catches them, but only in `shelterwood`, not in the crate the SPEC row now cites.

### 5. LOW–MEDIUM — SPEC R4's "per **unspawned** initial member" is pinned by nothing.

Dropping `!record.spawned_once` from the dynamic startup arm (so a `RestartPending` initial member re-attracts a `StartChild` from settlement) survives **all 647 tests at both widths**. The walk cannot see it: `startable()` holds, so R5's acknowledgeability passes, and re-settling reproduces the same start, so the fixed-point check passes. `start_effects_are_confined_to_the_spawn_transition` cannot see it either — its own comment notes `spawned_once` is false in every arm. Whether this is a latent double-construction hazard against the engine's restart path is worth a look independently of the test question.

### 6. LOW — the R6 exemption is broader than the divergence it exists for.

`ordered_cursor_may_wait` is true whenever *any* member of an ordered scope is `Removing`. The actual parking rule is about the cursor. Instrumented: of the 265 states the exemption covers, **118 would have satisfied the assertion anyway** — the exemption is silently switching off a live property in those. A tighter predicate passes the full walk unchanged:

```rust
&& transition.before.next_ordered_start().is_some_and(|cursor| {
    transition.before.children.iter().any(|(&key, record)| {
        key >= cursor && record.state.membership_status() == MembershipStatus::Removing
    })
});
```

That also makes the SPEC sentence precise: it is not "a member latched for removal" that withholds completion, it is a member latched for removal *at or ahead of the cursor*.

### 7. LOW — doc says "Two fields are deliberately not encoded"; four are.

`child_keys`, `keys`, `ordered_stop_inspections` (justified inline instead) and `ChildRecord::membership` (justified above `slot`). The justifications are all sound — `ordered_stop_inspections` is genuinely never read by `apply` (grepped), and the membership-per-slot argument holds because `seen` is local to one `explore` call — but the count in the sentence that sells the exhaustive destructure is wrong.

### 8. LOW — no bound assertion on the lifecycle fingerprint's 3-bit fields. See inline.

---

## Codex adjudications

### `exploration.rs:662` (S5 checker) — **CONFIRMED, with a correction to the suggested fix.**

Both halves reproduce.

```
CODEX1a_dynamic_never_finishes    :: SURVIVED (core)   :: WIDTH3-SURVIVED
CODEX1b_dynamic_finishes_undrained:: SURVIVED (core)   :: WIDTH3-SURVIVED
```

`CODEX1a` = `settle_finish` returns early for `ScopeFlavor::Dynamic` (settlement omits completion for a draining dynamic scope). `CODEX1b` = `finish_if_ready` drops its `flavor == Ordered` clause (a non-draining dynamic scope finishes once its children join). The walk passes both, so the checker does not enforce S5's "iff", exactly as the comment says.

**Correction to the severity:** the workspace does catch both, so this is a gap in the walk and in the SPEC citation, not an unguarded hole:
- `CODEX1a` → `shelterwood driver::tests::events::queued_admissions_yield_to_shutdown_without_forwarder_tasks` (fails at 10.0s — it hangs, so this is timeout coverage, not an assertion)
- `CODEX1b` → `shelterwood::integration dynamic::tombstones_occupy_ids_until_explicit_removal`

**Correction to the fix:** comparing the emitted effect against `lifecycle.finish_if_ready(...)` — the literal suggestion — would restate `settle_finish`'s own body and be near-tautological. Two independent state-level properties get the same kills without mirroring the implementation, and I verified they pass clean and fail under the mutations:

```rust
let finished = transition.effects.iter().any(|e| matches!(e, Effect::Finished { .. }));
// kills CODEX1a
if matches!(transition.event, Event::Settle)
    && !transition.before.finish_emitted
    && transition.before.lifecycle().is_draining()
    && transition.after.all_children_joined()
{
    assert!(finished, "a drained scope whose children have all joined finishes");
}
// kills CODEX1b
if finished {
    assert!(
        transition.after.lifecycle().is_draining()
            || (transition.after.flavor() == ScopeFlavor::Ordered && !transition.after.is_empty()),
        "only a draining scope, or a non-empty ordered scope, finishes"
    );
}
```

### `exploration.rs:468` (R2 checker) — **CONFIRMED, with a correction to the example.**

The comment's *reasoning* is exactly right: the assertion accepts the edge purely because the event is `Ready`, and E4 does not see the readiness bit.

**Correction:** the specific example it gives (`Unstarted`) is caught, but only *incidentally* and only in one configuration. Deleting the whole phase guard (`CODEX2_ready_any_phase`) fails the walk — at `exploration.rs:523`, R6's *progress* assertion, because an ordered scope with two `Unstarted { ready: true }` members withholds `StartupCompleted`. That is an accidental kill through an unrelated property.

Narrow the mutation to just one out-of-order phase and the kill disappears entirely. Both of these **pass all 647 tests, at width 2 and width 3**:

```
CODEX2b_ready_in_complete          (Active | Stopping | Complete)        :: 647 passed
CODEX2c_ready_in_restart_pending   (Active | Stopping | RestartPending)  :: 647 passed
```

So the conclusion "such an out-of-order readiness regression remains invisible" holds. `CODEX2c` is the more interesting one — it means R2's own headline, *"readiness is incarnation-local and monotone until restart"*, can be broken in the restart direction with no test noticing.

Fix — three lines inside the existing `if !before && after` block, verified to pass clean and to kill both `CODEX2b` (`got Some(Complete)`) and `CODEX2c` (`got Some(RestartPending)`):

```rust
assert!(
    matches!(
        incarnation(transition.before, child),
        Some(IncarnationState::Active | IncarnationState::Stopping)
    ),
    "readiness publishes only from an executing incarnation, got {:?}",
    incarnation(transition.before, child)
);
```

---

## Things I checked and found nothing wrong with

- **The environment model.** Offering every event in every state makes the safety properties strictly *stronger*, not weaker; I found no property that holds only because an unreachable state happens not to violate it. The one place adversarial reachability leaks into a weaker claim is the R6 exemption (finding 6), and even there the tighter form passes.
- **Coverage regression from the deleted test.** Everything it asserted is re-asserted somewhere — removal monotonicity (`E4c2` killed by `check_e4`), joined non-regression (E4's empty successor set), the derived completion fold (`S5e` killed), effects naming live keys, `startable` on starts — *except* the state-level readiness/removal property, which is finding 1.
- **`assert_eq!(after.keys.current(), minted)`.** Correctly guards the one excluded field whose justification is dynamic rather than structural.
- **`ordered_stop_inspections` exclusion.** `apply` never reads it (grepped); it is genuinely diagnostic.
- **Tautologies.** None found. `check_r5_settlement_reaches_a_fixed_point` looked like it might restate the reducer, but `S5d_finish_twice` proves it discriminates a real repeat-effect spin.
- **Numbers in the PR body.** 40.4× measured vs 40× claimed; 806,400 old schedules; 21.25s old debug vs 21.4s claimed; 1,396,417 / 53,063,846 at width 3. All accurate.

crates/shelterwood-core/src/supervisor/tests/exploration.rs
```diff
@@ -0,0 +1,707 @@
...
+/// R3 — removal is sampled at the publication transition. A true sample marks
+/// the membership `Removing` first, and readiness is then rejected, so the
+/// removal path can never manufacture the readiness edge it raced.
+fn check_r3_removal_is_sampled_at_publication(transition: &Transition<'_>) {
+    let Event::Ready {
+        child,
+        removal_latched: true,
+    } = transition.event
+    else {
+        return;
```
ralexstokes, line 449: **HIGH — R3 is enforced only on the latched event, not on the state.**

This early return means the checker never looks at a child that is *already* `Removing` and then receives `Ready { removal_latched: false }` — a pair the alphabet offers in every such state. SPEC R3 states the rule over the membership ("readiness is then rejected"), not over the latch bit the event happens to carry.

Mutation: narrow the reducer's guard to `(status == Removing && removal_latched)`.

```
SURVIVED            (shelterwood-core --lib)
WORKSPACE-SURVIVED  (all 647 tests)
WIDTH3-SURVIVED     (the #[ignore]d walk)
```

A `Removing` membership gaining readiness is invisible to the whole repository. This is the one property the deleted enumeration asserted (`"removal never manufactures a readiness edge"`, stated over the state) that has no successor here.

Fix — verified to pass clean and to kill the mutation:

```rust
for &child in transition.keys {
    if transition.before.contains(child)
        && transition.before.membership_status(child) == MembershipStatus::Removing
    {
        assert!(
            !transition.after.initial_ready(child) || transition.before.initial_ready(child),
            "a Removing membership never gains readiness, got {:?}", transition.event
        );
    }
}
```

crates/shelterwood-core/src/supervisor/tests/exploration.rs
```diff
@@ -0,0 +1,707 @@
...
+            continue;
+        }
+        let before = transition.before.initial_ready(child);
+        let after = transition.after.initial_ready(child);
+        if !before && after {
+            assert!(
+                matches!(transition.event, Event::Ready { child: key, .. } if *key == child),
+                "only a readiness edge sets the aggregate bit, got {:?}",
+                transition.event
+            );
```
ralexstokes, line 488: **Adjudicating the Codex comment on this line: CONFIRMED, with a correction to its example.**

The reasoning is right — this accepts the edge purely because the event is `Ready`, and E4 never sees the readiness bit. But the `Unstarted` case it names *is* caught, just not here and not on purpose: deleting the whole phase guard fails the walk at line 523, R6's **progress** assertion, because an ordered scope with two `Unstarted { ready: true }` members withholds `StartupCompleted`. An accidental kill through an unrelated property.

Narrow the mutation to a single out-of-order phase and the kill disappears. Both of these pass all 647 workspace tests, at width 2 and at width 3:

```
Ready accepted in Complete         (Active | Stopping | Complete)        :: 647 passed
Ready accepted in RestartPending   (Active | Stopping | RestartPending)  :: 647 passed
```

The second is the one that stings: R2's own headline is *"readiness is incarnation-local and monotone until restart"*, and it can be broken in the restart direction with nothing noticing.

Fix — three lines inside this same `if !before && after` block, verified to pass clean and to kill both (`got Some(Complete)` / `got Some(RestartPending)`):

```rust
assert!(
    matches!(
        incarnation(transition.before, child),
        Some(IncarnationState::Active | IncarnationState::Stopping)
    ),
    "readiness publishes only from an executing incarnation, got {:?}",
    incarnation(transition.before, child)
);
```

crates/shelterwood-core/src/supervisor/tests/exploration.rs
```diff
@@ -0,0 +1,707 @@
...
+        "the derived completion query cannot drift from child state"
+    );
+    for effect in transition.effects {
+        if let Effect::Finished { .. } = effect {
+            assert!(
+                !transition.before.finish_emitted,
+                "the finish edge is published once"
+            );
+            assert!(
+                transition.after.all_children_joined(),
```
ralexstokes: **Adjudicating the Codex comment here: CONFIRMED, with corrections to severity and to the fix.**

Both halves reproduce and both are also invisible at width 3:

```
settle_finish returns early for Dynamic (draining dynamic never finishes) :: SURVIVED core, WIDTH3-SURVIVED
finish_if_ready drops its `flavor == Ordered` clause (undrained dynamic finishes) :: SURVIVED core, WIDTH3-SURVIVED
```

**Severity correction:** the workspace does catch both, so this is a gap in the walk and in the SPEC S5 citation, not an unguarded hole — `driver::tests::events::queued_admissions_yield_to_shutdown_without_forwarder_tasks` (by hanging, 10.0s) and `integration dynamic::tombstones_occupy_ids_until_explicit_removal` respectively. Neither lives in the crate the S5 row now cites.

**Fix correction:** comparing the effect against `lifecycle.finish_if_ready(...)`, as suggested, would restate `settle_finish`'s body and be near-tautological. Two state-level properties get the same kills without mirroring the implementation; verified to pass clean and to fail under each mutation:

```rust
let finished = transition.effects.iter().any(|e| matches!(e, Effect::Finished { .. }));
if matches!(transition.event, Event::Settle)
    && !transition.before.finish_emitted
    && transition.before.lifecycle().is_draining()
    && transition.after.all_children_joined()
{
    assert!(finished, "a drained scope whose children have all joined finishes");
}
if finished {
    assert!(
        transition.after.lifecycle().is_draining()
            || (transition.after.flavor() == ScopeFlavor::Ordered && !transition.after.is_empty()),
        "only a draining scope, or a non-empty ordered scope, finishes"
    );
}
```

crates/shelterwood-core/src/supervisor/tests/exploration.rs
```diff
@@ -0,0 +1,707 @@
...
+            | u64::from(initial) << 5
+            | u64::from(ready) << 6
+            | u64::from(spawned_once) << 7;
+        word |= packed << (8 * slot(key));
+    }
+
+    let cursor = |key: &Option<ChildKey>| key.map_or(7, slot);
+    let (lifecycle_state, lifecycle_reason) = lifecycle.fingerprint();
+    word |= u64::from(lifecycle_state) << (8 * MAX_WIDTH);
+    word |= u64::from(lifecycle_reason) << (8 * MAX_WIDTH + 3);
```
ralexstokes, line 118: **LOW — nothing enforces the 3-bit budget these two `u8`s are packed into.**

The layout is correct today, and I verified it empirically: comparing `seen` against a lossless `String` projection of every visited state shows zero collisions at width 2 (35293/7725/5651) and at width 3 (1190378/206039).

But the safety argument in the `fingerprint` doc — a new field "cannot compile until it is classified" — does not extend here. `ScopeLifecycle::fingerprint`'s exhaustive `match` forces a new variant to *choose* a projection; it does not stop it choosing `8`. `lifecycle_state` currently maxes at 5 and `lifecycle_reason` (a `StopPrecedence`) at 4, so there are 2 and 3 spare values before the value silently aliases into the flavor bit at 38 — and a collision prunes the space rather than failing.

One line makes the budget structural instead of arithmetical:

```rust
debug_assert!(lifecycle_state < 8 && lifecycle_reason < 8, "the lifecycle projection outgrew its field");
```

crates/shelterwood-core/src/supervisor/tests/exploration.rs
```diff
@@ -0,0 +1,707 @@
...
+    // even when every initial member is already ready. That is a deliberate
+    // ordered-sequencing rule rather than a defect, but it is not what R6 says
+    // — see `ordered_startup_waits_for_a_removing_member_to_commit`, which
+    // pins the ordered behavior on its own.
+    let ordered_cursor_may_wait = transition.before.flavor() == ScopeFlavor::Ordered
+        && transition
+            .before
+            .children
+            .values()
+            .any(|record| record.state.membership_status() == MembershipStatus::Removing);
```
ralexstokes: **LOW — the exemption is broader than the divergence it exists for.**

The divergence itself is real; I confirmed it by deleting `&& !ordered_cursor_may_wait` and reading the witness the walk produces:

```
a settle that finds no unready initial member completes startup: Ordered {
  ChildKey(1): state: Removing(Active), startup: Initial { ready: true }, spawned_once: true,
  ChildKey(2): state: Resident(Active), startup: Initial { ready: true }, spawned_once: true }
```

But this predicate is true whenever *any* member is `Removing`, while the parking rule is about the cursor. Instrumented over the width-2 walk: the exemption is taken 265 times and is genuinely needed in **147** of them — in the other **118** the assertion would have passed, so a live property is being switched off for free.

A tighter predicate passes the whole walk unchanged:

```rust
&& transition.before.next_ordered_start().is_some_and(|cursor| {
    transition.before.children.iter().any(|(&key, record)| {
        key >= cursor && record.state.membership_status() == MembershipStatus::Removing
    })
});
```

It also sharpens the SPEC sentence this comment justifies: it is not "a member latched for removal" that withholds completion, it is one latched *at or ahead of the cursor*.

SPEC.md
```diff
@@ -1187,22 +1187,33 @@ integration suite and `supervisor::…` names the pure reducer suite.
...
-   `supervisor::exhaustive_small_scope_interleavings_preserve_reducer_invariants`.)
+   emits no acknowledgeable work is already at a fixed point, and re-entering
+   `Settle` on an unchanged state reproduces its start effects and nothing
+   else. This prevents the driver’s level-triggered settle loop from spinning.
+   (`supervisor::start_effects_are_confined_to_the_spawn_transition`,
+   `supervisor::exhaustive_reachable_states_preserve_the_reducer_invariants`.)
 6. **R6 — aggregate completion is derived.** `Settle` emits
-   `StartupCompleted` at most once, iff no resident initial record remains
-   unready. The empty initial set satisfies the predicate. (`integration::dynamic_startup_completes_after_removing_sole_unready_initial_member`,
+   `StartupCompleted` at most once, and never while a resident initial record
```
ralexstokes: **MEDIUM — "resident" is contradicted by the test this row now cites.**

The reducer's aggregate does not filter on residency:

```rust
.all(|record| !matches!(record.startup, StartupMembership::Initial { ready: false }))
```

and neither does the walk's `unready`. I made the code match this wording literally — skip `Removing` records in the aggregate — and the walk fails:

```
supervisor::tests::exploration::…preserve_the_reducer_invariants   FAILED  (exploration.rs:503)
supervisor::tests::sampled_removal_suppresses_start_effects_until_commit  FAILED
```

So a `Removing`-but-unready initial record *does* withhold completion — which is what R3's "shrinks the initial set only at `Reclaim`" implies, and is the right behavior. The word is just wrong, and R5 four lines down uses "resident" in the `ChildState::Resident` sense, so it reads as a real qualifier rather than filler. Drop it.

Pre-existing wording, but the row was rewritten and given this citation in this PR, so it is in scope for the claim that every rewritten row cites a test that discriminates the rule as stated.

SPEC.md
```diff
@@ -1187,22 +1187,33 @@ integration suite and `supervisor::…` names the pure reducer suite.
...
+   `supervisor::exhaustive_reachable_states_preserve_the_reducer_invariants`,
    `integration::queued_removal_suppresses_replayed_self_stop_readiness`,
    `integration::startup_removal_response_follows_aggregate_recomputation`.)
 4. **R4 — ordered start is one accepted edge at a time.** `Settle` emits
    `StartChild` only for the current initial cursor, and advances the cursor
-   only past a spawned-and-ready member or a reclaimed key. Dynamic startup
-   may emit one accepted start per unspawned initial member. (`supervisor::start_effects_are_confined_to_the_spawn_transition`,
+   only past a spawned-and-ready member or a reclaimed key, reaching every
+   initial member in declaration order. Dynamic startup may emit one accepted
+   start per unspawned initial member. (`supervisor::start_effects_are_confined_to_the_spawn_transition`,
```
ralexstokes, line 1197: **LOW–MEDIUM — the word "unspawned" in this clause is pinned by nothing.**

Mutation: drop `!record.spawned_once` from `settle_startup`'s dynamic arm, so a `RestartPending` initial member re-attracts a `StartChild` from settlement.

```
SURVIVED            (shelterwood-core --lib)
WORKSPACE-SURVIVED  (all 647 tests)
WIDTH3-SURVIVED
```

The walk structurally cannot see it: `startable()` holds so R5's acknowledgeability passes, and re-settling reproduces the same start so the fixed-point check passes. `start_effects_are_confined_to_the_spawn_transition` cannot see it either — its own comment notes `spawned_once` is false in every arm it builds.

Separately from the test question: worth confirming this is not a live double-construction hazard against whatever else drives a restart-pending member's start.

ralexstokes:
All eight findings applied in 4193c33, and both Codex comments confirmed and fixed. Every fix is mutation-verified against the exact mutation that motivated it — all six that survived the review now die in `shelterwood-core --lib` alone:

```
R3b   removing membership still accepts an unlatched Ready  -> killed by the walk
CODEX2b  Ready accepted in Complete                          -> killed by the walk
CODEX2c  Ready accepted in RestartPending                    -> killed by the walk
CODEX1a  a draining dynamic scope never finishes             -> killed by the walk
CODEX1b  a non-draining dynamic scope finishes               -> killed by the walk
R4c   dynamic settlement re-attracts a restart-pending member -> killed by
      dynamic_startup_leaves_a_restart_pending_member_to_the_restart_path
```

`just ci` green: 648 tests in 3.9s.

**1 (HIGH), R3 stated over the event not the membership.** Applied as suggested — the rule now runs over every roster key whose membership was already `Removing`, so the `RemovalSampled`-then-unlatched-`Ready` pair is covered. This was the one property the deleted enumeration asserted with no successor; thanks for catching it.

**2 (HIGH), R2's phase constraint.** Applied — readiness must come from `Active | Stopping`. See the Codex adjudication below.

**3 (MEDIUM), SPEC R6's "resident".** Fixed, and your reading is the right one: the row now says *"never while any initial record remains unready — including one already latched for removal, which leaves the initial set only at `Reclaim` (R3)"*, which is both what the code does and what R3 implies.

**4 (MEDIUM), S5's "iff".** Applied in your non-tautological form rather than Codex's literal suggestion, for the reason you gave — comparing against `finish_if_ready` would restate `settle_finish`'s body.

**5 (LOW–MED), R4's "unspawned".** Pinned by a new targeted test, since as you say no state-space property can see it: the record is `startable`, so acknowledgeability passes, and re-settling reproduces the start, so the fixed point holds. On the latent-hazard question — the guard is what keeps settlement and the restart deadline from both constructing the same incarnation, so it is load-bearing rather than incidental, and the test now says so.

**6 (LOW), the R6 exemption's breadth.** Tightened to your cursor-scoped predicate; the walk passes unchanged. It also sharpened the SPEC sentence, which now says a member latched for removal *at or ahead of the cursor* withholds completion.

**7 (LOW), "two fields" was four.** Corrected, with `ordered_stop_inspections` and `ChildRecord::membership` now named in the same list instead of only at their use sites.

**8 (LOW), the unenforced 3-bit budget.** Added `assert!(lifecycle_state < 8 && lifecycle_reason < 8)` at the pack site. Agreed this was the real risk in the fingerprint: the exhaustive `match` forces a new variant to *choose* a projection but not to choose one that fits.

## Codex comments

Both **CONFIRMED and fixed**, with the reviewer's two corrections adopted.

- **S5 (`exploration.rs:662`)**: right that the checker only inspected emitted effects. Fixed with two state-level properties instead of the suggested `finish_if_ready` comparison, which would have mirrored the implementation. Severity was lower than implied — the workspace caught both directions, though `CODEX1a` only by hanging — but the SPEC row cited `shelterwood-core`, so the gap was real where it was claimed.
- **R2 (`exploration.rs:468`)**: right in its reasoning. One correction: its `Unstarted` example *was* caught, but accidentally, through R6's progress assertion in the ordered configuration. Narrowed to `Complete` or `RestartPending`, both survived all 647 tests at both widths — so the conclusion held even though the example did not.

## Also in this push, unrelated to the review

The `#[ignore]`d width-3 lane is deleted. The measurement stands, but an ignored test is not evidence of the claim it exists to defend, and the structural argument is stronger than the empirical one: every property here is per-child, a fold over children, or cursor-versus-one-child, so a third membership adds combinations rather than cases. That reasoning now lives on `CONFIGURATIONS`, and widening stays a one-line change with `MAX_WIDTH` headroom in the fingerprint.

Not changed: the two thin surfaces you flagged — `StartChild` on 135 of 1.3M transitions, and `starts.len() > 1` exactly once. Both are inherent to settlement only emitting starts while `is_starting && !spawned_once`, and both still kill their mutations. Worth knowing, not worth contorting the walk for.

PR #288, opened 2026-08-16
Make snapshot watches transaction-consistent

### LINKED ISSUE #271: Adjudicate: SPEC §12's consistent cut vs the ungated watch surface (wait_for_child / borrow_latest)
Adjudication question from the 2026-08-16 review round (external claim, verified; mechanism confirmed, spec conformance split).

## Two observation surfaces, two consistency levels

- `SnapshotHub::publish` installs the new watch value synchronously under the gate; only the pulse is deferred to commit (`observe.rs:652-678`).
- `set_admitted_children` (`cells.rs:2217-2224`) performs one `clear_residents_locked` plus N `admit_child_locked` — each ending in `emit_locked` → publish for the scope **and every ancestor** — inside one held txn. The driver's initial admission passes the whole declared child set (`driver.rs:909-915`).
- `ScopeCell::snapshot()` and `subscribe_snapshots()` take the gate, so they block until commit and cannot tear. `SnapshotReceiver::borrow_latest`/`borrow_latest_and_closed` (`observe.rs:541-551`) and therefore `wait_for_child`'s loop head (`scope.rs:119`) read the watch with **no** gate — and two real wake paths (the timer arm at `scope.rs:138-143`; a pulse from an earlier committed txn) put the predicate in front of a mid-batch value. A waiter can see 0-of-N, then 1-of-N, then 2-of-N children during one admission.

The atomicity anchors both use the gated reader: `receiverless_config_state_is_atomic_under_concurrent_snapshots` reads `scope.snapshot()` (`driver/tests/observation.rs:999`), as does the events-vs-snapshots consistency test. Nothing exercises the watch path's view of a batch.

## The spec tension

SPEC §12's consistent-cut guarantee enumerates per-membership transitions (batch admission is not one of them) and B.9 explicitly permits watch conflation — so per-admit visibility does not violate §12 *as written*. But §12 also names `wait_for_child` as "an observer in this sense", i.e. bound by the cut, while the implementation gives it a strictly weaker view than `snapshot()`. And the codebase already leans on the gated reader for correctness (`tests/dynamic.rs:38-47`: "an absent resident is a sound signal the id is free again" — via `scope.as_scope().child(id)`; the same inference through `borrow_latest` would be unsound).

## Options

1. Strengthen the implementation: stage one snapshot per transaction and install at commit (per-hub coalescing in the txn), making both surfaces cut-consistent. Also collapses the M×(1+ancestors) redundant installs per batch.
2. Weaken the spec: restate §12 to say only the gated pull surfaces (`snapshot()`, `ScopeRef::child`/`descendant`) are cut-consistent; the watch path is per-publication with conflation.

## Acceptance criteria

- [ ] Adjudicate 1 vs 2; amend SPEC §12/B.9 accordingly.
- [ ] If 1: a test observing a batch admission through `borrow_latest`/`wait_for_child` sees all-or-nothing.
- [ ] If 2: a doc note on `borrow_latest`/`wait_for_child` naming the weaker guarantee.

ralexstokes:
## Ruling

Choose option 1: every snapshot-watch value is a complete observation transaction cut. `borrow_latest` and `wait_for_child` see either the preceding committed cut or the transaction's final cut, never an intermediate projection from a compound transition.

## Summary

- stage and coalesce snapshot publications per hub inside `ObservationTxn`
- install the final projection at commit while the observation gate is held
- defer retired snapshots, wakes, and generation-exhaustion panic until after unlock
- preserve one generation edge for publish-plus-close transactions
- align SPEC §12/B.9 and the public observation guidance with committed-cut conflation
- add a deterministic batch-admission regression

## Why

SPEC §12 explicitly includes `wait_for_child` in its consistent-cut guarantee. Weakening the watch surface would give one published snapshot two incompatible meanings depending on whether it was read through a gated pull or an ungated borrow.

## Impact

Initial admission and other compound transactions become all-or-nothing to watch readers. Watch conflation still skips intermediate committed cuts, but no longer exposes a transaction's internal publications. Redundant watch installs and generation mints within a transaction are eliminated.

## Validation

- focused cells/driver observation suites
- deterministic `borrow_latest` batch-admission regression
- `./tools/dev just ci` — 661 nextest tests plus Clippy, doctests, rustdoc/API reachability, packaged-doc sync, and Nix formatting

Closes #271


ralexstokes:
## Adversarial review

Reviewed head: `8dda2400f71cdccca0fc747e383bfb08c83e6eb0`

### Attack surface

I traced the full PR diff and the relevant base implementation through:

- `ObservationTxn` staging, per-hub coalescing, commit ordering, effect flushing, and unwind-time `Drop`
- `SnapshotPublication` same-channel identity, final-value replacement, generation minting/exhaustion, close dominance, and post-close laziness
- receiverless initialization/refresh and terminal initialization
- publish-plus-close generation/wake behavior and hostile-waker containment
- `RetainedScopeSnapshot` / `RetainedExit` retirement under the repository lock rule
- recursive snapshot propagation through every ancestor hub
- SPEC §12/B.9, public receiver/wait documentation, deterministic batch-admission coverage, and staging complexity

### Confirmed findings fixed

1. **Receiverless refresh retired a retained snapshot under the observation gate.** The retained exit wrapper prevented inline user destruction, but this still submitted critical disposal while an effects sink was available. Refresh retirement now goes through `ObservationTxn::defer`, so it necessarily occurs after unlock.
2. **Receiverless generation exhaustion still panicked under the gate.** The ordinary staged-publication path deferred this panic, but `subscribe` retained an in-lock `expect`. It now records exhaustion and resumes the panic from the transaction effects after unlock.
3. **First initialization performed a redundant refresh and generation mint.** The freshly initialized channel had zero receivers, so the old branch immediately replaced its just-installed value and consumed a generation. Initialization is now distinguished from a later receiverless refresh and starts at generation zero.

Added focused regressions for zero-mint initialization, deferred receiverless exhaustion, staged publication during unwind, and publish-after-staged-close laziness. The existing hostile failed-exit retirement test also covers the corrected receiverless path.

After those fixes, I found no remaining correctness issue in coalescing/close precedence, one-edge publish-plus-close behavior, wake ordering, same-hub identification, ancestor final-cut propagation, RetainedExit protection, documentation, test determinism, or the vector staging complexity relative to recursive snapshot construction.

### Checks

- `./tools/dev cargo test -p shelterwood-cells` — 22 passed
- `./tools/dev cargo test -p shelterwood --lib driver::tests::observation` — 42 passed
- `./tools/dev just ci` — green: formatting, both Clippy lanes, 661 nextest tests, doctests, rustdoc/API reachability, packaged-doc sync, and Nix formatting
- GitHub Actions `ci` is pending at the time of this comment: https://github.com/ralexstokes/shelterwood/actions/runs/31973280772/job/95228911608

ralexstokes:
## Independent review — mechanism is sound; no correctness bug. One P1 coverage gap, one incomplete self-review fix, three P2s

Reviewed head `8dda240`. Second-opinion pass, independent of the authoring agent's self-review. I found **no correctness defect** in staging, coalescing, close dominance, generation minting, unwind handling, or the lock rule as applied to the new paths. What follows is coverage and hygiene.

### F1 — P1 · `shelterwood-cells/src/cells.rs:966-978` — per-hub coalescing is entirely unpinned

The PR's second headline claim ("redundant watch installs and generation mints within a transaction are eliminated") has **zero coverage**. Mutation: change the coalescing probe to `.find(|staged| false && staged.same_hub(&publication))`, so every publication installs separately instead of coalescing. Result: **661/661 workspace tests pass.**

So `SnapshotPublication::same_hub` / `coalesce`, and the new public `WatchSender::same_channel` API added to `shelterwood-runtime` (`sync.rs:472-476`), can be deleted or regressed silently. The consistent-cut guarantee survives the mutation (installs still happen under the gate), so the *cut* tests do not protect coalescing — only the generation/wake economy differs, and nothing asserts it.

**Fix:** subscribe to a scope, run a batch admission, assert the hub's generation advanced by **exactly 1** and that `changed()` yields exactly one value for the whole batch.

### F2 — P2 · `shelterwood-cells/src/observe.rs:1119-1134` and `:1136-1166` — two of the four new unit tests pass on `origin/main` unchanged

I ported all four new unit tests verbatim onto an `origin/main` worktree. `initial_snapshot_subscription_does_not_mint_a_generation` and `receiverless_generation_exhaustion_is_deferred_until_commit` went **red** (good). `staged_snapshot_is_installed_during_unwind` and `publication_after_staged_close_is_not_built` went **green** — vacuous with respect to this PR.

On main, `publish`/`close` install synchronously, so the post-unwind value is present anyway, and `publish`'s `snapshot()` closure was already lazy behind `if state.closed { return; }` inside `modify_silently`. Neither test distinguishes "staged, installed at commit" from "installed immediately".

**Fixes (both verified — red on main, green on the PR):**
- *unwind test*: capture the mid-transaction value into a variable declared **outside** `catch_unwind` and assert afterwards — `let mut mid_txn = None; … mid_txn = Some(receiver.borrow_latest().state.clone()); panic!(…); … assert_eq!(mid_txn, Some(ScopeState::Unstarted));`
- *staged-close test*: assert `!receiver.borrow_latest_and_closed().1` between `hub.close(...)` and `hub.publish(...)`, proving the close is still staged so the laziness comes from `snapshot_hub_will_close` rather than an installed `state.closed`.

⚠️ **Worth recording:** my first attempt put that assert *inside* the `catch_unwind` closure and the test still passed on main — `catch_unwind` swallowed it. That is the same "assertion in a swallowed body" anti-pattern as an `assert!` inside an actor handler, just wearing a different hat. Evidence must be published outward and judged in the test body.

(The line itself is not unpinned: mutating `snapshot_hub_will_close` to always return `false` makes `publication_after_staged_close_is_not_built` the single workspace-wide failure. It pins the line — just not the property its name claims.)

### F3 — P2 · `shelterwood-cells/src/observe.rs:709-717` (`SnapshotHub::subscribe`) — self-review fix #1 is incomplete

On the *first-initialization* path (`initialized == true`), `get_or_init` consumes only `initial.clone()` and the `if !initialized && …` block is skipped — so the original `RetainedScopeSnapshot`, which owns `RetainedExit`s carrying user errors, is dropped at the end of `subscribe`, **under the observation gate and outside the transaction's effect list**.

That is exactly the class fix #1 addressed on the refresh path ("this still submitted critical disposal while an effects sink was available"). Reachable whenever a first subscription is taken on a scope whose current cut contains a failed child exit. Not a regression versus main (main dropped the displaced clone there instead), but the standard this PR sets is not met on the path it newly created.

**Fix:** `txn.defer(move || drop(initial));` on the `initialized` branch.

Adjacent, pre-existing, untouched: in the same function the `if state.closed { return; }` early return inside `modify_silently` drops `initial` while holding **both** the gate and the watch's internal write lock — worth routing through `txn` in the same pass.

### F4 — P2 · `shelterwood-cells/src/cells.rs:966-978` — new transient O(M²) memory retention during batch admission

`stage_snapshot` defers `drop(retired)` for every coalesced publication; on main each displaced projection was released immediately after its publish (peak: one). Measured with temporary instrumentation on a 64-child `set_admitted_children`:

```
MEASURE commit: staged_hubs=1 deferred_effects=191 coalesce_retired=63
```

63 full `RetainedScopeSnapshot` trees held live until commit, with `children` vectors of length 1…63 — ~2000 `ChildSnapshot` structs simultaneously, scaling as M²/2 per hub and ×(1+ancestors). The driver's initial admission passes the whole declared child set, so M is user-controlled. Separately, `publish` still evaluates `snapshot()` eagerly, so the M×(1+ancestors) full recursive snapshot *constructions* remain (unchanged from main, but this PR is the natural place to collapse them).

**Fix:** stage a **lazy producer** instead of a materialized value — `SnapshotPublication` holds `Box<dyn FnOnce() -> RetainedScopeSnapshot>` capturing `Arc<ScopeCell>`; coalescing replaces the closure (nothing to retire, no deferred drop); `install` builds the projection once per hub inside `commit` while the gate is still held. That removes both the retention and the redundant construction, delivering #271's "collapses the M×(1+ancestors) redundant installs" in full.

### F5 — P2 · `shelterwood-cells/src/cells.rs:970-973` (`ObservationTxn::commit`) — the install-before-unlock ordering is untested

Mutation: hoist `drop(self.guard.take())` above the install loop → **661/661 green**. The ordering is load-bearing and correctly written: with installs outside the gate, two transactions on one gate could stage T1's cut, release, let T2 stage+install a newer cut, then have T1 install its stale cut — leaving `borrow_latest` permanently behind the tree until the next publish. SPEC §12's new paragraph promises this explicitly ("installed at commit, before the observation gate is released") and nothing enforces it.

### Cleared

- **Lock rule inside `SnapshotPublication::install` (`observe.rs:657-689`).** Every user-bearing value is moved (`Option::take`, `mem::replace`) or pushed to `effects`; `retired` and the exhaustion panic are deferred; the pulse is deferred; only framework counters/flags mutate under the watch lock.
- **Unwind path.** `ObservationTxn::Drop → commit` installs, releases the guard (poison-tolerant by design), then runs effects through `PanicAccumulator`, whose `Drop` uses `resume_preferred_panic` — payloads are *contained* when `std::thread::panicking()`, so no double-panic abort. A hostile waker panicking in one effect cannot strand the others (each effect is individually `catch_panic`'d) and cannot leave the generation inconsistent, because generation and value install together under the gate before any effect runs. The `.expect()`s reachable from `commit` are structurally unreachable (a publication drains exactly once).
- **Close dominance / generation semantics.** `publish` and `close` both short-circuit on `snapshot_hub_will_close`, so a staged close can never be downgraded by a later publish — `coalesce`'s `self.closed = newer.closed` is safe *only* because of that guard (worth a comment, but correct). Publish-then-close keeps exactly one generation edge via `self.published |= newer.published`; mutating that to `=` correctly reddens `snapshot_publication_before_close_is_drained`. Close-only mints nothing and still pulses, so `changed()` resolves `Err(SnapshotClosed)`, matching main. No generation is skipped or reused; a receiver cannot see a torn value/generation pair — both are written in one `modify_silently`.
- **Receiverless init vs refresh.** The unstaged mid-transaction installs in `subscribe` are unobservable: they run only when `receiver_count() == 0`, and 0→1 requires the gate (`SnapshotReceiver: Clone` only goes 1→2). `wait_for_child`'s `borrow_latest_and_closed` reads snapshot and closed flag from one `borrow_cloned`, so the terminal payload cannot come from an earlier cut.
- **Ancestor propagation.** `emit_locked` stages self-then-ancestors and `stage_snapshot` preserves first-staged order, so at commit a descendant hub installs before its ancestor — an ancestor is never momentarily *newer* than its descendant. Staging is O(depth) per publish, O(M·D²) per batch; depth is small, no time-complexity issue. Cross-hub atomicity for a reader holding two receivers was never promised and is unchanged.
- **`SnapshotHub::is_closed()`** now lags a staged close within a transaction, but its only caller is the `Debug` impl. The `debug_assert!` in `subscribe_snapshots` (`cells.rs:1675-1679`) still holds, since `close`'s install completes inside `commit` before the gate releases.
- **SPEC vs code.** §12/B.9 and the `wait_for_child` prose now promise per-hub committed-cut values for both surfaces, and the code delivers. Nothing silently weakened. Precision nit: "the single final projection is installed at commit, before the observation gate is released" is literally false for `subscribe`'s two unstaged installs — true for observers, but the sentence describes installation timing.
- **Doc sync:** `docs/observation.md` and `crates/shelterwood/doctests/docs/observation.md` got identical edits.

### Verdict on the three claimed self-review fixes

| # | Claim | Verdict |
|---|---|---|
| 1 | Receiverless refresh retirement goes through `txn.defer` | **Real, but incomplete** — refresh path fixed (`observe.rs:733-735`); the sibling first-initialization path still drops `initial` under the gate outside the sink (**F3**) |
| 2 | Receiverless generation exhaustion deferred, not raised under the gate | **Real and complete** — the shipped test panics on main at `observe.rs:634` inside `subscribe`, passes on the PR |
| 3 | First initialization no longer refreshes/mints; starts at generation zero | **Real and complete for the mint** — shipped test fails on main (`left: 1, right: 0`). The removed refresh is precisely what leaves `initial` unconsumed, i.e. **F3** |

The new driver test `snapshot_watch_batch_admission_is_one_committed_cut` is **non-vacuous** — ported to `origin/main` it fails at the first assert, and its asserts sit in a plain closure whose panic propagates.

### Commands

```
./tools/dev cargo test -p shelterwood-cells --lib                      → 22 passed (PR, clean)
git worktree add .worktrees/review-288-main origin/main
  4 new unit tests ported verbatim → 2 FAILED / 2 PASSED               (F2)
  driver batch test ported verbatim → FAILED                           (non-vacuous)
  strengthened unwind + staged-close tests → FAILED                    (fix verified)
strengthened tests applied to PR worktree → 13/13 passed, reverted
mutation `published |= newer.published` → `=`            → 1 FAILED (pinned)
mutation `snapshot_hub_will_close` → always false        → 1 FAILED (pinned)
mutation disable coalescing in stage_snapshot            → 661/661 PASS (F1)
mutation hoist `drop(self.guard.take())` above installs  → 661/661 PASS (F5)
instrumented commit + 64-child set_admitted_children     → coalesce_retired=63 (F4)
./tools/dev cargo nextest run --workspace (clean tree)   → 661/661 PASS
```

Heavy runs under `nice -n 10 taskset -c 14-18`; review worktree removed; worktree clean at `8dda240`.

ralexstokes:
## Review resolution — `52eb29b`

All five findings addressed. Nothing was re-litigated; F4 landed in full, inside `cells.rs`/`observe.rs` only.

| # | Finding | Resolution |
|---|---|---|
| **F1** | per-hub coalescing entirely unpinned | Pinned twice. `observe::tests::repeated_publication_builds_and_mints_once_per_transaction` (three publications on one hub → **1** producer built, generation **1**). `driver::tests::observation::snapshot_watch_batch_admission_mints_one_generation_per_hub` (batch admission on a nested scope → **exactly one** generation edge on *both* the admitting hub and its ancestor, one `changed()` delivery carrying the whole batch, nothing pending after). Two hubs on purpose: `same_channel` returning constant-true is as wrong as constant-false, and only a multi-hub batch separates them. Generation counts needed a reader — delivery conflates edges, so `changed()` alone cannot count them — hence `SnapshotReceiver::current_generation`, `cfg(test, test-util)`. |
| **F2** | two new unit tests pass unchanged on `origin/main` | Strengthened exactly as prescribed. Unwind test captures the mid-transaction reading into a `mid_txn` declared **outside** `catch_unwind` and asserts after it returns (an assert inside that closure is swallowed by the unwind it rides — same anti-pattern as an assert in an actor handler). Staged-close test asserts `!borrow_latest_and_closed().1` **between** `close` and `publish`, so the laziness is attributed to `snapshot_hub_will_close` and not to an installed `state.closed`. Both verified red on an `origin/main` worktree — see evidence below. |
| **F3** | self-review fix #1 incomplete | Both sibling paths now retire through the transaction. `subscribe` holds the caller's projection in an `uninstalled` slot that only the install consumes, so the first-initialization path (which consumed just a clone) and the pre-existing `if state.closed { return; }` early return (which dropped it under the gate **and** the watch's write lock) both hand it to `txn.defer`. |
| **F4** | transient O(M²) retention | Landed. `SnapshotPublication` now stages a `SnapshotProjection` — a producer capturing `Arc<ScopeCell>` — instead of a materialized cut. Coalescing replaces the producer, so a superseded cut is never built and nothing is retired but a box. `install` runs the survivor once per hub inside `commit`, gate still held, before the watch's own lock is taken. Measured on a 64-child `set_admitted_children`: **64 recursive projections built → 1**; retained-until-commit **63 → 0**. |
| **F5** | install-before-unlock ordering untested | Pinned, not merely argued. `cells::tests::a_staged_cut_is_installed_before_its_gate_is_released` stages a producer that records `gate.is_held()` while it runs; because the producer now runs *inside* `install`, that reading is a direct question about the ordering. Also commented at the site with the interleaving it prevents. |
| — | SPEC precision nit | §12 now reads "…installed at commit, before the observation gate is released; the only install that may precede commit is a new subscription seeding its own starting value, which no receiver is yet attached to observe." |

### Lock rule, for the new lazy path

The producer is `FnMut` rather than `FnOnce` on purpose: running an `FnOnce` also destroys its captures, and the capture here is an `Arc<ScopeCell>` whose last drop would destroy a scope — a user-bearing value — under the gate. `install` therefore runs the producer, hands the spent producer to `effects` **first**, and only then takes the watch lock; every later drop in that function (the displaced cut, the sender clone) reaches its scope through the producer the effect list now owns. A superseded producer from `coalesce` goes to `txn.defer` for the same reason. The build itself runs before `modify_silently`, so walking the resident tree never nests its locks inside the watch's.

### Mutation evidence (all runs `nice -n 10 taskset -c 14-17`)

```
clean tree                                                → 664/664 PASS   (661 before, +3 new tests)
./tools/dev just ci                                       → EXIT 0

mutation: stage_snapshot .find(|staged| false && …)       → 2 FAILED, both new:
    shelterwood driver::…::snapshot_watch_batch_admission_mints_one_generation_per_hub
    shelterwood-cells observe::…::repeated_publication_builds_and_mints_once_per_transaction

mutation: hoist drop(self.guard.take()) above installs    → 1 FAILED, the new one:
    shelterwood-cells cells::…::a_staged_cut_is_installed_before_its_gate_is_released
  (this mutation was 661/661 green before — F5)

git worktree add .worktrees/review-288-main origin/main
  strengthened unwind test, ported verbatim              → FAILED
      left: Some(Running), right: Some(Unstarted)
  strengthened staged-close test, ported verbatim        → FAILED
      "the close is still staged, so the publication below is declined
       by the transaction rather than by an installed terminal state"
  (both were green on main before strengthening — F2)

instrumented snapshot_locked + 64-child set_admitted_children
  origin/main                                            → builds=64
  this commit                                            → builds=1
```

Instrumentation and the review worktree were removed; `just ci` and the 664-test run are on the committed tree.

### Deliberately not done

F3's defer has no test. The observable difference between dropping a `RetainedScopeSnapshot` under the gate and deferring it is a `dispose_critical` submission's thread start, not a user-visible drop — pinning it would need a probe deeper than the finding warrants. It is argued, in the code comment, not pinned.

PR #289, opened 2026-08-16
Make drain-entry straggler sampling atomic

### LINKED ISSUE #270: Adjudicate: zero-budget shutdown straggler verdict is schedule-dependent on multi-thread runtimes
Adjudication question from the 2026-08-16 review round (internal sweep).

## The nondeterminism

`begin_drain_transition` (`driver/shutdown.rs:101-136`) publishes `Draining` via `set_state` — its own gate txn, wakes committed immediately — **before** `flush_supervisor_effects` runs `StopChild → begin_terminal_disposal → set_terminal_disposal_pending(true)` (`driver/child.rs:1072`). On a multi-thread runtime, a zero-budget `shutdown_scope` waiter on another worker can wake on the `Draining` pulse, take the `Elapsed` arm, and run `collect_stragglers` (`shutdown.rs:49-56, 70-71`) while the same batch is mid-flight — observing a restart-window child as neither `Terminal` nor `terminal_disposal_pending` (`shutdown.rs:6-8`) and returning `Err(ShutdownTimeout)` where the single-thread schedule returns `Ok`.

Everything still tears down correctly; the verdict is spec'd as an instantaneous sample ("an escalation budget, not an observation opportunity", `shutdown.rs:59-61`). But the Ok/Err nondeterminism concerns a child the driver was already disposing **in the same logical step**, and the pinning test (`tests/disposal.rs:1026`, `restart_window_cleanup_is_not_a_shutdown_straggler`) runs only on the current-thread flavor — so it is also a latent flake if that test ever moves to multi-thread.

## The question

Is "instantaneous sample at budget expiry" intended to include sampling *mid-batch* of the very transition that started the drain? Options:

1. Accept and document: the sample may split a drain-entry batch; the current-thread pin is flavor-specific by design.
2. Close the window: set `terminal_disposal_pending` (or an equivalent marker) before, or atomically with, the `Draining` publication for the entering batch.

## Acceptance criteria

- [ ] Adjudicate 1 vs 2; record the ruling (SPEC or code comment).
- [ ] If 2: multi-thread variant of the straggler-skip test.
- [ ] If 1: a comment at the pinning test explaining why it must stay current-thread.

ralexstokes:
## Ruling

Choose option 2: a zero-budget shutdown may sample a committed drain-entry cut, but it may not split the logical driver step that publishes `Draining` from the inactive-child cleanup selected by that same step.

## Summary

- publish `Draining` and selected terminal-disposal markers under one observation-gate transaction
- pre-settle the pure supervisor reducer so ordered scopes expose the exact first stop selected at drain entry
- cover ordinary drain and initial force entry
- retain the pending marker until terminal publication commits, removing the corresponding trailing gap
- pin the ruling with a deterministic multi-thread regression and SPEC text

## Why

The public `Draining` edge wakes a zero-budget waiter. Letting that waiter observe the entry batch halfway through made the same restart-window child alternate between success and `ShutdownTimeout` based only on runtime scheduling.

## Impact

Restart-window cleanup already selected at drain entry is not reported as a straggler. Active children and ordered siblings whose stop has not yet been selected remain reportable, preserving the zero-budget budget semantics. The guarantee covers the entry step only — later ordered stop steps race the sample by design, which SPEC §12 now states explicitly.

## Validation

Pinned by tests:

- focused deterministic multi-thread straggler regression; deleting the marker store, or the `begin_drain_transition` pre-`Settle`, makes it fail
- full disposal target: 36 passed
- driver shutdown tests: 6 passed
- `./tools/dev just ci` — exit 0, 656 tests plus Clippy, doctests, rustdoc/API reachability, docs sync, and Nix formatting

Argued, not pinned (no test in the suite provokes either window; both are defended by a comment at the site):

- the `collect_stragglers` marker-before-record read order (`driver/shutdown.rs`) — reverting to record-first leaves the suite green
- retaining the marker until after `terminalize_child` commits (`driver/child.rs`) — clearing early leaves the suite green
- the statement order inside `publish_drain` (markers before the `Draining` record write) — moving the loop after `set_state_locked` leaves the suite green, yet the swap reopens #270

Closes #270

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -2771,6 +2771,14 @@ dynamic removal: the removed id becomes reusable (and a repeated `remove`
 reports `AlreadyAbsent`) only at the commit that withdraws the member
 from residency and publishes its `Removed` edge — §2's
 resident-membership uniqueness holds at every observable cut.
+Drain entry follows the same rule: publishing `Draining` includes the
+terminal-disposal intent for every inactive child selected to stop in that
+same driver step. A zero-budget shutdown's straggler sample cannot split that
+entry step, so restart-window cleanup already committed by it is not a
+straggler; an active child, or an ordered sibling whose stop has not yet been
+selected, remains reportable. This ruling is pinned on a multi-thread runtime
+by
+`integration::restart_window_cleanup_is_isolated_without_reclassifying_the_recorded_exit`.
 
 `tracing` spans emit from one choke point (the optional `metrics` surface
 is Part II §20). Everything else observational — peer monitoring, actor
--- a/SPEC.md
+++ b/SPEC.md
@@ -2775,9 +2775,12 @@ Drain entry follows the same rule: publishing `Draining` includes the
 terminal-disposal intent for every inactive child selected to stop in that
 same driver step. A zero-budget shutdown's straggler sample cannot split that
 entry step, so restart-window cleanup already committed by it is not a
-straggler; an active child, or an ordered sibling whose stop has not yet been
-selected, remains reportable. This ruling is pinned on a multi-thread runtime
-by
+straggler; an active child, or an ordered sibling whose stop has not yet
+been selected, remains reportable. The guarantee covers that entry step
+only: an ordered scope stops its children one step at a time, and every
+step after entry races the sample by design, so whether a later sibling
+has already terminalized when the sample runs is schedule-dependent and
+both outcomes conform. This ruling is pinned on a multi-thread runtime by
 `integration::restart_window_cleanup_is_isolated_without_reclassifying_the_recorded_exit`.
 
 `tracing` spans emit from one choke point (the optional `metrics` surface
--- a/crates/shelterwood-cells/src/cells.rs
+++ b/crates/shelterwood-cells/src/cells.rs
@@ -509,10 +509,41 @@ impl MemberCell {
         };
     }
 
+    /// Reads the drain-entry terminal-disposal marker.
+    ///
+    /// `Acquire` pairs with the `Release` store in
+    /// [`Self::set_terminal_disposal_pending`]. Two separate edges make a
+    /// zero-budget straggler sample (the driver's `collect_stragglers`)
+    /// correct, and both rest on these atomics alone: the sample is *not*
+    /// incidentally serialized by the observation gate, because
+    /// [`ScopeCell::resident_projections`] takes `observation.current_children`
+    /// — a different mutex from the gate `publish_drain` holds.
+    ///
+    /// * A load that observes the *clear* synchronizes with the store the
+    ///   driver issues after `terminalize_child` published
+    ///   `MemberStage::Terminal`, so the sampler's following record read is
+    ///   ordered after that publication. It cannot pair a stale nonterminal
+    ///   projection with an already-cleared marker.
+    /// * A load cannot return a stale `false` from *before* the marker was
+    ///   set. A sampler reaches this call only after reading
+    ///   `ScopeState::Draining` out of the scope record, and
+    ///   [`Self::set_terminal_disposal_pending`] stores every drain-entry
+    ///   marker *before* `publish_drain` writes that record. The record
+    ///   write/read pair supplies happens-before, and coherence then forbids
+    ///   this load from returning a value preceding `true` in the marker's
+    ///   modification order.
     pub fn terminal_disposal_pending(&self) -> bool {
         self.terminal_disposal_pending.load(Ordering::Acquire)
     }
 
+    /// Installs or clears the drain-entry terminal-disposal marker.
+    ///
+    /// `Release` publishes everything the caller sequenced before it. The
+    /// clearing store at terminal publication therefore carries
+    /// `MemberStage::Terminal` to any sampler whose `Acquire` load reads it;
+    /// the setting store is made visible by the `Draining` record write that
+    /// `publish_drain` performs afterwards. See
+    /// [`Self::terminal_disposal_pending`] for the full argument.
     pub fn set_terminal_disposal_pending(&self, pending: bool) {
         self.terminal_disposal_pending
             .store(pending, Ordering::Release);
@@ -1450,6 +1481,18 @@ impl ScopeCell {
     ) {
         debug_assert!(matches!(state, ScopeState::Draining));
         self.with_observation_gate(|txn| {
+            // Statement order is load-bearing: every marker must be stored
+            // *before* `set_state_locked` writes `Draining` into the scope
+            // record. A zero-budget sampler reads that record first, so the
+            // `Draining` read is the release edge that makes the markers
+            // visible to it. Publishing the state first reopens #270: the
+            // driver writes `Draining` with no marker yet stored, a sampler on
+            // another worker reads `Draining`, then reads `false` for a child
+            // this very step is disposing, then reads that child's still
+            // nonterminal record, and reports a spurious
+            // `Err(ShutdownTimeout)`. No test holds this order — moving the
+            // loop below `set_state_locked` leaves the suite green — so this
+            // comment is the only guard.
             for member in terminal_disposals {
                 debug_assert!(
                     self.current_observation_gate()
--- a/crates/shelterwood-cells/src/cells.rs
+++ b/crates/shelterwood-cells/src/cells.rs
@@ -1435,6 +1435,36 @@ impl ScopeCell {
         });
     }
 
+    /// Publishes drain entry together with terminal-cleanup intent selected by
+    /// the same driver step.
+    ///
+    /// A zero-budget shutdown waiter wakes from the state publication and
+    /// immediately samples descendants. Installing these markers under the
+    /// shared observation gate keeps that sample from splitting drain entry
+    /// between the scope transition and an inactive child's cleanup.
+    pub fn publish_drain(
+        &self,
+        state: ScopeState,
+        startup: Option<Result<(), StartupError>>,
+        terminal_disposals: &[Arc<MemberCell>],
+    ) {
+        debug_assert!(matches!(state, ScopeState::Draining));
+        self.with_observation_gate(|txn| {
+            for member in terminal_disposals {
+                debug_assert!(
+                    self.current_observation_gate()
+                        .same_gate(&member.current_observation_gate()),
+                    "a drain entry may mark only a resident member on its observation gate"
+                );
+                member.set_terminal_disposal_pending(true);
+            }
+            if let Some(startup) = startup {
+                self.set_startup_locked(startup, txn);
+            }
+            self.set_state_locked(state, txn);
+        });
+    }
+
     fn set_state_locked(&self, state: ScopeState, txn: &mut ObservationTxn<'_>) {
         let mut transient_retained = Vec::new();
         RetainedExit::retain_scope_state(&mut transient_retained, &state);
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -8,7 +8,7 @@ mod shutdown;
 mod startup;
 
 use std::{
-    collections::BTreeMap,
+    collections::{BTreeMap, BTreeSet},
     ops::{Index, IndexMut},
     sync::{Arc, OnceLock},
     time::Instant,
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -33,8 +33,8 @@ use crate::{
     ScopeState, ShutdownStraggler, ShutdownTimeout, StartupFailure, StartupFailureCause,
     admission::{NotAdmittingCause, ReserveError},
     cells::{
-        MemberStage, MemberTransition, ResidentProjection, RetainedExit, RetainedStopReason,
-        ScopeCell, ScopeControlEvent, StartupDisposition,
+        MemberCell, MemberStage, MemberTransition, ResidentProjection, RetainedExit,
+        RetainedStopReason, ScopeCell, ScopeControlEvent, StartupDisposition,
     },
     deadline::Deadline,
     engine::{
@@ -71,7 +71,7 @@ pub(crate) use admission_control::{
 };
 
 #[cfg(test)]
-use crate::cells::{GateCapture, MemberCell, RuntimeStorage};
+use crate::cells::{GateCapture, RuntimeStorage};
 #[cfg(test)]
 use admission_control::RemovalResponses;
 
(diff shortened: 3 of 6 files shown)


ralexstokes:
## Adversarial review — head `b74df063cd392977844b9cc5dccaeaa1ee1d2933`

I reviewed the issue #270 ruling, the relevant SPEC shutdown contract, and the complete PR diff. I attacked:

- option 2’s boundary against the existing zero-budget “immediate escalation” rule
- the pure-reducer pre-`Settle` timing and the exact `supervisor_effects[before..]` slice
- effect order for dynamic scopes, reverse-ordered scopes, ordinary drain entry, and first-entry `Force`
- active versus inactive membership classification, disposing/terminal exclusions, duplicate stop/force effects, and later ordered siblings
- pending-marker publication, lifetime, and the terminal-publication trailing edge
- observation-gate/waker behavior and `Arc<MemberCell>` clone/drop safety under the project lock rule
- the hostile blocking-waker regression’s ability to pin the driver after the gate unlock and let the second worker take the zero-budget sample
- normal nonzero-budget shutdown behavior and zero-budget determinism

### Confirmed findings fixed directly

1. **The trailing-gap fix still admitted a stale-record pairing.** The sampler read the member record before acquiring the pending marker. It could therefore read a stale nonterminal record, race terminal publication plus the later release-store clearing the marker, then observe `false` and report a straggler. Commit `b74df06` reads the acquire marker first; seeing the cleared value now orders the following record read after terminal publication, while seeing the set value excludes the still-pending child directly.

2. **Initial force entry regressed to quadratic scanning.** A fresh `Force` emits both `StopChild` and `ForceChild` for each dynamic child. Deduplicating with `Vec::contains` made marker selection O(n²) in child count. Commit `b74df06` uses a `BTreeSet<ChildKey>`, preserving framework-only key handling and bounding the scan to O(n log n).

### Verdict after fixes

No remaining correctness, lock-safety, or scope-flavor findings. The option-2 ruling is internally consistent: dynamic drain marks every selected inactive child; ordered drain pre-settles only the first reverse-order stop; fresh force marks every inactive forced child; active children and not-yet-selected ordered siblings remain reportable. The reducer pre-settle is pure and adds no wake/user-code edge. `publish_drain` mutates only framework-owned state while holding the shared observation gate, and the hostile waker runs from the transaction flush after unlock.

### Checks

- focused deterministic multi-thread regression: passed
- full `disposal` integration target: 36 passed
- `driver::tests::shutdown`: 6 passed
- `./tools/dev just ci`: passed — formatting, Clippy, 656 tests, doctests, rustdoc/API reachability, docs sync, and Nix formatting
- `git diff --check origin/main...HEAD`: clean

The review fixes were committed and pushed to this PR branch as `b74df06`.

ralexstokes:
## Independent review — memory ordering is **sound**; no P0/P1; five P2 hardening findings

Reviewed head `b74df06`. Second-opinion pass, independent of the authoring agent's self-review. The fix is correct as written — everything below is hardening, unpinned-claim, or test-robustness.

### Memory-ordering verdict: **sound, but under-documented**

- **Load:** `shelterwood-cells/src/cells.rs:513` — `terminal_disposal_pending.load(Acquire)`, called at `driver/shutdown.rs:12`, before the record read at `:13` (a later load cannot be hoisted above an Acquire load).
- **Matching store:** `cells.rs:518` — `store(pending, Release)`, called at `driver/child.rs:1141`, sequenced *after* `terminalize_child(...)` at `child.rs:1137`, which publishes `MemberStage::Terminal` into the member watch record (`child.rs:555-559`). An Acquire load reading that store synchronizes-with it, so the terminal write happens-before the subsequent `record()` read. **The comment's claim holds exactly as stated.**
- **The undocumented half.** Why the load cannot return a stale `false` from *before* the marker was set is nowhere written down, and it is what the whole fix rests on: `shutdown_scope` samples only after reading `ScopeState::Draining` (`shutdown.rs:58`), and `publish_drain` stores the markers (`cells.rs:1459`) *before* `set_state_locked` writes that record (`cells.rs:1476`, under the watch write lock). That write/read pair gives HB, and coherence then forbids the later Acquire load from returning a value preceding `true` in the modification order. Correct — but invisible to a maintainer (see P2-1).
- I also confirmed the sample is **not** incidentally serialized by the gate: `resident_projections()` takes `observation.current_children` (`cells.rs:1221-1226`), a *different* mutex, so the argument must stand on the atomics alone. It does.

### P2-1 — the load-bearing statement order inside `publish_drain` is undocumented and unpinned

`shelterwood-cells/src/cells.rs:1452-1465`. Markers must be stored *before* `set_state_locked` publishes `Draining`. Nothing says so and no test holds it: moving the loop to run *after* `set_state_locked` (still inside the same gate txn) leaves **493/493 green**.

Failure interleaving with that swap: driver holds the gate, writes `state = Draining` into the watch, has not yet stored the marker; sampler on another worker reads `Draining` (`shutdown.rs:58`), reads marker `false`, reads a non-terminal record → `Err(ShutdownTimeout)` for the very child the entry step is disposing. #270 reopens silently.

Fix: one comment line naming the constraint ("markers before the state write; the sampler's `Draining` read is the release edge that makes them visible"). A test is genuinely hard here; the comment is the cheap defence.

### P2-2 — `force_all`'s added pre-`Settle` is dead weight, and the "covers initial force entry" claim rests on something else

`driver/shutdown.rs:192`. `Event::Force` already pushes `ForceChild` for *every* non-joined child (`shelterwood-core/src/supervisor.rs:701-708`) — a strict superset of the single `StopChild` that `settle_ordered_stop` (`supervisor.rs:533-556`) can add — so the pre-`Settle` never contributes a member to `terminal_disposals`.

Evidence: deleting that line leaves **493/493 green**, whereas deleting the same line in `begin_drain_transition` (`shutdown.rs:169`) makes the new regression test **fail**. So the ordered-drain half is pinned; the force half is not. Not harmful (it only advances `ordered_stop_cursor`/`ordered_stop_waiting` one step earlier, and `settle_supervisor` re-converges), but unexplained cost. Either drop it and note that `Force` already selects every child, or keep it and mark it defensive.

### P2-3 — both self-review fixes are correct but entirely unpinned

Reverting the `collect_stragglers` read order (`shutdown.rs:12-13` → record first) and re-adding an early clear before `terminalize_child` (`child.rs:1141`) each leave **493/493 green**. Both close genuinely narrow windows, so I wouldn't demand tests — but the PR body presents them as validated. Suggest "argued, not pinned" in the body or the code comment.

### P2-4 — the regression's driver park point isn't asserted, so a future stray pulse degrades it to a vacuous pass

`crates/shelterwood/tests/disposal.rs:1058-1075`. `BlockingWake` is one-shot (`blocker.take()`), so it parks the driver at *whichever snapshot pulse comes first*. If any publication lands between the `snapshots.changed()` registration and the drain, the driver parks before `Draining`, the second `poll_once` returns `Pending` because the shutdown future is still waiting to *observe* `Draining` — not because it sampled — and the test passes having tested nothing. The assert message ("the zero-budget sample escalates before the blocked driver continues") cannot distinguish the two.

Today the child sits in a 60 s fixed backoff so the scope is quiet and the interleaving *is* reached (proved by mutation A below) — but the guard is circumstantial. Fix: after `wait_for_destructor`, assert the parked cut explicitly (`assert!(matches!(scope.state(), ScopeState::Draining))`) before the second `poll_once`.

### P2-5 — residual schedule-dependence outside the entry step is real; the new SPEC text doesn't say so

`SPEC.md:2774-2781` reads "an ordered sibling whose stop has not yet been selected remains reportable" — but "not yet selected" is itself schedule-relative. For an ordered scope with two inactive restart-window children only the last is marked at entry, so whether the earlier one is `Terminal` when the sampler runs depends on how fast the sampler wakes: `Ok` and `Err` are both reachable for the same program.

That is the pre-existing "instantaneous sample" semantics and is consistent with the ruling, but a reader of the new paragraph could reasonably infer the whole drain is atomic. Suggest one clause stating the guarantee covers *only* the entry step and that later ordered steps race the sample by design.

### Verdict on the two claimed self-review fixes

Both **confirmed present and correct**, both **unpinned** (P2-3):

- *Stale-record pairing* — `shutdown.rs:12-13` now reads the marker first. Real window closed: under the old order a sampler could read a non-terminal record, then read the marker after the whole set→clear cycle, and report a straggler that had already terminalized.
- *Quadratic force-entry scan* — `Vec::contains` → `BTreeSet` (`shutdown.rs:91/108`), and it is **behaviour-preserving**: the only order change is effect-order → ascending `ChildKey`, the sole consumer sets one independent `AtomicBool` per member, all inside a single gate transaction, so no observer can see a partial or ordered application. The `filter_map` re-lookup is a no-op (`children.get(key)` was already proved `Some` in the loop).

### Cleared

- **Marker lifetime.** Every marked key is inactive, non-terminal, non-disposing and present in `self.children` — exactly the predicate under which `begin_stop_child` (`child.rs:760-774`) takes the inactive branch into `begin_terminal_disposal`. Nothing mutates those fields between `publish_drain` and the synchronous `flush_supervisor_effects`. I injected an invariant probe (after flush+settle, every still-marked drain-entry member must be `is_disposing` / `membership_terminal` / `pending_terminal.is_some()`, or gone) across the whole suite: **no leak, 493/493**. The only new leak path is an unwind inside the publication txn (hostile waker panic), which kills the driver anyway.
- **Lock rule.** `publish_drain` holds the gate over per-member `AtomicBool` stores (framework data), `set_startup_locked`/`set_state_locked` (unchanged, `RetainedExit`-protected), and deferred pulses. Members arrive as a borrowed `&[Arc<MemberCell>]`, so **no `Arc` is dropped under the gate**; the owning `Vec` is dropped by the caller after return, and the driver + residency each hold another strong ref. No user callback, no user-value drop, no formatting. The one new locked read is the `debug_assert!`'s `member.current_observation_gate()` (member `RwLock` under the scope gate) — same direction as the existing `with_observation_txn` re-read, so no new lock order. The pre-`Settle` `reduce` runs entirely outside any lock and mutates only `SupervisorState`. **Reducer-purity claim holds.**
- **Pre-`Settle` semantics.** `settle_startup` no-ops (lifecycle is already `Draining`, so `is_starting()` is false — `engine.rs:825`, `889-908`); `settle_ordered_stop` produces exactly the first stop main produced one step later; `settle_finish` can only emit `Finished` earlier in the vec but strictly after the entry `StopChild`s in execution order; `settle_supervisor` still converges. No `StartupFailed`/`DrainStarted` can reach the `unreachable!()` arms from `Settle`.
- **Test quality.** No `assert!` inside an actor handler (the handler is `async { Err(...) }`; all assertions are in the test body). Deterministic — 20/20 at ~3 ms, no sleeps, no `assert_eventually` in the critical section; `wait_for_destructor` is a condvar handshake.

### Mutations run (each reverted)

| | Mutation | Result |
|---|---|---|
| A | drop marker store in `publish_drain` (`cells.rs:1459`) | target test **FAILED** (`ShutdownTimeout` straggler) — regression is non-vacuous |
| B | revert `collect_stragglers` read order | 44/44 passed — unpinned |
| C | drop pre-`Settle` in `begin_drain_transition` | target test **FAILED** — pinned |
| D | drop pre-`Settle` in `force_all` | 493/493 passed — unpinned / redundant (P2-2) |
| E | early-clear marker in `handle_construction_disposed` | 493/493 passed — unpinned |
| F | move markers after `set_state_locked` | 493/493 passed — load-bearing order unpinned (P2-1) |
| G | injected leaked-marker invariant probe | 493/493 passed — no marker leak |

Also: `cargo nextest run -p shelterwood --no-fail-fast` → 493/493; clippy on `shelterwood` + `shelterwood-cells` all-targets/all-features → clean; `just packaged-docs-check` → exit 0. Heavy runs under `nice -n 10 taskset -c 19-23`. Worktree clean at `b74df06`.

ralexstokes:
## Review resolution — all five P2 findings addressed (`0cb3fa2`)

| Finding | Resolution |
|---|---|
| **P2-0** memory-ordering argument under-documented | Documented at the atomics themselves: `MemberCell::terminal_disposal_pending` / `set_terminal_disposal_pending` (`shelterwood-cells/src/cells.rs`) now carry both halves — the clear-side acquire that orders the record read after terminal publication, and why a stale `false` is impossible (the sampler reads `Draining` first, `publish_drain` stores the markers before that record write; happens-before plus coherence). The doc also records that `resident_projections` takes `observation.current_children`, a *different* mutex from the gate, so the argument rests on the atomics alone. |
| **P2-1** load-bearing statement order in `publish_drain` unpinned | Comment added inside the gate transaction naming the constraint and spelling out the failure interleaving (driver writes `Draining` with no marker stored → sampler reads `Draining`, `false`, nonterminal record → spurious `Err(ShutdownTimeout)`), and stating plainly that no test holds the order. Re-ran **mutation F** on the committed tree first: moving the loop below `set_state_locked` → **493/493 green**, confirmed. |
| **P2-2** `force_all` pre-`Settle` is dead weight | **Dropped.** Re-ran **mutation D** before editing: deleting it → **493/493 green**, confirmed. The site now carries a comment recording why — `SupervisorEvent::Force` already pushes `ForceChild` for every non-joined child, a strict superset of the single `StopChild` an ordered `Settle` could add. Suite green after the deletion (656/656 in full CI). |
| **P2-3** both self-review fixes unpinned but presented as validated | Marked "argued, not pinned" in the comment at each site (`driver/shutdown.rs` read order, `driver/child.rs` marker retention), and the PR body's Validation section now separates *pinned by tests* from *argued, not pinned* — listing all three unpinned claims including P2-1's statement order. |
| **P2-4** regression can degrade to a vacuous pass | After `wait_for_destructor`, the test now asserts the parked cut explicitly (`matches!(scope.snapshot().state, ScopeState::Draining)`) with a comment explaining why, before the second `poll_once`. Verified non-masking: with **mutation A** (marker store dropped) the test still fails at the *shutdown* expectation with a `ShutdownTimeout` straggler, not at the new assert — so the intended interleaving is still the thing being exercised. |
| **P2-5** SPEC could read as "the whole drain is atomic" | `SPEC.md` §12 now adds: the guarantee covers the entry step only; an ordered scope stops one step at a time and every step after entry races the sample by design, so a later sibling's terminal status at sample time is schedule-dependent and both outcomes conform. Wrapped to the surrounding width. |

### Mutations re-run on the committed tree

| | Mutation | Result |
|---|---|---|
| F | move markers after `set_state_locked` | 493/493 green — confirmed unpinned (now comment-guarded) |
| D | drop pre-`Settle` in `force_all` | 493/493 green — confirmed redundant, line deleted |
| A | drop marker store in `publish_drain` (post-edit re-check) | target test **FAILED** with `ShutdownTimeout` at the shutdown assertion — the new parked-cut assert does not preempt it |

`./tools/dev just ci` → **exit 0**, 656 tests. Nothing else changed; no behaviour change beyond removing the redundant `force_all` pre-`Settle`.

Issue #284, opened 2026-08-16
Tracking: 2026-08-16 full-review round (bugs, waker containment, hardening, simplification, tests)


ralexstokes:
Tracking issue for the 2026-08-16 full-review round: a ten-agent internal sweep (per-crate, cross-crate, test suite) plus two external reviews with every claim adjudicated case-by-case (confirmed / partially confirmed / refuted, with ownership traces; the two headline bugs were reproduced empirically against the facade). Headline result: **no reachable memory-safety, deadlock, or lock-rule bug in ordinary operation** — the confirmed defects cluster around hostile wakers, panic timing, and teardown fidelity.

## Issues

**Confirmed bugs (repros attached)**
- #263 — pending offload panic not resumed on the stop path (`on_stop` runs, decorators see `Ok`)
- #264 — driver-future destruction publishes `Aborted` over the classified exit in `pending_terminal`

**Hostile-waker containment cluster** (CLAUDE.md's own threat model, applied to the wake infrastructure)
- #265 — tokio bulk notification strands co-waiters after one waker panic
- #266 — teardown loops stop at the first resumed waker panic (orphan tasks, stranded epoch retirement)
- #267 — user payloads cross panic-resuming effect flushes unprotected (spawn window, pre-acceptance call)

**Correctness residue**
- #268 — panicking interval `Clone` desynchronizes the timer index
- #269 — losing-terminalizer disagreement between `ScopeCell` and `MemberCell` (+ under-gate drop hardening)

**Adjudications** (decisions, then small diffs)
- #270 — zero-budget shutdown straggler verdict is schedule-dependent on multi-thread runtimes
- #271 — SPEC §12's consistent cut vs the ungated watch surface (`wait_for_child`/`borrow_latest`)

**Hardening**
- #272 — lock-rule hardening: the remaining convention-held sites, made structural
- #273 — public-surface holes: `from_latch`, installable seams, `Exit::new`, `spawn_blocking`

**Simplification**
- #274 — supervisor effect vector: half result channel, half command stream (highest-value structural cleanup)
- #275 — replace `RetainedLifecycleEventKind` with wrap-plus-guards
- #276 — shrink `DynamicRoute` to its one cross-crate method
- #277 — dead code and dead configurability sweep
- #278 — invalid-states-unconstructible backlog: multi-field encodings
- #279 — mechanical dedup and file splits (`raw.rs`, `cells.rs`)

**Tests**
- #280 — core `compile_fail` fences pass vacuously
[...]

Issue #273, opened 2026-08-16
Public-surface holes: from_latch, installable seams, Exit::new, spawn_blocking


ralexstokes:
Decision batch from the 2026-08-16 review round: places where the supported boundary is documented-but-not-structural, verified case by case.

## Findings

- [ ] **`CancellationToken::from_latch` threads every check** (`shelterwood-cells/src/cancellation.rs:13-19`): `#[doc(hidden)] pub`, takes a `shelterwood_runtime::sync::Latch` (a forbidden root), sits on a facade-public type. `doc(hidden)` items are absent from rustdoc JSON, so the api-reachability lane cannot see the signature, and the E0365 shim protects types, not methods. `SnapshotReceiver::borrow_latest_and_closed` (`observe.rs:547`) is the same hidden-method-on-public-type pattern with a benign signature. Decide: seal, `pub(crate)`-gate via a helper crate-side, or teach the checker about hidden methods on exported types.
- [ ] **The seam traits are installable, not merely implementable, by direct dependents.** `DynamicRoute`, `MailboxControl`, `MailboxTermination`, `ActorIdentity`, `MailboxRuntime` are pub and unsealed, and the installation paths are pub on pub types: `ScopeCell::set_dynamic_route` (`cells.rs:2275`), `MemberCell::attach_mailbox` (`cells.rs:628`), `actor_ref_from_parts` (`futures.rs:75`). Once installed, framework code runs the foreign impl **under the gate / member mailbox mutex** (`cells.rs:651`, `:720`, `:1442-1444`). The facade contains all of it (curated exports), and both inner crates document themselves as unsupported seams — but that is convention. Also: `actor_ref_from_parts` lets a direct dependent pair one identity with a different mailbox — handles that compare/hash equal while routing differently, and `SendError.actor_id` inconsistent with `id()`. Decide: seal (private supertrait), capability-gate construction, or accept and record the documented-convention stance.
[...]

PR #290, opened 2026-08-16
Define the supported implementation-seam boundary

ralexstokes:
## Ruling

- structurally crate-gate `CancellationToken::from_latch`
- structurally private-supertrait seal `MailboxControl` and `MailboxTermination`, whose legitimate implementations live in their defining mailbox crate
- explicitly accept `DynamicRoute`, `ActorIdentity`, and `MailboxRuntime` plus their installers as unsupported sibling-crate seams
- enforce the supported `shelterwood` façade boundary, which exports neither those traits nor their installation paths

Private-supertrait sealing cannot work for the remaining three traits because their legitimate implementations live in downstream sibling crates. A public installation capability would be equally obtainable by a direct dependent. Direct lower-crate implementation therefore remains unsupported and voids the lock-rule and identity-pairing contracts.

## Summary

- route legitimate façade token construction through the hidden parent-token capability
- seal both implementation traits whose production implementations are local to `shelterwood-mailbox`
- document the remaining cross-crate boundary in SPEC, project lock-rule guidance, lower-crate docs, trait docs, and constructor docs
- hide `actor_ref_from_parts` from generated documentation and state its identity-pairing invariant
- add a standalone default-feature external consumer
- make CI prove `from_latch` is private, all five seam traits and their installer-bearing types/functions remain absent from the façade, and foreign implementations of the two sealed mailbox traits fail
- mirror the consumer and formatting checks in the clean Nix enforcement lane

## Validation

- cancellation tests: 3 passed
- workspace all-target/all-feature check
- external-consumer positive and diagnostic-pinned negative probes
- default consumer feature tree contains neither `test-util` nor Tokio `rt-multi-thread`
- `./tools/dev just ci` — 656 tests and all local lanes passed
- `./tools/dev just ci-nix` — authoritative clean checks passed

## Scope

This is only the seal-vs-document ruling scheduled in group 2 of #284. The later #273 mechanical items — `Exit::new`, `spawn_blocking`, Tokio shutdown compatibility, and the remaining small leaks — intentionally remain for sequence step 4.

Refs #273


ralexstokes:
## Adversarial review — head `8ee830269696b737d1c1076ff960959acd9fd4a7`

I reviewed this as the sequence-step-2 seal-vs-document ruling from #284/#273, not as the later mechanical #273 batch.

### Surfaces attacked

- full PR diff and the #284/#273 scope/acceptance language
- normal/dev feature dependency graph and the curated `shelterwood` façade exports
- every production implementation and installation path for `DynamicRoute`, `ActorIdentity`, `MailboxRuntime`, `MailboxControl`, and `MailboxTermination`
- the gate/member-mailbox call sites that make foreign implementations relevant to the lock rule
- every `CancellationToken`/`ParentCancellationToken` construction path after crate-gating `from_latch`
- SPEC, AGENTS lock-rule guidance, lower-crate/trait/constructor documentation
- standalone-consumer diagnostic vacuity, installer reachability, and default-feature/no-`test-util` isolation
- local-`just` versus clean-Nix enforcement parity, including the nested consumer's formatting coverage
- scope containment: no `Exit::new`, `spawn_blocking`, Tokio shutdown-contract, event-capacity, channel-alias, or driver-loop mechanical work was absorbed

### Confirmed findings and fixes

1. **The original acceptance rationale was overbroad.** Private-supertrait sealing is genuinely unavailable without dependency restructuring for `DynamicRoute`, `ActorIdentity`, and `MailboxRuntime`, because their legitimate implementations live in downstream sibling crates. It was not unavailable for `MailboxControl` or `MailboxTermination`: both production implementations live in `shelterwood-mailbox`.

   Fixed directly in `8ee8302`: both locally implemented traits now have inaccessible supertraits, with their existing mailbox types as the only implementers. The SPEC, AGENTS rule, lower-crate docs, and PR ruling now distinguish the two sealed traits from the three convention-held cross-crate seams.

2. **The façade probe proved trait-name absence but not the stated installer boundary.**

   Fixed directly in `8ee8302`: the negative consumer now also requires `MailboxCell`, `MemberCell`, `ScopeCell`, `ParentCancellationToken`, and `actor_ref_from_parts` to remain absent from the façade. A separate direct-lower-crate probe implements complete foreign `MailboxControl`/`MailboxTermination` candidates and requires both sealed-supertrait diagnostics, so it cannot pass merely because some unrelated compile error occurred.

3. **The standalone nested crate was outside the workspace formatting lane.**

   Fixed directly in `8ee8302`: pinned-nightly formatting for its manifest is now present in both `just fmt` and the clean Nix enforcement derivation.

No further actionable finding remains on the reviewed head.

### Evidence

- `./tools/dev cargo tree --locked --manifest-path tools/external-consumer/Cargo.toml -e features`: façade defaults only; no `test-util` and no Tokio `rt-multi-thread`
- expected `from_latch` failure: E0624, exact private-associated-function diagnostic
- expected façade seam failure: E0432 for all five traits plus all installer-bearing types/functions
- expected lower-crate implementation failures: E0277 for both inaccessible sealed supertraits
- `./tools/dev cargo check --locked --workspace --all-targets --all-features`: passed
- `./tools/dev just ci`: passed; 656/656 tests, both clippy lanes, doctests/docs, all three API-reachability walks, packaged-doc sync, consumer probes, and formatting
- `./tools/dev just ci-nix`: passed; all 10 clean flake checks
- hosted GitHub Actions run [31973462231](https://github.com/ralexstokes/shelterwood/actions/runs/31973462231): passed

Review verdict: the narrowed documented acceptance is now justified for exactly the three cross-crate-implemented seams, while the two locally implementable seams and the supported façade boundary are structurally enforced.

ralexstokes:
## Independent review — one P1: the seal probe **fails open** under the exact mutation it exists to catch

Reviewed head `8ee8302`. Second-opinion pass, independent of the authoring agent's self-review. The ruling itself holds up well; the enforcement around it has one real hole.

### F1 — P1 — `tools/check-external-consumer.sh:44-52` (+ `crates/shelterwood-mailbox/src/lib.rs:44-48`)

The `sealed-mailbox-seams` probe greps the diagnostics for the bare strings `SealedMailboxControl` / `SealedMailboxTermination`. Those strings appear in the E0277 message **whether or not `mod private` is private** — so the probe pins "the consumer forgot to implement a supertrait", not "the supertrait is unreachable".

Failure scenario: a later change makes the seal module nameable (`pub mod private`, a `pub use private::*`, or moving the traits into an already-`pub` module). CI stays green, the "structurally sealed" claim in SPEC/AGENTS silently becomes false, and a downstream crate installs a foreign `MailboxControl` that the framework then invokes **while holding the member mailbox mutex** (`cells.rs:651`, `:720`, `:1442-1444`) — precisely the lock-rule violation the seal exists to prevent.

Mutation evidence (applied to the committed tree, reverted after):

- `#[doc(hidden)] pub mod private` at `crates/shelterwood-mailbox/src/lib.rs:44`
- `./tools/check-external-consumer.sh` → **EXIT 0, still green**
- a scratch crate *outside the workspace* implementing `shelterwood_mailbox::private::SealedMailbox{Control,Termination}` → **compiles cleanly** under the mutation
- after `git checkout --`, that same scratch crate fails with `error[E0603]: module 'private' is private` (×2)

So the seal is genuine today — only its enforcement is missing. This does refute self-review claim #2 ("cannot pass merely because some unrelated compile error occurred"): it can, and does, pass for the wrong reason.

**Fix.** rustc emits a distinguishing note only when the supertrait is unreachable: `` `MailboxControl` is a "sealed trait" … which is not accessible ``. Verified: the phrase `is not accessible` appears **twice** with the seal intact and **zero** times under `pub mod private`. Either retarget the two greps to that phrase, or add a fourth probe feature that does `use shelterwood_mailbox::private::SealedMailboxControl;` and pins `module 'private' is private`.

### F2 — P2 — `tools/check-external-consumer.sh:22-33` + `crates/shelterwood/src/mailbox.rs:3`

The façade-absence probe is a hand-maintained 10-name list where half the class could be structural.

`crates/shelterwood/src/cells.rs` is `pub(crate) use shelterwood_cells::*;`, so *every* cells name is protected by the E0365 shim — mutation `pub use cells::ScopeCell;` in `lib.rs` did not even compile. But `crates/shelterwood/src/mailbox.rs` is `pub use shelterwood_mailbox::*;`, so mailbox names have **no** structural protection: mutation `pub use mailbox::MailboxControl;` compiled fine and was caught *only* by the grep list. That list covers 10 names; the same glob also re-exports `MailboxSignal`, `MailboxSignalWatcher`, `ErasedOneShotSender/Receiver/Close`, `ErasedValue`, `BoxedSleep`, `MailboxDisposal` — none pinned.

Splitting `mailbox.rs` into an explicit `pub use` (the ~13 façade names) plus `pub(crate) use` (the seams) makes the whole class structural instead of list-maintained.

### F3 — P2 — `crates/shelterwood-mailbox/src/futures.rs:79`

This PR adds a new `#[doc(hidden)]` to a `pub fn`, which narrows the very enforcement lane #273 complained about. Direct proof from the Nix lane: with plain `pub fn from_latch` the reachability walk reports `shelterwood_cells::cancellation::CancellationToken -> shelterwood_runtime::sync::Latch`; with `#[doc(hidden)] pub fn from_latch` the walk prints **clean** and only the shell probe catches it. `actor_ref_from_parts` is now likewise invisible to `shelterwood-api-reachability`. Its signature names only local types so there is no live leak — but the PR doesn't record that it traded checker coverage for rustdoc tidiness.

### F4 — P2 — `crates/shelterwood-cells/src/cancellation.rs:44` + the new SPEC §13 paragraph

`ParentCancellationToken::from_latch(primary: Latch) -> Self` is still plain `pub` on a `#[doc(hidden)] pub struct` — still callable by any direct dependent of `shelterwood-cells`, and still invisible to the rustdoc-JSON walk (same proof as F3). That is *consistent* with the ruling (direct lower-crate use is unsupported) and it is the mechanism the façade now uses (`raw.rs:1177`, `task.rs:44-45`). But the SPEC sentence singles out `from_latch` as "crate-gated" without noting the identical capability survives one type over, and #273's third option ("teach the checker about hidden methods on exported types") is declined silently. One sentence in the ruling, or an explicit residual on #273.

### Nit

The PR body's "default features, no `test-util`, no Tokio `rt-multi-thread`" is a manual validation — I re-verified it (`cargo tree -e features` shows only `default, macros, rt, sync, time`), but nothing in `check-external-consumer.sh` pins it. #273's acceptance criterion is met by construction, not by enforcement.

### Verdict on "sealing unavailable for the other three" — **accurate, not overclaimed**

Enumerated every impl in the tree:

- `DynamicRoute` (defined `cells.rs:835`) → sole impl `driver/admission_control.rs:531` (façade, downstream)
- `ActorIdentity` (defined `mailbox/src/lib.rs:117`) → sole production impl `cells.rs:378` (downstream); one test double
- `MailboxRuntime` (defined `mailbox/src/capability.rs:65`) → sole production impl `shelterwood-runtime/src/mailbox.rs:27` (downstream); two test doubles

I specifically hunted for approaches the PR did *not* consider, given the standing "API breaks and dependency restructuring are acceptable" directive. None work:

- A `#[doc(hidden)] pub mod private` seal or a capability-token type is nameable by any direct dependent — exactly the actor being excluded.
- A `#[cfg(feature = "__internal")]` seal is defeated by any downstream `Cargo.toml` adding the feature (features are additive).
- **Defer the calls off-lock instead of sealing** — the `MailboxRuntime` treatment — is the one structurally different lever, and it is unavailable for `DynamicRoute`: five of its six methods take `&mut ObservationTxn<'_>` (`cells.rs:837-870`); they are *designed* to run inside the gate transaction and cannot be deferred past it.
- Crate merging seals at most one trait each, and each merge destroys a documented constraint (mailbox+runtime would put Tokio into the executor-free crate, violating the #242 pin).

The AGENTS.md/SPEC wording — "Rust cannot private-seal a trait in the lower crate while permitting its legitimate implementation in a downstream sibling" — is precisely true as written. Recommend accepting the ruling as-is.

### Mutation-check results for every probe

| Mutation | Probe | Result |
|---|---|---|
| `pub(crate) fn from_latch` → `pub fn` | `from-latch` | **RED** |
| same, via `nix build .#checks…api-enforcement` | reachability walk | **RED** earlier, at the rustdoc walk |
| restore pre-PR `#[doc(hidden)] pub fn from_latch` | Nix `api-enforcement` | **RED** at `check-external-consumer.sh` — proves the shell probe really runs in the clean lane |
| `pub use cells::ScopeCell;` in façade | `installable-seams` | **RED** (E0365, structural) |
| `pub use mailbox::MailboxControl;` in façade | `installable-seams` | **RED** (grep only — see F2) |
| `mod private` → `#[doc(hidden)] pub mod private` | `sealed-mailbox-seams` | **GREEN — F1** |
| same, counting `is not accessible` | — | 2 sealed / **0** unsealed → proposed grep discriminates |

### Nix source-filter / CI parity — **clean, verified empirically**

No `extraSourceFilter` change is needed and none is missing: `rust-nix-template`'s `crane-checks.nix` composes `craneLib.filterCargoSources`, which admits any `*.rs`/`*.toml` and `Cargo.lock` anywhere in the tree, so `tools/external-consumer/{Cargo.toml,Cargo.lock,src/main.rs}` are already in the sandbox, and the existing `.*\.sh` rule covers the script. Confirmed from the build log rather than by reading: the api-enforcement derivation printed `+++ command cargo fmt --manifest-path tools/external-consumer/Cargo.toml -- --check`, and under the restore-pre-PR mutation it printed the `from_latch` failure. `just ci-nix` passed on the unmodified head.

### Scope and doc fidelity

Scope is contained — `exit.rs` (`Exit::new`), `spawn.rs` (`spawn_blocking`, tokio shutdown), `observe.rs` (`LIFECYCLE_EVENT_CAPACITY`, `borrow_latest_and_closed`), `mailbox/src/cell.rs:771`, and `driver.rs` (`UnboundedMpsc*`) are all untouched; #284's step-4 batch is neither absorbed nor pre-empted, and the ruling is applied to all five enumerated seams.

Docs say what the code does (`CLAUDE.md` is a symlink to `AGENTS.md`, so the single edit covers both; nothing claims all five are sealed). One wording nit: AGENTS.md calls the three "the one convention-held public boundary", but `MailboxSignal`, `MailboxSignalWatcher` and the `ErasedOneShot*` family are also public unsealed cross-crate traits. They ride under `MailboxRuntime` as sub-capabilities it mints — the text just doesn't say so.

Seven mutations applied and reverted; no `result` symlink left behind; worktree clean at `8ee8302`.

ralexstokes:
## Review findings resolved (`7aadaab`)

| # | Finding | Resolution |
|---|---|---|
| F1 (P1) | `sealed-mailbox-seams` probe fails open — bare-name greps pin "forgot a supertrait", not "supertrait unreachable" | Belt and braces, both applied. The two greps now target rustc's seal-specific note, ``` `shelterwood_mailbox::private::Sealed…`, which is not accessible ``` — the phrase rustc emits *only* when the supertrait is out of reach. Plus a new fourth probe feature `private-seal-module` that names the module directly (`use shelterwood_mailbox::private::SealedMailboxControl;` / `…Termination;`) and pins `` module `private` is private `` (E0603). |
| F2 (P2) | `crates/shelterwood/src/mailbox.rs` is a `pub use …::*` glob, so mailbox seams had no structural protection — only the hand-maintained 10-name grep list | Split into an explicit `pub use` of the 13 supported façade types (`ActorRef`, the `Send*`/`Call*`/`Reply*` families) and a `pub(crate) use` of the 5 seams the façade actually consumes (`AcceptedSequence`, `MailboxCell`, `MailboxControl`, `MailboxReceiver`, `actor_ref_from_parts`). Everything else — `MailboxSignal`, `MailboxSignalWatcher`, `ErasedOneShot*`, `ErasedValue`, `BoxedSleep`, `MailboxDisposal`, `MailboxRuntime`, `MailboxTermination`, `ActorIdentity` — is no longer re-exported at all. The class is now structural, matching the `cells` shim. |
| F3 (P2) | New `#[doc(hidden)]` on `actor_ref_from_parts` narrows the rustdoc-JSON reachability lane #273 exists to strengthen | Recorded at the site (`crates/shelterwood-mailbox/src/futures.rs`) as a non-doc comment on the attribute: the walk skips hidden items, the signature names only local types so nothing leaks today, and `tools/check-external-consumer.sh`'s `installable-seams` probe is now the sole enforcement. Also called out below. |
| F4 (P2) | SPEC §13 singles out `CancellationToken::from_latch` as crate-gated without noting `ParentCancellationToken::from_latch` keeps the identical capability; #273's third option declined silently | SPEC ruling extended: the same capability survives one type over on a `#[doc(hidden)] pub struct`, reachable only through the same unsupported direct dependency, so crate-gating is a tidiness gain and not the enforcement; and the walk is *deliberately* left blind to hidden items rather than taught to descend into hidden methods on exported types — the façade-absence probe is the enforcement. Residual posted on #273. |
| Nit | AGENTS.md calls the three seams "the one convention-held public boundary" | Reworded: `MailboxSignal`, `MailboxSignalWatcher` and the `ErasedOneShot*` family are public unsealed cross-crate traits for the same reason and ride under the same ruling as sub-capabilities `MailboxRuntime` mints. (`CLAUDE.md` is a symlink to `AGENTS.md`; symlink intact.) |

### Mutation evidence — the seal probe now fails **closed**

Mutations applied to the committed tree at `7aadaab`, each reverted with `git checkout --`.

| Mutation | Probe | Before | After |
|---|---|---|---|
| `mod private` → `#[doc(hidden)] pub mod private` (`shelterwood-mailbox/src/lib.rs`) | `check-external-consumer.sh` (whole script) | **EXIT 0 — green** | **EXIT 1 — red**, `sealed-mailbox-seam probe failed for an unexpected reason`; `grep -c 'is not accessible'` on the diagnostics = **0** |
| same | `private-seal-module` feature alone | (did not exist) | `cargo check` **exits 0** — i.e. the seal module became nameable, which the script reports as `external consumers can name the mailbox seal module` |
| `pub use mailbox::MailboxControl;` in the façade root | `cargo check -p shelterwood` | compiled; caught by grep only | **exit 101**, `error[E0365]: `MailboxControl` is only public within the crate, and cannot be re-exported outside` |
| `pub use mailbox::MailboxSignal;` in the façade root | `cargo check -p shelterwood` | compiled (unpinned entirely) | **exit 101**, `error[E0432]: unresolved import `mailbox::MailboxSignal`` |

Unmutated: `./tools/check-external-consumer.sh` exits 0.

### Enforcement-coverage note (F3)

This PR's `#[doc(hidden)]` additions trade rustdoc-JSON checker coverage for rustdoc tidiness. `shelterwood-api-reachability` walks only non-hidden items, so `actor_ref_from_parts` and `ParentCancellationToken::from_latch` are now invisible to it. Neither leaks — both signatures name only types local to their crate — but the shell probe is the sole remaining enforcement for those items, and that is now written down at the `futures.rs` site and in the SPEC ruling.

### CI

- `./tools/dev just ci` → **exit 0**
- `./tools/dev just ci-nix` → **exit 0**, `all checks passed!`. `checks.x86_64-linux.api-enforcement` resolves to `maya99p81gfv3jf90ynhh9xlm9c9ljqm-shelterwood-0.1.0.drv`, which appears in the build list — the clean lane really re-ran the probe against these changes.
- No new files under `tools/`, so no `flake.nix` `extraSourceFilter` entry is needed; `filterCargoSources` already admits `tools/external-consumer/{Cargo.toml,Cargo.lock,src/main.rs}` and the existing `.*\.sh` rule covers the script. No `result` symlink left behind.

Not merging — leaving that to you.

Issue #272, opened 2026-08-16
Lock-rule hardening: make the remaining convention-held sites structural


ralexstokes:
Batch issue from the 2026-08-16 review round: every remaining site where the lock rule (or its travelling conventions) is held by discipline rather than structure. None is a reachable bug today; each was verified with an ownership trace. Grouped so one pass can close them.

## Confirmed deviations (fix)

- [ ] **User `Debug` under the watch read lock** — `Debug for WatchSender`/`WatchReceiver` (`shelterwood-runtime/src/sync.rs:545-562`) formats `&*self.0.borrow()` while tokio's watch `RwLock` read guard is held; `WatchSender<MemberRecord>` carries a user error, and derived `Debug` on cells types reaches it. Format a cloned snapshot (or metadata only); never format retained exits or watched values under the guard.
- [ ] **`unreachable!` under the member mailbox mutex** — `cells.rs:757-761` would poison the mutex the codebase `.expect()`s; the same function's other panic path (`cells.rs:657-666`) releases first. Use the documented fallback (`debug_assert!`) or restructure to release.
- [ ] **Mailbox exhaustion panics destroy a user message under the lock** — waiter-ID exhaustion (`cell.rs:473-477`, via `submit`'s last-owner `Arc<SendOperation>`) and accepted-sequence exhaustion (`cell.rs:1190-1196`, live `message: M` parameter in `accept_locked`/`promote_waiter_queue`): the unwind drops M strictly before `MailboxTxn::Drop` releases the guard. ~2^64 reachability, but contrary to compute-verdict-release-then-panic. Return exhaustion as transition data; unlock/isolate; then panic.

## Make structural (currently convention-held)

- [ ] **Wake-capable `MailboxControl` methods vs the gate** — `bind`/`freeze`/`close` flush effects (pulses, waker drops) on return; nothing stops a future driver edit calling them inside a gate closure, or a mailbox edit adding a pulse to `prepare_termination` (the one method cells calls under the gate, currently effect-free). Thread an effects-sink parameter (the `DynamicRoute` pattern) or add `debug_assert!(!gate.is_held())` at the flush.
[...]

PR #294, opened 2026-08-19
Make terminalizer races first-wins and lock-safe

### LINKED ISSUE #269: Reconcile losing-terminalizer handling between ScopeCell and MemberCell
Found in the 2026-08-16 review round (internal sweep).

## The disagreement

Two adjacent layers hold contradictory positions on losing terminalizers:

- `MemberCell::terminalize_locked` (`cells.rs:735-739`, `:771-773`) deliberately applies `StartupDisposition` even when a competing terminalizer already won the stage transition ("A supervised terminal edge updates `startup_aborted` even if a competing terminalizer won").
- Its only production route to a non-`Unchanged` disposition, `ScopeCell::terminalize_child`, returns `false` at the early already-`Terminal` check (`cells.rs:1553-1556`) before ever reaching it — silently dropping the `Aborted` verdict and skipping the `Exited` event.

Today single-driver serialization appears to make the early return unreachable for supervised edges (task-side `member.terminalize` calls are test-only), so either the inner losing-edge startup handling is dead code or the outer guard is wrong. Reconcile before a future caller makes the race real.

Related hardening from the same trace (adjudicated: not bugs today, both proofs are non-local):

- The early return drops the incoming `Exit` under the gate. On every reachable path this is provably refcount-only (the `handle_construction_disposed` local co-owns it across the gate section) or payload-free (`discharge` path) — but a future caller passing a sole-owner `Failed` exit would run a user destructor under the gate. Returning the losing value as transaction output (or `txn.defer`) makes it locally correct.
- The stale-epoch branch of `finish_incarnation_with_terminal` (`cells.rs:1962-1976`) drops an unused `StopReason` under the gate; safety currently rests on `retain_stop_reason` (`driver/shutdown.rs:111`) plus the `ScopeRuntime` field-ordering comment (`driver.rs:236-241`). A `txn.defer` there is cheap local correctness.
- The deferred losing-exit drop (`cells.rs:730-731`) runs synchronously on the committing thread; deferring `RetainedExit::new(losing_exit)` instead routes a possibly-blocking user destructor to `dispose_critical`, consistent with the #244/#254 destruction class.

## Acceptance criteria

- [ ] Reconcile `terminalize_child`'s early return with `terminalize_locked`'s losing-edge handling (delete the dead half or fix the guard); test whichever survives.
- [ ] Make the three under-gate/inline drops locally correct (`defer` / `RetainedExit`) rather than globally proven.

## Related

- #254 round — the Exit destruction class this extends

ralexstokes:
Closes #269.

## What changed

- make the first terminalizer own the complete terminal record, including startup classification
- prevent losing terminalizers from republishing a record edge
- route losing failed exits through `RetainedExit` so user payload destruction is isolated
- defer losing supervised exits and stale structured stop reasons until after the observation gate unlocks
- add direct regressions for the losing member path, the supervised outer guard, and stale scope epochs

## Why

`ScopeCell::terminalize_child` already rejected an already-terminal child, while `MemberCell::terminalize_locked` still contained dead logic that could update `startup_aborted` after a competing terminalizer won. The same trace exposed three paths whose safety depended on non-local ownership proofs for user-owning exits and stop reasons.

The resulting contract is explicit: the outer guard is authoritative, first terminalizer wins, and rejected values leave through post-unlock or isolated disposal.

## Validation

- mutation-checked all four repaired paths against their focused regressions
- `./tools/dev cargo test -p shelterwood-cells` (25 passed)
- `./tools/dev cargo test -p shelterwood --lib` (183 passed)
- `./tools/dev just ci` (678 nextest tests plus docs, API reachability, external consumer, and nix formatting)
- `git diff --check`

## Stack

This is the base for the #272 lock-rule hardening PR.

### DIFF
--- a/crates/shelterwood-cells/src/cells.rs
+++ b/crates/shelterwood-cells/src/cells.rs
@@ -2127,8 +2127,12 @@ impl ScopeCell {
                     self.close_observation_locked(wakes);
                 }
                 // `StopReason::StartupFailed` recursively owns the failed
-                // child's user error. A stale verdict is unused, but it still
-                // leaves only after the observation gate is released.
+                // child's user error. A stale verdict is unused, but the
+                // framework still owns this copy: retaining it before the
+                // deferred drop sends a possibly-blocking or panicking user
+                // destructor to `dispose_critical` instead of running it
+                // inline on the committing thread once the gate is released.
+                let reason = RetainedStopReason::new(reason);
                 wakes.defer(move || drop(reason));
                 return;
             }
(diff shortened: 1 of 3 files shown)


ralexstokes (commented):
Reviewed the complete immediate-base diff and its place in the #294 → #300 → #304 → #314 stack. I found no actionable issues in the first-wins terminalization logic, losing-exit retention, stale-epoch cleanup, or the added regressions. Targeted validation passed: `cargo test -p shelterwood-cells` (25 tests) and `cargo test -p shelterwood --lib` (183 tests).

ralexstokes:
Fixed in `a2e562d`. The stale-epoch branch of `finish_incarnation_with_terminal` retired the declined `StopReason` with a bare `wakes.defer(move || drop(reason))`. That leaves the observation gate, but a `StopReason::StartupFailed` recursively owns the triggering child's `Exit`, so the type-erased user error ran its destructor inline on the committing thread — a tokio worker on the `driver.rs` completion path (`driver.rs` clones `reason` out of `ScopeCompletion` and drops the `RetainedStopReason` guard before calling `finish_root_incarnation`), or the `shutdown_and_wait` caller. CLAUDE.md requires framework-retained `Exit` copies to reach `runtime::dispose_critical`, not merely to leave the critical section, and this PR's other two exit paths were already upgraded to `RetainedExit`. The reason is now wrapped in `RetainedStopReason` before the defer: construction is Arc refcount traffic under the gate, and the post-unlock drop hands the failed exit to the isolated disposal lane.

New test `cells::tests::a_stale_scope_verdict_disposes_its_nested_exit_off_the_finishing_thread`, modelled on the sibling `a_losing_failed_terminal_exit_disposes_off_the_retiring_thread`: the nested user error reports its own drop thread through a channel and the test body asserts it differs from the finishing thread. It pins the destination of the disposal, not just the timing — the pre-existing `a_stale_scope_verdict_destroys_its_nested_exit_outside_the_gate` only probes gate-held-ness, which a bare post-unlock inline drop already satisfies, so it passed with and without the fix. Mutation-checked: deleting only the `RetainedStopReason::new` line fails the new test (`left: ThreadId(2), right: ThreadId(2)`) and restoring it passes.

`cargo nextest run -p shelterwood-cells` passes 26/26; the driver-side gate/terminalization filter passes 19/19.

Issue #279, opened 2026-08-16
Mechanical dedup and file splits (raw.rs, cells.rs)


ralexstokes:
Mechanical refactor batch from the 2026-08-16 review round. No behavior changes; each item verified as textual duplication or a clean seam.

## Dedup

- [ ] `TimerStore` bucket scan written three times (`raw.rs:626-656`, `:699-720`, `:722-732`): a private `locate(hash, predicate)` + `unlink(entry)` pair collapses ~40 lines and centralizes the index-coherence invariant.
- [ ] Continuation-pop stage duplicated (`raw.rs:1617-1625` vs `:1669-1676`): one helper with an `allow_lead` flag, co-located with `record_continuation_delivery`'s debug_assert.
- [ ] Cells gate-acquire/retry loops duplicated (`with_observation_txn` vs `with_observation_gate`, `cells.rs:540-555` vs `:1404-1425`; the two `adopt_observation_gate` inner loops, `:584-598` vs `:1332-1357`): one member-level primitive parameterized on the probe hook.
- [ ] `begin_drain_transition` vs `force_all` DrainStarted dance (`shutdown.rs:113-133` vs `:140-159`): shared helper taking `Option<Result<(), StartupError>>`. (Subsumed if the supervisor-effects issue part 2 lands first.)
- [ ] `handle_construction_disposed` retention-prune duplication (`child.rs:1140-1152`): hoist the `Retention::Remove → prune_terminal` check after the branch.
- [ ] `with_dynamic_entry` generic helper with one caller (`removal.rs:9-27`): inline. Fix the stale comment naming nonexistent `dynamic_membership_is_removing` (`removal.rs:116-118`).
- [ ] Duplicated mailbox attach (`tree/builders.rs:48-54` vs `tree/dynamic_api.rs:68-77`): one helper.

## File splits (along existing types, no new traits)

- [ ] `raw.rs` (3064) → `raw/`: (1) disposal funnel + owned containers (`PanicSlot`/`RawDisposal`/`EventQueue`/`ContinuationQueue`/`TimerStore`), (2) offload machinery (`Guard`/`Blocking`/`CatchUnwindFuture`/`SharedOffloadState`), (3) `RawContext` + arbitration (`ReadyBatch`, `next_ready`) — this one stays whole, (4) definition/erasure plumbing. No shared private state beyond `RawDisposal`.
[...]

PR #295, opened 2026-08-19
Keep raw timer state coherent across caught panics

### LINKED ISSUE #268: A panicking interval Clone desynchronizes the timer index
Found in the 2026-08-16 review round (internal sweep; independently re-reported by an external review).

## The bug

A panicking interval `M::clone` desynchronizes the timer index, silently killing the interval:

- `deliver_timer` (`raw.rs:1747-1757`) writes `entry.deadline` (`:1751`) **before** running the user clone (`:1755`) and re-indexing `deadlines` (`:1756`). A clone panic leaves an entry whose `deadline` field claims an instant absent from the deadline index — the interval never fires again, with no diagnostic.
- Similarly, `next_ready` `take`s the `ReadyBatch` before running user code (`make_message` at `:1710`, the clone at `:1755`); a panic drops the batch while its armings were already removed from `deadlines` at `take_due` (`:1731`) — permanently unindexed live entries.

Reachable from safe code: a `RawActor::run` body wraps `context.recv()` in `catch_unwind` and keeps using the context — timers vanish. On the framework path the panic is incarnation-fatal and `freeze` clears everything, so the exposure is exactly the caught-unwind case.

## Fix shape

Either reorder `deliver_timer` (clone before mutating table state) and make `next_ready`'s user-code steps effect-last, or document on `recv`/`try_recv` that the context is unspecified after a panic escapes a receive call — plus, either way, the missing test: no panicking-`Clone` fixture exists anywhere in the suite, and the interval delivery clone (`raw.rs:1755`) is the one user-callback venue in the timer path with no pinning test. A second untested edge in the same area: interval rearm overflow at delivery (`raw.rs:1750` producing `None` → dormant interval); only arming-time overflow is covered (`events.rs:1027`).

## Acceptance criteria

- [ ] Decide reorder vs. document; implement.
- [ ] Panicking-`Clone` interval test (also pins whichever behavior is chosen).
- [ ] `TimerStore`-level test for interval rearm overflow at delivery.

ralexstokes:
Closes #268.

## What changed

- keep the authoritative `ReadyBatch` installed while offload continuations and timer delivery invoke user code
- leave due timer armings queued until delivery succeeds
- clone interval messages before mutating the timer entry or deadline index
- pin delivery-time rearm overflow leaving a live, dormant, clearable interval (already the behaviour on `main`; this PR adds the regression #268 asked for)
- add regressions for a caught interval-clone panic, a caught continuation panic with later fired work, and interval rearm overflow

## Root cause

`next_ready` removed its `ReadyBatch` before callback-capable work, while `deliver_timer` updated the entry deadline before invoking the user `Clone`. A caught panic could therefore discard the remaining fired cut or leave an interval claiming a deadline absent from the deadline index.

The callback step is now effect-last: the batch remains recoverable during unwind, and timer state changes only after cloning succeeds.

## Validation

- `./tools/dev cargo test -p shelterwood raw:: --lib` (23 passed)
- `./tools/dev cargo test -p shelterwood --test events` (13 passed)
- `./tools/dev just ci` (677 nextest tests plus clippy, docs, API reachability, external consumer, and nix formatting)
- `git diff --check`

This PR is independent of #269/#294 and should land before #279 splits `raw.rs`.

### DIFF
--- a/crates/shelterwood/src/raw.rs
+++ b/crates/shelterwood/src/raw.rs
@@ -749,6 +749,25 @@ impl<M> TimerStore<M> {
         }
     }
 
+    fn rearm_interval(&mut self, arming_order: ArmingOrder, now: Instant) -> Option<M> {
+        let (message, deadline) = {
+            let entry = self.entry_mut(arming_order)?;
+            let period = entry.period?;
+            let deadline = crate::deadline::Deadline::after(now, period).instant();
+            let TimerMessage::Interval(message, clone_message) = &entry.message else {
+                unreachable!("an interval timer must own a message factory")
+            };
+            // Cloning is user code. Keep the entry's prior deadline intact
+            // until it succeeds so the fired batch can retry this arming if
+            // the panic escapes `recv` and the raw actor catches it.
+            let message = clone_message(message);
+            entry.deadline = deadline;
+            (message, deadline)
+        };
+        self.arm_deadline(arming_order, deadline);
+        Some(message)
+    }
+
     fn next_deadline(&self) -> Option<Instant> {
         self.deadlines.first().map(|(deadline, _)| *deadline)
     }
@@ -1658,7 +1677,10 @@ impl<M: Send + 'static> RawContext<M> {
                     .events
                     .pop_through(&mut batch.offloads_remaining)
                 {
-                    if let Some(message) = self.materialize_event(event) {
+                    let (restored, message) = self
+                        .with_ready_batch_installed(batch, |this| this.materialize_event(event));
+                    batch = restored;
+                    if let Some(message) = message {
                         self.resources.continuation_needs_external = false;
                         self.resources.ready_batch = Some(batch);
                         return Some(message);
@@ -1693,8 +1715,13 @@ impl<M: Send + 'static> RawContext<M> {
             }
             batch.continuations_remaining = 0;
 
-            while let Some(arming) = batch.armings.pop_front() {
-                if let Some(message) = self.deliver_timer(arming) {
+            while let Some(arming) = batch.armings.front().copied() {
+                let (restored, message) =
+                    self.with_ready_batch_installed(batch, |this| this.deliver_timer(arming));
+                batch = restored;
+                let removed = batch.armings.pop_front();
+                debug_assert_eq!(removed, Some(arming));
+                if let Some(message) = message {
                     self.resources.continuation_needs_external = false;
                     self.resources.ready_batch = Some(batch);
                     return Some(message);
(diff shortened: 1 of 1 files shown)


crates/shelterwood/src/raw.rs
```diff
@@ -749,6 +749,25 @@ impl<M> TimerStore<M> {
         }
     }
 
+    fn rearm_interval(&mut self, arming_order: ArmingOrder, now: Instant) -> Option<M> {
+        let (message, deadline) = {
+            let entry = self.entry_mut(arming_order)?;
```
ralexstokes: **[P1] Preserve the stale-arming skip when the timer was cleared or replaced**

`None` here now means both “this entry is a one-shot” and “this captured arming no longer exists.” A fired batch persists across handler calls, so with two due timers an actor can receive the first and clear (or replace) the second; the next `try_recv()` reaches `deliver_timer`, treats this `None` as a one-shot, and panics on `remove_arming(...).expect(...)`. The pre-change `deliver_timer` began with `entry_mut(arming)?`, so stale captured armings were skipped, matching `clear_timer`’s documented support for elapsed-but-undelivered timers. Please preserve that distinction (for example with an explicit rearm outcome) and add the two-due-timers/clear-second regression. I reproduced this on the PR head; the added test fails at `raw.rs:1819`.

ralexstokes (commented):
Reviewed the full `main...agent/issue-268` diff and its interaction with stacked PR #310. The panic-recovery ordering and hostile `Hash`/`Eq` containment otherwise look sound, and `./tools/dev cargo test -p shelterwood raw:: --lib` passes 23/23 on the PR head. One merge-blocking stale-arming regression is called out inline; an isolated reproduction fails consistently.

crates/shelterwood/src/raw.rs
ralexstokes: Fixed in `16a339f`. `rearm_interval` now returns an explicit `Missing` / `OneShot` / `Interval` outcome, so a cleared or replaced captured arming is skipped instead of being treated as a one-shot. I added the two-due-timers regression that clears the second elapsed-but-undelivered timer. `./tools/dev cargo test -p shelterwood raw:: --lib` passes 24/24, and stacked PR #310 has been rebased onto this fix.

ralexstokes:
Review follow-up (3749863): documented the coherence guarantee this PR introduces, and corrected one PR-body bullet.

**Doc addition.** Keeping the fired batch installed across a caught panic is user-observable in one direction the PR did not record: on re-entry `begin_ready_batch` early-returns (the batch is fired), the mailbox/offload completion flags are already set and continuations are 0, so control reaches the *same* arming again. A raw `run` body that wraps `context.recv()` in `catch_unwind` and continues now loops forever on an interval whose user `Clone` panics deterministically, where before this PR the batch was discarded and the interval died silently while the actor kept running. The framework path is incarnation-fatal, so this is confined to hand-rolled `catch_unwind` loops — but it should be written down. Added the same paragraph to the `recv` and `try_recv` rustdoc:

> A panic escaping this call leaves the fired selection cut installed, so the next receive retries the same timer arming rather than discarding the remaining batch. A raw loop that catches such a panic and receives again therefore repeats an interval whose user `Clone` panics deterministically; clear that key with [`clear_timer`](Self::clear_timer) — or [`stop`](Self::stop) — before resuming the loop.

That closes the "reorder OR document" choice #268 offered: the reorder shipped, this is the other half.

**Body correction.** The bullet "make delivery-time rearm overflow leave a live, dormant, clearable interval" overstated the change. On `main`, `deliver_timer` already computed `Deadline::after(..).instant()` as an `Option<Instant>`, assigned it to `entry.deadline`, and passed it to an `arm_deadline` that already took `Option` and skipped `None` — a live, dormant, clearable interval, identically. This PR changes nothing there; it adds the regression #268 asked for. Reworded the bullet to "pin".

**Checks:** `cargo check -p shelterwood` clean; `RUSTDOCFLAGS=-D warnings cargo doc -p shelterwood --no-deps` clean (intra-doc links resolve); `cargo test -p shelterwood --doc` 3 passed; `cargo +nightly fmt --all` no diff.

Issue #282, opened 2026-08-16
Test-fixture consolidation: driver preambles, integration fixtures, library-crate duplication


ralexstokes:
Test-fixture consolidation batch from the 2026-08-16 review round. All duplication verified by location; the PR #251 rule (doubles delegate, never restate) is the standard throughout.

- [ ] **`ScopeRuntimeBuilder::build` restates the production `ScopeRuntime` literal field-for-field** (`driver/tests/support.rs:237-285` vs `driver.rs:873-898`), bypassing `ScopeEpochGuard::transfer()` and hand-recreating the route-before-Added publication ordering. A production-side `for_test` constructor that delegates to the real wiring removes the parallel copy — a compile error currently catches an added field but not changed initialization semantics.
- [ ] **The ~35-line lowered-plan preamble repeats at ~17 sites** (`shutdown.rs:48-81,122-154,203-229`; `exhaustion.rs` ×4; `startup.rs` ×10; `events.rs:157-174`): `running_dynamic_fixture` (`support.rs:313`) proves the collapse works; add an ordered-scope sibling (~300+ lines removed).
- [ ] **`driver/tests/dynamic.rs` reserve→define→insert block copy-pasted three times** (`:86-118`, `:372-400`, `:460-498`); the `recv_child_exit` timeout/panic match re-inlined for `Removal`/`ConstructionDisposed` at 7+ sites — extend the helper.
- [ ] **Integration-suite fixtures**: the gated-recorder actor defined independently in `mailbox.rs:35-95`, `deadlines.rs:28-90`, `delivery.rs:261-279` → one `common/` fixture. `signalled_waiting_task` re-inlined at 6+ sites in `readiness.rs`/`lifecycle.rs`/`dynamic.rs` despite the `common/waiting.rs` helper. The `StartupFailed→Child→ExitKind` destructuring ladder inline ×4 in `subtrees.rs` → `common/`. Lifecycle-stream scan loops hand-rolled with divergent lag policy at 8+ sites → promote `next_item`/`next_event` (from `tests/observation.rs:24-36`) into `common/` and unify lag policy. `tasks.rs:328-334` hand-rolled wait → `assert_eventually!`/`POLL_TIMEOUT`.
[...]

PR #296, opened 2026-08-19
Make regression assertions externally observable

### LINKED ISSUE #281: Assertion-quality fixes: PR #249 recurrences and overfit pins
Assertion-quality batch from the 2026-08-16 review round: recurrences of previously-fixed test pathologies, each verified.

## PR #249 recurrences (assert judged where supervision eats it)

- [ ] `tests/offloads.rs:111` `zero_budget_offload_never_polls_work_and_times_out_on_actor_task` — the "times out"/"on actor task" claims live in in-continuation asserts (`:63-64`); the body checks only `Finished` + `!polled`. A mutation delivering `Ok(7)` or resuming off-task passes.
- [ ] `tests/offloads.rs:1656` `offload_completion_flood_is_absorbed_and_fully_delivered` — exactly-once is an in-handler assert (`:1637-1640`); duplicates panic the actor, body still sees `Finished`.
- [ ] `tests/offloads.rs:1728` `sequential_offload_cycles_deliver_every_completion` — in-order assert in-handler (`:1710`); same swallow.
- [ ] (minor) `tests/offloads.rs:1801` — the `Err(DeadlineElapsed)` classification side-claim is in-continuation; the backlog property itself is soundly judged.
- [ ] `tests/drain.rs:259` — the `!is_draining()` assert placement can only fail by hanging into the harness timeout (release the gate first, judge in the body).
- [ ] `tests/mailbox.rs:309-331` — the "late replies are discarded" half is two assertion-free `send` calls; strengthen or delete in favor of `deadlines.rs:594-644`.

The model to copy is in the same file: `offloads.rs:745-751` publishes evidence and judges in the body.

## Overfit / fragile pins

- [ ] `driver/tests/dynamic.rs:3-74` — the mid-cycle `RuntimeStorage` literal demands `deadline_slots == 1`, i.e. the raw `BinaryHeap` length including tombstones (`engine.rs:1062-1065`), though compaction policy permits 2× slack. Pin `children`/leak-zero exactly; bound `deadline_slots` instead. (Blocks any deadline-queue representation change.)
- [ ] ~16 driver-test sites use ad-hoc 2s/5s real-clock bounds against `support.rs:292-301`'s own documented 10s constants (`terminalization.rs:614,633`; `dynamic.rs:612,793,822,841,1180,1219,1287,1469`; `observation.rs:32,58,87,115,463,656,708,772`). Switch to `CAPTURE_PROBE_WAIT`/`DRIVER_PROGRESS_WAIT`.
- [ ] `crates/shelterwood-mailbox/src/futures.rs:861-888` — raw `thread::sleep(5ms)` poll loops where sibling suites use the mpsc `recv_timeout` handshake.
- [ ] `driver/tests/dynamic.rs:218-222` — "reservations are terminalized on close" message; only emptiness is checked, `MemberStage::Terminal` is not.
- [ ] `shelterwood-runtime/src/timer.rs:174-192` — both tick-boundary tests silently `return` green on platforms where the boundary is unrepresentable; mark the skip.
- [ ] PR #242 regression: `tokio::runtime::Handle::current()` in `driver/tests/dynamic.rs:904,917` — replace with a shelterwood-runtime test-util fixture (the `DedicatedRuntime` pattern).

## Acceptance criteria

- [ ] Each listed test either strengthened (evidence published, judged in the body) or replaced; verify by mutating the code under test where practical (the #234/#245 discipline).

ralexstokes:
Closes #281.

## What changed

- move offload timeout, task-affinity, exactly-once, and ordering assertions out of actor continuations into test bodies
- publish and judge pre-drain state after releasing handler gates
- make late-reply disposal observable and exact
- preserve exact leak-zero storage assertions while bounding raw deadline heap slack to its documented policy
- use shared driver wait constants instead of ad-hoc shorter real-clock bounds
- replace mailbox disposal polling sleeps with a bounded channel handshake
- assert reservation terminalization directly
- route the admission race through the runtime test fixture instead of entering Tokio directly
- make platform-specific timer boundary skips explicit

## Why

Several regressions repeated the PR #249 failure mode: an assertion inside supervised user code could panic the actor while the outer test still passed. Other tests pinned implementation details more tightly than their contract or used timing/runtime fixtures inconsistent with the rest of the suite.

The tests now publish evidence outward and judge it where supervision cannot swallow the assertion.

## Validation

- mutation probes fail for a dropped offload completion, retained late reply, skipped reservation terminalization, and unbounded deadline storage
- affected integration suites: 45 tests
- driver unit suite: 135 tests
- mailbox unit suite: 28 tests
- runtime timer suite: 7 tests
- `./tools/dev just ci` (674 nextest tests plus clippy, doctests, docs, API reachability, external consumer, and Nix formatting)
- nightly format check and `git diff --check`

This should land before #282 consolidates fixtures and before #279 moves the deadline queue.

### DIFF
--- a/crates/shelterwood-mailbox/src/futures.rs
+++ b/crates/shelterwood-mailbox/src/futures.rs
@@ -700,6 +700,7 @@ mod tests {
         sync::{
             Arc, Mutex, Weak,
             atomic::{AtomicUsize, Ordering},
+            mpsc,
         },
         task::{Context, Poll, RawWaker, RawWakerVTable, Wake, Waker},
         thread::ThreadId,
@@ -712,6 +713,8 @@ mod tests {
         super::cell::tests::{actor, actor_for},
         ActorRef, MailboxCell, SendFuture, SendFutureState, SendOperation,
     };
+
+    const DISPOSAL_WAIT: Duration = Duration::from_secs(5);
 
     #[derive(Default)]
     struct WakerVtableCalls {
@@ -851,11 +854,15 @@ mod tests {
         );
     }
 
-    struct CountedDrop(Arc<AtomicUsize>);
+    struct CountedDrop {
+        drops: Arc<AtomicUsize>,
+        disposed: mpsc::Sender<()>,
+    }
 
     impl Drop for CountedDrop {
         fn drop(&mut self) {
-            self.0.fetch_add(1, Ordering::SeqCst);
+            self.drops.fetch_add(1, Ordering::SeqCst);
+            let _ = self.disposed.send(());
         }
     }
 
@@ -863,17 +870,18 @@ mod tests {
     fn dropping_a_never_polled_send_disposes_the_message_exactly_once() {
         let (mailbox, actor): (Arc<MailboxCell<CountedDrop>>, ActorRef<CountedDrop>) = actor_for();
         let drops = Arc::new(AtomicUsize::new(0));
+        let (disposed, disposal) = mpsc::channel();
 
-        drop(actor.send(CountedDrop(Arc::clone(&drops))));
+        drop(actor.send(CountedDrop {
+            drops: Arc::clone(&drops),
+            disposed,
+        }));
 
         // The never-submitted message routes through detached isolated
         // disposal on another thread, so acknowledge it with a bounded wait.
-        for _ in 0..1_000 {
-            if drops.load(Ordering::SeqCst) == 1 {
-                break;
-            }
-            std::thread::sleep(std::time::Duration::from_millis(5));
-        }
+        disposal
+            .recv_timeout(DISPOSAL_WAIT)
+            .expect("isolated disposal destroys the never-submitted message");
         assert_eq!(drops.load(Ordering::SeqCst), 1);
         let state = mailbox.state.lock().expect("mailbox mutex poisoned");
         assert!(state.queue.is_empty());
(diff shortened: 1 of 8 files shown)


ralexstokes (commented):
Reviewed the exact `main...agent/issue-281` diff, including the externally observed offload/drain assertions, bounded disposal handshakes, deadline-storage contract, and admission-race fixture. I found no actionable issues. The PR's GitHub CI check is passing, and the inherited changes were also exercised through the reviewed leaf branches.

ralexstokes:
Applied the review findings. All six addressed; two adjudicated differently
than the review framed them.

**1 (P2) — exact leak assertion restored.** `deadline_slots` is back inside
the single `assert_eq!` against the whole `RuntimeStorage`, replacing the
`deadlines..=deadlines * 2` band. The band was a loosening inside a
strengthening PR, and the value is deterministic: one arm pushes one entry
and the neighbouring post-removal assertion pins the queue empty first.

The review's *reason* for the exactness does not hold, so the comment states
a different one. Mutation-check, three runs on the committed tree:

- Weakening `compact_if_sparse` to `> len * 2 + 1`: **test still passes.**
- Removing compaction entirely (`if false`): **test still passes.**
  This fixture holds one deadline at a time, so every stale entry becomes the
  heap head and `prune_stale_head` reclaims it from the driver's
  `next_deadline()`. `compact_if_sparse` is not what keeps this at 1.
- Pushing one extra never-registered entry per `DeadlineQueue::push`, sorted
  after the live entry so `prune_stale_head` cannot reach it: **exact form
  FAILS** — `deadline_slots: 2` vs `1`, at cycle 0. Restoring the PR's band
  with the same mutation still applied: **PASSES.**

So the band does hide a real per-deadline heap leak; the leak just arrives
from the arm side rather than the cancel side.

**2 (P2) — confirmed; disposal evidence now waits in real time.** The
reasoning holds: `Reply::send` routes an unclaimed value through
`dispose` → `dispose_detached` → `submit_blocking_job`, so destruction costs
a blocking-pool or fallback thread start, while `assert_eventually!` sleeps
on the paused clock that auto-advances as soon as the runtime idles.
`CountedReplyDrop` now publishes a labelled notification on a
`std::sync::mpsc` channel, awaited with `recv_timeout` — the shape
`futures.rs` already uses.

**4 (P3) — `assert_quiet` made real rather than deleted.** Deleting it would
have dropped the only double-disposal probe. The two per-value counters are
now one shared counter, judged at `== 2`, followed by a
`recv_timeout(50ms)` on the same channel expecting `Timeout`. The test keeps
a sender alive so that is a real wait rather than an immediate
`Disconnected`, and the test's wall time went 0ms → 53ms, which is the
window actually elapsing.

**3 (P3) — both directions now judged in the test body.** A `DrainingProbe`
(`Arc<Mutex<Vec<bool>>>`) replaces the two new booleans and both surviving
in-handler `assert!(context.is_draining())`. Recording a *list* removes the
default-equals-asserted smell without a tri-state: an empty log satisfies no
expectation, and the expected lengths are tied to artifacts the test already
owns — `expected.len() - 1` drained `Value` deliveries, and exactly `drained`
fault visits, so a mode that never drains expects nothing and an absent
handler cannot pass.

Inverting all four `record(...)` arguments fails **8/8** drain tests,
including `SharedGrace`, which previously degraded to a 120s harness
terminate rather than a failure.

One placement note: the held-handler verdict is judged after the queued
payload has been handed to the actor. `hold_entered` publishes it strictly
earlier so nothing is weakened, but asserting before the send lets a failure
unwind through the live `DiscardPanicking`, whose panicking destructor turns
the failure into a SIGABRT.

**5 (P3) — both parts.** `system.wait()` is wrapped in
`tokio::time::timeout(POLL_TIMEOUT, ...)` like its flood and cycle siblings,
so a dropped completion fails instead of parking. `ZeroDeadlineObservation`
carries `actor_task` and `continuation_task` as `tokio::task::Id` and the
test compares them, dropping the `Debug`-string comparison this repo treats
as non-contractual; a failure now names both tasks.

**6 (P3) — skips made loud on supported targets.** `eprintln!` alone is
invisible: nextest prints captured output for failing tests only, so a
skipped boundary reported as a plain pass. Both sites now call
`skip_unrepresentable_tick_boundary`, which is `#[cfg(target_os = "linux")]
panic!` and `eprintln!` elsewhere. CI is `ubuntu-latest` only, and Linux's
`Instant` is a `timespec` whose seconds field spans the u64-millisecond
boundary, so on the supported target the skip is a defect, not a platform
fact. Targets that count nanoseconds in a `u64` keep the diagnostic skip.
Written as a `cfg`-gated helper rather than `assert!(!CONST)` because clippy
rejects the latter under `assertions_on_constants`. The two tests pass on
Linux, which proves the boundary is exercised rather than skipped.

**Tests** — `cargo nextest run -p shelterwood -p shelterwood-runtime
--test-threads=4`: **539 run, 539 passed, 0 skipped**, repeated four times.
`cargo clippy -p shelterwood -p shelterwood-runtime --all-targets --
-D warnings` clean; `cargo +nightly fmt --all` applied.

**Unrelated flake seen once.** On one run,
`driver::tests::shutdown::runtime_teardown_folds_an_arrived_factory_disposal_panic`
(untouched here, from #292) failed under heavy machine load. It then passed
200/200 in a direct stress loop and in every subsequent suite run. Flagging
it rather than chasing it in this PR.

Issue #266, opened 2026-08-16
Hostile wakers II: multi-item teardown loops stop at the first resumed waker panic


ralexstokes:
Part of the hostile-waker containment cluster from the 2026-08-16 review round (external claim, verified per site).

## The problem

Each `with_observation_gate` call correctly unlocks and then **resumes** an accumulated waker panic to its caller (`ObservationTxn::commit` → `PanicAccumulator::Drop` → `resume_preferred_panic`, which resumes when the thread is not already unwinding). Multi-item teardown loops do not re-contain that resume, so one hostile waiter aborts the rest of the loop. Verified sites:

- **`Drop for ScopeRuntime`** (`driver.rs:345`, loop at `:357-373`, epilogue `:376-396`) — the worst one. Panic escape vectors per iteration: `mailbox.freeze/close` (effects flush resumes), `active.shutdown.fire()` / `active.abort.fire()` (raw `notify_waiters`, no containment), `child.terminality.discharge()` (gate txn resume). A panic at child *k*: later children never get shutdown/abort fired or `abort_handle.abort()` — **orphan tasks keep running uncancelled** (no drop-glue fallback aborts them); `clear_residents` is skipped; and per the code's own comment at `driver.rs:381-384`, epoch retirement has no other owner — **permanently stranded**, `wait_stopped` waiters hang. (Child terminality is the one thing recovered, via `Obligation` drop glue with contained commits.)
- **`BuilderCore::terminalize`** (`plan.rs:431`, loop at `:437-440`), reached from `Drop for BuilderCore` on a normal drop of an armed never-lowered builder. Hostile wakers are installable pre-lowering (`wait_terminal`/`wait_stopped` on declaration-minted refs). First panic skips later slots plus `root.terminalize_never_started()` — permanent; no other owner terminalizes pre-runtime slots. Waiters on skipped members hang forever.
- **`terminalize_plan`** (`plan.rs:469`, loop at `:475-477` + root block `:478-484`), from `Drop for ScopePlan`. Same shape; a panic additionally skips residency withdrawal and root closure.
- **`RemovalResponses::complete`** (`admission_control.rs:40-44`) — different consequence: skipped senders are dropped by the `Vec` unwind, and tokio `Sender::drop` also wakes the receiver, so skipped removal waiters observe channel-closed (`None`) instead of `RemoveOutcome::Removed` — outcome lost, not a hang.

[...]

Issue #265, opened 2026-08-16
Hostile wakers I: tokio bulk notification strands co-waiters after one waker panic


ralexstokes:
Part of the hostile-waker containment cluster from the 2026-08-16 review round (external claim, verified against the pinned tokio 1.53.1 source). CLAUDE.md's own threat model names waker `wake`/`clone`/`drop` vtables as arbitrary user code; this issue is where that threat model and the wake infrastructure disagree.

## The problem

Tokio's bulk-notification primitives abandon remaining waiters after one waker panics, and every shelterwood containment layer sits *above* tokio — too late to restore the lost wakeups:

- `Notify::inner_notify_waiters` (tokio-1.53.1 `sync/notify.rs:744-830`): on a wake panic, `WakeList`'s drop guard **drops** the remaining in-batch wakers without waking them (`wake_list.rs:38-73`), and `NotifyWaitersList::Drop` unlinks the rest marked `Notification::All` with the explicit comment "We do not wake the waiters to avoid double panics" (`notify.rs:359-373`). Net: every remaining waiter is marked-notified-but-unwoken — a later poll returns `Ready`, but no wake is ever delivered. A task parked solely on that waiter never runs again.
- watch: `Sender::notify_rx` loops over 8 `Notify` shards with no panic protection (`watch.rs:406-410`) — a panic in shard *i* skips shards *i+1..7* entirely (those waiters are not even marked).

Shelterwood surfaces feeding these waiter sets with **user-installable wakers** (public futures pollable with a custom waker):

- `Latch::fired` / `Latch::notify` (`shelterwood-runtime/src/sync.rs:118-137`) — and `CancellationToken::cancelled()` is built from the very same `shutdown`/`abort` latches the driver fires (`task.rs:44-45`; fire sites `driver.rs:365-366, 757, 1040`, `driver/child.rs:800-814`). One latch's waiter set therefore mixes user wakers with the framework's own child shutdown observer (`driver.rs:989/996`).
- watch pulses (`sync.rs:480-482` `send_modify`) — `wait_started`, `wait_stopped`, `wait_terminal`, snapshot/lifecycle subscriptions all park here.

## Concrete failure scenarios

- A hostile waker panicking during a latch fire strands all other waiters of that latch, including the framework's shutdown observer — a child solely awaiting `cancelled()` never observes shutdown → hang.
[...]

PR #297, opened 2026-08-19
Isolate payloads across panic-resuming waker effects

### LINKED ISSUE #267: Hostile wakers III: user payloads cross panic-resuming effect flushes unprotected
Part of the hostile-waker containment cluster from the 2026-08-16 review round (external claim, verified with ownership traces). No lock is held in either window — these are `dispose_detached`-convention deviations, not lock-rule violations — but a resumed waker panic destroys a user payload inline on the driver/caller thread, and a blocking payload destructor then blocks that thread mid-unwind.

## Site 1: child launch (confirmed)

`dispatch_child_construction` (`driver/child.rs:345-433`, called at `:632-635`) extracts the sole one-shot user payload into `SpawnBody` — stripping the `runtime::Isolated<ChildConstruction>` protection the same payload has while retained (`driver/child.rs:160`). `body` is then a naked local until `spawn_child_tasks` consumes it at `:673`, across **two** panic-resuming flush points:

- `transition_child_stage(..Starting..)` at `:645-653` — Started publication pulses subscriber wakers synchronously post-unlock in `ObservationTxn::commit`; a hostile observer waker panic is resumed to the caller and unwinds through `spawn_child`, dropping `body` inline on the driver thread.
- `mailbox.bind(incarnation)` at `:643` — on a rebind/restart, parked-**sender** wakers (arbitrary user-executor wakers) flush through `MailboxEffects::Drop` with the same resume-on-normal-path behavior. Same window, second vector.

Fix shape: keep the payload isolated (`Isolated`/equivalent) until all fallible publication and binding completes, releasing it only at the `spawn_child_tasks` handoff.

## Site 2: pre-acceptance call timeout (confirmed residual)

On the pre-acceptance timeout path, `CallFuture` promises "the call surface has no way to hand the recovered message back; route the discard through isolated disposal" (`futures.rs:649`). But `withdraw_send` flushes waker effects (`futures.rs:443`) while the recovered `M` sits in a live `SendError<M>` local; a waker-drop panic resumes and the unwind drops `M` inline on the polling task — bypassing the promised isolated disposal.

For `SendTimeout` this same ordering is the **documented, deliberate** trade (`futures.rs:439-442`: the message "already belongs to `result`", the waker is the caller's own) — no change needed there. The fix is scoped to the call path: `WithdrawalDisposition::Isolated` for the pre-acceptance `CallFuture` withdrawal, matching what `SendFuture::drop` already does for cancellation.

## Acceptance criteria

- [ ] The one-shot spawn payload survives (reaches isolated disposal) when a hostile observer waker panics during Started publication, and when a hostile parked-sender waker panics during rebind.
- [ ] The pre-acceptance `CallFuture` timeout routes the recovered message through isolated disposal on the waker-panic unwind, honoring its own comment.
- [ ] Hostile-waker regression tests for both windows.

## Related

- Hostile wakers I (bulk notification) and II (teardown loops) — same cluster
- #250 round — the accepted-then-cancelled teardown-doc lesson this repeats

ralexstokes:
Closes #267.

## What changed

- keep one-shot raw, task, and scope spawn bodies inside `runtime::Isolated` from construction extraction through mailbox bind and Started publication
- release the isolated body only at the `spawn_child_tasks` ownership handoff
- give call acceptance-timeout withdrawal an isolated disposition while preserving `SendTimeout`'s documented inline semantics
- add hostile Started-observer, rebind sender, and call-timeout waker regressions

## Root cause

Two post-unlock effect flush windows could resume arbitrary waker panics while a sole user payload was held as a plain local. Unwinding then destroyed that payload inline on the driver or polling task, violating the detached-disposal convention even though no framework lock remained held.

The payload now stays behind isolated disposal until each risky flush is complete. Call withdrawal likewise isolates its registered waker before recovering a message the call surface cannot return.

## Validation

- focused spawn-payload regressions (2 passed)
- hostile pre-acceptance call-timeout regression
- all mailbox futures tests (9 passed)
- pre-existing teardown/factory-disposal regressions
- `./tools/dev just ci`
- nightly formatting and `git diff --check`

## Stack

This is the base of the hostile-waker series. #266 will target this branch; #265 will target #266.

### DIFF
--- a/crates/shelterwood-mailbox/src/futures.rs
+++ b/crates/shelterwood-mailbox/src/futures.rs
@@ -153,6 +153,12 @@ impl<M: Send + 'static> ActorRef<M> {
     /// actor handler deadlocks because the blocked handler is also the only code
     /// that can produce the reply; use an actor-local continuation or an
     /// incarnation-owned offload instead.
+    ///
+    /// A timeout before acceptance has no caller to hand the request back to,
+    /// so it destroys both the recovered request and the waker it registered
+    /// through isolated disposal, as cancelling a parked [`send`](Self::send)
+    /// does. A panicking waker destructor is contained there rather than
+    /// resuming on the awaiting task.
     pub fn call<T: Send + 'static>(
         &self,
         make_msg: impl FnOnce(Reply<T>) -> M + Send + 'static,
@@ -428,6 +434,13 @@ impl<M> fmt::Debug for SendTimeout<M> {
     }
 }
 
+/// Withdraws an unaccepted send and classifies its outcome under the
+/// caller's chosen waker disposition.
+///
+/// The outcome is taken before `finish()` releases the registered waker, so
+/// under either disposition the recovered message already belongs to
+/// `result` and a hostile waker destructor can neither divert nor destroy
+/// it.
 fn withdraw_send_with<M: Send + 'static>(
     send: &mut SendFuture<M>,
     disposition: WithdrawalDisposition,
@@ -454,12 +467,11 @@ fn withdraw_send_with<M: Send + 'static>(
 }
 
 fn withdraw_send<M: Send + 'static>(send: &mut SendFuture<M>) -> Result<Incarnation, SendError<M>> {
-    let result = withdraw_send_with(send, WithdrawalDisposition::Inline);
     // Unlike the cancellation path this is a normal return on the caller's own
-    // task, so the caller's waker destructor stays inline and its panic stays
-    // visible; no core lock is held and the recovered message already belongs
-    // to `result`.
-    result
+    // task, and the recovered message goes back to that same caller. No core
+    // lock is held, so the caller's waker destructor stays inline and its
+    // panic stays visible.
+    withdraw_send_with(send, WithdrawalDisposition::Inline)
 }
 
 impl<M: Send + 'static> DeadlineOperation for TimedSend<M> {
--- a/crates/shelterwood-mailbox/src/futures.rs
+++ b/crates/shelterwood-mailbox/src/futures.rs
@@ -428,9 +428,12 @@ impl<M> fmt::Debug for SendTimeout<M> {
     }
 }
 
-fn withdraw_send<M: Send + 'static>(send: &mut SendFuture<M>) -> Result<Incarnation, SendError<M>> {
+fn withdraw_send_with<M: Send + 'static>(
+    send: &mut SendFuture<M>,
+    disposition: WithdrawalDisposition,
+) -> Result<Incarnation, SendError<M>> {
     let actor_id = send.mailbox.actor_id.clone();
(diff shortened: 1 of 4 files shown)


ralexstokes (commented):
Reviewed the immediate `main` → `agent/issue-267` diff and its behavior in the full hostile-waker stack. I found no actionable defects. The one-shot payload remains isolated across mailbox-bind and Started-publication effects, and call timeout withdrawal preserves `SendTimeout`'s inline semantics while isolating call-only disposal. Targeted mailbox and driver payload regressions passed locally; the PR CI check is green.

ralexstokes:
Review follow-up (58f321a): two documentation fixes, no behaviour change.

**1. The propagation change is now documented.** Before this PR a `CallFuture` that timed out pre-acceptance destroyed the registered waker inline (`WakerAction::DropInline`), so a panicking waker `drop` propagated to the awaiting caller. With `WithdrawalDisposition::Isolated` it becomes `WakerAction::Dispose`, runs inside `DisposalJob::finish`'s `catch_panic` on a blocking worker, and is discarded. That is the right trade — it is what `SendFuture::Drop` already does — but it was recorded nowhere. Added to the `ActorRef::call` rustdoc:

> A timeout before acceptance has no caller to hand the request back to, so it destroys both the recovered request and the waker it registered through isolated disposal, as cancelling a parked [`send`](Self::send) does. A panicking waker destructor is contained there rather than resuming on the awaiting task.

One correction to the review that prompted this: the cancellation counterpart is *not* an existing user-facing note — it is only the code comment in `SendFuture::Drop`. The public API had no rustdoc on this at all, so this sentence is the first. Left SPEC §13.18 alone: its closing paragraph already scopes itself out of "where the destructor then runs", and it is a test-anchor list rather than an API-surface record.

Also worth noting in the PR body if it gets updated: this adds a second blocking-pool submission per pre-acceptance call timeout (message and waker are separate disposal jobs).

**2. Stale comment moved.** `withdraw_send` is now a one-line wrapper, but it kept the paragraph explaining the `finish()`-after-`result` ordering. That ordering now lives in `withdraw_send_with` and holds for *both* dispositions, so leaving the rationale on the `Inline` wrapper reads as if it were a property of that disposition. Moved the ordering half onto `withdraw_send_with` as a doc comment:

> The outcome is taken before `finish()` releases the registered waker, so under either disposition the recovered message already belongs to `result` and a hostile waker destructor can neither divert nor destroy it.

`withdraw_send` keeps only its own `SendTimeout`-specific justification for `Inline` (normal return on the caller's own task, message goes back to that caller, no core lock held, so the panic stays visible).

**Checks:** `cargo check -p shelterwood-mailbox` clean; `RUSTDOCFLAGS=-D warnings cargo doc -p shelterwood-mailbox --no-deps` clean; `cargo +nightly fmt --all` no diff. No SPEC.md change, so no SPEC-pinning test applies.

PR #298, opened 2026-08-19
Finish teardown after hostile waker panics

### LINKED ISSUE #266: Hostile wakers II: multi-item teardown loops stop at the first resumed waker panic

ralexstokes:
Closes #266.

## What changed

- add one outer `PanicAccumulator` to `ScopeRuntime::drop`, preserving all child cancellation, terminality, residency, dynamic-removal, child-clear, and epoch-retirement work before resuming the first panic
- contain every slot and root terminalization in `BuilderCore` and `ScopePlan` teardown
- separate never-started publication from identity eviction so a hostile publication wake cannot skip eviction
- contain every `RemovalResponses::complete` send independently
- keep arrived-disposal draining compatible with contained teardown
- add one hostile-waker regression for each loop, including two simultaneously hostile removal receivers

## Root cause

Each individual transaction correctly unlocked before resuming a waker panic, but the surrounding multi-item teardown loops had no outer containment boundary. The first hostile waiter could therefore skip later children, root closure, resident clearing, epoch retirement, or removal outcomes.

Each independent teardown action now runs to completion; only then does the first accumulated panic resume.

## Validation

- focused teardown-loop regressions, with all verdicts judged outside `catch_unwind`
- complete `shelterwood` library suite (186 passed)
- `./tools/dev just ci` (681 nextest tests plus clippy, formatting, doctests, docs, API reachability, external consumer, and Nix checks)
- `git diff --check`

One inherited arrived-disposal fixture was corrected to wait for the real child join before invoking its synthetic post-join handler, matching the production invariant.

## Stack

Targets #297 (`agent/issue-267`). #265 will target this branch.

### DIFF
--- a/crates/shelterwood/src/driver/shutdown.rs
+++ b/crates/shelterwood/src/driver/shutdown.rs
@@ -228,7 +228,8 @@ impl ScopeRuntime {
         // though its disposal event has not reached ordinary dispatch. Fold
         // every arrived completion before the hard-force fallback; the drain
         // is non-blocking, so still-running disposal remains detached.
-        self.drain_arrived_disposal_events();
+        let mut panics = runtime::PanicAccumulator::default();
+        self.drain_arrived_disposal_events(&mut panics);
         if self.supervisor.is_disposing(key) {
             // The incarnation has already exited; only its retained factory
             // remains, and the fold above found no completion reported for
--- a/crates/shelterwood/src/driver/shutdown.rs
+++ b/crates/shelterwood/src/driver/shutdown.rs
@@ -235,7 +235,7 @@ impl ScopeRuntime {
             // remains, and the fold above found no completion reported for
             // it. Hard escalation detaches that cleanup and keeps the
             // recorded verdict.
-            self.handle_construction_disposed(key, None);
+            panics.run(|| self.handle_construction_disposed(key, None));
         }
     }
 
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -347,16 +347,19 @@ struct ScopeCompletion {
 
 impl Drop for ScopeRuntime {
     fn drop(&mut self) {
-        let dynamic_entries = if let Some(dynamic) = &self.dynamic {
-            let entries = self.root.with_observation_gate(|txn| {
-                let entries = dynamic.close(&self.root, txn);
-                self.root.set_dynamic_route_locked(None, txn);
-                entries
+        let mut panics = runtime::PanicAccumulator::default();
+        let mut dynamic_entries = None;
+        if let Some(dynamic) = &self.dynamic {
+            panics.run(|| {
+                self.root.with_observation_gate(|txn| {
+                    // Retain the entries before the transaction flushes its
+                    // wakes. If one is hostile, removal completion must still
+                    // remain ordered after terminality and residency cleanup.
+                    dynamic_entries = Some(dynamic.close(&self.root, txn));
+                    self.root.set_dynamic_route_locked(None, txn);
+                });
             });
-            Some(entries)
-        } else {
-            None
-        };
+        }
         // An exited child can be waiting only for retained user construction
         // to finish disposal. Its exit is already classified, so driver death
         // must publish that verdict before the terminality fallback gets a
@@ -366,7 +369,7 @@ impl Drop for ScopeRuntime {
         // publication. A completion that has already been reported is
         // available without waiting, however, so fold everything reported
         // before falling back to the stored verdict.
-        self.drain_arrived_disposal_events();
+        self.drain_arrived_disposal_events(&mut panics);
         let child_keys: Vec<_> = self.children.iter().map(|(key, _)| key).collect();
         for key in child_keys {
             if self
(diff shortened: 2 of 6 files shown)


crates/shelterwood/src/driver/shutdown.rs
```diff
@@ -228,7 +228,8 @@ impl ScopeRuntime {
         // though its disposal event has not reached ordinary dispatch. Fold
         // every arrived completion before the hard-force fallback; the drain
         // is non-blocking, so still-running disposal remains detached.
-        self.drain_arrived_disposal_events();
+        let mut panics = runtime::PanicAccumulator::default();
+        self.drain_arrived_disposal_events(&mut panics);
```
ralexstokes, line 232: [P2] Keep the hard-force fallback inside this accumulator. `drain_arrived_disposal_events` can retain a hostile observer panic from one completed child, then the `handle_construction_disposed(key, None)` call below can independently panic while publishing this child's terminal state. Because that call is direct, its unwind drops `panics` while already panicking, so the earlier (documented first/primary) payload is discarded and the later panic escapes. Wrapping the fallback in `panics.run(...)` preserves first-panic precedence while still completing both terminalizations; a two-child hostile-terminal-observer regression would pin it.

ralexstokes (commented):
Reviewed the immediate `agent/issue-267` → `agent/issue-266` diff and its interaction with the surrounding stack. The teardown-loop containment is generally sound, but `force_child` leaves one post-drain terminalization outside the accumulator, which can replace the first retained panic. The full `shelterwood` library suite passed locally (186 tests); the PR CI check is green.

crates/shelterwood/src/driver/shutdown.rs
ralexstokes: Fixed in `925e346`. The hard-force fallback now runs through the existing `PanicAccumulator`, and the new two-child regression verifies that both children terminalize while the first hostile observer panic remains primary. `./tools/dev cargo test -p shelterwood --lib` passes 187/187.

PR #299, opened 2026-08-19
Consolidate driver test fixtures

ralexstokes:
Part of #282 and the #284 test-debt sequence.

## What changed

- Introduces the canonical `ScopeRuntime` construction path used by tests.
- Replaces repeated ordered-plan setup, dynamic admission setup, and event-wait loops with shared fixtures.
- Makes `ChildArena::insert` return its key directly.
- Preserves epoch transfer and route-before-`Added` publication ordering.

Net change: 247 lines removed across nine files.

Validation: driver tests passed 138/138, and `./tools/dev just ci` passed (674 tests plus all documentation/API/external-consumer lanes).

Stack: based on #296 / issue #281 so fixture cleanup cannot hide weak assertions.

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -55,7 +55,7 @@ use crate::{
     plan::{
         BuilderCore, ChildConstruction, ChildPlan, LowerError, ScopeFactory, ScopePlan, SlotCell,
     },
-    policy::{DefaultsInheritance, ResolvedDefaults, ScopeFlavor},
+    policy::{DefaultsInheritance, Intensity, ResolvedDefaults, ScopeFlavor},
     raw::{CatchUnwindFuture, RawRunContext, RawSpawn},
     runtime::{self, CompletionGatedLatch, Latch},
     task::{TaskContext, TaskContextLatches, TaskFactory},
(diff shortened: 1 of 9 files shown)


ralexstokes (commented):
Reviewed the exact `agent/issue-281...agent/issue-282-driver` diff. In particular, the `ScopeRuntime` construction refactor keeps epoch transfer last, installs the keyed dynamic route before `Added` publication, and preserves the fixture transition semantics. I found no actionable issues. GitHub CI is passing.

ralexstokes:
Merged `main` (2586bd9) into this branch as d638cb0 and adapted the call sites `main` added while this PR was open. Full local `just ci` is green (692 tests, exit 0).

What the integration adapted:

**1. `ScopeRuntimeWiring` vs #298 (`Separate supervisor results from effects`).** `main` deleted `SupervisorEffect::Admitted` and replaced the step-then-pop admission with `supervisor::admit`, which returns `Option<ChildKey>` directly. This PR had just moved `ScopeRuntime` construction behind `ScopeRuntime::new(ScopeRuntimeWiring, ScopeEpochGuard)`. Resolution: dropped the now-vestigial `supervisor_effects` field from `ScopeRuntimeWiring` (it can only ever be empty at construction now — `new` initialises it to `Vec::new()`), and rewrote `ScopeRuntime::for_test` to admit through `supervisor_admit` instead of `supervisor_step` + popping `Admitted`. Both sides keep their intent: admission stays a synchronous result, and construction stays behind the single wiring seam.

**2. `ChildArena::insert` at the new `main` call sites.** #306 and #297 added three fixtures still written against the old `Result<ChildKey, _>` return and the old `epoch: Epoch` builder parameter. Per this PR — the fixture arena cannot take the error branch — they were adapted forward, not reverted:
- `driver/tests/terminalization.rs::scope_runtime_drop_finishes_every_child_and_epoch_after_a_hostile_wake` (#306) and `driver/tests/shutdown.rs::hard_force_preserves_the_first_terminal_observer_panic_across_its_fallback` (#306) now build through `OrderedScopeFixture`, matching the sibling fixtures this PR already converted. Their hand-rolled `begin_incarnation` / `set_admitted_children` / `plan.finish_transfer()` prologue is exactly what the fixture already does — `publish_initial_children` publishes the same resident set in the same position.
- `driver/tests/payload.rs::one_shot_raw_scope` (#297) moves to `ScopeEpochGuard::begin` + `with_transferred_plan`.
- The terminalization test asserts `root.settled(Some(epoch))` after the scope is dropped, so it needs the raw epoch the guard transfers. Added `OrderedScopeFixture::epoch()` for that (the same `epoch.epoch()` idiom already used at terminalization.rs:140). Verified the converted fixture still admits both children (temporary arity assert, since an empty fixture would have made several of the loops vacuous).

**3. `shelterwood-mailbox/src/futures.rs` — pre-existing break on `main`, fixed here.** This one is not caused by the merge: `origin/main` at 2586bd9 does not compile `cargo check -p shelterwood-mailbox --all-targets` on its own. The #296 merge reshaped `DisposalThread` into the channel-based recorder but left `drop_panicking_drop_waker` / the `call_acceptance_timeout_*` test on the old `Arc<Mutex<Option<ThreadId>>>` shape. Fixed by routing the hostile call-waker probe through `record_disposal` and cloning `DisposalThread` directly. This preserves #296's channel handshake rather than restoring polling — mutation-checked: removing the `record_disposal` call makes `call_acceptance_timeout_isolates_message_before_hostile_waker_drop` fail on the `await_disposal` timeout.

Note: `origin/agent/fix-disposal-thread-integration` carries a byte-identical fix for (3); if that lands on `main` first the merge is a no-op.

No production semantics changed — everything above is `#[cfg(test)]` fixture code plus the two test-only wiring seams (`ScopeRuntimeWiring`'s dropped field, which was write-only, and `for_test`).

PR #300, opened 2026-08-19
Harden framework lock boundaries structurally

### LINKED ISSUE #272: Lock-rule hardening: make the remaining convention-held sites structural

ralexstokes:
Part of the #284 hardening sequence. Closes #272.

## What changed

- Makes watch debugging metadata-only and moves mailbox exhaustion panics/effects beyond framework locks.
- Threads mailbox effects through observation transactions and adds a mailbox-bound, single-use bind token.
- Defers resident retirement and lifecycle payload release past the observation gate.
- Makes observation closure retry-safe, interval charge policy self-contained, fallback disposal poison-tolerant, and raw receive shutdown freezing local.
- Documents the last-handle mailbox payload protocol and updates the lock-rule guidance.

## Evidence

Added targeted regressions for hostile Debug, all mailbox exhaustion paths, effect deferral, resident payload retirement, lifecycle exhaustion, disposal rejection/poisoning, bind-token ordering, and local raw freezing.

Validation: targeted crate suites and `./tools/dev just ci` passed (687 tests plus formatting, clippy, doctests, docs, API reachability, external consumer, and Nix formatting).

Stack: based on #294 / issue #269 because both touch terminalization lock boundaries.

### DIFF
--- a/crates/shelterwood-mailbox/src/cell.rs
+++ b/crates/shelterwood-mailbox/src/cell.rs
@@ -3,12 +3,16 @@ use std::{
     fmt,
     num::NonZeroUsize,
     ops::{Deref, DerefMut},
-    sync::{Arc, Mutex, MutexGuard, atomic::Ordering},
+    sync::{
+        Arc, Mutex, MutexGuard,
+        atomic::{AtomicBool, Ordering},
+    },
 };
 
 use crate::{
-    ChildId, Incarnation, MailboxControl, MailboxDisposal, MailboxRuntime, MailboxSignal,
-    MailboxSignalWatcher, MailboxTermination,
+    ChildId, Incarnation, MailboxBindToken, MailboxClose, MailboxControl, MailboxDisposal,
+    MailboxEffectQueue, MailboxEffectSink, MailboxRuntime, MailboxSignal, MailboxSignalWatcher,
+    MailboxTermination,
     capability::{dispose, dispose_value},
     identity::{AtomicPoisonedCounter, PoisonedCounter},
     panic::{PanicAccumulator, PanicPayload, resume_panic},
@@ -319,6 +323,7 @@ impl<M> SendOperation<M> {
 
 pub(super) struct MailboxState<M> {
     kind: Option<MailboxKind>,
+    bind_permit: Arc<AtomicBool>,
     binding: MailboxBinding<M>,
     last_bound: Option<Incarnation>,
     pub(super) queue: VecDeque<Envelope<M>>,
@@ -335,12 +340,12 @@ impl<M> MailboxState<M> {
         }
     }
 
-    fn park(&mut self, operation: &Arc<SendOperation<M>>) {
+    fn park(&mut self, operation: &Arc<SendOperation<M>>) -> bool {
         match &mut self.binding {
             MailboxBinding::Unbound(waiters)
             | MailboxBinding::Frozen { waiters, .. }
             | MailboxBinding::Bound(BoundState::Full { waiters, .. }) => {
-                waiters.park(operation);
+                return waiters.park(operation);
             }
             MailboxBinding::Bound(BoundState::Available(incarnation)) => {
                 let Some(MailboxKind::Queue(capacity)) = self.kind else {
@@ -349,7 +354,9 @@ impl<M> MailboxState<M> {
                 debug_assert_eq!(self.queue.len(), capacity.get());
                 let incarnation = *incarnation;
                 let mut waiters = WaiterQueue::default();
-                waiters.park(operation);
+                if !waiters.park(operation) {
+                    return false;
+                }
                 self.binding = MailboxBinding::Bound(BoundState::Full {
                     incarnation,
                     waiters,
@@ -359,6 +366,7 @@ impl<M> MailboxState<M> {
                 unreachable!("terminal submissions return their payload directly")
             }
         }
+        true
     }
 
     /// Detaches every waiter from a non-terminal binding before a transition.
@@ -426,6 +434,23 @@ pub(super) enum Submission<M> {
     },
 }
 
+enum SubmitTransition<M> {
+    Complete(Submission<M>),
+    WaiterIdentityExhausted(Arc<SendOperation<M>>),
+    AcceptedSequenceExhausted(M),
+}
+
+enum AcceptTransition<M> {
+    Accepted(Incarnation),
+    Full(M),
+    Exhausted(M),
+}
+
+enum TrySendTransition<M> {
+    Complete(Result<Incarnation, SendError<M>>),
+    AcceptedSequenceExhausted(M),
+}
+
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 struct WaiterId(u64);
 
@@ -470,20 +495,19 @@ impl<M> WaiterQueue<M> {
         self.entries.len()
     }
 
-    fn push_back(&mut self, operation: Arc<SendOperation<M>>) -> WaiterId {
-        let next = WaiterId(
-            self.ids
-                .mint()
-                .expect("mailbox waiter identity space exhausted"),
-        );
+    fn push_back(&mut self, operation: Arc<SendOperation<M>>) -> Option<WaiterId> {
+        let next = WaiterId(self.ids.mint()?);
         let replaced = self.entries.insert(next, operation);
         debug_assert!(replaced.is_none());
-        next
+        Some(next)
     }
 
-    fn park(&mut self, operation: &Arc<SendOperation<M>>) {
-        let registration = self.push_back(Arc::clone(operation));
+    fn park(&mut self, operation: &Arc<SendOperation<M>>) -> bool {
+        let Some(registration) = self.push_back(Arc::clone(operation)) else {
+            return false;
+        };
         operation.register(registration);
+        true
     }
 
     fn observe_all(&self, incarnation: Incarnation) {
@@ -520,24 +544,41 @@ impl<M> WaiterQueue<M> {
 /// of it: it never outlives the `MailboxTxn` that owns it, and every mailbox
 /// transition — including the per-message receive path — would otherwise pay
 /// two atomic refcount pairs to restate what the transaction already holds.
-struct MailboxEffects<'a, M: Send + 'static> {
+struct MailboxEffects<'a, 's, M: Send + 'static> {
     cell: &'a MailboxCell<M>,
+    external: Option<&'s mut dyn MailboxEffectSink>,
     pulse: bool,
     displaced: Vec<Envelope<M>>,
     isolate_displaced: bool,
     wakers: WakerEffects,
     returned: Option<M>,
+    accepted_sequence_exhausted: bool,
 }
 
-impl<'a, M: Send + 'static> MailboxEffects<'a, M> {
+impl<'a, 's, M: Send + 'static> MailboxEffects<'a, 's, M> {
     fn new(cell: &'a MailboxCell<M>) -> Self {
         Self {
             cell,
+            external: None,
             pulse: false,
             displaced: Vec::new(),
             isolate_displaced: false,
             wakers: WakerEffects::default(),
             returned: None,
+            accepted_sequence_exhausted: false,
+        }
+    }
+
+    fn deferred(cell: &'a MailboxCell<M>, external: &'s mut dyn MailboxEffectSink) -> Self {
+        Self {
+            cell,
+            external: Some(external),
+            pulse: false,
+            displaced: Vec::new(),
+            isolate_displaced: false,
+            wakers: WakerEffects::default(),
+            returned: None,
+            accepted_sequence_exhausted: false,
         }
     }
 
@@ -550,23 +591,56 @@ impl<'a, M: Send + 'static> MailboxEffects<'a, M> {
     }
 }
 
-impl<M: Send + 'static> Drop for MailboxEffects<'_, M> {
+impl<M: Send + 'static> Drop for MailboxEffects<'_, '_, M> {
     fn drop(&mut self) {
+        let batch = MailboxEffectBatch {
+            changed: Arc::clone(&self.cell.changed),
+            runtime: Arc::clone(&self.cell.runtime),
+            pulse: self.pulse,
+            displaced: std::mem::take(&mut self.displaced),
+            isolate_displaced: self.isolate_displaced,
+            wakers: std::mem::take(&mut self.wakers),
+            returned: self.returned.take(),
+            accepted_sequence_exhausted: self.accepted_sequence_exhausted,
+        };
+        if let Some(external) = self.external.take() {
+            external.defer_mailbox_effect(Box::new(move || {
+                batch.flush();
+            }));
+            return;
+        }
+        batch.flush();
+    }
+}
+
+struct MailboxEffectBatch<M> {
+    changed: Arc<dyn MailboxSignal>,
+    runtime: Arc<dyn MailboxRuntime>,
+    pulse: bool,
+    displaced: Vec<Envelope<M>>,
+    isolate_displaced: bool,
+    wakers: WakerEffects,
+    returned: Option<M>,
+    accepted_sequence_exhausted: bool,
+}
+
+impl<M: Send + 'static> MailboxEffectBatch<M> {
+    fn flush(mut self) {
         let mut panics = PanicAccumulator::default();
         if self.pulse {
-            panics.run(|| self.cell.changed.pulse());
+            panics.run(|| self.changed.pulse());
         }
         // Binding a latest-value mailbox historically handed displaced
         // payloads to isolated disposal before accepted senders were woken.
         // Preserve that observable ownership edge: a woken sender may run
         // immediately and must not race ahead of disposal submission.
         if self.isolate_displaced && !self.displaced.is_empty() {
-            let displaced = std::mem::take(&mut self.displaced);
+            let isolated = std::mem::take(&mut self.displaced);
             panics.run(|| {
                 dispose_value(
-                    self.cell.runtime.as_ref(),
+                    self.runtime.as_ref(),
                     MailboxPayload {
-                        queue: Some(displaced.into()),
+                        queue: Some(isolated.into()),
                         latest: None,
                         retired: Vec::new(),
                     },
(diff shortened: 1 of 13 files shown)


ralexstokes (commented):
Reviewed the complete immediate-base diff and cross-stack behavior. I found one actionable panic-containment issue in the new accepted-sequence promotion exhaustion path; details are inline. The rest of the mailbox transaction/effect-sink changes, resident/lifecycle retention, observation closure, runtime disposal, bind token, and raw receive changes looked consistent with the lock rule. Targeted validation passed: mailbox (29), runtime (38), cells (28), and façade lib (184) tests.

crates/shelterwood-mailbox/src/cell.rs
```diff
@@ -1293,14 +1442,21 @@ fn promote_waiter_queue<M: Send + 'static>(
...
         MailboxKind::Queue(capacity) => capacity.get().saturating_sub(queue.len()),
         MailboxKind::Latest => usize::MAX,
     };
     let mut accepted = 0usize;
     while accepted < available {
+        if waiters.is_empty() {
+            break;
+        }
+        let Some(accepted_sequence) = mint_accepted_sequence(accepted_sequence) else {
+            effects.accepted_sequence_exhausted = true;
```
ralexstokes, line 1513: [P2] Contain the received message before resuming this panic

When this runs from `MailboxCell::receive`, the just-popped message is already in `effects.returned`. `finish_returned` takes that value into its local `output` and then drops the transaction; this flag makes the effect batch resume the exhaustion panic while `output` is still live. Unwinding therefore drops the user message outside `PanicAccumulator`. A panicking message destructor causes a second panic and aborts the process, and a blocking destructor runs inline instead of through mailbox disposal. Please move/retain the returned value in a contained or isolated effect before raising the exhaustion panic (and cover the promotion case with a hostile `Drop`), rather than leaving it as the return local during unwind.

ralexstokes: Fixed in `3e3f500`. `finish_returned` now protects the received value with a disposal guard before flushing transaction effects; if exhaustion or another effect panics, the value is submitted for isolated disposal rather than being dropped as a live unwind local. The promotion-exhaustion regression verifies both disposal submission and that the message destructor observes the mailbox unlocked. `./tools/dev cargo test -p shelterwood-mailbox` passes 29/29; stacked PRs #304 and #314 were rebased onto this fix.

ralexstokes:
Applied the review. Three commits on top of `3e219ca`: `afe35d0`, `09a0b7b`, `14a252d`.

**1 (P2) — `close`'s unread payload across the effects flush.** Confirmed and fixed at the source rather than the call sites. `MailboxClose` now carries `Arc<dyn MailboxRuntime>` and holds its two halves in `Option`s; `Drop` routes an unconsumed `disposal` through the disposal capability, mirroring `ReturnedMessage`. `into_parts` takes both halves, so the happy path is unchanged and the ordering (pulse → wake → disposal submission) is preserved. Both callers — `finalize_child` and `ScopeRuntime::drop` — keep `closed` live across `drop(effects)` and now get the guard for free.

**2 (P2) — accepted-sequence exhaustion during `bind`.** Confirmed. `bind` now checks `effects.accepted_sequence_exhausted` before deriving its binding verdict and installs `Frozen { incarnation, waiters }`: no further acceptance is possible on that counter, the parked senders stay mailbox-owned instead of being destroyed under the guard, and the post-unlock effect raises the exhaustion panic as the primary diagnostic. `BoundState::Full`'s documented "only when a queue mailbox has reached capacity" invariant is left intact.

**3 (P3) — the "pinned" test failed open.** Confirmed. `RetainedExit::drop` never destroys the user error inline, so the `GateDropError` destructor runs on a worker thread in both arrangements and `gate.is_held()` only reported `true` when that worker beat the main thread out of `commit`. Rewrote it around the observable the change actually produces: hold the gate open across `emit_locked` and assert nothing reaches the probe channel inside that window. Nothing can arrive unless the submission already happened under the gate, so it fails closed.

**4 (P3) — `emit_locked`'s success path.** Confirmed; `wakes.defer(move || drop(event));` at the end. No dedicated test: on the success branch the published clones keep the shared `ExitError` alive, so only the submission — which has no hook in this crate — moves, and the destructor is unobservable.

**5 (P3) — `publish_terminal`'s `debug_assert!(false, …)`.** Confirmed; the verdict is now computed under the guard and raised after the block ends.

**6 (P3) — AGENTS.md seam list.** Added `MailboxEffectSink` with the same by-convention ruling as `MailboxRuntime`, noting that the framework calls `defer_mailbox_effect` under both the observation gate and `MemberCell::mailbox`, that `MailboxEffectQueue` is `pub` + `Default`, and that the façade re-exports neither (`pub(crate)` in `mailbox.rs`, absent for the trait) — which is what keeps the ruling true.

**7 (P3, suspected) — not reachable; left alone.** `child.mailbox_bind` can only be `None` at spawn if a prior `close` returned `None`, which needs the mailbox already `Terminal` at `finalize_child`. Every production terminalization of a driver-owned member runs through `ScopeRuntime::terminalize_child`, which `reduce`s `SupervisorEvent::Terminalized`; `spawn_child` returns early on `self.supervisor.membership_terminal(key)`, so no respawn can follow. The remaining terminalizers are `slot.terminalize_never_started*` (slots that never became driver children — rejected admissions, dropped plans) and the `ChildTerminality` obligation fallback, which fires during teardown. `reclaim_child` removes the entry outright. So the `.expect` cannot fire; the terminal-race no-op would be dead code.

**Mutation checks** (committed tree, then mutation, then revert):

- Finding 1 — replaced `MailboxClose::drop`'s `self.runtime.dispose(disposal)` with an inline `drop(disposal)`: `a_panicking_close_flush_isolates_the_unread_payload` FAILS on its own assertion (`left: ThreadId(2), right: ThreadId(2)` — the unread message destructor ran on the caller's thread). Restored → passes.
- Finding 2 — disabled the new exhaustion branch in `bind`: `bind_sequence_exhaustion_retains_parked_senders_after_unlock` FAILS, and the failure is exactly the reported shape — `unreachable!("a latest mailbox finishes all waiting submissions")` at `cell.rs:1295`, then `PoisonError` on the next `mailbox.state.lock()`. Restored → passes.
- Finding 3 — reverted `wakes.defer(move || drop(kind))` to `drop(kind)`: the new test FAILS in 0.002s on the in-gate assertion. With the same mutation, the *old* test body (`scope.emit(...)` + the destructor-thread probe) passed 5/5 — the fail-open claim reproduced.

**Tests** (`./tools/dev`, targeted):

- `cargo nextest run -p shelterwood-mailbox -p shelterwood-cells -p shelterwood` → **569 passed, 0 skipped**
- `cargo clippy -p shelterwood -p shelterwood-cells -p shelterwood-mailbox --all-targets` → clean
- `just fmt` → clean

Unrelated pre-existing flake seen while verifying: `driver::tests::shutdown::{runtime_teardown,hard_force}_folds_an_arrived_factory_disposal_panic` fails intermittently under the full `-p shelterwood` run. Reproduced on the unmodified branch tip (`3e3f500`) as well — 1 failure in 4 runs there, and it passes in isolation — so it is not from this change.

ralexstokes:
Integrated `main` (now at `d43516c`) in two merges: `f1c024c` covers #295/#297/#305/#296, `b23202f` covers #298/#306/#326 which landed while the first was being resolved. `cargo check --workspace --all-targets` was green after each, and `just ci` is green on the tip: fmt, clippy (`--all-targets --all-features` and `--lib`), 703/703 nextest, doctests, the API-reachability walk, packaged-docs sync, the external-consumer build and nixfmt.

**Per file, what survived from each side.**

- `crates/shelterwood-cells/src/cells.rs` — #305 deleted the `crate::{engine, exit, identity, mailbox, policy, runtime}` re-export shims in `shelterwood-cells/src/lib.rs`, so this branch's imports move onto the real crate paths: `MailboxEffectSink` now comes from `shelterwood_mailbox` alongside `MailboxControl`/`MailboxTermination`, and the two test-module `crate::runtime::mailbox_runtime()` calls became `shelterwood_runtime::mailbox_runtime()`. `observe::ScopeKind` is gone with #305 and was dropped from the import; `LifecycleEvent` is no longer named here because this branch routes publication through `RetainedLifecycleEvent`. The test module keeps this branch's additions (`LifecycleEventKind`, `ResidentProjection`, `MailboxCell`, `MailboxEffectQueue`, `MailboxReceiver`, `actor_ref_from_parts`, `ResolvedMailbox`) re-homed onto `shelterwood_core`/`shelterwood_mailbox`.
- `crates/shelterwood-mailbox/src/lib.rs` — kept #305's removal of `impl<T: ActorIdentity> ActorIdentity for Arc<T>` and this branch's `Arc` + `atomic::{AtomicBool, Ordering}` imports, which `MailboxBindToken` and `MailboxClose` need. `MailboxClose`'s `Drop`, `into_parts`, the `MailboxEffectSink` seam and `MailboxEffectQueue` are all intact.
- `crates/shelterwood-mailbox/src/cell.rs` — import composition only: #295's `time::Instant` plus this branch's `atomic::AtomicBool`. The `bind` exhaustion verdict (`effects.accepted_sequence_exhausted` → `Frozen { incarnation, waiters }`) and its test are unchanged.
- `crates/shelterwood/src/raw.rs` — #295 generified the fixture into `bound_raw_context_for<M>`; this branch needed the context's own shutdown latch. The generic builder now returns `(RawContext<M>, ActorRef<M>, Latch)`, `bound_raw_context()` stays the `u8` two-tuple convenience, and `recv_freezes_its_mailbox_when_it_observes_shutdown` calls `bound_raw_context_for::<u8>()`. #295's three new tests still use `bound_raw_context_for::<…>().0`.
- `crates/shelterwood/src/driver.rs` — the only conflict in the second merge. #306 wrapped every fallible step of `ScopeRuntime::drop` in a `PanicAccumulator`; this branch reshaped the same mailbox teardown around `MailboxEffectQueue`/`MailboxClose`. Composed: `freeze`, `close` and the effect flush each run under `panics.run`, so a panic in any one still leaves the others to run and still publishes the effects already collected, and the closed payload still reaches `dispose_detached` via `MailboxClose::into_parts`.
- `crates/shelterwood/src/driver/tests/payload.rs` (new in #296) drove the pre-token `MailboxControl` surface. Its prior-incarnation fixture now takes the child's `mailbox_bind` token, runs bind/freeze/close through a `MailboxEffectQueue`, and hands the token that `close` returns back to `ChildRuntime` so `spawn_child` still exercises the rebind edge rather than tripping the "configuration or close supplies each bind token" expectation.
- `AGENTS.md`, `crates/shelterwood-cells/src/observe.rs`, `crates/shelterwood-core/src/engine.rs`, `crates/shelterwood-mailbox/src/futures.rs`, `crates/shelterwood-runtime/src/{disposal.rs,sync.rs}`, `crates/shelterwood/src/{mailbox.rs,driver/child.rs}` — auto-merged. I diffed each against `origin/main` afterwards: the remaining delta is exactly this PR's contribution sitting on top of `main`'s version, including #297's `runtime::Isolated` one-shot bodies and #305's `intensity_trip()`/`is_closed`/`abort` changes.

**Nothing of #305's deletions was resurrected.** The adaptations forced by them were: the `crate::*` shim paths in `cells.rs`, `observe::ScopeKind`, and (already in this branch) `MailboxControl::bind_order_valid`, which the linear `MailboxBindToken` replaces structurally.

**One thing worth flagging.** `main` at `58d4946` did not build: #296 reworked the `DisposalThread` test fixture in `shelterwood-mailbox/src/futures.rs` into a channel-publishing struct while #297 independently added the panicking-drop waker probe against the old `Arc<Mutex<Option<ThreadId>>>` newtype, and the combination was never compiled — `probe.0.lock()` and three `Arc::clone` calls against a plainly-`Clone` type. I carried the same three-line fix in `f1c024c`; #326 landed the identical fix on `main` in the meantime, so the hunk merged with itself.

ralexstokes:
Re-integrated with `main` twice — `main` advanced again (PRs #299/#301/#302) while local CI was running, so this covers both rounds.

**Merge 1 — `46c1adc`** (into `07e8dce`, i.e. #307 + #315)

- `crates/shelterwood-cells/src/observe.rs` — kept **#307's** wrap-plus-guards `RetainedLifecycleEvent` (`event: LifecycleEvent` before `guards: Arc<Vec<RetainedExit>>`, the field-order argument intact) and its `into_public` retirement via `Arc::try_unwrap` + `drop(guard.into_exit())`. The mirrored `RetainedLifecycleEventKind` stays deleted rather than resurrected. Re-applied **#300's** `pub(crate)` visibility, `prepend_scope`, `From<LifecycleEvent>` and a `from_parts` constructor. #300 needed the retained form *before* the fallible sequence mint, which #307's `new(event)` cannot supply (no seq yet), so the retention `match` is split out unchanged as `RetainedLifecycleEvent::retain_guards(&kind)` and `new` calls it. **#307's exhaustive, wildcard-free `match` over `LifecycleEventKind` is untouched and still the single forcing point** — a new exit-bearing variant still fails to compile until its retention is declared.
- `crates/shelterwood-cells/src/cells.rs` (no conflict, but `main` deleted the type it referenced) — `emit_locked` now mints `guards` before the mint; the exhaustion path defers `drop(kind)` then `drop(guards)` so the guards are the last owner and destruction is submitted to isolated disposal off the observation gate. Same observable behaviour as before the merge; `lifecycle_sequence_exhaustion_retires_the_exit_after_unlock` still passes.
- `crates/shelterwood-runtime/src/sync.rs` — took **#315's** `Debug` bodies for `WatchSender`/`WatchReceiver` wholesale after establishing they already carry #300's hardening: both sides independently dropped the `T: fmt::Debug` bound and the `.field(\"value\", &*self.0.borrow())` that formatted a user value under the watch mutex. #315's versions additionally match its new layout (`self.receiver_count()`, `seen`). #300's two probe tests now assert the *omission* (`!rendered.contains(\"DebugProbe\")` + zero format calls) instead of an exact rendering, since `Debug` output is not contractual and #315 changed it. #315's `#[must_use]` wait structs and the `unwrap_or_else(PoisonError::into_inner)` value helper are untouched.
- `AGENTS.md` — no conflict; `main` had not touched it, so both sides' edits are present (`MailboxEffectSink` seam entry and the `clear_residents_locked`/`prune_child_locked` deferral note).
- `driver.rs` / `driver/child.rs` — auto-merged cleanly and verified by hand: #300's `MailboxBindToken` threading, `MailboxEffectQueue` plumbing and the `ScopeRuntime::drop` teardown (`panics.run` around freeze/close/flush, payload to `dispose_detached` via `into_parts`) coexist with #315's `framework_shutdown` latch and `fire_shutdown_edges`.

**Merge 2 — `821d423`** (into `65a9ea3`, the #282 test-fixture consolidation)

- `crates/shelterwood-mailbox/src/cell.rs` — adopted `main`'s `test_support::{mint_actor_incarnation, mint_actor_membership}` at every bind site while keeping #300's `configure`/`bind` token helpers and the `Incarnation` import. The two bind-order `should_panic` contract tests (`binding_before_configuration_trips_the_driver_contract`, `rebinding_before_close_trips_the_driver_contract`) stay **deleted**: `MailboxBindToken` makes both states unconstructible, so `main`'s fixture-rewritten versions have nothing left to assert. Also converted the fixture-shaped sites #300 added outside the conflict regions.
- `crates/shelterwood-mailbox/src/futures.rs` — `main`'s fixture import plus #300's `MailboxEffectQueue` and four-argument `MailboxControl::bind`.
- `crates/shelterwood-runtime/src/disposal.rs` — kept #300's `rejected_fallback_job_is_released_after_unlock` and `pending_query_tolerates_a_poisoned_disposal_job` (and the `enqueue_fallback_disposal_with` import they need); dropped the local `ReportingDrop` helper and the now-unused `Condvar` import that `main` relocated into `test_support`.

**Verification:** `just ci` green end to end on `821d423` — fmt, clippy (both feature lanes), `711 tests run: 711 passed, 0 skipped`, doc-check, runtime-API reachability (clean for all three crates), packaged-docs check, external-consumer check, nixfmt. No occurrence of the #325 flake in either run.

PR #301, opened 2026-08-19
Consolidate integration test fixtures

ralexstokes:
Part of #282 and the #284 test-debt sequence.

## What changed

- Adds shared gated recorders, lifecycle receive helpers, and startup-child exit extraction.
- Reuses signalled waiting helpers across readiness, lifecycle, and dynamic tests.
- Unifies unexpected lifecycle-lag handling.
- Replaces the remaining polling-style task wait with `assert_eventually!`.

Net change: 114 lines removed across the affected integration fixtures.

Validation: 251 affected integration tests passed; the stacked branch passed full serial `just ci`, and the final combined #282 stack passes ordinary `./tools/dev just ci` with 674/674 tests.

Stack: based on the driver-fixture PR.

### DIFF
--- a/crates/shelterwood/tests/acceptance_assistant.rs
+++ b/crates/shelterwood/tests/acceptance_assistant.rs
@@ -643,6 +643,11 @@ async fn assistant_control_plane_composes_nested_recovery_redelivery_streaming_a
     let mut saw_control_restart = false;
     let mut saw_tool_restart = false;
     let mut saw_gateway_restart = false;
+    // Lag here is a fixture defect, not load: this trace publishes a bounded
+    // number of edges — three restarts and one temporary child on top of a
+    // fixed tree — against the 128-event subscriber capacity, and nothing but
+    // this drain and the wait above reads it. A marker would mean dropped
+    // edges silently deciding the flags below, so it fails loudly instead.
     while let Ok(item) = lifecycle.try_recv() {
         let LifecycleItem::Event(event) = item else {
             panic!("the bounded assistant trace must not lag")
--- a/crates/shelterwood/tests/acceptance_assistant.rs
+++ b/crates/shelterwood/tests/acceptance_assistant.rs
@@ -6,12 +6,12 @@ use std::{
     time::Duration,
 };
 
-use crate::common::{POLL_TIMEOUT, ReleaseGate, assert_eventually, assert_quiet};
+use crate::common::{POLL_TIMEOUT, ReleaseGate, assert_eventually, assert_quiet, next_event};
 use shelterwood::{
     Actor, ActorDef, ActorOnceDef, ActorRef, ChildState, Context, DeadlineElapsed, DynamicScopeRef,
-    DynamicTree, ExitError, ExitResult, LifecycleEvent, LifecycleEventKind, LifecycleEvents,
-    LifecycleItem, Mailbox, Membership, RemoveOutcome, Reply, ReserveError, RestartCount,
-    RestartPolicy, ScopeState, StopContext, SubtreeOnceDef, Tree,
+    DynamicTree, ExitError, ExitResult, LifecycleEvent, LifecycleEventKind, LifecycleItem, Mailbox,
+    Membership, RemoveOutcome, Reply, ReserveError, RestartCount, RestartPolicy, ScopeState,
+    StopContext, SubtreeOnceDef, Tree,
 };
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender, unbounded_channel};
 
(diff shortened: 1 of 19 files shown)


ralexstokes (commented):
Reviewed the exact `agent/issue-282-driver...agent/issue-282-integration` diff, including recorder drain parity, lifecycle lag handling, startup-failure extraction, and the waiting-task substitutions. I found no actionable issues. GitHub CI is passing.

ralexstokes:
Applied the review findings in 4003dcf.

**1 (P2) — factory-time vs first-poll signal.** Confirmed: the inline definition
stored `sibling_started` in the factory closure body, which the driver runs
synchronously while starting the child; `signalled_waiting_task` stores at the
task future's first poll. With resolved-`Immediate` readiness publishing at
spawn, `wait_started()` can return before the sibling has been polled, so the
bare `assert!` became scheduler-dependent.

Took the second option, which preserves the original evidence exactly: added
`construction_signalled_waiting_task(constructed)` in `common/waiting.rs`, which
stores in the factory, and used it at `readiness.rs:1288`. The bare `assert!`
stands. The helper's doc says why the factory-time store is the point, so a
future substitution back to `signalled_waiting_task` has to argue with it.

Audited the other four `signalled_waiting_task` call sites for the same
dependency — `readiness.rs:574`, `readiness.rs:806`, `dynamic.rs:876` and
`lifecycle.rs`'s cancellation-only uses. All of them either poll the flag
through `assert_eventually!` or observe only the cancelled half, so none
depends on factory-time semantics. No other change needed.

Mutation-checked the assertion on the committed tree: dropping the store from
`construction_signalled_waiting_task` fails
`immediate_raw_construction_panic_classifies_post_ready`, so the flag is load-bearing
rather than decorative.

**2 (P3) — one-shot silently became restartable.** Confirmed and reverted:
`signalled_waiting_tree` builds its worker through `add_task_once` again, with
the completion claim dropped as before. To keep the consolidation, the shared
body moved into a private `signalled_waiting_body`, and
`signalled_waiting_task` / `liveness_probed_waiting_task` /
`signalled_waiting_once_task` are now three definitions over that one body — the
liveness branch included, so the duplicated probe block is gone too. Net effect:
the fixture's child kind is back to what `actor_and_subtree_slots_preserve_fused_and_split_drop_ownership`
was written against, and the file is shorter than before the revert.

**3 (P3) — first-match vs last-match exit scanning.** Added
`last_panic_message(events, id)` to `common/lifecycle.rs`: it drains the
lifecycle stream to closure, keeps the last `Panicked` exit for the id, treats a
lag marker as a fixture defect, and bounds the drain with `POLL_TIMEOUT` (the
old loops were unbounded, so a stream that stopped closing would have hung the
runner instead of failing it).

All five sites now use it — `offloads.rs` x3 and `raw.rs` x2 — so last-match-wins
and the implicit proof that the stream closes are both restored, and the three
offloads sites did not have to be adjudicated separately. `raw.rs`'s
`live_try_recv_resumes_a_retained_offload_panic` gains the `Panicked` filter it
lacked: its child is a `RawOnceDef`, so exactly one exit is reachable today, but
the filter is what makes that unambiguous rather than incidental. Its message
check is now `assert_eq!` on the drained message, which reports the wrong
message rather than just "matches! failed".

**4 (P3, judgement) — kept the strictness.** Lag on these streams is a function
of what the fixture publishes, not of machine load: nothing reads the
subscription concurrently, so a marker requires the fixture to emit more than
the 128-event per-subscriber capacity between two reads. For
`assistant_control_plane_composes_nested_recovery_redelivery_streaming_and_remount`
that budget is three restarts plus one temporary child on top of a fixed tree —
far short of it — and events dropped by a lag would silently decide the
`saw_*_restart` flags the drain exists to set. So degrading is the worse
behaviour here, and tolerance stays where the PR put it. Recorded the reasoning
at the shared `next_event` helper and at the assistant drain itself.

**Tests** (all via `./tools/dev`, `nice -n 10 CARGO_BUILD_JOBS=4`):

- `cargo nextest run -p shelterwood --test readiness --test dynamic --test offloads --test raw --test lifecycle --test acceptance_assistant --test-threads=4` — 122 passed, 0 failed.
- Every other binary that consumes the touched `common` modules:
  `--test observation --test deadlines --test actor --test disposal --test drain --test subtrees` — 111 passed, 0 failed.
- Finding 1 repeat run: `immediate_raw_construction_panic_classifies_post_ready`
  6x (1 + 5 repeats), 6 passed, 0 failed.
- `cargo clippy --tests -p shelterwood --all-features` clean.

Touched test files were formatted with `rustfmt +nightly --edition 2024
--config-path rustfmt.toml` directly, since `cargo fmt` does not reach
`crates/shelterwood/tests/**`.

PR #302, opened 2026-08-19
Consolidate library test fixtures

ralexstokes:
Completes the fixture-deduplication work for #282. Closes #282.

## What changed

- Shares the latest representable deadline helper between core and runtime tests without adding a production dependency.
- Centralizes runtime destructor/thread, blocking-pool, and saturated-shutdown fixtures.
- Centralizes mailbox actor-membership and incarnation minting fixtures.
- Keeps every helper behind test/test-util boundaries.

Validation: core, runtime, and mailbox suites passed independently. After stacking all three #282 commits, `./tools/dev just ci` passed with 674/674 tests plus formatting, clippy, doctests, docs, API reachability, external consumer, and Nix formatting.

Stack: based on the integration-fixture PR.

PR #303, opened 2026-08-19
Pin facade contracts and specification claims

### LINKED ISSUE #283: Coverage-gap batch: cells transaction core, panic containment, SPEC §18 claims
Coverage-gap batch from the 2026-08-16 review round. The hostile-waker regressions live with their own issues; this collects everything else, ordered roughly by value.

## The structural gap

- [ ] **shelterwood-cells is the transitive-only outlier**: of 2625 lines, in-crate tests cover only the destruction class. Zero in-crate tests for `ObservationTxn` defer/commit/unwind-flush (the doc claims at `cells.rs:946-961` — Drop flushes queued pulses during an unwind; one hostile pulse cannot suppress the rest), the parent-to-child `adopt_observation_gate` retry (the documented sole exception to gate-outermost), `ScopeCell` resident management, `attach_mailbox` double-attach rejection, the losing-terminalizer path, `publish_stopped_locked`'s precedence lattice, `take_shutdown_request`/`take_force_request` epoch semantics, and `RetainedExit::install`'s keep-in-place branch. Nothing in the crate fails if `ObservationTxn::commit` stops running on Drop.
- [ ] **`panic.rs` double-panic containment untested** (`panic.rs:60-125`): the `thread::panicking()` discard branch, accumulator Drop-resume, and `take` disarm — the abort-prevention paths both txn types lean on.

## Per-module gaps

- [ ] Mailbox in-crate: `freeze`/`close` (waiter carry, frozen-park, payload return, wrong-incarnation no-ops); the `AcceptedSequence` `<=` boundary and `LiveOnly`-vs-frozen refusal; `MailboxTermination::finish` with an accumulated waker panic; terminate-beats-withdrawal ordering (`WithdrawalOutcome::Terminated` evidence); `CallFuture`/`AcceptanceTimedOut` (no unit test names `CallFuture`); the construction-overdue disposal path; `OneShotClose::Pending` (`sync.rs:359` — the one uncovered close arm, relied on by `reply.rs:123`).
- [ ] Runtime: `dispose_all` (empty-vec fire, aggregate counting, the panic-isolation doc claim); `wait_scope` `Deadline`/`ParentShutdown` arms and `select_two`; broadcast wrappers (`Lagged` count mapping); `Signal::pulse` wakes a parked watcher; `runtime/mailbox.rs` adapter (zero direct tests; the only crate allowed to test against tokio per #242).
- [ ] Engine: `StopLadder::force` in `Idle`; `acknowledge_framework_abort` phase guard (a mutant accepting any phase survives); `ReadinessGate` premature deadline; `stop_reason_precedence` direct table test (currently pinned only transitively — a variant-reorder in the derived `Ord` survives).
- [ ] Facade: static `SubtreeSlot` behavioral test (signature-only today); `ReserveError::IdentityExhausted` (zero tests; `near_exhaustion` hooks exist — note two different exhausted resources map to this one variant); admission fail-closed value pin (mirror `LOST_REMOVAL_RESPONSE_OUTCOME`); `JitterSample` facade test.
- [ ] SPEC §18 claims that would stay green if broken (from the external review, spot-checked plausible): Always-restart × parked send/call; parked mailbox futures under owner-drop/nested-driver abort (§18.16); `shutdown_and_wait(Duration::ZERO)` (§16.2); public `Exited` once-ness (tests set a boolean — duplicates would pass); runtime children of a restarted nested DynamicTree (§14.2); dropped builder slot → `UnfilledReservations`; a `should_restart`/`dispatch_exit` policy × ExitKind table.
- [ ] A minimal external-consumer test on default features (no test-util) — currently every behavioral test enables test-util.

## Acceptance criteria

- [ ] Each box gets a test or a recorded decision not to; mutation-check the load-bearing ones (#245 discipline).

## Related

- #152 — loom modeling (the adversarial-scheduling axis all three reviews flagged)
- #238 — SPEC citation mutation-checking (same verification philosophy)

ralexstokes:
Final facade/SPEC slice of #283 and the #284 behavioral-pinning sequence. Merge after #313 and #314. Closes #283.

## Coverage added

- Behavioral `SubtreeSlot`, both public identity-exhaustion routes, fail-closed admission, and `JitterSample` boundaries.
- SPEC §18 restart/parked-future, owner-drop, zero-shutdown, exact `Exited`, restarted dynamic-tree cleanup, dropped reservation, and complete policy/exit dispatch claims.
- A default-feature external-consumer runtime test, now included in the shared external-consumer lane.

Mutation checks prove the new admission, restart-policy, and exact-once lifecycle assertions are load-bearing.

Validation: 99 affected facade tests, the external-consumer lane, and `./tools/dev just ci` all passed (683/683 tests plus formatting, clippy, docs and API checks).

Stack: based on #296 / issue #281. The lower-crate #283 coverage is provided by #313 and #314; this facade/SPEC draft is the designated closing PR.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -219,9 +219,11 @@ normative so that "pure" is not misread into pessimization:
 
 - `step` MAY take `&mut State` — pure means no I/O, clocks, randomness, or
   awaits; it does not mean persistent data structures or defensive clones.
-- Effects MAY flow through a sink (`step(&mut state, event, &mut impl
-  EffectSink)`) rather than a returned `Vec` — tests pass a `Vec`,
-  production passes an inline executor; hot steps allocate nothing.
+- Effects MAY flow through a caller-owned buffer
+  (`step(&mut state, event, &mut Vec<Effect>)`) rather than a returned `Vec`,
+  which a caller MAY reuse across steps. Reuse is a permission, not a
+  guarantee: a caller that drains the buffer by moving out of it gives the
+  allocation back on every flush.
 - The actor loop's next-action decision is a small `Copy` enum, not a
   boxed plan.
 
@@ -260,8 +262,9 @@ Scope flavors:
 - **Ordered** — declared membership fixed at build time; sequential,
   readiness-gated startup in declaration order; reverse-order teardown.
 - **Dynamic** — runtime membership; concurrent start/stop; per-child
-  fate-sharing only (strategy is structurally `OneForOne`; the strategy knob
-  does not exist on dynamic scopes [#371]).
+  fate-sharing only (strategy is structurally `OneForOne`, and a dynamic
+  scope carries no strategy at all — not in its builder, its config, or its
+  snapshot [#371]; §9.1).
 
 **Child ids.** Every child has an id: a non-empty UTF-8 string, unique among
 the **resident** memberships of its containing scope — live *or*
@@ -840,8 +843,8 @@ runs under either mutex. The one carve-out is a framework invariant break
 the payload still under the guard, and they poison the mutex regardless, so the
 transition is abandoned rather than completed.
 The registered-waker slot exposes no operation that returns or replaces a
-`Waker` without an effects sink, making the #202 failure shape structurally
-unrepresentable rather than a call-site convention. Cancellation returns one
+`Waker` without an effects sink, making an under-lock caller-code drop
+structurally unrepresentable rather than a call-site convention. Cancellation returns one
 withdrawal outcome plus its post-unlock effects object, never a tuple carrying
 a raw waker across the boundary.
 
@@ -1083,8 +1086,8 @@ zero-cost reborrow of the same underlying context — no boxing, no
 mapping layer (Part II §17's `project` is the paying, boxed cousin; the
 `Msg`-equality bound is what makes the free identity possible). The
 returned context therefore *is* the outer actor's: same incarnation and
-identity (`id()`, `incarnation()`, `myself()`), same mailbox and shared
-resources, same incarnation-owned timer table (§5.3's heterogeneous keys
+identity (`id()`, `incarnation()`), same mailbox and shared resources, same
+incarnation-owned timer table (§5.3's heterogeneous keys
 keep an inner actor's keys collision-free by type), and every operation
 through it — sends, timers, offloads, `stop()` — attributes to the one
 outer actor (one mailbox, one identity, one lifecycle: §17's attribution
@@ -1289,7 +1292,7 @@ enum Readiness { Immediate, AfterInit, Manual }   // mode only — the deadline 
   exception: their readiness is structural (below), so the subtree veneer
   carries no mode override — only the deadline.
 - The deadline is not part of the mode: it lives on the shared options
-  record (§8) as `ReadinessDeadline: Inherit | Bounded(Duration) |
+  record (§8) as `ReadinessDeadline: Inherit | Bounded(NonZeroDuration) |
   Unbounded`, defaulting to `Inherit` — resolution is declaration → scope
   default → library default (Appendix A). Unbounded gating exists only via
   the explicit `Unbounded` value; no `None` does double duty as both
@@ -2021,7 +2024,14 @@ single largest block of deferred engine complexity, and §10's mode-based
 exit funnel is designed so they land without restructuring. The `Strategy`
 type is non-exhaustive from day one, is a property of **ordered** scopes
 only (§2), and does not exist on dynamic scope builders, configs, or
-snapshots.
+snapshots. While `OneForOne` is the sole variant, ordered builders carry no
+strategy setter either: a knob whose only value is its default selects
+nothing, and offering it would pin a call shape no caller can vary. The
+setter is deferred until a second `Strategy` variant lands with Part II §19;
+until then the type is reachable only where it is *read* — `ScopeSnapshot`
+reports `Some(OneForOne)` for an ordered scope and `None` for a dynamic one
+(B.6). Adding the setter back is an additive change, and the type staying
+non-exhaustive is what keeps it one.
 
 ### 9.2 Restart policy and intensity [#371]
 
@@ -2502,8 +2512,9 @@ or lifetime paragraph is unmapped.
 - **Builder operation inventory** (content-normative; Appendix B's
   naming latitude applies). One constructor per flavor — `Tree` for an
   ordered root, `DynamicTree` for a dynamic one. Per-scope
-  configuration setters: `strategy` (ordered only, §9.1), `intensity`
-  (§9.2), `defaults(ScopeDefaults)` (§9.3). Membership declaration:
+  configuration setters: `intensity` (§9.2), `defaults(ScopeDefaults)`
+  (§9.3) — there is no strategy setter on either flavor, per §9.1.
+  Membership declaration:
   the eight `add_*` entry points and the `reserve_*` slot family (§8) —
   on an ordered builder, declaration order is start order (§2); nested
   scopes declare through `add_subtree` / `add_subtree_once`, whose
@@ -2681,7 +2692,7 @@ or lifetime paragraph is unmapped.
   cells terminalize exactly as a dropped unspawned tree's do (§3.2),
   and the incarnation exits `Failed` carrying the structured
   startup-failure payload with a **lowering cause** naming the
-  undefined slots' child-id paths — the same data
+  undefined slots' child ids relative to that subtree root — the same data
   `BuildError::UnfilledReservations` carries, reached through B.5's
   `startup_failure()` accessor (§7 provenance, non-forgeable). The
   scope publishes `Stopped { reason: StartupFailed }` (B.6) and exits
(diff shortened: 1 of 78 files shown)


ralexstokes (commented):
The immediate-base façade/SPEC diff looks correct, and this branch merges cleanly with #313. There is one cross-stack integration issue to address before the full #282/#283 sequence lands: combining this head with #302 produces a conflict in `crates/shelterwood/tests/lifecycle.rs`, where #301's shared `next_event` refactor and this PR's exact-`Exited` extension edit the same receive loop. Please restack whichever branch lands later, preserve both changes, and run CI on the combined head. This PR's own `./tools/dev just ci` run passed (683 tests plus the remaining lanes), as does GitHub CI.

ralexstokes:
Addressed the stack-integration finding in `296982e`: this PR is now based directly on PR #302’s `agent/issue-282-crates` branch. The lifecycle conflict preserves #302’s shared `next_event` fixture and this PR’s exact exited-event accounting; the consolidated test imports were also updated. `./tools/dev just ci` passes, and this tip merges cleanly with the independently restacked PR #313 tip in a three-way merge simulation.

ralexstokes:
Applied the review findings in fcdfe71.

**Finding 4 (P2, admission fail-closed pin).** Confirmed. `LOST_ADMISSION_RESPONSE_ERROR` guarded only the receiver-side `unwrap_or_else` in `PendingAdmission::start` — unreachable by design — while the value a caller actually observes came from an independent literal in the driver obligation's drop fallback (`admission_control.rs:383`).

Re-homed both constants next to the obligations that publish them, in `driver/admission_control.rs`, and re-exported through `driver`:

- `LOST_ADMISSION_RESPONSE_ERROR` now serves the drop fallback **and** the facade's receiver-side branch, so one literal backs the whole documented contract.
- The removal side had the same shape: `complete_removals` restated `RemoveOutcome::Removed`. It is now `LATCHED_REMOVAL_OUTCOME`, shared with `tree::Removal`'s fallback. (Renamed from `LOST_REMOVAL_RESPONSE_OUTCOME` because the driver-side site is the publication, not the loss; the existing `lost_removal_response_policy_fails_closed` test carries over unchanged in substance.)

Added `driver::tests::dynamic::a_dropped_admission_request_publishes_the_fail_closed_response`: it reserves, defines, starts a real admission, drops the `AdmissionRequest` without completing it, and asserts the published response. That is the reachable path, not the unreachable branch.

*Mutation check (committed first):* changing the drop-fallback literal to `NotAdmittingCause::Draining` makes the new test **FAIL** (`left: Some(Err(NotAdmitting(Draining))), right: Some(Err(NotAdmitting(Terminal)))`) while `lost_admission_response_policy_fails_closed` still passes — exactly the gap you identified. Restored, green.

**Finding 5 (P3, "joins" unpinned).** Confirmed. Mirrored the root-level sibling's `Clears(Arc<AtomicBool>)` drop guard and assert `!live.load(SeqCst)` on the line immediately after `shutdown_and_wait` returns, before any `.await`.

*Mutation check:* deleting the `wait_for_incarnation(&scope, epoch).await` join from the escalation tail in `driver/shutdown.rs` makes the new test **FAIL**; the **previous** version of the same test **PASSES** under that identical mutation. Restored, green.

**Finding 6 (P3, self-comparing clamp assertion).** Confirmed: mutating the clamp to `clamp(0.0, 0.5)` left the whole test green. Added both remedies — `assert_ne!(JitterSample::new(1.0), JitterSample::new(0.5))` and the stronger public-surface assertion that `JitterSample::new(1.0)` drives the 100ns equal-jitter backoff to its full 100ns band rather than the 75ns an interior ceiling would give.

*Mutation check:* `clamp(0.0, 0.5)` now makes the test **FAIL**. Restored, green.

**Finding 7 (P3, missing SPEC citations).** Added both, in the surrounding format: `integration::restarted_dynamic_subtree_does_not_recreate_runtime_children` on §11's dynamic-restart bullet, and `integration::subtree_shutdown_and_wait_zero_escalates_and_joins_its_target_incarnation` on the zero-width immediate-escalation row (with a parenthetical naming what the scope-handle sibling adds over the root-level one, matching that row's existing discriminating-pin note).

**Tests:** `cargo nextest run -p shelterwood` — 515 passed, 0 failed; `-p shelterwood-core` — 77 passed. `cargo clippy -p shelterwood --all-targets` clean; `cargo +nightly fmt --all` applied, plus a direct `rustfmt --check` on the two `crates/shelterwood/tests/**` files that `cargo fmt` does not reach.

ralexstokes:
Merged `main` (909e89c, PR #308) into this branch at 5445d2a. Full local `just ci` is green (709 tests, external-consumer default-feature test included); the tree that CI ran on is byte-identical to what was pushed.

The branch had branched at `c50a930` and carried its **own, superseded copies** of the fixture-consolidation work that landed on `main` as #299/#301/#302, so nearly every conflict was two variants of the same refactor rather than a genuine disagreement. `main`'s variant is the reviewed one and wins throughout; every #303-specific contribution survives.

Per conflicting file:

- **`crates/shelterwood/tests/common/lifecycle.rs`** (add/add) — took `main`'s file wholesale: it is a strict superset (`next_item`, a documented `next_event`, plus the new `last_panic_message`).
- **`crates/shelterwood/tests/common/mod.rs`** — `main`'s re-export list, which adds `last_panic_message` to the branch's.
- **`crates/shelterwood/tests/common/waiting.rs`** (auto-merged, but defectively) — both sides had independently added `start_signalled_waiting_task` and `cancellation_signalled_waiting_task`, so the auto-merge produced duplicate definitions. Dropped the second copy.
- **`crates/shelterwood/tests/offloads.rs`** (4 hunks) and **`crates/shelterwood/tests/raw.rs`** (3 hunks) — took `main`'s `last_panic_message(&mut events, id)` over the branch's break-at-first-match loops. `main`'s helper drains the stream to closure and keeps the *last* panic exit, which is the stronger claim: a child with more than one incarnation cannot pass by matching an earlier exit.
- **`crates/shelterwood/tests/readiness.rs`** (2 hunks) — took `main`'s `construction_signalled_waiting_task`. This is a real semantic difference: the branch's `start_signalled_waiting_task` stores from the task body at first poll, while the base fixture (and `main`'s helper) stores synchronously from the *factory*, which is what the sibling-startup assertion depends on. Import list is `main`'s union.
- **`crates/shelterwood/tests/dynamic.rs`** — took `main`'s `signalled_waiting_once_task`. The branch's consolidation had silently changed `signalled_waiting_tree` from `add_task_once` to a restartable `add_task`; `main`'s helper keeps the base's one-shot child kind. All 93 lines of this PR's own additions to the file (including `restarted_dynamic_subtree_does_not_recreate_runtime_children`, cited from SPEC §11) auto-merged untouched.
- **`crates/shelterwood/tests/lifecycle.rs`** — kept **this PR's** side: the exited-incarnation accounting is the deliverable, and it was already expressed against `main`'s shared `next_event`, so no adaptation was needed.
- **`crates/shelterwood/tests/policy_validation.rs`** — union import. Kept this PR's `JitterSample` and `RestartAttempt`; adopted `main`'s `NonZeroDuration` in place of `BoundedReadinessDeadline`, which #305 deleted. The strengthened `JitterSample` ceiling assertion (`assert_ne!` against an interior point, plus the observable `Backoff::fixed` delay) is intact.
- **`crates/shelterwood/tests/subtrees.rs`** — kept **this PR's** import line, which is `main`'s plus `POLL_TIMEOUT`; `main` had dropped `POLL_TIMEOUT` because it does not have this PR's new parked-mailbox subtree test, which still needs it. The `Clears` drop-guard join assertions and `subtree_shutdown_and_wait_zero_escalates_and_joins_its_target_incarnation` (cited from the zero-width immediate-escalation row) survive.
- **`crates/shelterwood/src/driver.rs`** (3 hunks) — took `main`'s side on all three. Every hunk was #306's separation of supervisor results from command effects (`supervisor_effects` off `ScopeRuntimeWiring`, `supervisor_admit(...) -> Option<ChildKey>` in `for_test`); none of this PR's commits touch that code. This PR's only `driver.rs` change — re-exporting `LOST_ADMISSION_RESPONSE_ERROR` and `LATCHED_REMOVAL_OUTCOME` — auto-merged.
- **`crates/shelterwood/src/driver/tests/support.rs`** — took `main`'s new `epoch()` accessor.
- **`crates/shelterwood/src/driver/tests/shutdown.rs`** — took `main`'s `OrderedScopeFixture` construction, which binds `mut event_receiver` for the `recv_child_exit` handshake #299 introduced.

`crates/shelterwood/src/driver/admission_control.rs` and `crates/shelterwood/src/tree/admission.rs` did **not** conflict in the end (they conflicted against the intermediate `main` tip and resolved cleanly against 909e89c). The single-literal dedup this round introduced is intact and is what the diff against `main` now shows: `LOST_ADMISSION_RESPONSE_ERROR` serves both the obligation drop fallback in `start_dynamic_admission` and the facade's unreachable `PendingAdmission::start` branch, and `LATCHED_REMOVAL_OUTCOME` likewise serves `complete_removals` and `Removal`'s fallback (the old `LOST_REMOVAL_RESPONSE_OUTCOME` is gone).

Nothing was weakened on `main`'s side: #307's exhaustive retention `match`, #315's `#[must_use]` waiter registry, #296's disposal channel handshakes and #300's disposal routing are all untouched by this merge.

No #303 claim was lost, and no conflict had two plausible resolutions with differing observable semantics that I had to guess at — the three that *did* differ semantically (drain-to-closure vs first match, factory-time vs first-poll store, one-shot vs restartable child) all resolve toward the base behaviour that `main`'s reviewed fixtures preserve.

PR #304, opened 2026-08-19
Close residual public surface holes

### LINKED ISSUE #273: Public-surface holes: from_latch, installable seams, Exit::new, spawn_blocking

ralexstokes:
Continues the #284 hardening stack. Closes #273.

## What changed

- Replaces the open `Exit::new(ExitKind, Cancellation)` constructor with per-kind constructors; `NeverStarted` can now only be `NotObserved`.
- Restricts the raw runtime blocking-spawn wrapper to test/test-util use.
- Exact-pins Tokio to `=1.53.1`.
- Removes lifecycle capacity from the facade, replaces Tokio MPSC aliases with opaque adapters, and records the closed-lane sender-lifetime invariant.
- Documents the hidden lower-crate mailbox constructor boundary and the deliberately retained snapshot bridge.
- Extends the external-consumer negative fences for the retired API.

Validation: the rebased #269 → #272 → #273 stack passes `./tools/dev just ci` with 687/687 tests, formatting, clippy, doctests, docs, API reachability, packaged docs, external consumer, and Nix formatting.

Stack: based on #300 / issue #272 to avoid duplicating constructor migrations in its new regressions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,5 +38,7 @@ shelterwood-core = { path = "crates/shelterwood-core", version = "0.1.0" }
 shelterwood-mailbox = { path = "crates/shelterwood-mailbox", version = "0.1.0" }
 shelterwood-runtime = { path = "crates/shelterwood-runtime", version = "0.1.0" }
 thiserror = "2.0.19"
-tokio = { version = "1.53.1", features = ["macros", "rt", "sync", "time"] }
+# Blocking-pool rejection ownership is verified against this exact Tokio
+# release. Any upgrade must re-audit and re-run the adapter regressions.
+tokio = { version = "=1.53.1", features = ["macros", "rt", "sync", "time"] }
 tracing = "0.1.44"
--- a/SPEC.md
+++ b/SPEC.md
@@ -3112,6 +3112,18 @@ integration toolkit for the driver shell and the end-to-end invariants.
     hidden items rather than taught to descend into hidden methods on
     exported types; the façade-absence probe in
     `tools/check-external-consumer.sh` is what enforces this boundary.
+    `SnapshotReceiver::borrow_latest_and_closed` is the explicit benign
+    exception: the downstream façade needs this lower-crate bridge, its
+    signature contains only supported observation data, and it grants no
+    construction or installation capability. It therefore remains callable
+    on the façade-public receiver but is hidden from generated documentation;
+    the JSON walk deliberately does not gain a second hidden-item graph for
+    it. `MailboxCell::new` is likewise a hidden cross-crate constructor, but
+    the façade exports neither `MailboxCell` nor the `MailboxRuntime`
+    capability required to call it, so direct lower-crate construction remains
+    outside the supported boundary. The driver event lane's public adapter
+    types are opaque wrappers rather than Tokio aliases, and the lifecycle
+    ring capacity is an internal implementation choice rather than façade API.
 14. **Event-woken observers see consistent-or-newer snapshots.** Subscribe
    to lifecycle events; *synchronously inside the event arm*, read the
    snapshot and assert it already reflects the event — at both ends of the
@@ -3948,6 +3960,12 @@ pre-release, which does make the payload values themselves constructible by
 an application; that is deliberate and costs nothing, because authentication
 lives in the provenance structure rather than in the payload's privacy.
 Adding a cause or field must update every façade match.
+Construction is one named constructor per kind: `completed`, `failed`,
+`panicked`, `readiness_timed_out`, and `aborted` take the kind payload plus
+the orthogonal cancellation observation; `never_started` fixes cancellation
+to `NotObserved`. There is no public constructor that accepts an arbitrary
+`ExitKind`, so applications cannot construct the semantically impossible
+`NeverStarted`/`Observed` pair. The external-consumer probe pins that absence.
 Scope-level shutdown-timeout errors carry the affected children as
 structured data: child-id paths plus membership tokens (§7) — never
 bare ids, which sibling scopes may reuse (§2).
--- a/crates/shelterwood-cells/src/cells.rs
+++ b/crates/shelterwood-cells/src/cells.rs
@@ -2631,8 +2631,7 @@ mod tests {
     };
 
     use shelterwood_core::{
-        Cancellation, Exit, ExitError, ExitKind, ScopeState, StartupFailure, StartupFailureCause,
-        StopReason,
+        Cancellation, Exit, ExitError, ScopeState, StartupFailure, StartupFailureCause, StopReason,
     };
 
     use crate::{
@@ -2719,8 +2718,8 @@ mod tests {
     fn retained_failed_exit_disposes_off_the_retiring_thread() {
         let retiring_thread = std::thread::current().id();
         let (dropped, observed) = mpsc::sync_channel(1);
-        let retained = RetainedExit::new(Exit::new(
-            ExitKind::Failed(ExitError::from(ThreadProbe(dropped))),
+        let retained = RetainedExit::new(Exit::failed(
+            ExitError::from(ThreadProbe(dropped)),
             Cancellation::NotObserved,
         ));
 
@@ -2747,14 +2746,14 @@ mod tests {
                 .expect("membership is available"),
         );
         member.terminalize(
-            Exit::new(ExitKind::Completed, Cancellation::NotObserved),
+            Exit::completed(Cancellation::NotObserved),
             StartupDisposition::Unchanged,
         );
         let (dropped, observed) = mpsc::sync_channel(1);
 
         member.terminalize(
-            Exit::new(
-                ExitKind::Failed(ExitError::from(ThreadProbe(dropped))),
+            Exit::failed(
+                ExitError::from(ThreadProbe(dropped)),
                 Cancellation::NotObserved,
             ),
             StartupDisposition::Unchanged,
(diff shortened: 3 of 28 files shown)


ralexstokes (commented):
Reviewed the complete immediate-base diff and its interaction with the preceding lock-hardening PRs. I found no actionable issues in the per-kind `Exit` constructors, runtime surface restrictions, exact Tokio pin, opaque MPSC adapters, lifecycle-capacity removal, or external-consumer fences. Targeted core/runtime tests and `just external-consumer-check` passed.

ralexstokes:
Applied three P3 review findings (94d16f0):

1. `driver.rs:1119` — the `debug_assert!` message stated the invariant, which reads as true text when it fires. Reworded to the violation: "a driver event lane closed before its receive loop exited".
2. `spawn.rs` — deleted the free `unbounded_mpsc_send` / `_try_recv` / `_is_empty` forwarders and converted every call site to the inherent methods this branch introduced. 18 sites across `driver.rs`, `driver/child.rs`, `driver/events.rs`, `driver/admission_control.rs`, five test modules and `spawn.rs`'s own unit tests — mechanical, and the resulting builder-style calls read better than the two-argument free-function form. One spelling remains.
3. `driver/tests/observation.rs:262,353,417` — dropped the three stray trailing commas inside the `assert!` arguments left by the `Exit::new` → `Exit::aborted` rewrite (rustfmt does not reach inside macro arguments, so CI would not have caught them).

Tests: `cargo nextest run -p shelterwood -p shelterwood-runtime` — 547 passed, 0 skipped. `cargo clippy --all-targets` clean on both packages.

ralexstokes:
Merged `main` (a46685f) into this branch as 5ff42ef. Full local `just ci` green: 710/710 nextest, clippy/fmt/doc/API-reachability/external-consumer fences all clean.

**Conflict resolutions (6 hunks over 6 files)**

- `crates/shelterwood/src/lib.rs` and `crates/shelterwood/src/observe.rs` — both sides shrank the same `shelterwood_cells` re-export list. Kept **both** deletions: this round's `LIFECYCLE_EVENT_CAPACITY` removal and main's `ScopeKind` removal.
- `crates/shelterwood/tests/observation.rs` (2 hunks) — took main's `ScopeKind` → `ScopeFlavor` rename and its shared `common::{next_event, next_item}` fixtures, deleting the duplicate local helpers this branch still carried (rule 4). Kept this round's local `const LIFECYCLE_EVENT_CAPACITY: usize = 128;` and its comment, so the black-box overflow regression stops compiling against façade API.
- `crates/shelterwood/src/driver/tests/support.rs` — main's shrunk `engine::{ChildKey, Epoch, Event as SupervisorEvent, ScopeLifecycle, StopLadder, arbitrate}` list (the `SupervisorEffect`/`SupervisorState`/`step` imports main deleted were **not** resurrected, rule 2) plus this round's `cells::LIFECYCLE_EVENT_CAPACITY` path through the `pub(crate)` shim.
- `crates/shelterwood/src/driver.rs` — main's `drain_arrived_disposal_events(&mut self, panics: &mut runtime::PanicAccumulator)` signature combined with this round's `self.disposal_event_receiver.try_recv()` inherent call. The batch-collected `panics.run(...)` fold from #292/#296 is preserved.
- `crates/shelterwood-cells/src/cells.rs` — main's test-module imports (`ChildId`, `identity::ScopeIdentity`, `policy::{ResolvedMailbox, ScopeFlavor}`, and the `shelterwood_mailbox::{MailboxCell, MailboxControl, MailboxEffectQueue, MailboxReceiver, actor_ref_from_parts}` block from #300). `ExitKind` was dropped from that list only after its last use in the module became `Exit::failed`.

**`Exit::new` conversion**

`main` had **1** `Exit::new` call site this branch had not already converted — `cells.rs::a_declined_stop_reason_disposes_its_nested_exit_off_the_finishing_thread`, converted to `Exit::failed(ExitError::from(ThreadProbe(dropped)), Cancellation::NotObserved)`, preserving `ExitKind::Failed` + `Cancellation::NotObserved` exactly. Every other site on `main` was one this branch had already converted, so the merge carried the conversion through textually. `git grep -E "(^|[^a-zA-Z_])Exit::new" -- crates/ tools/` now matches only the negative fence in `tools/external-consumer/src/main.rs` and its diagnostic grep string in `tools/check-external-consumer.sh`, as designed.

No call site needed a provenance-changing approximation: the per-kind constructor set covers every `(ExitKind, Cancellation)` pair that appears. The one gap worth naming for the record is structural rather than reachable — `Exit::never_started()` fixes `Cancellation::NotObserved`, so an `ExitKind::NeverStarted` exit with `Cancellation::Observed` is unconstructible. Nothing on either side wanted one (a membership with no incarnation has no token to observe), which is exactly the mismatched provenance this PR set out to make unrepresentable.

**Free-forwarder conversion**

`main` added two new `runtime::unbounded_mpsc_*` call sites after this round deleted the free functions; both converted to the newtype inherent methods — `driver/tests/payload.rs` (`self.dropped.send(...)`) and `driver/tests/shutdown.rs` (`scope.disposal_event_receiver.is_empty()`). `git grep unbounded_mpsc_ -- crates/` is now empty.

**Fences re-verified non-vacuously** — `cargo check --features exit-new` fails with `` no associated function or constant named `new` found for struct `Exit` ``, and `--features lifecycle-capacity` fails with `` no `LIFECYCLE_EVENT_CAPACITY` in the root ``. Both diagnostics are the ones the script greps for.

Nothing from this round's co-landing work was weakened: #307's wildcard-free retention match, #315's `#[must_use]` wait structs and poison-tolerant watch, #296's channel handshakes, and #300's `MailboxClose` disposal routing / `MailboxBindToken` all came through untouched.

Issue #274, opened 2026-08-16
Supervisor effect vector is half result channel, half command stream


ralexstokes:
Simplification from the 2026-08-16 review round (external claim, verified and found broader than claimed). Highest-value structural cleanup on the board.

## The shape problem

The supervisor effect vector is half result channel, half command stream: 4 of 9 `Effect` variants are results extracted out-of-band by the caller before the flush, and the flush declares them `unreachable!`:

- `flush_supervisor_effects` (`driver.rs:405-430`): `Admitted => unreachable!(..)`, `StartupFailed | DrainStarted => unreachable!(..)`.
- Four extraction sites: `insert_child` snapshots `before = effects.len()`, reduces, `get(before)` expecting `Admitted{child}`, `remove(before)` (`driver.rs:445-469`); a bare `pop()` for `Admitted` in the scope constructor loop (`driver.rs:862`); the same before/get/remove dance for `StartupFailed` (`driver/startup.rs:127-134`); two `position`/`remove` scans for `DrainStarted` (`driver/shutdown.rs:112-127`, `:139-152`).

Two `unreachable!` arms in the driver event loop are a process abort waiting for the next "who publishes this" refactor.

## The fix, part 1 (cheap, decisive)

Have `admit` return `Option<ChildKey>` directly (additive alongside `step`; `reduce` currently returns `()`). This deletes both `Admitted` extraction sites, one `unreachable!` arm, and — verified — costs the #257 exhaustive reachable-state walk **nothing**: `Admit` is deliberately outside the walk's alphabet (`exploration.rs:131-134`) and `Effect::Admitted` is already its no-op arm.

Fold in the adjacent dead code found independently: `Event::Admit::start_immediately` is `false` at every production and test-support call site (`driver.rs:455`, `:858`, `support.rs:247`; only core's own unit test passes `true`), and its guard (`supervisor.rs:446-448`) duplicates the facade's admission gate (`driver.rs:490-497`) while emitting `StartChild` without consulting `startable()` — an untested hole in the reducer's acceptance discipline. Delete the field and branch. (`admit` itself has no draining guard — the facade is the policy authority; keep it that way and say so.)

## The fix, part 2 (separable)

`StartupFailed`/`DrainStarted` carry state the transition owner must publish before the flush proceeds, so they want return values from *their* events rather than the same treatment — a distinct design step; also collapses the duplicated DrainStarted-extraction dance between `begin_drain_transition` and `force_all`.

## Acceptance criteria

[...]

PR #305, opened 2026-08-19
Delete dead code and accidental configurability

### LINKED ISSUE #277: Dead code and dead configurability sweep
Deletion batch from the 2026-08-16 review round. Every item verified caller-by-caller; none is load-bearing.

## Dead code / dead configurability

- [ ] `sleep_deadline` (`shelterwood-runtime/src/timer.rs:46`) — zero callers; fix the `timer.rs:91` comment that references it.
- [ ] `RetainedExit::kind()`/`cancellation()` (`cells.rs:64-70`) — zero callers; consumers use `as_exit()`/`into_exit()`.
- [ ] `Tree::strategy` setter (`tree/builders.rs:244-247`) — `Strategy` has one variant, so the setter can only write the default while threading state through `ScopeConfig`/`set_observation_config`/`ScopeSnapshot`. Delete until a second strategy exists (per the #234 principle and the no-compat-machinery stance).
- [ ] Scope-readiness override resolution (`plan.rs:266-267`) — no public setter reaches it (`SubtreeDef`/`SubtreeOnceDef` expose none; `ScopeDefaults` has no readiness field). Delete the branch or add the setter — decide which.
- [ ] `Vec<Vec<ChildId>>` in `BuildError::UnfilledReservations` / `StartupFailureCause::Lowering` — the sole constructor builds exactly one segment (`plan.rs:359`); nested lowering reports relative to its own root. `Vec<ChildId>` suffices.
- [ ] `ChildKeyArg` trait + four `Index` impls (`driver.rs:252-266`, `:312-338`) — ~50 lines for `&key` ergonomics at five call sites; standardize on by-value.
- [ ] `ActorIdentity` blanket `impl for Arc<T>` (`shelterwood-mailbox/src/lib.rs:91-93`) — satisfies no bound anywhere; adds a forwarding hop.
- [ ] `OneShotSender::is_closed` dead `None` branch (`sync.rs:309`) — `channel` is only `None` inside consuming `send`; unobservable.
- [ ] `ActorWork.abort: AbortHandle` duplicates the retained `JoinHandle`'s capability; the `Cancelled` arm in `join` is unreachable (`spawn.rs:100-131`).
- [ ] `EffectSink` trait (`supervisor.rs:237-241`) — one impl (`Vec<Effect>`), never named outside core, always static-dispatched. `&mut Vec<Effect>` is equivalent. (Interacts with the supervisor-effects issue; do them together.)
- [ ] `CallOperation.runtime` / `SendFuture.runtime` restate values reachable through the sibling field (`futures.rs:208-209`, `:490-501`).
- [ ] Dead defensive branch in `CallOperation::poll_deadlined` (`futures.rs:568-571`) — unreachable by `Deadlined::poll`'s pass ordering; delete or mark defensive.
- [ ] `redundant post-split shims`: cells' module-rename shims (`shelterwood-cells/src/lib.rs:32-54`, ~6 importing sites); duplicate `BoxedSleep` alias (`capability.rs:11` vs `timer.rs:40`).
- [ ] Dead-in-practice pub items with zero cross-crate references (the class `unreachable_pub` can't catch because inner crates allow it): `deadline` fn (`timer.rs:42`), core's `ChildCompletionState`/`DrainEffect`/`RestartDecision`(public form)/`IncarnationState`/`StopPrecedence` names, `DEFAULT_*` constants, `structured_intensity_trip_error`, mailbox's `MailboxReceiver::changed`, cells' `MemberRecord`/`ScopeRecord` type names, `ObservationGate::same_gate`. Tighten visibility.
- [ ] `BoundedReadinessDeadline` (`policy.rs:561-569`) — enforces nothing beyond the `NonZeroDuration` it wraps (its own doc concedes it); the sibling `Shutdown::Graceful{grace: NonZeroDuration}` uses the bare type. Replace with `ReadinessDeadline::Bounded(NonZeroDuration)`. (SPEC writes plain `Duration` at ~1275 — amend spec to `NonZeroDuration` while there.)
- [ ] `TaskSlot::define_once` missing `#[must_use]` (`tree/slots.rs:286-292`) — the tuple return defeats the field-level attribute every sibling has.
- [ ] `StopContext::myself` — its own doc says posting work is futile in the only context the type exists; identity uses (Eq/Hash/id) remain real. Trim to identity or re-document; don't leave the send surface implying delivery.
- [ ] `resolve_and_take_defined_with(before_claim)` test-only closure parameter on a production signature (`plan.rs:223-240`) — move behind `#[cfg(test)]`/test-util seam.
- [ ] Historical comments that name tickets instead of invariants (`#202`, "historically handed displaced payloads", "the wake-start scan this path replaced") — keep the invariant sentence, drop the archaeology.
- [ ] Naming pairs where one name is a strict alias: `membership_terminal` ≡ `joined`; `removal_requested` ≡ `removal_latched` (`admission_control.rs:119-127` vs `:159-167`); `ScopeKind` vs `ScopeFlavor`; `JoinVerdict` vs `JoinOutcome` (core vs runtime re-export). Keep one name each.

## Acceptance criteria

- [ ] Each box either deleted/tightened or explicitly kept with a recorded reason.

ralexstokes:
Starts the #284 simplification sequence. Closes #277.

This audits every issue checkbox and either deletes the named seam or records the one real cross-crate reason it remains.

Highlights:

- Removes dead timer, retained-exit, tree strategy, scope readiness, identity, child-key, actor-work, timeout-arbitration, and context-forwarding paths.
- Flattens undefined-child paths and replaces the bounded readiness wrapper with `NonZeroDuration`.
- Deletes future disposal/withdraw function pointers using structural `Send + 'static` bounds.
- Tightens lower-crate visibility/reexports and unifies duplicate state aliases.
- Adds `#[must_use]` to one-shot task definition.

Two explicit ownership decisions: `EffectSink` is completed by the following #274 PR, and the two still-public cross-crate names have documented live consumers.

Validation: focused crate/API suites and `./tools/dev just ci` passed (674/674 tests plus formatting, clippy, docs, reachability, external consumer, and Nix formatting).

Merge order: review after the #281–#283 behavioral pins; this draft is based on current main to keep the deletion diff readable.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -840,8 +840,8 @@ runs under either mutex. The one carve-out is a framework invariant break
 the payload still under the guard, and they poison the mutex regardless, so the
 transition is abandoned rather than completed.
 The registered-waker slot exposes no operation that returns or replaces a
-`Waker` without an effects sink, making the #202 failure shape structurally
-unrepresentable rather than a call-site convention. Cancellation returns one
+`Waker` without an effects sink, making an under-lock caller-code drop
+structurally unrepresentable rather than a call-site convention. Cancellation returns one
 withdrawal outcome plus its post-unlock effects object, never a tuple carrying
 a raw waker across the boundary.
 
@@ -1083,8 +1083,8 @@ zero-cost reborrow of the same underlying context — no boxing, no
 mapping layer (Part II §17's `project` is the paying, boxed cousin; the
 `Msg`-equality bound is what makes the free identity possible). The
 returned context therefore *is* the outer actor's: same incarnation and
-identity (`id()`, `incarnation()`, `myself()`), same mailbox and shared
-resources, same incarnation-owned timer table (§5.3's heterogeneous keys
+identity (`id()`, `incarnation()`), same mailbox and shared resources, same
+incarnation-owned timer table (§5.3's heterogeneous keys
 keep an inner actor's keys collision-free by type), and every operation
 through it — sends, timers, offloads, `stop()` — attributes to the one
 outer actor (one mailbox, one identity, one lifecycle: §17's attribution
@@ -1289,7 +1289,7 @@ enum Readiness { Immediate, AfterInit, Manual }   // mode only — the deadline 
   exception: their readiness is structural (below), so the subtree veneer
   carries no mode override — only the deadline.
 - The deadline is not part of the mode: it lives on the shared options
-  record (§8) as `ReadinessDeadline: Inherit | Bounded(Duration) |
+  record (§8) as `ReadinessDeadline: Inherit | Bounded(NonZeroDuration) |
   Unbounded`, defaulting to `Inherit` — resolution is declaration → scope
   default → library default (Appendix A). Unbounded gating exists only via
   the explicit `Unbounded` value; no `None` does double duty as both
@@ -2680,7 +2680,7 @@ or lifetime paragraph is unmapped.
   cells terminalize exactly as a dropped unspawned tree's do (§3.2),
   and the incarnation exits `Failed` carrying the structured
   startup-failure payload with a **lowering cause** naming the
-  undefined slots' child-id paths — the same data
+  undefined slots' child ids relative to that subtree root — the same data
   `BuildError::UnfilledReservations` carries, reached through B.5's
   `startup_failure()` accessor (§7 provenance, non-forgeable). The
   scope publishes `Stopped { reason: StartupFailed }` (B.6) and exits
@@ -2965,7 +2965,7 @@ integration toolkit for the driver shell and the end-to-end invariants.
    Provoke each verdict (readiness expiry, grace-expiry abort, cancelled
    completion) and match the typed variant — never a stringly `Failed`.
    Enforce the mechanism structurally: core classification consumes typed
-   `RecordedOutcome` and `JoinVerdict` values and contains no `Any` or
+   `RecordedOutcome` and `JoinOutcome` values and contains no `Any` or
    downcast path; the runtime adapter converts its join result before core
    sees it, while user error erasure stays at the façade boundary. The former
    grep-level exit-path check is retired with the crate split because the
@@ -3664,7 +3664,8 @@ the same series during shutdown drain; **Stop** = `StopContext<'_, A>` in
 
 | Operation | Raw | Live | Drain | Stop |
 |---|---|---|---|---|
-| `id()`, `incarnation()`, `myself()`, `scope()`, `shutdown_token()` | ✓ | ✓ | ✓ | ✓ |
+| `id()`, `incarnation()`, `scope()`, `shutdown_token()` | ✓ | ✓ | ✓ | ✓ |
+| `myself()` | ✓ | ✓ | ✓ | — |
 | `request_scope_shutdown()` (fire-and-forget; awaiting your own scope's shutdown deadlocks — documented) | ✓ | ✓ | ✓ | ✓ |
 | `run_blocking(f)` | ✓ | ✓ | ✓ | ✓ |
 | `recv()` / `try_recv()` (the merged event source: mailbox, offload completions, fired timers, queued continuations, §5.2 priority; `recv` yields `None` on stop request, biased; `try_recv` ignores the stop token — the drain primitive for raw loops: under `Drain`, exhaust the frozen prefix via `try_recv` after `recv` yields `None`, §10's raw-loop obligation) | ✓ | — | — | — |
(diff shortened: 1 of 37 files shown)


ralexstokes (commented):
Reviewed as the `main → agent/issue-277` slice of the ten-PR #284 stack, including the lock-rule-sensitive disposal/API simplifications and their downstream uses. I found no actionable defects. GitHub CI is green, and the complete posted stack passed `./tools/dev just ci` in an isolated checkout.

ralexstokes:
Review findings applied in `ddebcd1`.

**1 — SPEC still advertised the deleted `Tree::strategy` setter (P1, confirmed).** The §11 builder operation inventory is content-normative and still listed `strategy` among the per-scope setters, and §9.1 read as asserting the knob exists on ordered builders. Both amended: the inventory now names `intensity` and `defaults` only, §9.1 explains that a knob whose only value is its default selects nothing and defers the setter until a second `Strategy` variant lands with Part II §19 — the type stays `#[non_exhaustive]`, and it stays readable on `ScopeSnapshot` (`Some(OneForOne)` ordered, `None` dynamic, B.6). §2's dynamic-flavor note was reworded to match, since "the knob does not exist on dynamic scopes" no longer distinguishes the flavors. §21's outline contract is untouched: it describes the resolved declaration surface, which still carries a structural per-scope strategy.

**2 — the atomicity test exercised a `#[cfg(test)]` copy, not production (P2, confirmed).** `resolve_and_take_defined` had been inlined, leaving `..._with` a `#[cfg(test)]` duplicate whose body only happened to match. Restored the delegation, so there is one body again, with a doc comment naming the test as the closure's sole consumer.

Mutation-check (committed tree first, `plan.rs` reverted with `git checkout` after each):

- *Control, PR as reviewed (`7de924a`)* — inject `drop(state); let mut state = self.definition.lock()…` between resolution and claim in the **production** `resolve_and_take_defined`, leave the test double untouched: `dynamic_policy_resolution_and_definition_claim_are_atomic` **PASSES**. The invariant was unpinned.
- *After the fix (`ddebcd1`)* — the same injection into the now-single body: **FAILS**, `unreachable!("the definition lock keeps resolution and claim atomic")` at `plan.rs:247`, the competing remover having claimed the definition through the reopened gap.
- One weaker placement is worth recording: injecting the unlock *after* the `before_claim()` hook rather than across it still passes, because the test's `yield_now` has already run by then. The hook marks the resolution→claim seam, so that is the placement the invariant is stated at.

**3 — `ScopeFlavor` regains `Hash` and variant docs (P3, confirmed).** The deleted `ScopeKind` carried both; `ScopeFlavor` inherited its role on the public `ScopeSnapshot` without them, and `shelterwood-core` allows `missing_docs`, so both variants rendered bare in the façade's rustdoc. Derives and doc comments restored.

**4 — new `.expect()` on an unreachable branch under two mutexes (P3, confirmed).** `OneShotSender::is_closed` is reached from `RemovalResponses::subscribe` inside `remove_dynamic_impl`, which holds the observation gate *and* `control.state`. Per CLAUDE.md, a verdict that cannot release its lock gets `debug_assert!`, not a panic that poisons the mutex for every later caller. Restored `is_none_or(Sender::is_closed)` behind a `debug_assert!` and a comment naming the callers. `ActorWork::abort` got the same treatment — no framework mutex there, but the same shape, and aborting an already-joined handle is the harmless reading. `ActorWork::join` keeps its `.expect()`: it consumes `self`, so no lock and no later caller.

**5 — `ChildCompletionState` restored as `pub(crate)` (P3, confirmed).** `finish_if_ready(flavor, has_children, all_terminal)` took two adjacent same-typed bools; transposing them at the sole call site compiles and silently changes the finish predicate, so the struct was enforcing something. `pub(crate)` in `shelterwood-core` is what #277 asked for — the visibility, not the type. `finish_if_ready` narrowed to `pub(crate)` with it.

`DrainEffect` stays flattened, deliberately: its two members are `bool` and `ScopeState`, so a transposition is a compile error and the struct enforced nothing the types don't. The engine test that read `drain.0`/`drain.1` now destructures `let (startup_pending, state) = …` for the readability the struct was carrying.

Checks (all on `ddebcd1`):

```
cargo nextest run -p shelterwood -p shelterwood-cells -p shelterwood-mailbox   557 passed, 0 skipped
cargo nextest run -p shelterwood-core -p shelterwood-runtime                   111 passed, 0 skipped
cargo test --doc -p shelterwood-core -p shelterwood                              9 passed (compile-fail fences)
cargo clippy --workspace --all-targets -- -D warnings                          clean
just fmt / doc-check / runtime-api-check                                       clean
```

PR #306, opened 2026-08-19
Separate supervisor results from command effects

### LINKED ISSUE #274: Supervisor effect vector is half result channel, half command stream

ralexstokes:
Continues the #284 simplification stack. Closes #274 and completes #277's `EffectSink` item.

## What changed

- Makes `Vec<Effect>` a pure command stream by deleting `Admitted`, `StartupFailed`, and `DrainStarted`.
- Adds typed admission, startup-failure, drain, and force entry points shared with the total `Event` reducer.
- Deletes `Event::Admit::start_immediately` and every driver-side result scan/pop/remove/unreachable arm.
- Returns drain metadata directly as the structural tuple retained by #277, without resurrecting its deleted wrapper type.
- Updates the stale `EffectSink` specification text.

The exhaustive event alphabet and transition graph are unchanged: 48,669 states and 1,314,063 transitions.

Validation: focused supervisor and driver suites passed; the rebased #277 → #274 stack passes `./tools/dev just ci` with 675/675 tests and all documentation/API/external-consumer checks.

This also subsumes #279's proposed `DrainStarted` extraction helper.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -220,8 +220,10 @@ normative so that "pure" is not misread into pessimization:
 - `step` MAY take `&mut State` — pure means no I/O, clocks, randomness, or
   awaits; it does not mean persistent data structures or defensive clones.
 - Effects MAY flow through a caller-owned buffer
-  (`step(&mut state, event, &mut Vec<Effect>)`) rather than a returned `Vec`;
-  tests and production can reuse that buffer, so hot steps allocate nothing.
+  (`step(&mut state, event, &mut Vec<Effect>)`) rather than a returned `Vec`,
+  which a caller MAY reuse across steps. Reuse is a permission, not a
+  guarantee: a caller that drains the buffer by moving out of it gives the
+  allocation back on every flush.
 - The actor loop's next-action decision is a small `Copy` enum, not a
   boxed plan.
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -219,9 +219,9 @@ normative so that "pure" is not misread into pessimization:
 
 - `step` MAY take `&mut State` — pure means no I/O, clocks, randomness, or
   awaits; it does not mean persistent data structures or defensive clones.
-- Effects MAY flow through a sink (`step(&mut state, event, &mut impl
-  EffectSink)`) rather than a returned `Vec` — tests pass a `Vec`,
-  production passes an inline executor; hot steps allocate nothing.
+- Effects MAY flow through a caller-owned buffer
+  (`step(&mut state, event, &mut Vec<Effect>)`) rather than a returned `Vec`;
+  tests and production can reuse that buffer, so hot steps allocate nothing.
 - The actor loop's next-action decision is a small `Copy` enum, not a
   boxed plan.
 
--- a/crates/shelterwood-core/src/supervisor.rs
+++ b/crates/shelterwood-core/src/supervisor.rs
@@ -137,10 +137,6 @@ impl ChildRecord {
 /// One input to the structural reducer.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum Event {
-    Admit {
-        membership: Membership,
-        initial: bool,
-    },
     Spawned {
         child: ChildKey,
     },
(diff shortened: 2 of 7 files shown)


ralexstokes (commented):
Reviewed as the `agent/issue-277 → agent/issue-274` slice of the #284 stack. The typed admission/startup/drain results preserve the prior reducer transitions while keeping the command vector pure. I found no actionable defects. GitHub CI is green, and the complete posted stack passed `./tools/dev just ci` in an isolated checkout.

ralexstokes:
Applied three P3 review findings (a7ee925):

4. **Deleted `Event::Admit`.** Confirmed no production constructor remains — this PR made `supervisor::admit` the sole entry point, and `alphabet()` in the exploration walk excludes admission by design (every admission mints a fresh key, so an explorable `Admit` makes the state space infinite). The arm was live enough to be dangerous: it called `self.admit(membership, initial)` and dropped the returned `ChildKey`, so a future caller routing an admission through `step` would insert a `ChildRecord` + `child_keys` entry the runtime holds no key for — an orphan that gates `next_ordered_start` and can never be spawned, stopped or reclaimed. Deletion cost no coverage: `duplicate_admission_and_stale_events_are_total_noops` already asserts the duplicate case through `super::admit(...) == None` plus `state.len() == 1`; the `step` call was redundant with it. Nothing in SPEC.md names the event.

5. **SPEC.md effects-buffer permission.** Confirmed the claim was false: `flush_supervisor_effects` (`driver.rs:469-487`) does `std::mem::take(&mut self.supervisor_effects)` then `for effect in effects`, consuming by `IntoIter`, so the allocation is freed on every flush and the next `push` reallocates. Reworded to a MAY-reuse permission and added the explicit caveat that a caller draining by move gives the allocation back. Did not touch the flush itself.

6. **`begin_drain` / `force` doc comments.** Added a line to each naming the tuple as `(startup_pending, state)` — `startup_pending` reporting whether startup was still in flight when the drain opened — and what `None` means (already draining; only the reason may have been upgraded). No newtype, per #277.

Tests: `cargo nextest run -p shelterwood-core -p shelterwood` — 583 passed, 0 skipped (including the exhaustive reachable-state walk). `cargo clippy --all-targets` clean.

PR #307, opened 2026-08-19
Simplify retained lifecycle events

### LINKED ISSUE #275: Replace RetainedLifecycleEventKind with wrap-plus-guards
Simplification from the 2026-08-16 review round (external claim, verified feasible).

## The duplication

`RetainedLifecycleEventKind` (`observe.rs:103-139`) mirrors all 7 variants of the public `LifecycleEventKind` (`observe.rs:278-340`), plus `RetainedLifecycleEvent` mirroring `LifecycleEvent` — but only 2 of 7 variants carry anything retained (`Exited{RetainedExit}`, `ScopeState{retained_exits}`). The other five are byte-identical restatements, and the two exhaustive conversion matches (`new` at `:142-203`, `into_public` at `:205-270`) make every future event variant a **4-edit** tax. The public enum is `#[non_exhaustive]` but same-crate, so no wildcard arm protects either match.

## The fix

Use the shape the snapshots already use (`RetainedScopeSnapshot`, `observe.rs:489-492`): wrap the public value plus a guard list — `{ event: LifecycleEvent, guards: Arc<Vec<RetainedExit>> }`, public value declared first. Verified against the two objections:

- Delivery already operates on an owned broadcast clone (`recv` at `observe.rs:821` calls `into_public()` on a value cloned out of the ring), so destructuring the clone — move `event` out, drop the cloned guards Arc as refcount traffic — works.
- Eviction safety is the identical field-order argument `RetainedStopReason` documents (`cells.rs:172-181`): the raw event drops first (refcount-only, guards still hold the `Arc<ExitErrorInner>`), then guards submit to isolated disposal.

Removes 7 variants, 2 exhaustive matches, and the per-variant tax, for one field-order comment.

The external review's longer-term "store `Exit` once behind `Arc`" is largely already true — `ExitError` is `Arc`-backed and `Eq` is `Arc::ptr_eq`; the guard machinery exists to control *when* the last Arc dies, which another Arc layer would not change. No action there.

## Acceptance criteria

- [ ] `RetainedLifecycleEventKind`/`RetainedLifecycleEvent` replaced by wrap-plus-guards; conversions deleted.
- [ ] Field-order comment stating the eviction argument; existing hostile-destructor eviction tests still pass.

ralexstokes:
Continues the #284 simplification stack. Closes #275.

`RetainedLifecycleEvent` now owns the public `LifecycleEvent` directly, followed by one shared guard vector. Construction retains exits only for `Exited` and recursively for `ScopeState`; all seven mirrored retained variants and both exhaustive conversion matches are gone.

The public event field is deliberately declared before its guards so ring eviction drops the raw projection while retained guards still protect user error payloads. Delivery moves the event out and releases the cloned guard arc after leaving the ring.

Validation: both hostile lifecycle-ring eviction regressions, all cells observation tests, formatting, and full `just ci` passed on the original branch; the tracker-ordered rebased branch passes the focused eviction suite 2/2.

### DIFF
--- a/crates/shelterwood-cells/src/observe.rs
+++ b/crates/shelterwood-cells/src/observe.rs
@@ -87,12 +87,24 @@ struct RetainedLifecycleEvent {
 
 impl RetainedLifecycleEvent {
     fn new(event: LifecycleEvent) -> Self {
+        // Exhaustive with no wildcard arm on purpose: `LifecycleEventKind` is
+        // non-exhaustive for downstream crates but not here, so a new variant
+        // fails to compile until its retention is declared. A variant that
+        // carries an `Exit` and slipped through with no guard would let ring
+        // eviction drop a user error under the observation gate.
         let mut guards = Vec::new();
-        if let LifecycleEventKind::Exited { exit, .. } = &event.kind {
-            RetainedExit::retain_exit(&mut guards, exit);
-        }
-        if let LifecycleEventKind::ScopeState { state } = &event.kind {
-            RetainedExit::retain_scope_state(&mut guards, state);
+        match &event.kind {
+            LifecycleEventKind::Exited { exit, .. } => {
+                RetainedExit::retain_exit(&mut guards, exit);
+            }
+            LifecycleEventKind::ScopeState { state } => {
+                RetainedExit::retain_scope_state(&mut guards, state);
+            }
+            LifecycleEventKind::Added { .. }
+            | LifecycleEventKind::Started { .. }
+            | LifecycleEventKind::Ready { .. }
+            | LifecycleEventKind::RestartScheduled { .. }
+            | LifecycleEventKind::Removed { .. } => {}
         }
         Self {
             event,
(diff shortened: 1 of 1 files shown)


ralexstokes (commented):
Reviewed as the `agent/issue-274 → agent/issue-275` slice of the #284 stack, with particular attention to field-drop order and retained-exit coverage for `Exited` and recursive `ScopeState` events. I found no actionable defects. GitHub CI is green, and the complete posted stack passed `./tools/dev just ci` in an isolated checkout.

ralexstokes:
Applied both review findings in 5693b2b.

**P2 — exhaustive-match forcing function restored.** Replacing the deleted `RetainedLifecycleEventKind` match with two independent `if let`s dropped the compile-time gate on retention. `LifecycleEventKind` is `#[non_exhaustive]` with documented intent that Part II adds kinds, so a future variant carrying an `Exit` (or `ScopeState`/`StopReason`/`StartupFailure`) would have built cleanly with an empty `guards` vector, and ring eviction inside `LifecycleHub::publish`'s `modify_silently` would drop that raw `Exit` as last owner under the observation gate — exactly what `lifecycle_ring_eviction_isolates_a_failed_exit_payload` and `..._an_exit_nested_in_scope_state` exist to prevent, with no compile error and no failing test. The dispatch is now a `match` with no `_` arm: `Exited`/`ScopeState` retain, the other five explicitly retain nothing. Mutation-checked on the committed tree by adding a `Probe { exit: Exit }` variant — `E0004: non-exhaustive patterns: `&LifecycleEventKind::Probe { .. }` not covered`. The PR's simplification is otherwise preserved; the deleted enum stays deleted, and the load-bearing `event`-before-`guards` field order is untouched.

**P3 — `into_public` no longer submits a pointless disposal job.** `drop(guards)` released the `Arc<Vec<RetainedExit>>` blind. In the common single-subscriber case tokio's `RecvGuard::drop` has already cleared `slot.val`, so the receiver's clone is the last arc owner, `RetainedExit::drop` runs, and `dispose_critical` gets an `Exit` whose payload the returned `event` still owns — a possible blocking-worker thread start per received failed-exit event, and under a torn-down runtime a job kept forever, pinning a clone of the user error for the process lifetime. It now unwraps the guard allocation and retires the copies inline (`drop(guard.into_exit())`), mirroring `RetainedScopeSnapshot::into_public`. The comment claiming the guard arc "is released as refcount traffic" is replaced with the actual argument.

Tests: `-p shelterwood-cells` 24/24 pass; `-p shelterwood` observation filter 50/50 pass, including both ring-eviction regressions.

PR #308, opened 2026-08-19
Shrink the dynamic route seam

### LINKED ISSUE #276: Shrink DynamicRoute to its one cross-crate method
Simplification from the 2026-08-16 review round (internal sweep; independently proposed by an external review).

## The seam is five-sixths a self-call

Of `DynamicRoute`'s six methods (`cells.rs:826-867`), only `close_admission` is called cross-crate (cells' drain/startup-failure path, `cells.rs:1444`). `reserve`, `start_admission`, `cancel_reservation`, `signal_fused_cancel`, `remove` are invoked exclusively from the facade — the impl's own crate — through the `Arc<ErasedDynamicRoute>` the facade itself installed, with each impl body downcasting the erased slot back to the concrete type it never really lost (`plan.rs:86-96`, `downcast.expect`). The `Slot` associated type has exactly one instantiation (`type Slot = ErasedDynamicSlot`, `admission_control.rs:531-532`), and the trait declared in cells names concrete runtime types in its signatures (`cells.rs:843,866`).

## The fix

Shrink the cells-side seam to the one thing cells needs: a close-admission callback (one-method trait or an `FnOnce`-shaped hook) on `ScopeCell`. Keep `DynamicControl`'s other five operations as inherent methods in the facade. Delete `ErasedDynamicRoute`, `ErasedDynamicSlot`, and the `concrete_dynamic_slot*` downcast helpers. Note the surviving callback still runs under the gate — keep the `&mut ObservationTxn` parameter so gate-correctness stays in the signature (that property is the one thing the current design gets right; see the lock-rule hardening issue for the `MailboxControl` counterpart).

Adjacent micro-item from the same seam family, separable: `MailboxSignalWatcher::changed` returns `Pin<Box<dyn Future + '_>>` and re-boxes **per park** (`capability.rs:48-51`, `runtime/mailbox.rs:112-118`; 0 allocations on the queued-message fast path, 1-2 per park). A `poll_changed(&mut self, cx)` method — precedented by `ErasedOneShotReceiver::poll_receive` in the same file — removes the allocation without reintroducing tokio into shelterwood-mailbox.

## Acceptance criteria

- [ ] Cells exposes only the close-admission hook; the five facade-only operations become inherent; erasure/downcast helpers deleted.
- [ ] (Optional) `poll_changed` replaces the boxed `changed()`.

ralexstokes:
Continues the #284 simplification stack. Closes #276.

`DynamicRoute` is now only the cells-side `close_admission(&mut ObservationTxn)` hook. Reservation, admission, cancellation, fused-cancel, and removal operations are inherent `DynamicControl` methods; reservations retain the concrete control.

The restart-stable `ScopeCell` still owns an `Arc<dyn DynamicRoute>`, but the facade recovers `DynamicControl` once. Slot values remain concrete end-to-end. The associated slot type, erased route/slot aliases, five per-call downcast helpers, and erasure-only test are deleted.

Validation: dynamic driver tests 20/20 and workspace all-target/all-feature check pass on the rebased stack; the original focused integration, gate-handoff, shutdown, and full-CI lanes were also green. The optional mailbox watcher allocation change remains deliberately separate.

### DIFF
--- a/crates/shelterwood-cells/src/cells.rs
+++ b/crates/shelterwood-cells/src/cells.rs
@@ -33,12 +33,10 @@ use shelterwood_core::{
 use shelterwood_mailbox::{ActorIdentity, MailboxControl, MailboxTermination};
 use shelterwood_runtime as runtime;
 
-use crate::{
-    observe::{
-        ChildSnapshot, ChildState, LifecycleEvent, LifecycleEventKind, LifecycleEvents,
-        LifecycleHub, LifecycleSeq, RetainedScopeSnapshot, ScopeSnapshot, SnapshotHub,
-        SnapshotPublication, SnapshotReceiver,
-    },
+use crate::observe::{
+    ChildSnapshot, ChildState, LifecycleEvent, LifecycleEventKind, LifecycleEvents, LifecycleHub,
+    LifecycleSeq, RetainedScopeSnapshot, ScopeSnapshot, SnapshotHub, SnapshotPublication,
+    SnapshotReceiver,
 };
 
 /// An exit copy retained by framework state.
--- a/crates/shelterwood-cells/src/cells.rs
+++ b/crates/shelterwood-cells/src/cells.rs
@@ -3,8 +3,8 @@
 //! These cells are the neutral synchronization and observation projection
 //! layer shared by public handles, mailboxes, actor/task contexts, and the
 //! mutable supervision driver. In particular, this module does not depend on
-//! mutable driver state. Its dynamic-route interface names declaration slots
-//! only as opaque capability payloads; their state remains owned elsewhere.
+//! mutable driver state. Its dynamic-route interface retains only the one
+//! close-admission hook needed by restart-stable scope transitions.
 
 use std::{
     any::Any,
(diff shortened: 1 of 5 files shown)


ralexstokes (commented):
Reviewed as the `agent/issue-275 → agent/issue-276` slice of the #284 stack, including route recovery, observation-gate transaction scope, reservation cancellation, fused removal, and teardown. I found no actionable defects. GitHub CI is green, and the complete posted stack passed `./tools/dev just ci` in an isolated checkout.

ralexstokes:
Applied the review finding in 4e37c24.

**P3 — downcast `expect` under the observation gate.** `DynamicControl::in_scope` takes an `&ObservationTxn<'_>` by signature, so it is only ever reachable inside `with_observation_gate` — the resident tree's outermost mutex. An `Arc::downcast(route).expect(..)` there unwinds with that guard held and poisons it, after which every later `.expect("observation gate poisoned")` aborts the whole tree rather than failing one operation. Only an out-of-contract direct dependency on `shelterwood-cells` can install a foreign `DynamicRoute`, so it is unreachable in-tree, but the convention is compute-the-verdict-then-panic and this `.expect()` cannot release its lock first.

Now `Arc::downcast(route).ok()`, reading a foreign route as no live incarnation, plus a `debug_assert!` recording the invariant — the documented fallback where releasing is impossible (`MailboxState::take_waiters` is the precedent). Verified both call sites handle `None` gracefully before changing it: `reserve_dynamic_slot` (admission_control.rs:353) maps it to `ReserveError::NotAdmitting(NotAdmittingCause::NoLiveIncarnation)`, and the removal path (:527) returns `completed_removal(RemoveOutcome::AlreadyAbsent)`. No caller unwraps or otherwise depends on the `Some`.

Tests: `-p shelterwood` 504/504 pass (dynamic/admission/reservation/removal filter: 104/104).

Issue #278, opened 2026-08-16
Invalid-states-unconstructible backlog: multi-field encodings


ralexstokes:
Batch from the 2026-08-16 review round: multi-field encodings where invalid combinations are constructible and the invariant lives in parallel code — the invalid-states-unconstructible backlog.

- [ ] **`ReadyBatch` triple-encodes its phase** (`raw.rs:763-838`): `mailbox_remaining: Option<usize>` doubles as the fired/steady discriminant, `mailbox_complete`/`offloads_complete` shadow the counters, `continuations_remaining` is meaningful only when fired. A two-variant enum (`Steady { mailbox_budget } / Fired { armings, continuations_remaining }`) makes `promote_to_fired`'s field-by-field reset (`:792-808`) and the debug_asserts structural.
- [ ] **`ReceiveMode` × `accepted_through` encodes 4 states, 2 reachable** (`cell.rs:59-62`, `:865-906`): only `(IncludeFrozen, None)` and `(LiveOnly, Some(_))` occur. A two-variant enum (`Drain` / `LiveThrough(AcceptedSequence)`).
- [ ] **`ResolvedDefaults` redundant state** (`policy.rs:774-867`): `mailbox: ResolvedMailbox` and `queue_capacity` must agree ("if Queue(c) then capacity == c") but `overlay` recomputes them with independent match arms. Store kind + capacity; derive `mailbox` in an accessor.
- [ ] **`adopt_or_mint_membership` returns `Option<Option<MintedMembership>>`** (`identity.rs:411-427`): three positional states read as chained `.expect`s. A three-variant enum.
- [ ] **`SendFuture`/`CallOperation` fn-pointer fields** (`futures.rs:207-243`, `:497-501`): `withdraw_operation` has exactly one value program-wide; `dispose`/`dispose_constructor` exist only so unbounded Drop impls reach `M: Send + 'static`. Declare the bound on the structs; delete the fields and the `WithdrawOperation` alias.
- [ ] **`ChildConstruction` Task/TaskOnce merge** (`plan.rs:30-35`): Raw and Scope already carry `DefinitionSource`; Task is the one family split at this level, and the typed completion channel is erased before this layer (`TaskOnceDef::erase` captures the sender inside the closure). Merging deletes `OnceTask`'s `Option` — `definition.rs:5-7` says `DefinitionSource::Spent` exists exactly to avoid that. (The external review's `SpawnBody` 5→3 half just relocates branches; skip unless it falls out.)
[...]

PR #309, opened 2026-08-19
Make construction states structural

ralexstokes:

- Merges `ChildConstruction::Task` and `TaskOnce` into one task construction backed by `DefinitionSource`.
- Deletes `OnceTask` and its mirrored optional body; one-shot extraction transitions structurally to `Spent`.
- Replaces parallel handler argument/actor options with `HandlerState::Uninit | Running` and documents the one-run contract.

The inner `Option<Args>` remains only as the consumed sentinel needed while async initialization owns the arguments; the invalid independent args-vs-actor combinations are gone. The optional `SpawnBody` reshaping was deliberately skipped.

Validation: actor/tasks/ownership, driver, shutdown, plan, and full-CI suites passed; the combined structural stack passes `./tools/dev just ci` with 676/676 tests.

Stack: based on the mailbox-encoding PR. The ReadyBatch phase PR completes #278 separately atop its required #268 panic-coherence base.

### DIFF
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -325,19 +325,20 @@ pub struct Handler<A: Actor> {
 }
 
 enum HandlerState<A: Actor> {
-    // The inner Option is the one-run sentinel while initialization owns the
-    // arguments and after any path that returns without constructing an
-    // actor. Unlike the former parallel Options, the enum makes the handler's
-    // phase structural.
-    Uninit(Option<A::Args>),
+    Uninit(A::Args),
     Running(A),
+    // The one-run sentinel. Initialization moves the arguments out and leaves
+    // `Spent` behind, so every path that returns without constructing an
+    // actor is already observably spent and a second run has exactly one
+    // panic site to reach.
+    Spent,
 }
 
 impl<A: Actor> Handler<A> {
     /// Wraps owned args using the callback-actor default `AfterInit` readiness.
     pub fn new(args: A::Args) -> Self {
         Self {
-            state: HandlerState::Uninit(Some(args)),
+            state: HandlerState::Uninit(args),
         }
     }
 }
@@ -362,12 +363,10 @@ impl<A: Actor> RawActor for Handler<A> {
     /// returns, and it must not observe still-live incarnation resources;
     /// every error exit therefore freezes and joins them here first.
     async fn run(&mut self, raw: &mut RawContext<Self::Msg>) -> ExitResult {
-        let HandlerState::Uninit(args) = &mut self.state else {
+        let HandlerState::Uninit(args) = std::mem::replace(&mut self.state, HandlerState::Spent)
+        else {
             panic!("handler actor initialization invoked more than once");
         };
-        let args = args
-            .take()
-            .expect("handler actor initialization invoked more than once");
         let initialized = {
             let mut context = Context::<A>::new(raw, DeliveryStage::Live);
             A::init(args, &mut context).await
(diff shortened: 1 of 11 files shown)


ralexstokes (commented):
Reviewed as the `agent/issue-278-mailbox → agent/issue-278-construction` slice of the #284 stack. One-shot task ownership still transitions exactly once to `Spent`, and the handler state preserves the teardown/drop ordering contract. I found no actionable defects. GitHub CI is green, and the complete posted stack passed `./tools/dev just ci` in an isolated checkout.

ralexstokes:
Applied both review findings in d5a5bb3.

**`HandlerState`'s inner `Option` (actor.rs)** — confirmed: `Uninit(None)` was a state whose only meaning was "spent", and the two panic sites (`let`-else at :365, `.expect()` at :370) were one contract split by whether the previous run got past `A::init`. Added `HandlerState::Spent` and took the args with `std::mem::replace`, matching `DefinitionSource::take_one_shot`. One panic site now, message unchanged.

**Deleted `should_panic` coverage** — restored both:

- `driver::tests::construction::dispatching_a_spent_one_shot_task_construction_panics` pins `"one-shot task construction invoked more than once"` at `driver/child.rs:393`. `spawn_child` drops a spent construction before it could be dispatched again, so the panic is only reachable through the dispatch helper directly; added a `#[cfg(test)] pub(super)` entry point for exactly that, with a comment saying why. Terminality is discharged before the second dispatch so the obligation's fallback does not run under the unwind.
- `raw::tests::handler_initialization_runs_at_most_once` pins the surviving `actor.rs` panic against the `bound_raw_context()` fixture that module already owns, using an actor whose `init` always fails — that is the path that previously left `Uninit(None)` behind and now leaves `Spent`.

`cargo nextest run -p shelterwood`: 506 tests, 506 passed. Clippy clean on `--all-targets`.

ralexstokes:
Rebased onto `main` (`--onto origin/main 9b4c7bb`) rather than merged. The branch was stacked on #306/#307/#308/#311/#312, all of which have since landed on main in reviewed form; replaying only this PR's own two commits drops those duplicated ancestor copies and cuts the conflict set from 34 across 15 files to 5 across 3. Force-push is the rewritten history, not new work — the diff against main is exactly `Make construction states structural` + `Collapse the handler's one-run sentinel and re-pin its panics`.

Conflict resolutions:

- `driver/child.rs`, `dispatch_child_construction` — kept **both** sides. main's `PendingSpawnBody::one_shot(..)` / `::restartable(..)` wrapper (#297's isolation of one-shot bodies across panic-resuming waker effects) is preserved, and this PR's `TaskConstruction::restartable() -> Option<&TaskFactory>` accessor now selects the arm and derives `construction_spent`. The merged arm is byte-for-byte the shape the `Scope` arm on main already uses.
- `driver/tests/dynamic.rs` (4 hunks) — took main's `insert_dynamic_fixture(..)` consolidation (#282) throughout and applied this PR's API change inside it (`ChildConstruction::Task(TaskDef::new(..).erase())`). No inline `reserve_dynamic` + `define` + `resolve_and_take_defined` + `from_plan` + `insert_child` sequences were reintroduced.
- `driver/tests/support.rs` (2 hunks) — both were import lists. Took main's list (which had moved `LIFECYCLE_EVENT_CAPACITY` under `cells::`, dropped `SupervisorState`/`supervisor_admit`/`ChildResources`, and added `ScopeRuntimeTestWiring`) and added this PR's two entries: `TaskOnceDef` and `child::dispatch_child_construction_for_test`.
- `driver/tests/exhaustion.rs` — not a conflict; a `ChildConstruction::Task` construction site added on main after this branch forked, updated for the `.erase()` signature.

What this PR contributes is intact: `TaskConstruction` with private fields over `DefinitionSource` (so a spent one-shot is `Spent`, not `OnceTask { body: None }`); `HandlerState::Spent` replacing the `Uninit(Option<A::Args>)` sentinel with one panic site, message text unchanged; and both restored `#[should_panic]` tests (`driver::tests::construction::dispatching_a_spent_one_shot_task_construction_panics`, `raw::tests::handler_initialization_runs_at_most_once`).

`just ci` green on the rebased head `06f4aa7` — 724/724 tests, both restored tests passing, no known flake hit.

PR #310, opened 2026-08-19
Make ready-batch phases structural

### LINKED ISSUE #278: Invalid-states-unconstructible backlog: multi-field encodings

ralexstokes:
Completes the remaining #278 state-encoding work after the other three focused #278 drafts land. Closes #278.

`ReadyBatchPhase` is now exactly:

- `Steady { mailbox_budget }`
- `Fired { armings, continuations_remaining }`

The optional discriminant, mailbox/offload completion booleans, and phase-invalid combinations are gone; exhaustion derives from budgets and counters. Promotion swaps the phase atomically and refreshes shared cutoffs.

This PR is intentionally based on #295 / issue #268 because caught callback panics must keep the authoritative fired batch installed. Duplicating that prerequisite into the broader simplification stack would obscure both reviews.

Validation: raw 23/23, events 13/13, offloads 31/31, and `./tools/dev just ci` passed (677 tests plus all documentation/API/external-consumer lanes).

### DIFF
--- a/crates/shelterwood/src/raw.rs
+++ b/crates/shelterwood/src/raw.rs
@@ -790,31 +790,33 @@ impl<M> Drop for TimerStore<M> {
 }
 
 struct ReadyBatch {
-    armings: VecDeque<ArmingOrder>,
-    // Only fired batches constrain continuations to a captured prefix.
-    // Steady-state continuations stay live so one queued by an external
-    // handler retains `continue_with`'s next-message priority.
-    continuations_remaining: usize,
+    phase: ReadyBatchPhase,
     mailbox_through: AcceptedSequence,
-    // Steady state takes one mailbox delivery before its captured offload
-    // prefix. A timer promotion removes that budget and drains the entire
-    // pre-fire mailbox prefix before delivering the timer armings.
-    mailbox_remaining: Option<usize>,
-    mailbox_complete: bool,
     offloads_remaining: usize,
-    offloads_complete: bool,
+}
+
+enum ReadyBatchPhase {
+    // Steady state takes one mailbox delivery before its captured offload
+    // prefix. Steady-state continuations stay live so one queued by an
+    // external handler retains `continue_with`'s next-message priority.
+    Steady {
+        mailbox_budget: usize,
+    },
+    // Fired batches drain the entire pre-fire mailbox prefix, constrain
+    // continuations to their captured prefix, and retain every due arming
+    // until its delivery commits.
+    Fired {
+        armings: VecDeque<ArmingOrder>,
+        continuations_remaining: usize,
+    },
 }
 
 impl ReadyBatch {
     fn steady(mailbox_through: AcceptedSequence, offloads_remaining: usize) -> Self {
         Self {
-            armings: VecDeque::new(),
-            continuations_remaining: 0,
+            phase: ReadyBatchPhase::Steady { mailbox_budget: 1 },
             mailbox_through,
-            mailbox_remaining: Some(1),
-            mailbox_complete: false,
             offloads_remaining,
-            offloads_complete: false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


ralexstokes (commented):
Reviewed the complete `agent/issue-268...agent/issue-278-ready` immediate-base diff and the combined stack. No independent findings in this PR: the `Steady`/`Fired` phase encoding preserves the former eligibility and budget behavior, and `./tools/dev cargo test -p shelterwood raw:: --lib` passes 23/23 on the PR head. The stack does inherit the merge-blocking stale captured-arming defect reported inline on #295, so this branch should be refreshed/revalidated after the base fix.

PR #311, opened 2026-08-19
Make core state encodings structural

ralexstokes:

- `ResolvedDefaults` now stores one mailbox kind plus its nearest capacity and derives `ResolvedMailbox`, eliminating contradictory parallel state.
- Membership reconciliation is the named three-way `Adopted | Minted | Exhausted` result throughout identity, cells, and lowering.
- Documents the fail-closed missing-membership status ruling.

Existing tests pin every reconciliation arm, capacity inheritance, and partial-adoption exhaustion. The complete core/mailbox/construction stack passes `./tools/dev just ci` with 676/676 tests and all documentation/API lanes.

Stack: based on #308.

### DIFF
--- a/crates/shelterwood-cells/src/cells.rs
+++ b/crates/shelterwood-cells/src/cells.rs
@@ -27,7 +27,10 @@ use shelterwood_core::{
         ExitKind, StartupError, StartupFailure, StartupFailureCause, StopReason,
         stop_reason_precedence,
     },
-    identity::{AtomicPoisonedCounter, IncarnationCounter, MintedMembership, ScopeIdentity},
+    identity::{
+        AtomicPoisonedCounter, IncarnationCounter, MembershipReconciliation, MintedMembership,
+        ScopeIdentity,
+    },
     policy::{ResolvedCommonOptions, ScopeFlavor},
 };
 use shelterwood_mailbox::{ActorIdentity, MailboxControl, MailboxTermination};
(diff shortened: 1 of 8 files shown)


ralexstokes (commented):
Reviewed as the `agent/issue-276 → agent/issue-278-core` slice of the #284 stack. The mailbox-default and membership-reconciliation encodings preserve all prior semantic cases while removing contradictory states. I found no actionable defects. GitHub CI is green, and the complete posted stack passed `./tools/dev just ci` in an isolated checkout.

ralexstokes:
Applied both findings in 72eecf9.

**Dropped `#[must_use]` on `MembershipReconciliation`** — confirmed. The former `Option<Option<MintedMembership>>` carried `must_use` for free; the bare enum carried none, and a discarded `Minted(identity)` means `rebase_membership` never runs, so the slot keeps its provisional lineage while the stable scope has already issued the successor. Marked the enum `#[must_use]`, with a doc note saying what a dropped outcome costs. Both `adopt_or_mint_membership` signatures (core and the `ScopeCell` forwarder) inherit it.

**`ResolvedMailbox::resolve`** — confirmed as the last spot where `inherited = Queue(a)` could pair with `inherited_queue_capacity = b`, `a != b`. Not reachable today (the sole caller passes `defaults.mailbox()` and `defaults.queue_capacity`), but the parameter pair is what made it writable. Folded into `ResolvedDefaults::resolve_child_mailbox(&self, Option<Mailbox>)`, which reads the kind and the capacity from one value exactly as `overlay` does inline, and deleted the old helper.

`cargo nextest run -p shelterwood-core -p shelterwood-cells`: 102 tests, 102 passed. Clippy clean on `-p shelterwood-core -p shelterwood-cells -p shelterwood --all-targets` (no new `must_use` fallout).

PR #312, opened 2026-08-19
Make mailbox states structural

ralexstokes:

- Replaces `(ReceiveMode, Option<AcceptedSequence>)` with exactly `Drain | LiveThrough(sequence)`.
- Centralizes all live binding replacement through one invariant-checking helper.
- Adds direct cutoff/frozen-mode and live-waiter-domain regressions.

The debug assertion is deliberate under the project lock rule: it inspects only framework state at the point where releasing the mailbox mutex is structurally impossible.

Validation: mailbox 30/30, raw integration/lib suites, and full CI passed on the original branch; the combined #278 stack passes full CI with 676/676 tests.

Stack: based on the core-encoding PR.

### DIFF
--- a/crates/shelterwood-mailbox/src/cell.rs
+++ b/crates/shelterwood-mailbox/src/cell.rs
@@ -55,11 +55,20 @@ impl<M> BoundState<M> {
     }
 }
 
-/// Which mailbox binding states may yield accepted messages.
+/// The structurally valid receive domains.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 enum ReceiveMode {
-    LiveOnly,
-    IncludeFrozen,
+    Drain,
+    LiveThrough(AcceptedSequence),
+}
+
+impl ReceiveMode {
+    fn accepts(self, sequence: AcceptedSequence) -> bool {
+        match self {
+            Self::Drain => true,
+            Self::LiveThrough(limit) => sequence <= limit,
+        }
+    }
 }
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


ralexstokes (commented):
Reviewed as the `agent/issue-278-core → agent/issue-278-mailbox` slice of the #284 stack, including live cutoff versus frozen drain behavior, waiter-domain replacement, and post-unlock effect handling. I found no actionable defects. GitHub CI is green; the complete posted stack passed `./tools/dev just ci`, and the mailbox crate's release-mode tests also passed in an isolated checkout.

ralexstokes:
Applied in 52b5a2e.

**`replace_binding` discarded its verdict in release** — confirmed. `self.binding = replacement` ran unconditionally, so only the `debug_assert!` kept the old binding; the doc comment described only that path. In release a violated invariant dropped a non-empty `WaiterQueue` under the mailbox mutex, releasing its `Arc<SendOperation<M>>`s — a last-owner release runs a user `M` destructor in the critical section. Added `if !replaceable { return; }` and rewrote the comment to cover both builds.

Two notes:

- The check stays a `debug_assert!` rather than an `assert!` for the reason the original comment gives: every caller holds the mutex, and panicking there poisons it for every later caller.
- The early return still drops the *replacement* under the lock. Every real caller detaches with `take_waiters()` before replacing, so the replacement's queue is the empty one and the old binding's is the one that can hold live payloads — the trade is strictly in the right direction. Removing the residual entirely needs the sink, i.e. the module restructure being filed separately.

No new test: the surviving behavior is release-only, the existing `binding_replacement_rejects_a_live_waiter_identity_domain` is `#[cfg(debug_assertions)]`-gated, and there is no release test lane in the `justfile` for a `not(debug_assertions)` sibling to run in.

`cargo nextest run -p shelterwood-mailbox`: 30 tests, 30 passed. Clippy clean on `--all-targets`.

PR #313, opened 2026-08-19
Pin mailbox runtime and engine contracts

ralexstokes:
Part of #283 and the #284 behavioral-pinning sequence.

Adds direct, mutation-checked coverage for:

- mailbox freeze/close, accepted-sequence boundaries, termination/withdrawal, and the complete call-future state matrix;
- runtime disposal aggregation/isolation, wait/select semantics, broadcast/signal behavior, and every mailbox-adapter operation;
- idle hard-force, framework-abort acknowledgement, premature readiness deadlines, and exact stop precedence.

The only non-test seam is a hidden test-util staged one-shot helper needed to deterministically reach `ONESHOT_SENDING`. Hostile multi-waker coverage remains in #265–#267 as the tracker assigns.

Mutation checks killed early completion, boundary exclusivity, frozen-live-read, idle-force, abort phase, readiness, precedence, pending-reply, and overdue-submission mutants.

Validation: core 80/80, runtime 47/47, mailbox 38/38, and `./tools/dev just ci` passed with 700/700 tests plus all formatting, documentation, API, external-consumer, and Nix lanes.

Stack: based on #296 / issue #281.

### DIFF
--- a/crates/shelterwood-core/src/engine.rs
+++ b/crates/shelterwood-core/src/engine.rs
@@ -1216,6 +1216,24 @@ mod tests {
     }
 
     #[test]
+    fn force_in_idle_arms_immediate_escalation_on_the_first_advance() {
+        let forced_at = Instant::now();
+        let mut ladder = StopLadder::new(
+            Shutdown::graceful(Duration::from_secs(30)).expect("test grace is non-zero"),
+        );
+
+        ladder.force(forced_at);
+        assert_eq!(ladder.deadline(), None, "an idle ladder is not armed yet");
+        assert_eq!(ladder.advance(forced_at), Some(StopAction::Cancel));
+        assert_eq!(ladder.deadline(), Some(forced_at));
+        assert_eq!(
+            ladder.advance(forced_at),
+            Some(StopAction::Escalate),
+            "force before the first advance skips the cooperative grace"
+        );
+    }
+
+    #[test]
     fn force_preserves_an_already_due_deadline() {
         let start = Instant::now();
         let grace = Duration::from_secs(30);
@@ -1284,6 +1302,33 @@ mod tests {
             Some(StopAction::HardAbort {
                 phase: GracePhase::WithinGrace
             })
+        );
+    }
+
+    #[test]
+    fn framework_abort_acknowledgement_is_ignored_before_the_abort_phase() {
+        let start = Instant::now();
+        let mut ladder = StopLadder::for_framework_driver(Shutdown::Abort);
+
+        ladder.acknowledge_framework_abort();
+        assert_eq!(ladder.advance(start), Some(StopAction::Cancel));
+        ladder.acknowledge_framework_abort();
+        assert_eq!(ladder.advance(start), Some(StopAction::Escalate));
+        ladder.acknowledge_framework_abort();
+        let tidy = ladder.deadline().expect("abort policy keeps a tidy beat");
+        assert_eq!(
+            ladder.advance(tidy),
+            Some(StopAction::AbortFramework {
+                phase: GracePhase::WithinGrace
+            })
+        );
+        let framework_tidy = ladder.deadline().expect("framework abort has a tidy beat");
+        assert_eq!(
+            ladder.advance(framework_tidy),
+            Some(StopAction::HardAbort {
+                phase: GracePhase::WithinGrace
+            }),
+            "acknowledgements before AbortingFramework must not suppress the hard-abort fallback"
         );
     }
 
@@ -1610,6 +1655,37 @@ mod tests {
             Some(ReadinessEffect::TimedOut { deadline })
         );
         assert!(!timed_out.needs_signal_watch());
+
+        let configured_deadline = deadline + Duration::from_secs(2);
+        let mut premature = ReadinessGate::new();
+        assert_eq!(
+            premature.step(ReadinessEvent::Configure {
+                readiness: crate::Readiness::Manual,
+                deadline: Some(configured_deadline),
+            }),
+            Some(ReadinessEffect::ArmDeadline {
+                deadline: configured_deadline
+            })
+        );
+        assert_eq!(
+            premature.step(ReadinessEvent::Deadline {
+                now: deadline,
+                signal_seen: false,
+            }),
+            None,
+            "a deadline event before the configured instant cannot time readiness out"
+        );
+        assert!(premature.needs_signal_watch());
+        assert_eq!(
+            premature.step(ReadinessEvent::Deadline {
+                now: configured_deadline,
+                signal_seen: false,
+            }),
+            Some(ReadinessEffect::TimedOut {
+                deadline: configured_deadline
+            }),
+            "the original deadline remains armed after a premature event"
+        );
 
         let mut immediate = ReadinessGate::new();
         assert_eq!(
--- a/crates/shelterwood-core/src/exit.rs
+++ b/crates/shelterwood-core/src/exit.rs
@@ -1149,7 +1149,22 @@ mod tests {
     }
 
     #[test]
-    fn stop_reason_precedence_is_an_explicit_exhaustive_table() {
+    fn stop_reason_precedence_is_severity_ascending() {
+        use super::StopPrecedence;
+
+        // `StopPrecedence` derives `Ord` from declaration order, and both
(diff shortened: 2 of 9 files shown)


ralexstokes (commented):
The immediate-base mailbox/runtime/engine coverage diff looks correct, including the test-gated staged one-shot seam; I found no lock-rule violation in the added paths. This branch also merges cleanly with #303. The cross-stack integration issue is #302: combining the heads produces conflicts in `crates/shelterwood-mailbox/src/cell.rs` and `futures.rs`, where #302's consolidated test-support minting helpers overlap this PR's expanded mailbox fixtures. Please restack whichever stack lands later and run CI on the resolved combined head. This PR's own `./tools/dev just ci` passed (700 tests plus the remaining lanes), as does GitHub CI.

ralexstokes:
Addressed the stack-integration finding in `61955b3`: this PR is now based directly on PR #302’s `agent/issue-282-crates` branch. The mailbox conflicts preserve this PR’s runtime-aware fixtures and coverage while using #302’s consolidated identity helpers; the disposal-test imports were updated accordingly. `./tools/dev just ci` passes (700/700 nextest tests plus the remaining CI mirror), and this tip merges cleanly with the independently restacked PR #303 tip in a three-way merge simulation.

ralexstokes:
All four applied in 069064f.

**Inert frozen-case assertion (`futures.rs`)** — confirmed. A frozen submission parks as a waiter (`cell.rs:812-817`) and `receive` promotes waiters only after popping an envelope while `Bound` (`cell.rs:899-905`), so on a frozen empty mailbox `try_recv()` reads `None` either way. Replaced with a waiter-queue probe through `MailboxState::waiters`, asserted both before the timeout (the call is parked) and after (it was withdrawn) — the "before" leg is what proves the probe is live, which is exactly what `try_recv` could not do. The unbound and full sub-cases are untouched; they were real.

**`stop_reason_precedence` identity table (`exit.rs`)** — confirmed: it restated `stop_reason_precedence`'s five lines and asserted nothing about the ordering the `StopPrecedence` doc claims. Now asserts the ascending chain over `stop_reason_precedence`, plus the two boundaries the doc names (`StartupFailed < ShutdownRequested`, `ShutdownRequested < NeverStarted`), and keeps an exhaustive `match` over `StopReason` so a new variant still cannot skip the list. Mutation-checked: swapping the `ShutdownRequested`/`NeverStarted` declaration order fails the new test and passed the old one.

**`biased;` unpinned (`spawn.rs`)** — confirmed for the two new tests. Worth noting the suite was not fully blind: `scope_wait_prefers_signal_when_both_control_futures_are_ready` already pinned signal over parent-shutdown and `scope_wait_prefers_a_primary_event_over_a_control_backlog` pinned message over control. Genuinely unpinned were `parent_shutdown > message` and `control > deadline`. Added `scope_wait_resolves_simultaneous_arms_in_declaration_order`, which walks all five arms with every weaker one simultaneously ready. Mutation-checked: swapping the message and control arms fails it.

**`OneShotSending::publish` restatement (`sync.rs`)** — confirmed byte-for-byte. Took the better option: extracted `publish_oneshot(channel, state, value)`, which both `OneShotSender::send` and the staged half call, so the two now differ only in how they enter the `ONESHOT_SENDING` window (CAS from `ONESHOT_OPEN` vs. constructed inside it) — which was the part the review noted is genuinely unshareable.

`cargo nextest run -p shelterwood-mailbox -p shelterwood-core -p shelterwood-runtime`: 166 tests, 166 passed. Clippy clean on `--all-targets`.

ralexstokes:
Rebased onto `main` (`0d87809`) rather than merged: the branch was stacked on the #282 fixture-consolidation commits, which reached `main` in modified form, so only this PR's own two commits were replayed (`git rebase --onto origin/main 61955b3^`). Head is now `d541c25`; `git log --oneline origin/main..HEAD` shows exactly `cf54419 Cover mailbox runtime and engine contracts` and `d541c25 Make four new coverage claims actually load-bearing`. The force-push drops the duplicated ancestors.

## Conflict resolutions

**`shelterwood-core/src/exit.rs` — test import list.** Union. `main` renamed `JoinVerdict as JoinOutcome` to a real `JoinOutcome`; this PR added `stop_reason_precedence`. Kept both.

**`shelterwood-core/src/exit.rs` — the precedence assertion (2nd commit).** Took this PR's ascending-chain form. `main` still returns `u8` from `stop_reason_precedence` (the PR base returned `StopPrecedence`), which the chain assertion is indifferent to — `StopPrecedence` derives `Ord` from declaration order and `as u8` preserves it, so the two named-boundary asserts and the `windows(2)` chain both hold unchanged. No assertion weakened.

**`shelterwood-runtime/src/disposal.rs` — test import list.** Union of `AtomicBool` (main) and `Condvar` (this PR).

**`shelterwood-mailbox/src/futures.rs` — two test import lists.** Union, then pruned by the compiler.

**`shelterwood-mailbox/src/cell.rs` — test import lists (×2).** Union for `std::sync`; dropped this PR's `identity::ScopeIdentity` in favour of #302's `mint_actor_membership()`, which `two_incarnations()` now delegates to.

**`shelterwood-mailbox/src/cell.rs` — four new tests.** Ported forward to #300's API rather than resurrecting the old one:

- `MailboxControl::configure/bind` now go through the existing `configure(...) -> MailboxBindToken` / `bind(..., token, ...)` test helpers, and the bind token is threaded through the close that returns it — `freeze_and_close_preserve_waiters_payloads_and_incarnation_boundaries` now takes the rebind permission from `MailboxClose::into_parts()`, which is a strictly stronger statement of the same claim (a rebind is unavailable without the matching close).
- The payload-transfer assertions kept their inline-drop semantics by dropping the `MailboxDisposal` from `into_parts()`. Dropping the whole `MailboxClose` would have routed the payload to the runtime's disposal worker and made `drops == 1` a race, so the test takes it apart deliberately.
- Added two `freeze`/`close` test helpers in the same shape as `main`'s existing `configure`/`bind`/`prepare_termination` wrappers; `close` keeps its result live across the effect flush, matching the driver shape `a_panicking_close_flush_isolates_the_unread_payload` documents. These five helpers are now `pub(crate)` so `futures.rs` reuses them instead of restating the effects plumbing.

## Two duplications the rebase exposed, both collapsed

- `main`'s `PanickingPulseRuntime` (#300) and this PR's `PanicPulseRuntime` were near-identical delegating fixtures. Unified into `PanickingPulseRuntime` with a disposal counter; `termination_finish_completes_waiters_and_isolates_payload_after_signal_panic` now expects `main`'s `"injected mailbox pulse panic"` and still observes `wakes == 2` and one isolated disposal.
- `main` grew its own `CallMessage` in `futures.rs` tests. This PR's call fixture is a different shape (`reply: Option<Reply<u8>>`, so a filler can be sent with no reply), so it was renamed `CallRequest` and `main`'s type left untouched.

`super::unbounded_mpsc_send(&sender, v)` became `sender.send(v)` — `main` moved it onto `UnboundedMpscSender`.

## The four review fixes all survive, and each was re-confirmed by mutation on the rebased tree

1. **Frozen waiter probe.** Stranding the frozen waiter (`remove_waiter` hands the operation back but re-parks it) fails on `"the frozen call was withdrawn"`. The old `try_recv().is_none()` form passes under that same mutation.
2. **Ascending-chain precedence.** Swapping `StartupFailed`/`ShutdownRequested` in `StopPrecedence` fails `stop_reason_precedence_is_severity_ascending`.
3. **`biased;` precedence.** Deleting `biased;` from `wait_scope`'s `select!` fails `scope_wait_resolves_simultaneous_arms_in_declaration_order`.
4. **Shared publication tail.** Corrupting `publish_oneshot`'s success store fails *both* `closing_oneshot_waits_for_a_send_that_won_before_publication` and `closing_oneshot_distinguishes_value_sender_drop_and_receiver_win`, i.e. `publish` and `OneShotSender::send` really do share the tail.

Nothing on `main` was weakened: #296's channel handshakes, #300's disposal routing and bind token, #307's exhaustive match, #312's `replace_binding` guard and #315's `#[must_use]`/poison-tolerant watch are all untouched. The only non-test change remains the `#[cfg(any(test, feature = "test-util"))] #[doc(hidden)]` one-shot staging seam plus the `publish_oneshot` extraction.

Local `just ci` is green end to end (exit 0), including the #325 flake.

PR #314, opened 2026-08-19
Pin cells transactions and panic containment

ralexstokes:
Completes the cells/core-panic subset of #283.

Coverage directly exercises:

- observation transaction unlock/order/drain-once/unwind and first-panic accumulation;
- obsolete-gate retry and recursive descendant rehoming;
- resident admit/prune/clear projections;
- mailbox double-attach and later terminalization;
- losing terminalizer first-wins behavior;
- the full stopped-precedence lattice;
- shutdown/force epoch isolation;
- retained-exit installation identity and isolated disposal;
- preferred-panic and accumulator drop/take behavior.

Six mutation checks prove the load-bearing branches fail when disabled; the active-unwind double-panic branch is instead pinned without intentionally aborting the test process.

After rebasing onto #294 → #300 → #304 and migrating its new tests to the per-kind exit constructors, core 80/80 and cells 39/39 pass. The original branch also passed full `./tools/dev just ci`.

This is one of three focused #283 coverage drafts; merge it with #303 and #313 before closing the issue.

### DIFF
--- a/crates/shelterwood-cells/src/cells.rs
+++ b/crates/shelterwood-cells/src/cells.rs
@@ -2593,8 +2593,8 @@ mod tests {
         future::Future,
         panic::{AssertUnwindSafe, catch_unwind},
         sync::{
-            Arc,
-            atomic::{AtomicBool, Ordering},
+            Arc, Mutex,
+            atomic::{AtomicBool, AtomicUsize, Ordering},
             mpsc,
         },
         task::{Context, Poll, Wake, Waker},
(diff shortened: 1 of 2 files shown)


ralexstokes (commented):
Reviewed the complete immediate-base, test-only diff and the behaviors it pins across the stack. I found no actionable issues in the transaction/unwind, gate-adoption, residency, terminalization, epoch, retained-exit, or panic-accumulator coverage. Targeted validation passed: cells (39) and core (80) tests, plus doctests.

ralexstokes:
Review follow-up, `7f93362`. Still test-only.

**The recursive rehome was not pinned.** `adopting_a_resident_subtree_rehomes_every_descendant_gate` built a subtree of depth 1: root ← nested ← {leaf-scope, leaf-member}, and leaf-scope had no children. `adopt_descendant_observation_gates_locked` re-homes each child through `descendant.member.install_observation_gate_locked`, and `ScopeCell::observation_gate` reads its own member, so leaf-scope was already correct after the one-level loop — the recursive call had nothing to walk. The tree now carries a `grandchild` member admitted into `leaf_scope`, reachable only through the recursion, and `grandchild.observation_gate()` joins the asserted set.

Mutation check on the committed tree:

- delete only the recursive call (`cells.rs:1511-1513`) → strengthened test **FAILS** (`every descendant joins the adopting parent's observation gate`); the same mutation against the pre-fix test **PASSES**, which is the hole.
- empty the whole `for descendant` body → strengthened test **FAILS** on the same assertion.
- restored → passes.

**Three assertion messages claimed more than they detect.**

- `observation_txn_flushes_during_unwind_...`: the `assert!(!gate.is_held())` calls sat inside the deferred closures, so `PanicAccumulator` captured them and `resume_preferred_panic` discarded the captures while the primary unwind was already in flight — they could not fail the test, and the unlock property survived only because `push` happened to follow the assert. Each closure now publishes `is_held()` alongside its ordinal and the test body judges `[(1, false), (2, false)]`. Verified live: moving the flush in `commit` ahead of `drop(self.guard.take())` turns the assertion red with `left: [(1, true), (2, true)]`.
- `observation_txn_commit_unlocks_before_effects_and_drains_once`: the post-`drop(txn)` message claimed the Drop "sees an already-drained effect list", but a `Box<dyn FnOnce()>` can only be consumed once, so no mutation of `commit` could re-run it for that reason. Reworded to what it pins — a second commit is a no-op. The first half (effects run after the unlock, in defer order) was already a real pin and is unchanged.
- `retained_exit_install_keeps_an_equal_shared_guard_set_in_place`: the empty-probe assertion claimed guards "retire as refcount traffic". It cannot detect that. `Exit` compares failed payloads by `Arc::ptr_eq`, so an equal incoming guard necessarily shares the error `Arc` and `ThreadProbe::drop` cannot fire whatever `install` does — the probe is empty on every path. Making it real would need a disposal-lane submission counter in `shelterwood-runtime`, i.e. a non-test change, so the message now states the precondition it actually holds. The other two assertions in that test are genuine and untouched.

```
CARGO_BUILD_JOBS=6 nice -n 10 ./tools/dev cargo nextest run -p shelterwood-cells --test-threads=4
Summary [0.025s] 39 tests run: 39 passed, 0 skipped
```

`just fmt` and `cargo clippy -p shelterwood-cells --all-targets` are clean.

Issue #152, opened 2026-08-10
Loom model checking for the hub/gate/watermark ordering arguments


ralexstokes:
Carried out of #151 §5; originally noted at the closure of #91: the hub/gate/watermark memory-ordering arguments are analytic — argued in comments and pinned only by ordinary tests — and the crate has no loom-style model checking to exercise the interleavings those arguments rule out.

## Sites whose ordering arguments deserve a model

- **`runtime::Latch`** (`runtime.rs`): `fired: AtomicBool` with an `AcqRel` CAS in `fire`, plus the create-notified-then-recheck protocol in `fired()` that leans on Tokio's guarantee that a pre-created `Notified` observes a later `notify_waiters`.
- **`runtime::CompletionGatedLatch`**: the `AtomicU8` gate that linearizes `fire` against `complete` so a retained capability can neither publish after completion nor vanish between a sample and the completion notification.
- **`observe::SnapshotHub`**: `closed: AtomicBool` `Acquire` loads paired with watch-channel `send_modify` on close — the close/subscribe race is the interesting interleaving.
- **`cells` lifecycle watermark mint** (`emit_locked`): the `Release` `try_update` that serves as both counter and published watermark under the observation gate, including the `u64::MAX` saturation path.
- Possibly the `raw` event-queue watermark (`concurrent_pushes_and_timer_watermark_share_one_linearization_point` pins one linearization claim already; a loom model would cover the rest).

## Design questions to settle first

- Tokio does not build under `cfg(loom)`, and `Latch` is built on `tokio::sync::Notify`. Either abstract the primitives behind `cfg(loom)` type aliases with a loom-compatible stand-in for `Notify` (modeling its create-then-recheck contract explicitly), or scope the models to the pure-atomic protocols and keep the `Notify` interaction argued analytically.
- Where the models live: `#[cfg(loom)]` unit modules next to each primitive, run via a dedicated `just` recipe / CI lane (`RUSTFLAGS="--cfg loom" cargo test ...`), consistent with the flake-defined CI.

## Acceptance

- Loom models exist for at least `Latch`, `CompletionGatedLatch`, and the watermark mint, exercising the documented ordering claims.
- A `just` recipe runs them locally and CI runs them in some lane (they are slow; a separate non-blocking or scheduled lane is acceptable).

ralexstokes:
Context from #214 track B, which is being retired as an inline checklist.

[...]

PR #315, opened 2026-08-19
Contain waiter notification and separate shutdown edges

### LINKED ISSUE #265: Hostile wakers I: tokio bulk notification strands co-waiters after one waker panic

ralexstokes:
Completes the #284 hostile-waker stack. Closes #265.

## What changed

- Replaces Tokio bulk notification in the affected latch/watch paths with a framework-owned waiter registry.
- Clones wakers before locking, moves only inert registry state under lock, and wakes/drops every waiter after unlock under individual panic containment.
- Preserves register-then-recheck lost-wake protection, level-triggered latch behavior, watch conflation/versioning/cloning/closure semantics, and final unseen values.
- Gives nested scopes a framework-only shutdown edge distinct from the user-facing cancellation latch; parent cancellation and teardown publish both independently.

Regressions combine hostile and ordinary latch/watch waiters, cover pulse and closure, pin latch identity separation, prove a hostile user cancellation waiter cannot strand the nested driver, and prove non-scope children allocate no extra latch. Assertions remain outside `catch_unwind`.

Validation: runtime sync 14/14, cells 24/24, driver 138/138, facade suites, and `./tools/dev just ci` passed with 687/687 tests plus all formatting, clippy, documentation, API, external-consumer, and Nix lanes.

Stack: based on #298 / issue #266; no loom work is included because #152 remains the dedicated future target.

### DIFF
--- a/crates/shelterwood/src/driver/child.rs
+++ b/crates/shelterwood/src/driver/child.rs
@@ -102,6 +102,7 @@ pub(super) struct ActiveChild {
     pub(super) readiness: ReadinessGate,
     pub(super) readiness_deadline: Option<DeadlineHandle>,
     pub(super) ready_signal: CompletionGatedLatch,
+    pub(super) framework_shutdown: Option<Latch>,
     pub(super) framework_abort: Option<Latch>,
     pub(super) framework_abort_ack: Option<Latch>,
     pub(super) stop_deadline: Option<DeadlineHandle>,
@@ -308,8 +309,10 @@ struct SpawnDispatch {
 /// Latches have deliberately separate ownership:
 ///
 /// - `shutdown`/`abort` are the child-facing cooperative ladder;
-/// - `framework_abort`/`framework_abort_ack` bound a nested scope driver's
-///   recursive drain before its task is aborted;
+/// - `framework_shutdown` is the nested scope driver's private observation
+///   edge, separate from user-installable shutdown-token waiters;
+/// - `framework_abort`/`framework_abort_ack` bound that driver's recursive
+///   drain before its task is aborted;
 /// - `ready` also carries the completion edge that makes readiness and
 ///   self-stop watcher tasks finite.
 struct SpawnLatches {
@@ -317,6 +320,7 @@ struct SpawnLatches {
     abort: Latch,
     ready: CompletionGatedLatch,
     local_stop: Latch,
+    framework_shutdown: Option<Latch>,
     framework_abort: Option<Latch>,
     framework_abort_ack: Option<Latch>,
 }
@@ -328,6 +332,7 @@ impl SpawnLatches {
             abort: Latch::default(),
             ready: CompletionGatedLatch::default(),
             local_stop: Latch::default(),
+            framework_shutdown: scope_child.then(Latch::default),
             framework_abort: scope_child.then(Latch::default),
             framework_abort_ack: scope_child.then(Latch::default),
         }
@@ -344,8 +349,12 @@ impl SpawnLatches {
     fn nested_scope(&self) -> NestedScopeLatches {
         NestedScopeLatches {
             parent_ready: self.ready.clone(),
+            child_shutdown: self.shutdown.clone(),
             ancestor: AncestorCommandLatches {
-                shutdown: self.shutdown.clone(),
+                framework_shutdown: self
+                    .framework_shutdown
+                    .clone()
+                    .expect("scope incarnations own a framework-shutdown latch"),
                 abort: self
                     .framework_abort
                     .clone()
@@ -357,6 +366,18 @@ impl SpawnLatches {
             },
         }
     }
+}
+
+pub(super) fn fire_shutdown_edges(shutdown: &Latch, framework_shutdown: Option<&Latch>) {
+    let mut panics = runtime::PanicAccumulator::default();
+    if let Some(framework_shutdown) = framework_shutdown {
+        panics.run(|| {
+            framework_shutdown.fire();
+        });
+    }
+    panics.run(|| {
+        shutdown.fire();
+    });
 }
 
 struct ChildTaskLaunch {
@@ -662,7 +683,9 @@ impl ScopeRuntime {
         // - ready and local_stop flow from application code back to helpers;
         // - ready's completion edge terminates those helpers when the child
         //   exits first and orders late retained readiness capabilities;
-        // - framework_abort/ack join nested-scope escalation before exit.
+        // - framework_shutdown keeps the nested driver observer separate from
+        //   user waiters, while framework_abort/ack joins escalation before
+        //   exit.
         // Each edge is level-triggered, so helper startup cannot lose a pulse.
         let scope_child = matches!(child.construction.get_mut(), ChildConstruction::Scope(_));
         let latches = SpawnLatches::new(scope_child);
@@ -735,6 +758,7 @@ impl ScopeRuntime {
             readiness,
             readiness_deadline: None,
             ready_signal: latches.ready,
+            framework_shutdown: latches.framework_shutdown,
             framework_abort: latches.framework_abort,
             framework_abort_ack: latches.framework_abort_ack,
             stop_deadline: None,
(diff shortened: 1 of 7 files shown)


crates/shelterwood/src/driver/child.rs
```diff
@@ -359,6 +368,18 @@ impl SpawnLatches {
     }
 }
 
+pub(super) fn fire_shutdown_edges(shutdown: &Latch, framework_shutdown: Option<&Latch>) {
+    let mut panics = runtime::PanicAccumulator::default();
+    if let Some(framework_shutdown) = framework_shutdown {
+        panics.run(|| {
+            framework_shutdown.fire();
```
ralexstokes, line 380: [P1] Commit the child-facing cancellation bit before waking the framework observer. `framework_shutdown.fire()` may schedule the nested driver on another worker; an empty/fast nested scope can observe that edge, finish, and let `ReportCompletion::fill` sample `child_shutdown == false` before the later `shutdown.fire()` runs. The parent then publishes a `Cancellation::NotObserved` exit for an ancestor-requested stop, contrary to the §11 invariant relied on in `run_nested_tree_with_epoch`. Fire `shutdown` silently first, then fire the framework edge, then notify the user latch (under the same accumulator) so hostile user waiters remain last without opening this race; add a concurrent nested-completion regression.

ralexstokes (commented):
Reviewed the immediate `agent/issue-266` → `agent/issue-265` diff plus the complete stack. The waiter registry follows the lock rule and the targeted runtime/latch suites passed, but the split shutdown edges currently introduce a cancellation-evidence race when a nested driver runs on another worker. The PR CI check is green.

crates/shelterwood/src/driver/child.rs
ralexstokes: Fixed in `ab6db60`. `fire_shutdown_edges` commits the child-facing bit with `fire_silently()` first, then wakes the framework edge, and only then notifies user waiters through the accumulator. The new deterministic regression has the framework waker sample the child cancellation bit and verifies it is already committed. `./tools/dev cargo test -p shelterwood --lib` passes 191/191.

ralexstokes:
Applied the review findings in 131e8f9 and 7a043d2.

**Finding 1 (P2, must_use).** Confirmed: `Latch::fired`, `WatchReceiver::changed` and `WatchReceiver::changed_or_closed` moved from `async fn` to hand-written future structs, dropping the implicit `#[must_use]` the opaque `async fn` return type carried. Added `#[must_use = "futures do nothing unless you `.await` or poll them"]` to `LatchWait`, `WatchChanged` and `WatchChangedOrClosed`. I swept the whole diff for other async-fn-to-struct conversions (`grep` for new `pub struct` / `impl Future for` across the range): these three are the only ones — `shelterwood-mailbox/src/futures.rs` added no future types, only the `withdraw_send_with` split.

*Probe:* a temporary test calling `latch.fired(); receiver.changed(); receiver.changed_or_closed();` as statements produced **0** warnings before the change and **3** `unused ... that must be used` warnings after. Probe removed.

**Finding 2 (P3, poisoning).** Confirmed, and I took the poison-tolerant option. All five `.expect("framework watch mutation does not panic")` sites now go through one private `WatchShared::value()` accessor using `unwrap_or_else(PoisonError::into_inner)`, matching `ObservationGate::lock` — the guarded data is plain framework state and no invariant spans the closure. I preferred this to hoisting the `debug_assert`/`mint` verdicts out of `SnapshotPublication::install`, because it also covers `LifecycleHub::publish`, whose `broadcast::send` under the guard cannot be hoisted, and because it is the ruling the gate already made for the same reason.

Added `a_panicking_watch_closure_leaves_the_channel_usable`, which panics inside a `modify_silently` closure and then asserts `read_cloned`, `read_with`, `borrow_cloned`, `borrow_and_update_cloned`, a later `modify_silently` and `pulse` all still work.

*Mutation check:* reverting the accessor to the prior `.expect(...)` makes that test **FAIL** (`PoisonError { .. }`); restored, green.

**Finding 3 (P3, self-satisfied assertion).** Confirmed. Rather than only narrowing the assertion to the two self-stop sources, I withheld the parent-published edge from the `AncestorShutdown` fixture and assert `!child_shutdown.is_fired()` there, which turns a vacuous assertion into one that pins the deliberate `if self_shutdown` = false branch and the code comment about not coupling the framework observer back to user-installable cancellation waiters.

*Mutation check:* removing the `if self_shutdown` guard in `driver.rs` makes `pre_loop_ancestor_shutdown_upgrades_a_nested_lowering_failure` **FAIL**; restored, green. (The previous form passed under that mutation.)

**Not fixed, as directed:** the `WatchSender::drop` / `wake_all`-under-the-gate concern — pre-existing and behaviour-preserving, being filed separately.

**Tests:** `cargo nextest run -p shelterwood-runtime -p shelterwood` — 554 passed, 0 failed. `cargo clippy -p shelterwood-runtime -p shelterwood --all-targets` clean; `cargo +nightly fmt --all` applied.

PR #316, opened 2026-08-19
Deduplicate raw cells and driver transitions

ralexstokes:
Starts the #279 cleanup after the #278 structural stack.

- Centralizes timer bucket lookup/unlink with panic-coherent pre-mutation validation.
- Centralizes continuation delivery with the explicit lead gate.
- Unifies observation and adoption retry loops through member-level primitives.
- Hoists construction-disposal pruning after the required removing flush.
- Inlines the single dynamic-entry inspection helper and fixes its stale race comment.
- Shares static/dynamic actor mailbox attachment.

New timer equality-panic coverage proves index coherence and lookup-probe accounting remain intact. `DrainStarted` work is deliberately absent because #306 fully subsumes it.

Validation: focused timer, observation, continuation, construction, and dynamic suites passed. On the full simplification stack, `./tools/dev just ci` hit the known unrelated parallel arrived-disposal race once; the exact test passed immediately and serial full CI passed all 677 tests plus every formatting/docs/API/external-consumer lane.

Stack: based on #309 and also logically depends on #310 before merge. The raw/cells split drafts follow this PR and will be rebased over the final ReadyBatch shape.

### DIFF
--- a/crates/shelterwood-cells/src/cells.rs
+++ b/crates/shelterwood-cells/src/cells.rs
@@ -564,9 +564,18 @@ impl MemberCell {
     }
 
     fn with_observation_txn<R>(&self, operation: impl FnOnce(&mut ObservationTxn<'_>) -> R) -> R {
+        self.with_observation_txn_probed(|| {}, operation)
+    }
+
+    fn with_observation_txn_probed<R>(
+        &self,
+        mut report_capture: impl FnMut(),
+        operation: impl FnOnce(&mut ObservationTxn<'_>) -> R,
+    ) -> R {
         let mut operation = Some(operation);
         loop {
             let gate = self.current_observation_gate();
+            report_capture();
             let guard = gate.lock();
             if gate.shares_gate(&self.current_observation_gate()) {
                 let mut txn = ObservationTxn::new(guard);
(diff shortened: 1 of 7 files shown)


ralexstokes (commented):
Reviewed as the `agent/issue-278-construction → agent/issue-279-dedup` slice of the #284 stack, including timer panic coherence, gate handoff retry semantics, terminal pruning, continuation fairness, and shared mailbox attachment. I found no actionable defects. I treated the declared #295/#310 dependency as planned ancestry rather than a defect in this posted slice. GitHub CI is green, and the complete posted stack passed `./tools/dev just ci` in an isolated checkout.

ralexstokes:
Applied three P3 review findings (de2264b):

7. **`unlink`'s cross-index `assert_eq!` → `debug_assert_eq!`, with the lookup hoisted.** Confirmed unreachable: `replace` (`raw.rs:610-618`) pushes into `keyed[h]` and does `armings.insert(ao, h)` in one window with no fallible step between them — the user `Hash`/`Eq` callbacks run strictly before — and entries never migrate buckets except through `unlink`. Also inconsistent with the symmetric invariant `replace` itself checks (`debug_assert!(previous.is_none())`). The release path is now one `keyed.get_mut` + `swap_remove` again; the debug check reads `entry.arming_order` off the value already owned, so it costs nothing extra even in debug. Confirmed `keyed_timer_churn_has_one_lookup_probe_per_removal` counts only bucket probes inside `take_hashed`'s predicate and would not have caught the regression.

8. **`pop_continuation(.., allow_lead)` → `is_lead_slot`.** Confirmed inverted: `allow_lead: true` selects the branch that applies `continuation_needs_external` (restricting delivery), and `false` is the unrestricted remainder drain. Renamed; both call sites now read correctly.

9. **Split the two corruption modes.** Took the shared shape of `remove_arming` and `entry_mut` into a new `locate_arming`, which resolves the arming hash and the bucket index in one place: `None` only for an unknown arming order, `expect("an arming index must reference a key bucket")` for a missing bucket, `expect("an arming index must reference a timer")` for a missing entry. That restores the distinction `locate`'s single `None` had conflated without re-inlining the lookup at both sites, so it deduplicates rather than fighting this PR's direction. Same map-operation count as before.

Tests: `cargo nextest run -p shelterwood raw::` — 21 passed. Also ran the full `-p shelterwood` suite: 505 passed, 0 skipped. `cargo clippy --all-targets` clean.

PR #317, opened 2026-08-19
Split the raw actor implementation by responsibility

ralexstokes:
Continues #279 after the deduplication PR.

The first commit is an exact 100% `raw.rs → raw/mod.rs` move. The second leaves a 13-line module/re-export shell and extracts:

- `disposal.rs`: panic containment and raw disposal;
- `offload.rs`: guards, blocking work, and shared offload state;
- `context.rs`: event/timer/continuation queues, ready batching, resources, and `RawContext`;
- `definition.rs`: actor definitions, erasure, construction, spawn, and run context.

Public `raw::*` paths and the crate-private driver seam remain unchanged. Visibility widens only to `pub(super)`/existing `pub(crate)` inside the private module.

History was inspected with color-move and copy detection; Git keeps the canonical original lineage through the 100% move, while the extraction commit reports the large context/definition copies. This avoids promising impossible `--follow` ancestry for every child file.

Validation: raw unit 21/21, raw/events/offloads/deadlines integrations 73/73, workspace all-target/all-feature check, and `./tools/dev just ci` passed with 677/677 tests after one exact rerun of the known timing-sensitive shutdown test.

Dependency note: rebase over #295/#310 before merge so the final split carries the panic-coherent ReadyBatch phase shape.


ralexstokes (commented):
Reviewed as the `agent/issue-279-dedup → agent/issue-279-raw-split` slice of the #284 stack. I checked the exact preliminary move, compared the full definition/function inventory across the extraction, and audited the widened names as private-module visibility only. I found no actionable defects. I treated the declared #295/#310 dependency as planned ancestry rather than a defect in this posted slice. GitHub CI is green, and the complete posted stack passed `./tools/dev just ci` in an isolated checkout.

ralexstokes:
All three applied in ca8e4b8.

**`crate::raw::RawActor` doc link** — confirmed. `mod raw;` is private (`lib.rs:20`) and the type ships as `shelterwood::RawActor` (`lib.rs:52`), and with the bare `[\`path\`]` form rustdoc uses the path as the rendered text. Changed to `[\`RawActor\`](crate::RawActor)`. Verified against the generated HTML: it now renders `RawActor` and links to `trait.RawActor.html`.

**Dropped `TaskPollContext` alias** — confirmed. `git show HEAD~2:crates/shelterwood/src/raw.rs` has `task::{Context as TaskPollContext, Poll}` and three `poll` signatures using it; the split replaced all three with bare `Context<'_>` while `shelterwood::Context` (the actor context, `lib.rs:26`) is still exported. Restored the alias in `offload.rs` and `disposal.rs`.

**Missing module docs** — confirmed; `context.rs` was the only one of the four with a `//!` header. Added one line each, describing what each file actually holds rather than repeating its filename:

- `definition.rs` — the `RawActor` trait, its definitions, and their erasure into a spawnable construction
- `disposal.rs` — panic containment (`CatchUnwindFuture`, `PanicSlot`) and isolated disposal
- `offload.rs` — offloaded blocking work and its shared per-incarnation resource

`cargo nextest run -p shelterwood raw::`: 21 tests, 21 passed. `RUSTDOCFLAGS="-D warnings" cargo doc -p shelterwood --no-deps` clean; clippy clean on `--all-targets`.

ralexstokes:
Re-derived the split from current `main` and force-pushed; the branch is now a single commit on top of `30f9c99`.

**Why not a rebase.** The old split commit deleted a ~118KB body while `main` kept editing that same body (#295 timer coherence, #300 mailbox teardown, #304 Exit constructors, #316 raw dedup, and #310 `ReadyBatchPhase` which landed mid-run). Replaying it would have silently reverted those. So the partition was re-derived: `main`'s `raw.rs` was cut into the same five files by the same responsibility boundaries, taking every item body verbatim from `main`.

**Mechanical no-content-change evidence.** Both sides were parsed into top-level items (an item = its doc comments + attributes + body). Each item was normalised by stripping `pub(super)`/`pub(crate)` prefixes and collapsing all whitespace, so rustfmt re-wrapping of the widened signatures cannot register as a difference, and the two multisets were compared:

```
old items: 87  new items: 87
IDENTICAL: every top-level item matches after normalisation
```

`origin/main:crates/shelterwood/src/raw.rs` has 87 top-level items; the five new files have 87; every one matches. The only normalisations applied beyond whitespace were the `pub(super)`/`pub(crate)` prefixes and the one intended doc-link edit below. Nothing else differs — no dropped item, no edited body, no reordered field.

**Layout.** `raw/mod.rs` carries the original `//!` header, four `mod` declarations and the previous re-export surface, so nothing outside `raw` changes.

* `context.rs` — `RawContext`, `Rejected`, `RawResources`, event queue, timer store (incl. `ReadyBatchPhase` from #310), and both inline `#[cfg(test)]` modules
* `definition.rs` — `RawActor`, `RawDef`, `RawOnceDef`, `RawConstruction`, `RawRunContext`, `RawSpawn`
* `disposal.rs` — `PanicSlot`, `RawDisposal`, `Contained`, `CatchUnwindFuture`
* `offload.rs` — shared offload state, `Guard`, `Blocking`, `DeadlineElapsed`

**Visibility.** `mod raw;` stays private in `lib.rs`; nothing gained `pub` or `pub(crate)` that did not already have it. The `pub(super)` set was derived by the compiler rather than copied: every candidate was narrowed back to private and re-checked, and only the ones that actually failed to build were widened. That dropped four of the widenings the earlier revision carried (`type SharedWork`, `OffloadFutureState::{future, cancelled}`, `SharedOffloadState::finished`). The 37 that remain:

* `context.rs` (2): `RawContext::new`, `RawContext::take_resource_panic`
* `disposal.rs` (12): `PanicSlot` + `record`/`take`; `RawDisposal` + fields `panic`/`signal` + `record`/`dispose`; `Contained` + `new`/`get`/`into_inner`
* `offload.rs` (23): `Guard` fields `cancellation`/`finished`/`armed`; `Blocking` fields `future`/`cancellation`/`completed`; `OffloadFutureState` (the type, required by `private_interfaces`) + field `polling`; `OffloadPoll`; `SharedOffloadState` + field `state` + `new`/`take_for_poll`/`finish_poll`/`dispose`/`cancel`; `SharedOffloadFuture` + its tuple field; `OffloadResource` + fields `cancellation`/`finished`/`state`/`task` + `cancel`

**Review fixes carried forward.** The `RawContext` doc link is written as ``[`RawActor`](crate::RawActor)`` (the public path — `crate::raw` is private); `use std::task::{Context as TaskPollContext, Poll};` keeps `std::task::Context` distinct from the crate's `Context` in `offload.rs` and `disposal.rs`; all four new files carry a `//!` header. The test module's `#[cfg(test)] use super::offload::OffloadPoll;` is present and cfg-gated, and both inline test modules still compile.

**CI.** `just ci` green, exit 0, 739 tests passed. `RUSTDOCFLAGS="-D warnings" cargo doc -p shelterwood --no-deps` clean, confirming the doc link resolves.

PR #318, opened 2026-08-19
Split restart-stable cells by responsibility

ralexstokes:
Completes the required #279 file splits. Closes #279 after #316 and #317 land.

The first commit is an exact 100% `cells.rs → cells/mod.rs` move. The second extracts:

- `retained.rs`: retained exit/stop-reason ownership;
- `member.rs`: member record, transition, mailbox, and terminalization;
- `gate.rs`: observation gate/transaction;
- `scope.rs`: residency, epochs, control-plane storage and transitions;
- `scope/projection.rs`: scope record plus snapshot/lifecycle projection.

Projection remains nested so it can access `ScopeCell` internals without externally widening them. All six original tests moved beside their owning implementation unchanged; a definition inventory confirms all 190 original definitions remain exactly once. New visibility is private-module `pub(super)` only.

History retains the canonical 100% rename; individual one-to-many extracts remain traceable with color-move/`--find-copies-harder`, with the usual Git limitation that every child cannot own `--follow` ancestry.

Because this independent stack does not yet contain #298, it carries the same test-only real-join synchronization from #266. The teardown fixture now waits for the aborted task's actual join event before injecting its synthetic post-join verdict, so it cannot inspect publication while the task still owns a factory clone.

Validation: cells 24, driver observation/dynamic/shutdown 73, facade observation/dynamic 52, full local `./tools/dev just ci` with 677/677 tests, authoritative `./tools/dev just ci-nix`, and GitHub CI all pass. The combined raw+cells split also passes workspace all-target/all-feature check.

The optional engine and child splits are explicitly deferred: both are internally sectioned and under active semantic churn, while the issue's required raw/cells split is complete.

This is the designated final merge in the full-review sequence. Merge it only after the other #284 drafts; then it closes #284.

PR #328, opened 2026-08-19
Make arrived disposal-panic teardown tests deterministic

### LINKED ISSUE #325: Flaky: *_folds_an_arrived_factory_disposal_panic fails intermittently under load
Reported independently by two agents during the #294-#318 review round (2026-08-19), on different branches, and reproduced on an unmodified branch tip — so this is pre-existing, not introduced by that round.

## Symptom

`crates/shelterwood/src/driver/tests/shutdown.rs`:
- `runtime_teardown_folds_an_arrived_factory_disposal_panic`
- `hard_force_folds_an_arrived_factory_disposal_panic`

fail intermittently under a full `-p shelterwood` run on a loaded machine, and pass in isolation.

Observations:
- One agent: ~1 failure in 4 full-suite runs on the unmodified branch tip; passes 200/200 in a direct stress loop.
- Another agent, on a different branch: one failure under heavy load, then passed in every later suite run and in a 200-iteration stress loop.

The common factor is whole-suite contention, not the code under test, which points at a timing assumption in the fixture rather than a runtime defect.

## Where to look

These tests came in with #292 (arrived-disposal-panic teardown work). The fixture arms a factory whose disposal panics and then waits for the panic to be *folded* into the teardown accumulator. PR #318 already had to add a real join to the sibling fixture `scope_with_arrived_factory_disposal_panic` for a related ordering reason:

> `ChildEvent::Exited` is emitted from a separate task that awaits `runtime::join(handle)`, and tokio resolves a `JoinHandle` only after the task future — and hence the factory clone — has been dropped.

So the likely shape is the same: the test observes the disposal outcome before the owning task has actually been joined, and under load the ordering it depends on is not guaranteed. Worth checking whether these two tests need the same real-join synchronization #318 applied to their sibling.

## Why it matters

A load-sensitive failure in a *panic-folding* test is the kind that gets re-run and dismissed. It should either be synchronized properly or, if the flake reflects a genuine ordering hole in teardown, escalated.

ralexstokes:
## Summary

- replace the fixture's untyped disposal-lane `is_empty` polling with an exact `ConstructionDisposed` receive
- verify that the completion belongs to the expected child and carries the expected factory-drop panic
- pin that the terminal reports exactly one construction disposal, then replay the validated completion — pulse included — so runtime teardown, batch collection and hard force still exercise the production arrived-event paths

## Why

The reported tests depended on a timing observation: any non-empty disposal lane was treated as proof that the intended factory-disposal panic had arrived. It is not. `SpawnBody::TaskRestartable` holds `Arc::clone(factory)`, so the spawned task and the retained `ChildConstruction` co-own the factory; if the aborted task has not released its clone by the time `handle_terminal_disposal` runs, the retained drop is a non-last refcount decrement, `ConstructionDisposed` still arrives, and it carries `panic: None`. The old fixture accepted that and failed later, at the teardown assertion, while the real panic surfaced on an unrelated worker thread.

The post-join synchronization that landed through #298 removes that ownership race — #325 was filed 92 seconds before #298 merged, so both reports were against trees without it. This change does not alter any test outcome; it makes the arrival boundary exact so a recurrence names itself at the fixture instead of at teardown.

Closes #325.

## Validation

- affected three disposal-panic shutdown tests
- mutation check: with the `recv_child_exit` join deleted, the new assertion fires 117/200 runs (2 CPUs under load) at `the retained factory disposal panics`; with the join restored, 0/200
- `main` unmodified: 250 consecutive full `-p shelterwood` runs on 2 CPUs under load, 0 failures — the flake mode is already gone, this is the guard
- 100 consecutive full `cargo nextest run -p shelterwood` passes after the change (53,600 tests)
- `./tools/dev just ci`
- `./tools/dev just ci-nix`

### DIFF
--- a/crates/shelterwood/src/driver/tests/shutdown.rs
+++ b/crates/shelterwood/src/driver/tests/shutdown.rs
@@ -74,13 +74,28 @@ async fn scope_with_arrived_factory_disposal_panic() -> (ScopeRuntime, ChildKey,
     );
     assert!(scope.children[key].pending_terminal.is_some());
 
-    let arrived = crate::runtime::timeout(DRIVER_PROGRESS_WAIT, async {
-        while scope.disposal_event_receiver.is_empty() {
-            crate::runtime::yield_now().await;
-        }
-    })
+    let (disposed_child, panic) = recv_construction_disposed(
+        &mut scope.disposal_event_receiver,
+        DRIVER_PROGRESS_WAIT,
+        "the retained factory disposal panic to arrive",
+    )
     .await;
-    assert!(matches!(arrived, crate::runtime::Timeout::Completed(())));
+    assert_eq!(disposed_child, key);
+    let panic = panic.expect("the retained factory disposal panics");
+    assert_eq!(
+        panic.message.as_deref(),
+        Some(ARRIVED_FACTORY_DISPOSAL_PANIC)
+    );
+    assert!(
+        scope
+            .disposal_events
+            .send(DriverEvent::Child(ChildEvent::ConstructionDisposed {
+                child: disposed_child,
+                panic: Some(panic),
+            }))
+            .is_ok(),
+        "the validated disposal completion returns to the live lane"
+    );
     (scope, key, member)
 }
 
--- a/crates/shelterwood/src/driver/tests/shutdown.rs
+++ b/crates/shelterwood/src/driver/tests/shutdown.rs
@@ -87,6 +87,13 @@ async fn scope_with_arrived_factory_disposal_panic() -> (ScopeRuntime, ChildKey,
         Some(ARRIVED_FACTORY_DISPOSAL_PANIC)
     );
     assert!(
+        scope.disposal_event_receiver.is_empty(),
+        "a terminal disposes its retained construction exactly once"
+    );
+    // Replay the validated completion, pulse included, so runtime teardown,
+    // batch collection and the hard-force fallback still meet it exactly as
+    // `handle_terminal_disposal` leaves it.
+    assert!(
         scope
             .disposal_events
             .send(DriverEvent::Child(ChildEvent::ConstructionDisposed {
@@ -96,6 +103,7 @@ async fn scope_with_arrived_factory_disposal_panic() -> (ScopeRuntime, ChildKey,
             .is_ok(),
         "the validated disposal completion returns to the live lane"
     );
+    scope.root.signal().pulse();
     (scope, key, member)
 }
 


ralexstokes:
Reviewed, reproduced the underlying race, and pushed two small follow-ups to the branch (e88563d). The change is correct and worth landing, with one framing correction that I already applied to the PR body.

## The mechanism, confirmed

`SpawnBody::TaskRestartable` carries `factory: Arc::clone(factory)` (`driver/child.rs:445`), so the spawned task and the retained `ChildConstruction` are two owners of the same factory `Arc`. If the aborted task has not released its clone by the time `handle_terminal_disposal` calls `dispose_then`, the retained drop is a non-last refcount decrement: `PanickingFactoryDrop` never runs, `ConstructionDisposed` still lands on the lane, but with `panic: None`. The old `is_empty()` wait accepted that event as proof of arrival, so the test ran on and failed at `assert_arrived_disposal_panic_published`, while the real panic surfaced later on an unrelated worker thread — exactly the "gets re-run and dismissed" shape #325 describes.

Mutation check, on this branch with the `recv_child_exit` join deleted, three tests, 2 CPUs under load, 200 runs:

- **117/200 failed**, every one at `shutdown.rs:79: the retained factory disposal panics`, each paired with a stray `tokio-rt-worker ... panicked at shutdown.rs:17: arrived factory disposal panic`
- join restored (this PR as written): **0/200**

So the new assertion is not decorative — it names the reported failure mode precisely — and the post-join wait is the load-bearing synchronization.

## It was #298, not #318

The join came in with cf2b90d ("Contain teardown waker panics"), i.e. #298. The timing settles it: #325 was filed at 03:07:46Z and #298 merged at 03:09:18Z, 92 seconds later. Both reporters were on trees that predated the join, which is why neither has seen it since.

Consequence: after the fixture returns, the lane holds the same single event as before — received, validated, replayed — so nothing downstream can observe a difference and **this change cannot alter a test outcome**. It converts a silent wrong verdict into an exact, early failure. That is worth having, but "removes a load-sensitive fixture assumption" overstated it, so I rewrote the PR body to say what it does and to record the mutation evidence.

`Closes #325` still holds, and it closes out the issue's escalation branch too: this is not an ordering hole in teardown, it is factory-`Arc` co-ownership, already fixed.

## Pushed in e88563d

1. `assert!(scope.disposal_event_receiver.is_empty(), ...)` after the receive — pins "exactly one construction disposal per terminal", which is the invariant `drain_arrived_disposal_events`' `debug_assert!` already assumes.
2. `scope.root.signal().pulse()` beside the replayed send. `handle_terminal_disposal` pairs a pulse with every lane send (`child.rs:1213`); omitting it is harmless for these three tests, which drive the scope directly, but leaves a fixture that looks production-faithful and would hang a future test that waits on the signal.

I left `assert!(... .is_ok(), "...")` alone — `DriverEvent` has no `Debug`, so `.expect` will not compile there.

`./tools/dev just ci` passes on the branch.

## Unrelated, filed separately

Stressing unmodified `main` (full suite, 4 CPUs, 120 runs) turned up a different flake, 1/120: `completion_during_the_tidy_beat_is_not_reclassified_as_abort` — `tests/lifecycle.rs:777: the policy does not pre-decide the classification: Exit { kind: Aborted { phase: WithinGrace }, cancellation: Observed }`. Filed as #330; not touched here.

Issue #348, opened 2026-08-19
Tracking: 2026-08-19 full review — zero bugs; ranked test-gap/quality/simplification backlog


ralexstokes:
Tracking issue for the 2026-08-19 full-codebase review (13 parallel reviewers: 10 module reviewers covering every source file, plus api-simplicity, test-coverage, and lock-rule auditors, with an adversarial verification stage).

**Headline: zero bug claims.** For the first review round, no finding rose to a reachable correctness bug — every suspicious path traced resolved to documented, test-pinned, or structurally-enforced design. The lock-rule audit of every production lock site found one acyclic order and no ABBA path; the test audit found no vacuous test. All 44 findings are test-gaps, quality items, and simplifications.

## Filed issues, ranked

P2 — significant:
- [ ] #332 publish_drain marker-before-Draining order: the #270 regression guard is prose only (structural fix preferred)
- [ ] #333 mutation-invisible mailbox cell behaviors (receive isolation, latest displacement, observe_all)
- [ ] #334 delete ResidentChild's unreachable armed Drop + disarm choreography (~50 lines + deadlock trap)
- [ ] #335 adjudicate + pin finish_if_ready from Starting
- [ ] #336 untested policy branches: OneShot resolve_common (silent restart override), resolve_child_mailbox arms
- [ ] #337 collapse admission_control three-layer forwarding
- [ ] #338 consolidate mailbox test doubles (~100 lines)
- [ ] #339 withdrawn-arm panics under held guards (verdict-release-panic convention)

P3 — batches:
- [ ] #340 cells gate-held retirement deviations (5 items)
- [ ] #341 runtime residuals (5 items)
- [ ] #342 raw context residuals (2 items)
- [ ] #343 facade cleanups (4 items)
- [ ] #344 mailbox residuals (4 items)
- [ ] #345 cells residuals (4 items)
- [ ] #346 core residuals (4 items)
- [ ] #347 test-style consistency (2 items)

## Recorded with no action proposed

[...]

PR #354, opened 2026-08-19
Address core findings from the full review

### LINKED ISSUE #335: finish_if_ready can return Finished while the lifecycle is still Starting — adjudicate the edge and pin it
**Priority: P2 (adjudicate, then pin).** From the 2026-08-19 full review.

The non-draining arm of `ScopeLifecycle::finish_if_ready` (`crates/shelterwood-core/src/engine.rs:905`) gates only on `!startup_failed() && Ordered && has_children && all_terminal`. In the `Starting` state, `startup_failed()` is false, so an Ordered scope whose registered children are all terminal before `complete_startup()` runs would get `Some(StopReason::Finished)` mid-startup — e.g. child 1 of an ordered spawn sequence completes with `RestartCondition::Never` before later children register. Whether that interleaving is reachable depends on the driver, which is outside this file.

The embedded tests exercise `finish_if_ready` from `StartupFailed`, `Draining`, and `Running`, but never from `Starting`, so this edge of the finish predicate is entirely unpinned: if finishing mid-startup is intended, no test documents it; if it is not, nothing catches the driver ever exposing it.

**Remedy.** Adjudicate the intended behavior (SPEC is adjustable per the standing protocol), then add a test asserting `finish_if_ready` in the `Starting` state with `has_children && all_terminal` — and if finishing mid-startup is not intended, add `!self.is_starting()` to the gate.

### LINKED ISSUE #336: resolve_common's ChildMode::OneShot branch and resolve_child_mailbox's override arms are untested
**Priority: P2.** From the 2026-08-19 full review. Two policy-resolution branches in `crates/shelterwood-core/src/policy.rs` have zero coverage.

- [ ] **`resolve_common`'s `ChildMode::OneShot` branch (~line 960).** It forces `RestartPolicy::Never` for OneShot children (silently discarding any explicit `options.restart` the caller supplied) and defaults retention to `Remove` instead of `Retain`. Both tests that call `resolve_common` (`default_overlay_and_child_resolution_share_inheritance_rules`) use only `ChildMode::Restartable`, so the OneShot branch — including the silent override of an explicit restart option, the kind of quiet policy substitution a refactor could easily break — is untested. Test: `resolve_common` with `ChildMode::OneShot` asserting restart is `Never` even when `options.restart` is `Some(Always)`, and retention defaults to `Remove` while an explicit `Retain` override still wins.
- [ ] **`resolve_child_mailbox`'s explicit-override arms (~line 906).** Of its four arms, tests only exercise the `None` arm (through `resolve_common` with `mailbox: None`) plus capacity-by-kind inheritance via overlay. The `Some(Queue(Some(cap)))`, `Some(Queue(None))`, and `Some(Latest)` child-level override arms — including the case where a child declares queue_inherit under a Latest scope default and must pick up the retained `queue_capacity` — have no direct coverage.

### LINKED ISSUE #346: shelterwood-core residuals: redundant IntensityTrip parameter, unsatisfiable check_invariants assert, sibling-crate-only panic pins, stale exit.rs boundary doc
**Priority: P3 batch.** From the 2026-08-19 full review of `crates/shelterwood-core`.

- [ ] **`IntensityTrip::new` takes a redundant policy parameter** (`engine.rs:288`): `IntensityCharge` stores its own `policy` field and the only production call site passes `self.charge.policy` alongside `self.charge`, so the two arguments can never legitimately disagree; the extra parameter creates a spurious degree of freedom tests also have to thread. Change to `IntensityTrip::new(charge: IntensityCharge)` and read `charge.policy` internally.
- [ ] **`check_invariants` asserts an unsatisfiable-by-construction predicate** (`supervisor.rs:696`): `assert!(!child.startable() || !child.state.joined())` can never fire for any value of `ChildRecord` — both predicates derive from the single `state` field, `startable()` requires `Resident(Unstarted | RestartPending)`, `joined()` requires incarnation `Joined`. The comment claims it checks the emission/acceptance agreement, but that agreement is actually checked non-vacuously by the exploration walk's `check_r5_effects_are_acknowledgeable` and `start_effects_are_confined_to_the_spawn_transition`. Per the repo's recorded stance on vacuous proof checks, a check that cannot fail misleads readers about where the invariant is enforced. Delete it (and its comment) or replace with a pointer comment to check_r5.
- [ ] **Core panic-resumption variants pinned only by a sibling crate** (`panic.rs:81`): panic.rs's own tests cover during-unwind containment, hostile-payload discard, and PanicAccumulator drop/take, but not `resume_preferred_panic_outside_unwind`, nor `resume_preferred_panic`'s primary-over-cleanup precedence outside an unwind — those are pinned only in `crates/shelterwood/src/raw/context.rs` tests, so a `cargo test -p shelterwood-core` run would pass a regression. Move or duplicate the two precedence tests (~20 lines, no runtime machinery) into panic.rs's own test module.
- [ ] **Stale `#[doc(hidden)]` and boundary doc on `structured_intensity_trip_error`** (`exit.rs:173`): the fn is private, so `#[doc(hidden)]` is inert, and its doc describes the situation of its pub twin — for the intensity variant the authentication boundary IS a privacy rule again (only `stop_reason_into_nested_result`/`stop_reason_root_exit` can mint it), strictly stronger than the comment claims. Only `structured_startup_failure_error` genuinely needs `pub` (used at `crates/shelterwood/src/driver.rs:899`). Drop the attribute and reword.

ralexstokes:
## Summary

- require ordered natural completion to wait until aggregate startup has completed, and pin the `Starting` edge in both the lifecycle unit test and reducer exploration
- cover every child mailbox override plus the one-shot restart/retention resolution branches
- complete the core residual cleanup: collapse the intensity-trip constructor input, remove the vacuous supervisor assertion, add local panic-precedence pins, correct structured-error boundary docs, and make `NeverStarted`'s dispatch exclusion assert-loud

### #335: `Starting` does not naturally finish

I ruled that an ordered scope must not report `Finished` until aggregate startup has completed. The production driver currently makes the suspicious state structurally difficult to reach: it admits the entire initial plan before the first settle, `Settle` runs startup progression before finish progression, and a terminal pre-ready exit leaves `Starting` through startup failure. However, accepting natural completion directly from `Starting` would still contradict the readiness barrier and would make a future driver refactor able to publish `Stopped { reason: Finished }` without ever publishing `Running`.

The predicate now requires `startup_complete()`, SPEC section 11 states that ordering explicitly, and the exhaustive reducer property rejects a non-draining `Finished` edge unless the lifecycle has completed startup.

### #346: `NeverStarted` is outside dispatch's domain

I chose the assert-loud hygiene option from the issue comment. `dispatch_exit` now has the same `debug_assert!` domain guard as the stop-reason twin, and its matrix tests only the ten constructible incarnation-exit/cancellation combinations. A direct debug assertion test pins the boundary. No defensive terminal arm was added because that would silently accept an invalid membership-level input.

No checklist items were omitted.

## Validation

- `./tools/dev cargo test -p shelterwood-core` (89 unit tests and 9 doc tests passed)
- `./tools/dev just ci` (format, clippy, 771 nextest tests, doc tests/docs, API reachability, packaged docs, external consumer, and Nix formatting passed)

Closes #335
Closes #336
Closes #346

Tracking: #348

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -2243,7 +2243,8 @@ cooperative cancel → grace expiry → tidy-abort beat → hard abort
 5. **S5 — completion is derived and level-triggered.** `all_children_joined`
    is derived from child states. `Settle` emits `Finished` once iff the
    lifecycle’s flavor-specific finish predicate accepts that derived value.
-   (`supervisor::derived_completion_property_matches_the_child_states`,
+   (`engine::scope_lifecycle_owns_first_failure_drain_status_and_finish_policy`,
+   `supervisor::derived_completion_property_matches_the_child_states`,
    `supervisor::exhaustive_reachable_states_preserve_the_reducer_invariants`.)
 6. **S6 — shutdown requests are sampled latches.** Scope shutdown and removal
    are synchronous, idempotent latches sampled into reducer events at step
@@ -2646,13 +2647,13 @@ or lifetime paragraph is unmapped.
   was requested concurrently elsewhere);
   `wait_started()` resolves once at startup and cannot observe a
   later trip, so `wait()` is the post-startup observation point. Natural
-  completion is pinned exactly: an **ordered** scope *finishes* when it
-  has at least one membership and every membership is terminal (a
-  retained terminal child counts — §8's retention is observability, not
-  liveness); a root parked in `StartupFailed` is exempt — it never
-  finishes (the park rule above). The finishing test runs **at each membership's
-  terminalization, strictly before retention-based pruning** removes it
-  (§8's remove-on-terminal default), and its result is latched — so
+  completion is pinned exactly: once aggregate startup has completed, an
+  **ordered** scope *finishes* when it has at least one membership and every
+  membership is terminal (a retained terminal child counts — §8's retention
+  is observability, not liveness); a root parked in `StartupFailed` is exempt —
+  it never finishes (the park rule above). The finishing test runs **at each
+  membership's terminalization, strictly before retention-based pruning**
+  removes it (§8's remove-on-terminal default), and its result is latched — so
   pruning the final one-shot membership can never turn a finished
   workload into an idling empty scope, and pruning order is otherwise
   unobservable. The scope then publishes `Stopped { reason: Finished }` and, when
--- a/crates/shelterwood-core/src/engine.rs
+++ b/crates/shelterwood-core/src/engine.rs
@@ -7,7 +7,7 @@ use std::{
 };
 
 use crate::{
-    Exit, GracePhase, Intensity, IntensityTrip, JitterSample, Readiness, RestartAttempt,
+    Exit, ExitKind, GracePhase, Intensity, IntensityTrip, JitterSample, Readiness, RestartAttempt,
     RestartCount, RestartPolicy, Shutdown, TotalRestarts,
     deadline::Deadline,
     exit::{StopReason, stop_reason_precedence},
@@ -251,6 +251,10 @@ pub fn dispatch_exit(
     scope: ScopeMode,
     membership: MembershipStatus,
 ) -> ExitDispatch {
+    debug_assert!(
+        !matches!(exit.kind(), ExitKind::NeverStarted),
+        "NeverStarted is a membership outcome outside incarnation dispatch"
+    );
     if scope == ScopeMode::Draining || membership == MembershipStatus::Removing {
         return ExitDispatch::Terminal;
     }
@@ -285,7 +289,8 @@ struct IntensityCharge {
 }
 
 impl IntensityTrip {
-    fn new(policy: Intensity, charge: IntensityCharge) -> Self {
+    fn new(charge: IntensityCharge) -> Self {
+        let policy = charge.policy;
         debug_assert_eq!(charge.tripped, charge.in_window > policy.max_restarts());
         Self {
             max_restarts: policy.max_restarts(),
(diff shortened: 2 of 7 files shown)


ralexstokes:
Adversarial review complete: no confirmed findings and no code changes required.

I checked #335 end to end: natural completion now requires aggregate startup completion; reducer Settle advances startup before finish evaluation; driver effect flushing publishes StartupCompleted before recording Finished; the SPEC exact rule and exploration invariant agree with that ruling.

I checked #336 branch-by-branch: None, Queue(inherit), Queue(explicit), and Latest child mailbox resolution are all pinned, including retained queue capacity across a Latest default; OneShot pins explicit restart override to Never, default Remove retention, and explicit Retain precedence.

I checked every main-body and comment item in #346: the IntensityTrip constructor degree of freedom is removed; the vacuous supervisor assertion is gone while the non-vacuous checks remain; both panic precedence behaviors are locally pinned in core; the structured-error boundary docs are accurate; the inverted NeverStarted matrix commentary is removed; and dispatch excludes NeverStarted with the chosen debug-assert hygiene. The panic test is correctly cfg(debug_assertions), so release builds do not expect a debug assertion; the production funnel remains structurally closed as documented.

Validation:
- ./tools/dev cargo test -p shelterwood-core — 89 unit + 9 doc tests passed
- ./tools/dev cargo test -p shelterwood-core --release — 88 unit + 9 doc tests passed (expected debug-only domain test excluded)
- ./tools/dev just ci — passed, including 771 nextest tests
- GitHub CI — passed

PR #355, opened 2026-08-19
Contain raw teardown panics and freeze drain boundaries

### LINKED ISSUE #342: raw context residuals: try_recv skips the local receiver freeze recv insists on; materialize_event's cancelled-after-queued disposal is covered only racily
**Priority: P3 batch.** From the 2026-08-19 full review of `crates/shelterwood/src/raw`.

- [ ] **`try_recv`'s drain path relies on remote freeze ordering that `recv` refuses to depend on** (`raw/context.rs:1130`): `recv()` locally calls `self.receiver.freeze()` on both stop branches, with a comment that correctness of the receive boundary must not depend on the driver freezing the mailbox before firing the shutdown latch. `try_recv()`'s stopping branch calls only `freeze_and_report()` (resources), never `self.receiver.freeze()` — its Drain-mode read is bounded to the accepted prefix only because (a) `local_stop` is fired solely by `stop()`, which freezes the receiver first, and (b) the driver freezes the mailbox (driver/child.rs ~853) before `advance_ladder` fires the latch. If a future driver change fired the latch before freezing, `try_recv` would silently drain messages accepted after the stop observation, widening SPEC §5.1's close point — `recv` would not. The `try_recv` doc also overstates: it says the drain "freezes first" and reads "the frozen accepted mailbox prefix," but the function freezes only resources. Remedy: add `self.receiver.freeze()` to `try_recv`'s stopping branch (idempotent), making the doc claim locally true.
- [ ] **`materialize_event`'s cancelled-after-queued disposal branch has no deterministic test** (`raw/context.rs:1409`): the branch where a queued offload completion's cancellation latch is already fired at delivery time — so the event is disposed through `RawDisposal` instead of its `make_message` running — is only exercised racily (`tests/offloads.rs dropping_scoped_guard_suppresses_the_continuation` is nondeterministic about which suppression path runs). Unit tests hit event disposal only via freeze/clear, never via the materialize path, and none asserts that a hostile destructor inside a cancelled queued event is contained and retained at that site. Test: using `bound_raw_context`, push a `QueuedEvent` whose cancellation latch is pre-fired and whose `make_message` would panic/record if invoked; call `try_recv`; assert the event is disposed (destructor ran, message not delivered, panic contained in the disposal slot if hostile).

### LINKED ISSUE #350: A hostile finished-latch waker panic escapes RawResources::freeze, stranding later offloads and parking the incarnation in join_offloads
**Confirmed bug (external claim, adversarially verified with an end-to-end repro; P2 — blast radius is the offending incarnation plus whoever awaits its exit).** Belongs with the hostile-waker cluster (#265-#267 lineage).

`RawResources::freeze` (`crates/shelterwood/src/raw/context.rs:699-715`) clears `accepting` (:703), then walks `for offload in &mut self.offloads { offload.cancel(); }` (:707-709) with no containment. `OffloadResource::cancel` (`raw/offload.rs:266-280`) fires the cancellation latch, `state.cancel()` (which fires `finished`), `task.abort()`, and `finished.fire()`. `Latch::fire` -> `WaiterRegistry::wake_all` contains each waker behind a `PanicAccumulator` — but the accumulator's Drop *resumes* the first payload on a non-unwinding thread (`panic.rs:118-125`). So one panicking waker registered on the first offload's public `Guard::finished()` (`offload.rs:59`, supported API) unwinds out of `Latch::fire`, out of `cancel`, and out of the freeze loop.

**Consequences (all reproduced):**
- Every later `OffloadResource` is left uncancelled and un-aborted, and `continuations.clear()`, `timers.clear()`, `ready_batch = None`, `events.clear()`, and the dropped-continuation report after the loop are all skipped.
- The Drop retry cannot save it: `Drop for RawResources` (:762-775) calls `catch_panic(|| self.freeze())`, and freeze returns immediately because `accepting` was already cleared — exactly as the external claim stated.
- Worse than the claim stated: the very next epilogue step, `join_offloads` (:741-759), awaits `task.join()` on the never-aborted task, so **the incarnation parks and publishes no exit at all** until something fires the *parent* shutdown latch. A self-exiting child stalls its own termination indefinitely. (Repro: hostile-waker fixture shows no `Exited` edge for 1.5s and a still-alive second offload; only an external `system.shutdown()` unblocks it. Benign-waker control passes in 3ms.)
- One limb of the original claim is overstated: "futures can miss RawDisposal" is wrong — when the parent shutdown eventually cancels, the work drops inside `SharedOffloadFuture::poll`'s `catch_panic` and still routes through `RawDisposal` (`offload.rs:229-254`). Stalled cancellation is the failure mode, not uncontained disposal.

**Spec basis:** SPEC.md:991 and :1180-1182 require offloads to be cancelled and joined at the freeze; SPEC.md:428-431 and §13.18 require waker panics to be accumulated and resumed only after cleanup completes. Freeze currently resumes the wake panic before the rest of cleanup. The comment at context.rs:705-706 ("a failure cannot prevent later offloads from being cancelled") is true only of *destructor* panics (those go through `RawDisposal::dispose` -> `catch_panic`); it says nothing about latch fires, and the existing freeze tests all inject destructor panics, so none covers this path.

**Remedy (three places):**
- [ ] `OffloadResource::cancel`: run `cancellation.fire()`, `state.cancel()`, `task.abort()`, `finished.fire()` each behind `PanicAccumulator::run`, taking the payload rather than resuming (the `Latch::fire_silently`/`notify` split in sync.rs:196-210 is the other idiomatic tool).
- [ ] `RawResources::freeze`: run the whole loop and the four collection clears behind one accumulator, then hand the payload to `self.disposal.panic` (the epilogue already drains it via `take_resource_panic`) instead of unwinding out of freeze — this also removes the dependence on the `accepting`-disarmed Drop retry.
- [ ] Fix the stale containment comment at context.rs:705-706.

Regression test: the two-offload fixture — first offload's `finished()` polled once with a panicking `Wake` impl and the future forgotten, second offload's work future flagging its Drop; assert the second offload is cancelled at freeze and the incarnation still publishes its own exit.

ralexstokes:
## Summary

- contain every offload-cancellation effect independently and accumulate failures across the entire raw-resource freeze
- retain the first freeze failure in the incarnation disposal slot only after all later offloads and raw collections have been cleaned up
- make `try_recv` establish its own mailbox freeze boundary before draining
- add deterministic coverage for a cancelled queued completion with a hostile captured destructor
- add an end-to-end hostile `Guard::finished()` waker regression proving a later offload is cancelled and the incarnation publishes its exit without parent shutdown

Closes #350.
Closes #342.

## Design decisions and omissions

- `OffloadResource::cancel` returns its first accumulated panic payload to its owner. `RawResources::freeze` owns the cleanup-wide accumulator and therefore controls when that payload becomes observable.
- The completed freeze records its first direct cleanup panic into the existing `PanicSlot` rather than pulsing `RawDisposal`. The actor is already synchronously entering its owned epilogue, which immediately drains the slot before joining; another wake is unnecessary and would introduce a new hostile-waker surface during teardown.
- Existing `PanicSlot` first-wins behavior preserves an earlier offload/destructor diagnostic ahead of a later wake failure.
- `try_recv` freezes the receiver before resources. The mailbox operation is idempotent and matches `recv`'s local containment boundary without depending on driver/latch ordering.
- The hostile-waker regression intentionally forgets one polled `Guard::finished()` future so its waker remains registered until teardown; this models the supported abandoned-waiter path deterministically and leaks only that bounded test fixture allocation.
- No disposal routing was changed for cancelled offload work: as noted in #350, `SharedOffloadFuture` already routes those futures through `RawDisposal`; the defect was stalled cleanup and joining, not missed disposal.

## Validation

- `./tools/dev cargo test -p shelterwood raw::context::tests --lib`
- `./tools/dev cargo test -p shelterwood --test offloads hostile_finished_waker_cannot_strand_later_offload_or_exit_publication -- --exact --nocapture`
- `./tools/dev cargo test -p shelterwood --test offloads`
- `./tools/dev just ci`

### DIFF
--- a/crates/shelterwood/src/raw/context.rs
+++ b/crates/shelterwood/src/raw/context.rs
@@ -702,15 +702,24 @@ impl<M> RawResources<M> {
         }
         self.accepting = false;
         let dropped_continuations = self.continuations.len();
-        // Cancellation catches each future's destructor independently, so a
-        // failure cannot prevent later offloads from being cancelled/signalled.
+        // Treat the whole freeze as one cleanup transaction. Cancellation
+        // contains each latch wake, future disposal and task abort
+        // independently, while this outer accumulator keeps one failure from
+        // skipping later offloads or any of the collection drains.
+        let mut panics = PanicAccumulator::default();
         for offload in &mut self.offloads {
-            offload.cancel();
+            panics.record(offload.cancel());
         }
-        self.continuations.clear();
-        self.timers.clear();
-        self.ready_batch = None;
-        self.events.clear();
+        panics.run(|| self.continuations.clear());
+        panics.run(|| self.timers.clear());
+        panics.run(|| self.ready_batch = None);
+        panics.run(|| self.events.clear());
+        if let Some(payload) = panics.take() {
+            // The owned raw-incarnation epilogue drains this slot after the
+            // freeze and before joining, so publication is delayed until all
+            // synchronous cleanup has completed without losing the diagnostic.
+            self.disposal.panic.record(payload);
+        }
         dropped_continuations
     }
 
@@ -1129,6 +1138,10 @@ impl<M: Send + 'static> RawContext<M> {
     /// resuming the loop.
     pub fn try_recv(&mut self) -> Option<M> {
         if self.is_stopping() {
+            // Establish the receive boundary locally just as `recv` does. Do
+            // not rely on the driver's mailbox-freeze ordering relative to
+            // the shutdown latch this call observes.
+            self.receiver.freeze();
             self.freeze_and_report();
             self.resources.resume_pending_panic();
             self.receiver.try_recv()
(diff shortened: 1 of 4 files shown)


ralexstokes:
Adversarial follow-up review found and fixed one precedence defect in the draft: an early offload latch-wake panic could be displaced by a later collection destructor retained directly in PanicSlot, and a future destructor could likewise lose to its subsequent finished wake. Commit 9dd3804 folds slot-retained failures into the cleanup-wide accumulator at each boundary, restores the established winner after cleanup without dropping a displaced payload under the slot mutex, and adds deterministic tests for both orderings plus repeated-freeze/drop behavior. Lock-rule audit found no guarded user drop/wake in the resulting paths; cancellation still fires, cancels state, aborts, and finishes independently, and the existing end-to-end test proves later-offload disposal and exit publication. Validation: raw context unit tests (24/24), offloads integration tests (32/32), targeted hostile-waker exit test, and ./tools/dev just ci (772/772 plus all other lanes) passed. No canonical review threads existed to resolve.

PR #356, opened 2026-08-19
Harden cells observation transactions

### LINKED ISSUE #332: publish_drain's marker-before-Draining order — the #270 regression guard — is a comment, not a test or structure
**Priority: P2 (regression guard for a shipped bug).** From the 2026-08-19 full review; flagged independently by two auditors.

`ScopeCell::publish_drain` (`crates/shelterwood-cells/src/cells/scope.rs:526`) must store every terminal-disposal marker (`set_terminal_disposal_pending(true)`) before `set_state_locked` writes `ScopeState::Draining`. The order is load-bearing: a zero-budget shutdown sampler on another thread reads the record value through the watch mutex (visible before the gate is released), then samples the markers; the inverted order reopens #270 — the sampler pairs `Draining` with a false marker and a nonterminal child record and reports a spurious `Err(ShutdownTimeout)`.

The comment at scope.rs:532-534 states outright: "No test holds this order — moving the loop below `set_state_locked` leaves the suite green — so this comment is the only guard." By this repo's own bar (mutation-check pinned claims), a comment-only guard on a race that was a shipped bug is a coverage hole.

**Remedy (preferred: structural).** Fold the marker loop into `set_state_locked` (or a single `publish_drain_locked` helper that takes the `terminal_disposals` slice), so the store-markers-then-write-Draining sequence lives in one function no call site can reorder.

**Remedy (weaker: pin).** A deterministic test — e.g. a test-util seam invoked by `set_state_locked` before the record write, mirroring the existing `gate_capture_probe` pattern — that lets a sampler thread read the scope record and each marker between the two writes, asserting any observation of `Draining` implies every terminal-disposal marker is already set. Verify the test fails when the loop is moved below `set_state_locked`.

### LINKED ISSUE #334: ResidentChild's armed Drop fallback is unreachable yet forces the disarm choreography and the gate self-deadlock trap
**Priority: P2 (simplification, ~50 lines plus a documented deadlock trap).** From the 2026-08-19 full review; structure spot-checked.

The gate-taking branch of `ResidentChild::drop` (`crates/shelterwood-cells/src/cells/scope.rs:106`) can never execute. `ResidentChild` is private to scope.rs and lives only in `observation.current_children`; the exits from that Vec are:

- `prune_child_locked` — consumed through `complete_removal`, disarmed first;
- `clear_residents_locked` — `mem::take`, all disarmed before any fallible emit, then deferred drop;
- `ScopeCell`'s own drop glue — where `completion.parent` is `Arc::downgrade` of the very cell being dropped, so the upgrade returns `None` and the branch no-ops.

The only hypothetical armed drop (`Vec::push` failing in `admit_child_locked`) is an allocation abort, not an unwind.

What the fallback costs: the `ResidencyCompletion` struct with its duplicate projection clone, the `disarm_removal`/`complete_removal`/`publish_removal` choreography, and the documented hazard (scope.rs:161-165) that dropping a resident anywhere the gate is held self-deadlocks — a defense whose failure mode (deadlock) is worse than the omission it defends against (a missing `Removed` event).

**Remedy.** Delete the `Drop` impl and `ResidencyCompletion`; emit `Removed` directly in `prune_child_locked` and `clear_residents_locked` (the two real removal paths), keeping the deferred drop of displaced residents for last-mailbox-owner release. If a tripwire against future forgotten removal paths is still wanted, a `debug_assert` in `ResidentChild::drop` (removal already disarmed) preserves the check without ever taking the gate.

### LINKED ISSUE #340: Gate-held retirement deviations in cells: inline drops despite a sink in hand, a convention-held subscribe, an uncommented eviction drop
**Priority: P3 cluster (quality — all legal today, all deviations from the codebase's own stated preferences).** From the 2026-08-19 full review.

- [ ] **`publish_stopped_locked`'s suppressed-repeat branch** (`crates/shelterwood-cells/src/cells/scope.rs:1292-1297`): `drop(transient_retained)` runs each `RetainedExit::drop` inline while the observation gate is held; on this branch the guards are the last owner of the incoming reason's failed exit, so each drop submits a `dispose_critical` job — which can start a native thread — under the gate. CLAUDE.md classifies submission-under-lock as legal but states the preference: "a caller that already owns an effects sink should still flush it." The caller has `wakes` in hand. Remedy: `wakes.defer(move || { drop(state); drop(transient_retained); })`, preserving the raw-projection-first field order.
- [ ] **`set_startup_locked`'s `drop(incoming)`** (`scope.rs:773`) is the same shape for a rejected startup result. Remedy: `txn.defer(move || drop(incoming))`.
- [ ] **`LifecycleHub::publish` inline drop of an undelivered event** (`crates/shelterwood-cells/src/observe.rs:950`): when the hub is closed or receiverless, the `RetainedLifecycleEvent` is dropped inline under the gate. Benign today only because the sole production caller, `emit_locked`, defers a sibling clone of the event past the gate, so the guards Arc is provably not last-owner — a distant invariant a future direct caller would silently break. Remedy: `txn.defer(move || drop(undelivered))`.
- [ ] **`LifecycleHub::subscribe` holds its gate requirement by convention** (`observe.rs:919`) while `SnapshotHub::subscribe` made the identical requirement structural via a `&mut ObservationTxn` witness parameter ("a static property of every caller rather than a convention"). It snapshots `signal.explicit_lag` then subscribes to the broadcast as two non-atomic steps; an ungated future caller racing `publish_lagged`/`publish` would report a spurious `Lagged` marker for events dropped before it subscribed (diagnostic-only wrongness). Remedy: give it the same txn witness parameter (unused is fine).
- [ ] **Broadcast ring eviction drop is uncommented** (`observe.rs:947`): `events.send(event)` inside `signal.modify_silently` runs under both the gate and the watch value mutex; when the 128-slot ring is full, the evicted `RetainedLifecycleEvent` drops right there and can be the last owner of a failed exit's guard, submitting `dispose_critical` (legal — a submission runs no user code — but it can start a native thread under two locks, and it is the one publish-path drop that cannot reach the effects sink). The safety argument spans three files and is stated at none of the sites where the eviction happens. Remedy: a comment at the send site naming the eviction drop, why it is refcount-plus-submission only, and that in-place submission is the accepted deviation because drop glue inside tokio's send has no sink to reach.

### LINKED ISSUE #345: shelterwood-cells residuals: LifecycleChannels wrapper, double clone in SnapshotReceiver::changed, take_request duplication, RetainedExit::install local pin
**Priority: P3 batch.** From the 2026-08-19 full review of `crates/shelterwood-cells` (beyond the issues filed separately for publish_drain, ResidentChild, and the gate-held drop conventions).

- [ ] **`LifecycleChannels` is a needless single-use wrapper** (`observe.rs:894`): `LifecycleHub`'s only field is `channels: LifecycleChannels`, which only bundles `events` and `signal` — no method, no separate ownership, no reuse; every access pays the `self.channels.` hop. Inline the two fields.
- [ ] **`SnapshotReceiver::changed` clones the hub state twice per resolution** (`observe.rs:471`): the loop does `borrow_cloned()` to test the generation and a second `borrow_and_update_cloned()` to fetch the returned value, cloning the full `SnapshotHubState` twice. One `borrow_and_update_cloned()` per iteration is equivalent — the review verified no pulse exists on the snapshot channel that bumps the version without minting a generation or closing, so eagerly marking the watch version seen cannot strand a change.
- [ ] **`take_shutdown_request`/`take_force_request` duplicate a 20-line check-then-gate-then-consume block** (`scope.rs:997-1017` and `1033-1053`): identical except for projecting `control.shutdown` vs `control.force`; a future tightening of the recheck can silently miss one copy. Extract a private `take_request` helper both delegate to.
- [ ] **`RetainedExit::install`'s replacing branch has no local test** (`retained.rs:80`): the in-file test pins only the equal-set branch. The replacing branch — a changed guard set replaces the Arc, and the displaced set, when last-owned, must route each failed payload through isolated disposal — is exercised only indirectly via member.rs transition tests (which do catch both an always-equal and an always-replace mutation, so no mutant escapes entirely), but the module that owns the invariant does not pin it. Add a retained.rs test: install a changed guard set over a last-owned previous set containing a ThreadProbe-backed failed exit; assert the new Arc differs and the displaced payload disposes off-thread.

ralexstokes:
## Summary

- make drain-entry marker-before-record publication and straggler marker-before-record sampling structural
- remove the unreachable `ResidentChild` armed-drop fallback while preserving post-unlock last-owner retirement
- defer rejected retained values through `ObservationTxn` when an effects sink is available
- tighten lifecycle subscription, simplify observation state, and add focused retained-exit/order tests
- make member diagnostics avoid formatting user errors under framework locks and document snapshot-install unwind requirements

## Design decisions

- `ScopeCell::set_state_locked` now accepts the terminal-disposal slice and stores every marker before touching the scope record. This keeps the #270 ordering inside the writer rather than behind a test-only timing seam.
- `MemberCell::terminal_or_disposal_pending` owns the inverse sampling order. The shutdown driver can no longer independently reverse the marker and record reads.
- lifecycle broadcast `send` deliberately remains under the observation gate. Deferring it would permit transaction flush order to diverge from lifecycle `seq`, invert `Added` against descendant events, and expose closure before the final event. Undelivered values are deferred; in-ring eviction remains inline and its refcount/disposal tradeoff is documented at the send site.
- resident removal schedules the displaced projection in transaction effects before emitting `Removed`, including on unwind. There is no armed `Drop` fallback or gate reacquisition.
- no checklist items are omitted.

## Validation

- `./tools/dev cargo check --workspace --all-targets --all-features`
- `./tools/dev cargo test -p shelterwood-cells --all-features`
- `./tools/dev cargo test -p shelterwood --all-features shutdown`
- `./tools/dev just ci` (769 tests; formatting, clippy, docs, API reachability, and external consumer all green)

Closes #332
Closes #334
Closes #340
Closes #345

### DIFF
--- a/crates/shelterwood-cells/src/cells/gate.rs
+++ b/crates/shelterwood-cells/src/cells/gate.rs
@@ -122,6 +122,11 @@ impl<'a> ObservationTxn<'a> {
         // "T1 unlocks, T2 stages and installs a newer cut, T1 installs its
         // stale one", leaving every ungated borrow behind the tree until some
         // later publication corrected it. SPEC §12 promises this ordering.
+        // This install segment must remain panic-free apart from unreachable
+        // invariant assertions. A panic raised here runs from `Drop`, escapes
+        // before the post-unlock accumulator, and would strand later snapshot
+        // installations plus every queued effect; the current path invokes no
+        // user code and defers generation exhaustion into `effects`.
         for publication in self.snapshots.drain(..) {
             publication.install(&mut self.effects);
         }
--- a/crates/shelterwood-cells/src/cells/member.rs
+++ b/crates/shelterwood-cells/src/cells/member.rs
@@ -8,7 +8,7 @@ use std::{
 };
 
 use shelterwood_core::{
-    ChildId, Exit, Incarnation, Membership, RestartCount,
+    ChildId, Exit, ExitKind, Incarnation, Membership, RestartCount,
     engine::MembershipStatus,
     identity::{IncarnationCounter, MintedMembership},
     policy::ResolvedCommonOptions,
@@ -27,6 +27,20 @@ pub enum MemberStage {
     Restarting,
     Stopping,
     Terminal(Exit),
+}
+
+impl MemberStage {
+    fn tag(&self) -> &'static str {
+        match self {
+            Self::Reserved => "Reserved",
+            Self::Admitted => "Admitted",
+            Self::Starting => "Starting",
+            Self::Running => "Running",
+            Self::Restarting => "Restarting",
+            Self::Stopping => "Stopping",
+            Self::Terminal(_) => "Terminal",
+        }
+    }
 }
 
 /// One non-terminal member-record transition owned by the cell layer.
@@ -105,16 +119,16 @@ impl MemberRecord {
             MemberTransition::Admitted => {
                 debug_assert!(
                     matches!(self.stage, MemberStage::Reserved),
-                    "admission must consume a fresh reservation, not {:?}",
-                    self.stage
+                    "admission must consume a fresh reservation, not {}",
+                    self.stage.tag()
                 );
                 self.stage = MemberStage::Admitted;
             }
             MemberTransition::Starting { incarnation } => {
                 debug_assert!(
                     matches!(self.stage, MemberStage::Admitted | MemberStage::Restarting),
-                    "a spawn must start an admitted or restarting member, not {:?}",
-                    self.stage
+                    "a spawn must start an admitted or restarting member, not {}",
+                    self.stage.tag()
                 );
                 self.stage = MemberStage::Starting;
                 self.incarnation = Some(incarnation);
@@ -125,16 +139,16 @@ impl MemberRecord {
                 debug_assert!(
                     matches!(self.stage, MemberStage::Starting | MemberStage::Reserved),
                     "readiness must promote a starting member (or the root scope's own \
-                     never-admitted reservation), not {:?}",
-                    self.stage
+                     never-admitted reservation), not {}",
+                    self.stage.tag()
                 );
                 self.stage = MemberStage::Running;
             }
             MemberTransition::Stopping => {
                 debug_assert!(
                     matches!(self.stage, MemberStage::Starting | MemberStage::Running),
-                    "a stop ladder must begin on a starting or running member, not {:?}",
-                    self.stage
+                    "a stop ladder must begin on a starting or running member, not {}",
+                    self.stage.tag()
                 );
                 self.stage = MemberStage::Stopping;
             }
@@ -148,8 +162,8 @@ impl MemberRecord {
                         self.stage,
                         MemberStage::Starting | MemberStage::Running | MemberStage::Stopping
                     ),
-                    "a restart must be scheduled from an active incarnation's exit, not {:?}",
-                    self.stage
+                    "a restart must be scheduled from an active incarnation's exit, not {}",
+                    self.stage.tag()
                 );
                 self.stage = MemberStage::Restarting;
                 self.incarnation = None;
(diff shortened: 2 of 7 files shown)


ralexstokes:
Adversarial review complete: no confirmed findings.

I reviewed the full base-to-head patch and the live bodies/comments for #332, #334, #340, and #345, including both external-review adjudication follow-ups. In particular:

- #332: drain entry now makes marker-before-record publication structural in `set_state_locked`; the shutdown sampler likewise owns marker-before-record sampling in one method. The Release/Acquire argument is consistent with the record-lock happens-before edge.
- #334: the unreachable armed `ResidentChild::Drop` path is gone. Prune queues the displaced resident before its fallible publication path, and clear queues the complete displaced set before emitting any `Removed` edge, so unwind still retires last mailbox/member owners after gate release.
- #340: rejected startup/stopped projections and undelivered lifecycle values leave through transaction effects. Lifecycle subscription now requires the transaction witness. Broadcast insertion correctly remains under the gate: moving it to effects would break sequence order, child `Added` causal order, and final-event-before-closure atomicity; the in-place retained eviction submission is documented as the accepted trade.
- #345 plus follow-ups: the wrapper/double clone/request duplication are removed, the replacing retained-exit branch has a direct off-thread-destruction test, member diagnostics no longer format user errors under the mailbox/watch locks, and the snapshot-install unwind asymmetry is documented.

I also checked current PR review threads (none), all call sites, retained raw-value/guard drop order, and the relevant lifecycle close/lag behavior. I made no code changes, so there is no review-fix commit to push.

Validation:

- `./tools/dev cargo test -p shelterwood-cells --all-features` — 43 passed
- `./tools/dev cargo test -p shelterwood --all-features` — passed
- `./tools/dev just ci` — 769 passed; fmt, clippy, docs, API reachability, external consumer, and nixfmt all clean
- GitHub Actions CI — passed

Reviewed head: 47dc79e35fd94cdf3e3bc804ded95c5c5afd5ce8.

PR #357, opened 2026-08-19
Order AfterInit readiness before init self-stop

### LINKED ISSUE #352: Adjudicate: stop() during init swallows AfterInit readiness, so a successful init reports scope-wide StartupFailed
**Confirmed spec/code edge needing adjudication (external claim, adversarially verified with an end-to-end repro; P2 — deterministic and loud, no corruption, one-line workaround exists).**

An `AfterInit` (default `Actor`) child whose `init` returns `Ok` after calling `context.stop()` never publishes readiness and fails the whole scope's startup — with a payload that says the child completed successfully.

**Trace (every step verified):** `Context::stop` is live during init and fires `local_stop` (`raw/context.rs:907-912`). The post-init AfterInit mark (`actor.rs:453-455`) is swallowed because `RawContext::mark_ready` (`raw/context.rs:892-895`) no-ops while `is_stopping()`. The self-stop watcher then disarms the readiness gate (`driver/child.rs:945-969` -> `ReadinessEvent::Shutdown` -> `Disarmed`), the actor exits `Completed`+`Observed`, and the pre-ready terminal predicate (`child.rs:1068-1076`, duplicated at :707-714 — see #343's second item) classifies it `StartupDisposition::Aborted` -> §6 R7 `FailStartup`: ordered suffix `NeverStarted`, root parks in `StartupFailed`.

**Repro:** ordered tree, `StopInInit` actor + sibling task: `wait_started()` -> `Err(StartupFailed(... exit: Exit { kind: Completed, cancellation: Observed } ...))`, sibling never started. Control without `stop()`: `Ok`, sibling started. Workaround verified: `mark_ready(); stop()` in init makes startup succeed (SPEC's "earliest mark wins"). With `RestartCondition::Always` the spin is bounded by intensity (`IntensityTripped` after 6 restarts).

**Why this is an adjudication, not an open-and-shut bug:** SPEC §6 says AfterInit means "ready when `init` returns `Ok`" and the blanket loop "reports readiness through the same public `mark_ready` mechanism after `init` returns `Ok`" — here init returned Ok and no readiness was reported, and Appendix B.1's `mark_ready` no-op enumeration does not include a local-stop-suppressed no-op (neither do the method docs). But §6 also aborts startup on a "terminal pre-ready exit", and this membership did end before any readiness edge. `stop()` is listed as available in the Live stage, so stop-during-init is a supported operation. The two clauses collide exactly here: this is the one mode where a *successful* init yields a scope-wide `StartupFailed` reporting success.

**Decision + remedy (in order):**
1. Adjudicate: does a pre-ready **clean self-stop** (`Completed`+`Observed`) abort startup?
2. If NO: fix race-free — either defer the *effect* of a stop taken during init until after the post-init readiness mark (so the ready latch fires strictly before `local_stop`, making `handle_self_stop`'s existing `ready_before_stop` replay deterministic), or exclude `Completed`+`Observed` self-stops from the Aborted predicate **at both child.rs:707 and :1068**.
3. If YES (current behavior intended): document the local-stop no-op on `RawContext::mark_ready`/`Context::mark_ready` and in SPEC B.1, and pin with an ordered-tree test.

**Do NOT** apply the external reviewer's proposed fix ("fire AfterInit readiness even if local_stop is already set") alone — it is racy: the driver disarms the gate when the independently-spawned self-stop watcher's event is processed, so the late latch fire wins only when the actor task beats the watcher task, making the outcome nondeterministic.

Also refuted from the original claim: "stalls later siblings" (the failure is immediate and deterministic; the suffix terminalizes `NeverStarted` per §6 R7).

ralexstokes:
Closes #352.

Tracking issue: #348.

## Semantic decision

An effective `AfterInit` callback actor whose initializer calls `Context::stop()` and then returns `Ok` is ready before its clean self-stop becomes observable to supervision. Returning `Ok` therefore keeps the `AfterInit` promise and the stop is post-ready for startup classification.

This ruling is intentionally narrow:

- intake and incarnation resources freeze at the `stop()` call;
- the pending stop counts as stopping for actor-side operations;
- after the outermost initializer returns successfully and installs actor state, the blanket handler uses the ordinary `mark_ready()` path and then fires `local_stop`, with no await between them;
- a `Manual` override preserves terminal pre-ready self-stop behavior;
- an initializer that returns `Err`, panics, or is cancelled publishes the pending stop without manufacturing readiness;
- projected decorator contexts share the pending request, while the outermost initializer owns the return boundary.

This also makes the existing `Context::stop()` wording (effective after the current callback) literal for an `AfterInit` initializer. A non-returning initializer does not arm its self-stop ladder until that callback ends; intake is already frozen, and readiness deadline, parent shutdown, and escalation remain available.

## Rejected alternatives

- **Fire readiness after `local_stop`.** Rejected as racy: the independently scheduled self-stop watcher may disarm the readiness gate before the late mark is processed.
- **Treat every `Completed + Observed` pre-ready exit as non-aborting.** Rejected because it changes unrelated `Manual` semantics and would require a second structural rule to advance an ordered cursor past a terminal-but-unready member.
- **Defer all initializer stops.** Rejected because `Manual` is explicitly how a declaration asks for pre-ready gating and failure.
- **Use a private readiness signal.** Rejected because SPEC §6 requires `AfterInit` sugar to flow through the same public `mark_ready` mechanism.

## Implementation

- add an explicit callback initialization stage and outer-boundary ownership to `Context`;
- freeze immediately but defer only `local_stop` publication for effective `AfterInit` initializers;
- close successful initialization with `mark_ready()` followed by the deferred stop;
- preserve stop publication on error, cancellation, and unwind through the context drop fallback;
- update SPEC §6/R7 and the public `Context::stop` rustdoc.

## Regression coverage

- successful stop-in-init through a projected decorator waits for the outer initializer and then advances an ordered suffix;
- the clean self-stop terminalizes as `Completed + Observed`, not `StartupAborted`;
- a `Manual` stop-in-init still aborts startup and leaves the ordered suffix `NeverStarted`;
- an `AfterInit` initializer that returns `Err` after requesting stop does not publish readiness;
- repeated successful stop-in-init incarnations are post-ready and end through restart intensity rather than `StartupFailed`.

## Validation

- `./tools/dev cargo test -p shelterwood --test readiness` — 30 passed
- `./tools/dev just ci` — passed (formatting, clippy, 770 nextest tests, doctests, rustdoc, API reachability, packaged-doc sync, external consumer, nixfmt)

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1256,7 +1256,9 @@ and `supervisor::…` names the pure reducer suite.
    override receives no such edge and the same pre-ready stop remains R7.
    (`integration::ordered_terminal_pre_ready_exit_parks_the_root_and_marks_suffix_never_started`,
    `integration::dynamic_startup_failure_keeps_other_initial_members_supervised`,
-   `integration::nested_dynamic_startup_failure_rolls_back_and_preserves_inner_cause`.)
+   `integration::nested_dynamic_startup_failure_rolls_back_and_preserves_inner_cause`,
+   `integration::after_init_stop_waits_for_success_then_advances_the_ordered_suffix`,
+   `integration::manual_stop_in_init_remains_a_terminal_pre_ready_failure`.)
 
 No rule above is flagged unverified: each has direct checked evidence. The
 remaining prose in this section specifies declaration defaults, user-facing
@@ -2397,7 +2399,14 @@ framework contract**. No other normative shutdown paragraph is unmapped.
   bound on its drain-plus-`on_stop` window (§5.2), so a wedged self-stop
   escalates — grace expiry, tidy beat, hard abort — exactly like a
   supervisor-initiated one. Self-stop changes who started the clock,
-  never which ladder runs.
+  never which ladder runs. One case defers the clock rather than
+  changing it: an effective `AfterInit` initializer publishes its
+  `stop()` only when it returns (R7, §6), so an initializer that
+  requests a stop and then never returns has not yet armed the ladder.
+  Intake is already frozen and the readiness deadline still bounds it,
+  but a `ReadinessDeadline::Unbounded` declaration leaves that wedge
+  bounded only by parent or scope shutdown. Declaring an unbounded
+  readiness deadline is what accepts that.
 - Shutdown requests are **level-triggered, not queued**: owner drop,
   `request_shutdown()` / `request_scope_shutdown()`, and parent
   escalation each set an idempotent per-scope latch (a token edge, like
--- a/SPEC.md
+++ b/SPEC.md
@@ -1249,7 +1249,11 @@ and `supervisor::…` names the pure reducer suite.
    position-dependent.** The exit funnel applies restart policy first. A
    terminal pre-ready failure emits `FailStartup`; an ordered suffix becomes
    `NeverStarted`, while a dynamic scope has no unstarted suffix. Root failure
-   parks in `StartupFailed`; nested failure begins ordinary rollback.
+   parks in `StartupFailed`; nested failure begins ordinary rollback. For an
+   effective `AfterInit` callback actor, a `stop()` requested during an
+   initializer that returns `Ok` becomes observable only after the automatic
+   readiness edge, so its clean exit is post-ready. An effective `Manual`
+   override receives no such edge and the same pre-ready stop remains R7.
    (`integration::ordered_terminal_pre_ready_exit_parks_the_root_and_marks_suffix_never_started`,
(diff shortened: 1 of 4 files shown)


ralexstokes:
Fresh adversarial review found the production readiness/self-stop ordering sound. It added mutation-sensitive coverage for Immediate and Manual pre-return stop publication and the panic Drop fallback in follow-up commit 13c049c. Validation: full local CI (772 tests), 100 repeated AfterInit restart runs, and 50 repeated stop-in-init group runs all passed.

PR #358, opened 2026-08-19
Contain one-shot sender wake panics and harden runtime synchronization

### LINKED ISSUE #341: shelterwood-runtime residuals: waiter-protocol duplication, oneshot SENDING-window close paths, waker-drop ordering, dispatch_disposal testability
**Priority: P3 batch.** From the 2026-08-19 full review of `crates/shelterwood-runtime`.

- [ ] **Waiter-protocol duplication** (`sync.rs:234-257` and `sync.rs:802-836`): `LatchWait::poll` and `WatchWait::poll` write out the safety-critical register/recheck/deferred-drop protocol twice (~30 structurally identical lines each, including a duplicated copy of the load-bearing ordering comment about hostile displaced-waker destructors), differing only in the readiness predicate and the Ready payload. A future fix to one copy silently leaves the other wrong. Remedy: one helper on `WaiterRegistry` parameterized on a readiness check, encoding clone-before-lock and drop-after-recheck once.
- [ ] **`close()`/`close_and_take()` during the ONESHOT_SENDING window are untested** (`sync.rs:553`): the `oneshot_sending_for_test` staging helper exists precisely for this window and `close_and_poll_receive`'s SENDING branch is pinned, but plain `close()` (racing publish gets `Err(value)` back, state stored ONESHOT_RECEIVER_CLOSED) and `close_and_take()` (returns `None`, value bounces to the sender) in that window have no test. That bounce is what `DisposingReceiver::drop` relies on when it races an in-flight publish (the value must reach exactly one owner). Tests: (a) `close()` then `publish()` returns `Err(value)` and the receiver is empty; (b) `close_and_take()` during SENDING returns `None` and the subsequent `publish()` returns `Err(value)`.
- [ ] **Drop-after-recheck ordering for hostile displaced-waker destructors is comment-only** (`sync.rs:249-251`, `833-834`): the crate's hostile-waker tests (`CountPanicWake`) only panic in wake/wake_by_ref, never in the waker's Drop; RawWaker fixtures with panicking drop vtables exist only in shelterwood-mailbox. Mutation check from the review: moving `WaiterRegistry::drop_registered([displaced, ...])` above the recheck in either poll passes this crate's entire suite. Test: a RawWaker fixture whose drop vtable panics, displaced by a second poll racing a fire/pulse, asserting the fired transition is still consumed despite the destructor panic.
- [ ] **`dispatch_disposal`'s last-resort inline `finish()` is untestable as written** (`disposal.rs:296-303`): the final degradation path (blocking pool rejected AND the shared fallback thread cannot start, so the non-critical job's user destructor runs inline with its panic contained) can never execute under test — `enqueue_fallback_disposal` hardcodes the `FALLBACK_DISPOSALS` static and the real `spawn_fallback_worker`; the injectable `*_with` variants stop one layer below. Remedy: thread the disposals mutex and spawner through `dispatch_disposal` so a test can inject thread exhaustion end-to-end, or explicitly document the path as accepted-untested with the reasoning.
- [ ] **Stale `DisposingReceiver` doc** (`sync.rs:592`): the fn-pointer-erasure justification names `ReplyReceiver`, `ReplyReceive`, `CallFuture`, `OneShotTaskRef` as holders, but only `OneShotTaskRef` holds this type; the other three hold the separate twin `DisposingReceiver` in `shelterwood-mailbox/src/capability.rs:160`. Correct the holder list and cross-reference the sibling type (or fold into the #241 docs pass).

### LINKED ISSUE #349: OneShotSender::drop runs a hostile receiver waker uncontained — dropping an unanswered Reply during a panic aborts the process
**Confirmed bug (external claim, adversarially verified with an end-to-end repro; consequence is process abort, trigger requires a caller-supplied hostile waker — severity adjudicated P2).**

`OneShotSender::drop` (`crates/shelterwood-runtime/src/sync.rs:499-510`) only CASes the framework state word and lets field-drop glue drop the inner `tokio::sync::oneshot::Sender`. Tokio's sender drop (tokio 1.53.1, `sync/oneshot.rs:871` -> `Inner::complete()` -> `rx_task.with_task(Waker::wake_by_ref)`) runs the registered receiver waker inline on the dropping thread — foreign waker vtable code, uncontained.

The waker is the caller's, unfiltered: `ReplyReceiver::recv` -> `Deadlined::poll` (`deadline.rs:107`) -> `poll_reply` -> `DisposingReceiver::poll_receive` -> `OneShotReceiver::poll_receive` (`sync.rs:513`) passes the caller `Context` straight through. Polling a public future with a user-built `Waker` is safe Rust on the supported surface.

**Failure scenario (reproduced through the supported façade):** an `Actor` whose message carries a `Reply<u8>` panics in `handle`; the unwind drops the message and therefore the unanswered `Reply`; the sender drop wakes the registered hostile waker; the waker panics inside cleanup — "panic in a destructor during cleanup" -> non-unwinding panic -> SIGABRT (exit 134). The supervisor never publishes the exit. Repro'd both at the sync.rs unit level and end-to-end (`Reply` + supervised handler panic + one poll of `receiver.recv(30s)` with a panicking `Wake` impl); a control probe confirmed the same hostile waker parked on a `Signal` is contained (exit 0).

**Why this is a defect, not an accepted trade:** the project has already paid for containment at every sibling site — `WaiterRegistry::wake_all` runs each wake inside a `PanicAccumulator` (`sync.rs:69-81`); `BlockingJob::drop` does `discard_panic(catch_panic(|| drop(completion)))` for the *other* production one-shot sender with a comment saying exactly why (`spawn.rs:293-295`); `DisposalJob::finish` contains its completion (`disposal.rs:108-117`); `SendFuture::drop` fixed the sibling case with a test stating "Before containment this aborted the process" (`futures.rs:363-397`, test at :1348). SPEC §1 promises a hostile waker is "an ordinary, testable outcome instead of a liveness failure", and the docs/shutdown-and-resources.md:160-162 exclusion does not cover this: no user destructor panics here — the framework's channel invokes third-party waker code from its own drop glue.

**Scope:** every `shelterwood_runtime::oneshot` sender shares the gap — the internal `CallFuture` reply channel (`futures.rs:580`) has the same exposure, as does an actor that stores a `Reply` in its state and later unwinds. Framework-retirement paths are already safe (undelivered mailbox payloads go through `dispose_detached`; disposal contains the wake). The live exposure is specifically "a user frame owning a `Reply` unwinds".

**Remedy (verified against the sibling precedents):** contain in `OneShotSender::drop` itself so all one-shot users get it uniformly — `let mut panics = PanicAccumulator::default(); panics.run(|| drop(self.channel.take()));` — which surfaces a hostile wake to a non-unwinding dropper (matching `Latch::fire`/`WatchSender::pulse` and their existing tests) while discarding it during an unwind. Do NOT route the sender to `dispose_detached`: that would defer the `ReplyError::Dropped` discharge and, under the accepted thread-exhaustion fail-safe, never deliver it — contradicting SPEC §1's "synchronous fallback, never an await, never a join".

**Regression test requirements:** assert the child's exit status via a subprocess (an in-process catch cannot distinguish abort), and keep the receiver alive across the sender drop — otherwise tokio's `complete()` short-circuits on `is_closed()` and the bug hides.

ralexstokes:
Closes #349.
Closes #341.
Part of #348.

## Summary

- contain Tokio one-shot receiver-waker panics when `OneShotSender` closes from drop glue
- centralize the latch/watch register-recheck-deferred-drop protocol in `WaiterRegistry`
- cover both one-shot `ONESHOT_SENDING` close paths and hostile displaced-waker destruction
- make the final disposal degradation path injectable and test it end to end
- correct the stale runtime `DisposingReceiver` holder documentation
- add a supported-façade subprocess regression proving an unanswered `Reply` dropped during handler unwind cannot abort the process and still publishes the actor's `Panicked` exit

## Design decisions and omissions

- `OneShotSender::drop` keeps synchronous channel closure and wraps it in `PanicAccumulator`. It deliberately does not use `dispose_detached`: receiver discharge remains immediate, an ordinary dropper still observes a hostile wake panic, and an existing unwind contains it instead of double-panicking.
- The shared waiter helper accepts an `Option<T>` readiness check. It clones the caller waker before locking, performs registration and the publication recheck, removes any successful registration, and only then destroys displaced/removed wakers through the accumulator.
- The watch ordering regression advances the same atomic version edge as `pulse` from a clone-vtable hook without draining the registry. That is the deterministic way to exercise a displaced waker; calling `pulse` itself would consume the old registration through its wake path before replacement.
- The private disposal dispatcher now receives the fallback queue and spawner. Production still supplies the single static queue and real worker; the test supplies a local queue plus an exhausted spawner from a non-Tokio thread, which reaches the formerly untestable inline `finish()` branch.
- No #341 checklist items were omitted: waiter duplication, both SENDING-window close cases, latch/watch hostile drop ordering, disposal fallback injection, and stale documentation are all addressed.

## Validation

- `./tools/dev cargo test -p shelterwood-runtime` (62 passed)
- `./tools/dev cargo test -p shelterwood --test reply_waker_containment -- --nocapture` (1 subprocess regression passed)
- `./tools/dev just ci` (773 nextest tests plus formatting, clippy, doctests, docs, API reachability, packaged-doc, external-consumer, and Nix formatting checks)

### DIFF
--- a/crates/shelterwood-runtime/src/disposal.rs
+++ b/crates/shelterwood-runtime/src/disposal.rs
@@ -174,13 +174,6 @@ static FALLBACK_DISPOSALS: Mutex<FallbackDisposals> = Mutex::new(FallbackDisposa
     worker_live: false,
 });
 
-/// Queues a disposal job for the shared fallback thread, lazily starting it.
-/// Returns `false` when no worker exists and none could be started; the
-/// caller must then finish the job itself.
-fn enqueue_fallback_disposal(job: Arc<dyn BlockingPoolJob>) -> bool {
-    enqueue_fallback_disposal_with(&FALLBACK_DISPOSALS, job, spawn_fallback_worker)
-}
-
 /// Queues a disposal that must never fall back to the submitting thread.
 ///
 /// The `()` return is the contract: every path transfers ownership. If native
@@ -279,6 +272,17 @@ where
     T: Send + 'static,
     C: FnOnce(Option<DisposalPanic>) + Send + 'static,
 {
+    dispatch_disposal_with(job, &FALLBACK_DISPOSALS, spawn_fallback_worker);
+}
+
+fn dispatch_disposal_with<T, C>(
+    job: Arc<DisposalJob<T, C>>,
+    disposals: &Mutex<FallbackDisposals>,
+    spawn: impl FnOnce() -> std::io::Result<std::thread::JoinHandle<()>>,
+) where
+    T: Send + 'static,
+    C: FnOnce(Option<DisposalPanic>) + Send + 'static,
+{
     // A rejected submission falls through so the fallback thread, rather than
     // this runtime-teardown thread, owns user destruction.
     if submit_blocking_job(&job) {
@@ -293,7 +297,11 @@ where
     // user destructors, exactly what isolation must prevent. Serialization is
     // the accepted trade: one blocking destructor delays later fallback
     // disposals instead of consuming another native thread.
-    if enqueue_fallback_disposal(Arc::clone(&job) as Arc<dyn BlockingPoolJob>) {
+    if enqueue_fallback_disposal_with(
+        disposals,
+        Arc::clone(&job) as Arc<dyn BlockingPoolJob>,
+        spawn,
+    ) {
         return;
     }
 
@@ -379,8 +387,8 @@ mod tests {
     };
 
     use super::{
-        DisposalJob, DisposalPanic, FallbackDisposals, Isolated, dispose_detached,
-        enqueue_fallback_disposal_with, retain_fallback_disposal_with,
+        DisposalJob, DisposalPanic, FallbackDisposals, Isolated, dispatch_disposal_with,
+        dispose_detached, enqueue_fallback_disposal_with, retain_fallback_disposal_with,
     };
     use crate::{
         spawn::BlockingPoolJob,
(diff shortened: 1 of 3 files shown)


ralexstokes:
Fresh adversarial review: no findings. The reviewer mutation-checked the #349 subprocess regression (removing containment reproduces SIGABRT), mutation-checked both waiter drop-order regressions, verified RawWaker Arc ownership and all five #341 checklist items, and reran the runtime tests plus targeted clippy. No follow-up commit was required; reviewed head is eff5e43.

Issue #353, opened 2026-08-19
Design question: generic add<D: Definition> under the eight nominal entry points (declaration-matrix ruling)


ralexstokes:
**P3 design question (external proposal, premises adversarially verified; the broad "collapse the matrix" version is refuted, a narrower option is genuinely open).**

The external review proposed collapsing the 4-kind x Once/Def definition matrix and the three builder surfaces into one `add` + `reserve` + `define`. Verification refuted the load-bearing premises:

- **The Once/Def split buys three compile-time properties beyond ".restart() absence":** (1) construction-source shape — restartable forms take `Fn() -> X + Send + Sync` or `Args: Clone + Sync`, one-shot forms take an owned value with NO bound or `FnOnce`; an `Fn` closure cannot yield a moved-in unique resource (E0507, probe-verified), so the `_once` flavor is the only way to declare a child owning a non-re-mintable resource — the entire subject of `one_shot_resource_ownership.rs` and SPEC §4.2 ("The twins differ in what they can accept ... not in how long the child runs"); (2) `TaskOnceDef<T>` carries a typed completion channel and its define returns `(TaskRef, OneShotTaskRef<T>)` — SPEC lines 570-579 already adjudicated this unmergeable; (3) the missing `.restart()` method is what stops users setting a value `resolve_common` silently discards for OneShot (see #336).
- **The duplication is smaller than claimed:** Tree/DynamicTree already share one macro (`impl_common_builder_surface!`, zero hand copies); `DynamicScopeRef`'s methods are shaped differently (`Admission<H>`), not copied; the six slot types are 1-3-line forwarders over three generic cores the SlotEndpoint GAT already deduplicated across flavors (the GAT collapsed the *flavor* axis 12->6; nobody has collapsed kind or mode).
- The readiness-`Result` precedent (`definition.rs:97-106`) is real but not analogous: it is a value restriction on a shared enum where a type split would be worse — not evidence for preferring Results over unrepresentability.

**The genuinely open question worth a ruling:** SPEC line 1774-1775 says "Adding a new child kind or mode extends the shared record, not a hand-maintained matrix", while SPEC 1725-1728 pins eight nominal entry points for docs/autocomplete. A sealed `Definition` trait with `type Handles` would satisfy the first at some cost to the second:

[...]

PR #359, opened 2026-08-19
Address façade review findings

### LINKED ISSUE #337: Collapse the three-layer pure-forwarding delegation in driver/admission_control.rs
**Priority: P2 (simplification, ~35 lines of indirection).** From the 2026-08-19 full review.

Each dynamic control operation in `crates/shelterwood/src/driver/admission_control.rs` exists three times with identical signatures: a `pub(crate)` free function (`reserve_dynamic`/`start_admission`/`cancel_dynamic_reservation`/`signal_fused_cancel`/`remove_dynamic`), an inherent method on `DynamicControl` (lines ~269-314), and an implementation free function (`reserve_dynamic_slot`/`start_dynamic_admission`/`cancel_dynamic_reservation_impl`/`signal_fused_cancel_impl`/`remove_dynamic_impl`).

The inherent-method middle layer is pure forwarding with no other callers (verified during review: only the same-file wrappers and one test call them), no dynamic dispatch, and no visibility purpose — everything lives in one file. Readers chase two hops to find any body.

**Remedy.** Collapse to two layers: keep the `pub(crate)` entry points (the tree layer's API) and move the `*_impl` bodies into the `DynamicControl` methods, or delete the methods and have the entry points call the impl functions directly.

### LINKED ISSUE #343: Facade cleanups: mode() trio, duplicated StartupDisposition::Aborted predicate, misplaced shutdown_and_wait, silent drain-stage stop()/mark_ready()
**Priority: P3 batch.** From the 2026-08-19 full review of the shelterwood façade.

- [ ] **`mode()` wrapper trio duplicated verbatim**: `TaskConstruction::mode` (`task.rs:208`), `ScopeConstruction::mode` (`plan.rs:556`), `RawConstruction::mode` (`raw/definition.rs:297`) are the identical 6-line `is_one_shot -> ChildMode` body, with `restartable()`/`take_one_shot()` delegations repeated alongside two of them. Move `fn mode(&self) -> ChildMode` onto `DefinitionSource` itself (it already owns `is_one_shot`) and delete the copies.
- [ ] **`StartupDisposition::Aborted` predicate duplicated**: the 3-condition predicate `is_initial(key) && !lifecycle().startup_complete() && !initial_ready(key)` appears identically in `spawn_child`'s incarnation-exhaustion path (`driver/child.rs` ~707) and `handle_exit`'s Terminal dispatch (~1068), each with its own copy of the SPEC §6 rationale comment. A future adjudication of what counts as a startup abort must be applied twice. Extract a `ScopeRuntime` helper holding the predicate and comment once.
- [ ] **`ScopeRef::shutdown_and_wait` is misfiled** (`tree/dynamic_api.rs:40`): it is a plain ordered-scope method (no dynamic capability involved) yet is the only `ScopeRef` impl block outside scope.rs, and the module name actively suggests it is dynamic-only. Move it to scope.rs next to `request_shutdown`/`wait_stopped`.
- [ ] **Drain-stage `stop()`/`mark_ready()` silently no-op while every sibling op returns `Rejected`** (`actor.rs:160`, `:167`): during DrainingFrozenPrefix, `continue_with`/`set_timeout`/`set_interval`/`clear_timer`/`offload`/`offload_scoped` all return `Err(Rejected)` (each pinned in tests/drain.rs), but `Context::stop` and `Context::mark_ready` silently do nothing. The runtime outcome is correct, but a drain-stage handler cannot distinguish an accepted stop from an ignored one, and no test pins the inertness, so a refactor could route a drain-stage `stop()` through to raw state without any test failing. Either document the deliberate silence on both methods or make them return the same `Rejected` shape as their siblings; add a drain.rs assertion either way.

### LINKED ISSUE #347: Test-style consistency: raw.rs compares task ids as Debug strings; drain.rs judges recovered payloads inside the handler
**Priority: P3 (test-style consistency — neither test is vacuous; failures do propagate).** From the 2026-08-19 full review.

- [ ] **`tests/raw.rs` compares task ids as Debug strings** (~line 34): `FactoryTaskActor` stores `format!("{:?}", tokio::task::id())` at factory time and asserts equality against a second Debug rendering inside `run()`, with the verdict in an in-actor `assert_eq`. Failure is still detectable (the panic trips intensity, so the body's `system.wait() != StopReason::Finished` assertion fails), but it diverges from the corpus's own standard: `offloads.rs`'s `ZeroDeadlineObservation` deliberately carries `tokio::task::Id` whole "rather than as a pre-collapsed comparison" because Debug output is non-contractual in this repository, and it publishes both ids for body-side judgment. Store the `Id` itself (Copy + Eq + Send), publish it, assert in the test body.
- [ ] **`tests/drain.rs` judges recovered payloads inside the supervised handler** (~line 96): `DrainActor`'s Value arm asserts the rejected continuation/timer/interval/offload payloads (`assert!(matches!(continuation.into_inner(), ...))` etc.) inside `handle()`. Not vacuous — a panic there truncates the log so the body's exact-history assertion fails, and `deferred_ran` is judged in the body — but the recovered-payload claims surface only as a generic history mismatch rather than a named failure, unlike the file's own `DrainingProbe` pattern. `StoppingRejectionRaw` in raw.rs shows the stronger form (a body-checked flag gated behind every assert). Adopt that pattern or publish the payloads for body-side judgment.

### LINKED ISSUE #351: Close the last unwrapped window in the Isolated-always convention: definition payloads across the Into<ChildId> and readiness() hooks
**P3 hardening (residual of a refuted external P1 claim; the defect framing was refuted, the mechanism is real and was reproduced).**

The claim: definition payloads are unprotected around two user-controlled callbacks — `RawDef::erase`/`RawOnceDef::erase` invoke `RawActor::readiness()` while still owning the factory/actor (`raw/definition.rs:91`, `:141`), and static/dynamic add methods run the user's `Into<ChildId>` while the frame owns the definition (`plan.rs:332-345` via `checked_id`; `tree/dynamic_api.rs:68/81/93`). A panic in either hook drops the definition during the unwind; a hostile definition destructor then double-panics (SIGABRT, reproduced) or blocks.

**Why the P1 framing was refuted:** no lock is held at either point; both panics are user code invoked synchronously inside the user's own public call on the user's own thread (`add_*`/`reserve_*` have zero internal callers), destroying the user's own values; and SPEC §7's containment boundary explicitly carves this out (SPEC.md:1651-1669: a destructor panic inside the same unwind "is a genuine double panic and aborts the process; that is Rust's contract ... not something the runner can contain"; restated at :1480-1482 and :1497-1499). Reachability needs two simultaneously hostile user impls (a panicking `readiness()`/hand-written `From<_> for ChildId` AND a panicking/blocking definition destructor). The high-level `Actor` API cannot trigger the readiness half at all (`Handler::readiness` is a framework constant). The isolation contract that does exist (`dispose_rejected`, builders.rs:35-47, commit ab81c28) is scoped to framework *rejection-return* paths and holds there, with named tests.

**What remains worth doing:** `add_*`/`erase` form the one window in which a definition/`ChildConstruction` is a bare owned value in a framework frame — from `SlotCell::define` (`plan.rs:100`) onward it always lives inside `Isolated`. Closing the window completes the convention at near-zero cost:
- [ ] In `define_or_reject` and each static `add_*`, bind `let definition = Isolated::new(definition);` before evaluating the reservation; `take()` on the Ok path (bounds already hold: `D: Send + 'static`).
- [ ] Have `erase` take/hold `Isolated<Self>`, read `options.readiness` via `get()`, call `R::readiness()`, then `take()`.
- [ ] Amend SPEC §7's carve-out and the `dispose_rejected` doc comment in lockstep to say the definition is isolated across the id/readiness hooks too — otherwise the code silently promises more than the spec, which is exactly the drift this project tracks (#238's concern class).

Caveat recorded during verification: this buys a specific window, not a guarantee — a user's own `RawOnceDef::new(HostileDrop)` local on a stack that panics for any other reason aborts identically. That is why this is hardening, not a bug fix.

ralexstokes:
## Summary

- collapse dynamic admission control from three forwarding layers to the public façade entry points plus concrete `DynamicControl` methods
- complete the façade cleanup batch: centralize definition mode, centralize startup-abort classification, move `shutdown_and_wait`, document/test drain-stage silent controls, fuse lowering's undefined/policy pass, explicitly discharge startup removals on driver Drop, and import the supervisor reducer honestly
- move raw/drain test verdicts into the test body and add the missing init-error/live-offload teardown regression
- isolate supplied definitions across eager `Into<ChildId>` and `RawActor::readiness` hooks, with SPEC wording and hostile-panic regressions

Closes #337
Closes #343
Closes #347
Closes #351

## Decisions and boundaries

- `Context::stop` and `Context::mark_ready` remain unit-returning, deliberately silent no-ops during frozen-prefix drain. Changing them to `Result` would be a public API break for behavior that is already semantically inert; docs and drain-body assertions now pin that choice.
- Static lowering now resolves and classifies every slot in one pass into `Vec<Result<ResolvedCommonOptions, ChildId>>`, then claims definitions only after undefined validation and identity adoption. A dedicated state-aware resolver preserves `Lowered` as the existing panic rather than misreporting it as `Undefined`.
- `pending_startup_removals` is explicitly discharged by the synchronous Drop epilogue after residency cleanup. The field can reach Drop only through a resumed user-code panic because the retain-to-orderly-publish path contains no await.
- Declaration-hook isolation is intentionally narrow: it protects the supplied definition while the framework invokes the eager id/readiness hook. It does not extend the supervised runner boundary to arbitrary locals in a user's synchronous call; SPEC §7 and the `dispose_rejected` docs say so explicitly.
- No #353 declaration-matrix design was implemented. The eight nominal `add_*` entry points and all three reserve families remain intact. This PR only makes the existing internal `DefinitionSource::mode` the shared mode authority and widens the erased raw source seam crate-privately; there is no public declaration-surface change for the follow-on ruling.

## Validation

- `./tools/dev just ci`
  - rustfmt and clippy with warnings denied
  - 770/770 nextest tests
  - doctests and rustdoc warnings
  - public runtime reachability checks
  - packaged docs and external consumer checks
  - Nix formatting check
- targeted façade library and `actor`, `raw`, `drain`, `disposal`, and `one_shot_resource_ownership` test targets

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1667,6 +1667,15 @@ terminates the process by contract.
   `Drop` — is a genuine double panic and aborts the process; that is
   Rust's contract, documented alongside §11's `panic = "unwind"`
   precondition, not something the runner can contain.
+- **Declaration-hook isolation.** The static and dynamic `add_*` entry points
+  isolate the supplied definition before invoking the caller's
+  `Into<ChildId>` conversion, and raw-definition erasure keeps the definition
+  isolated while invoking `RawActor::readiness`. A panic from either eager
+  hook therefore cannot unwind through and destroy that definition on the
+  caller's thread. This is a narrow ownership guarantee, not an extension of
+  the runner boundary: other locals in the user's synchronous declaration
+  call remain subject to Rust's ordinary unwinding and double-panic contract,
+  as do user values that the framework never accepted or wrapped.
 
 ## 8. Child specification and options [#368]
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -1216,7 +1216,7 @@ and `supervisor::…` names the pure reducer suite.
    before the removal response resolves. (`supervisor::sampled_removal_suppresses_start_effects_until_commit`,
    `supervisor::exhaustive_reachable_states_preserve_the_reducer_invariants`,
    `integration::queued_removal_suppresses_replayed_self_stop_readiness`,
-   `integration::startup_removal_response_follows_aggregate_recomputation`.)
+   `integration::startup_removal_response_follows_recomputation_and_drop_discharges_it`.)
 4. **R4 — ordered start is one accepted edge at a time.** `Settle` emits
    `StartChild` only for the current initial cursor, and advances the cursor
    only past a spawned-and-ready member or a reclaimed key, reaching every
@@ -1458,7 +1458,7 @@ One classification, produced at one point, used by every consumer.
 3. **E3 — cancellation is orthogonal sampled state.** The exit records
    `Observed` iff the incarnation cancellation latch had fired when the
    outcome was recorded. Restart eligibility never inspects this field.
-   (`engine::funnel_dispatch_depends_on_mode_and_membership_state`,
+   (`engine::funnel_dispatch_covers_every_policy_exit_and_suppression_combination`,
    `integration::locally_requested_subtree_shutdown_reads_cancelled`.)
 4. **E4 — one authoritative membership/incarnation state.** Incarnation
    phases advance only through `Unstarted → Active → Stopping? → Complete →
@@ -1470,7 +1470,7 @@ One classification, produced at one point, used by every consumer.
 5. **E5 — restart suppression is state-derived.** Exits schedule restart only
    while the scope is running and the membership is resident. Draining or
    `Removing` records schedule nothing and charge no intensity.
-   (`engine::funnel_dispatch_depends_on_mode_and_membership_state`,
+   (`engine::funnel_dispatch_covers_every_policy_exit_and_suppression_combination`,
    `integration::same_batch_removal_suppresses_pending_restart_shutdown`.)
 6. **E6 — publication fences replacement.** A replacement spawn follows the
    predecessor’s terminal publication; stale incarnation evidence cannot
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -157,6 +157,10 @@ impl<'a, A: Actor> Context<'a, A> {
     actor_context_forwarders!(A);
 
     /// Releases this incarnation's readiness gate while live.
+    ///
+    /// During frozen-prefix drain the incarnation is already stopping, so
+    /// readiness can no longer change the startup outcome and this call is a
+    /// deliberate no-op.
     pub fn mark_ready(&self) {
         if self.stage == DeliveryStage::Live {
             self.raw.mark_ready();
@@ -164,6 +168,10 @@ impl<'a, A: Actor> Context<'a, A> {
     }
 
     /// Requests a clean local stop after the current callback.
+    ///
+    /// During frozen-prefix drain the incarnation is already stopping, so
+    /// this call deliberately does not inject a second local-stop edge and
+    /// returns silently.
     pub fn stop(&mut self) {
         if self.stage == DeliveryStage::Live {
             self.raw.stop();
--- a/crates/shelterwood/src/definition.rs
+++ b/crates/shelterwood/src/definition.rs
@@ -1,5 +1,7 @@
 //! Shared private machinery for declaration definitions.
 
+use crate::policy::ChildMode;
+
 /// Repeatable or consuming definition payload state.
 ///
 /// The explicit `Spent` state lets consumers move a one-shot payload without
@@ -14,6 +16,14 @@ pub(crate) enum DefinitionSource<R, O> {
 impl<R, O> DefinitionSource<R, O> {
     pub(crate) fn is_one_shot(&self) -> bool {
         matches!(self, Self::OneShot(_) | Self::Spent)
+    }
+
+    pub(crate) fn mode(&self) -> ChildMode {
+        if self.is_one_shot() {
+            ChildMode::OneShot
+        } else {
+            ChildMode::Restartable
+        }
     }
 
     pub(crate) fn restartable(&self) -> Option<&R> {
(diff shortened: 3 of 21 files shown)


ralexstokes:
Adversarial review complete; I found no material correctness, lock-rule, behavior-preservation, or API-surface defect.

Checked the full #337/#343 (including both follow-up comments)/#347/#351 checklist and the #353 boundary against the complete diff. In particular: the admission-control change is a literal body move; every new lock site/effect ordering remains unchanged; `Isolated` is armed before both eager `Into<ChildId>` paths and re-armed across `RawActor::readiness`, with the value taken only after the relevant hook; fused lowering keeps `Lowered` distinct from `Undefined` and preserves slot/options order; startup-removal completion is explicitly flushed after terminality/residency cleanup and before the remaining dynamic-entry/child/final-scope teardown; and all public declaration entry points/reserve families remain unchanged.

I also mutation-tested the drain assertions by temporarily forwarding drain-stage `stop` and `mark_ready` to the raw context: the drain target still passes because those raw primitives are already idempotent/inert once stopping. I do not consider that a blocker—the added test pins the public silent-return/no-perturbation contract, while the docs and stage gate carry the narrower implementation choice—but it is worth recording that the test does not structurally prove absence of the raw calls.

Validation reviewed/run: GitHub `ci` passed; local targeted nextest run covered the façade library plus `actor`, `disposal`, `drain`, `one_shot_resource_ownership`, and `raw` (295/295 passed). Worktree remains clean; no review-fix commit was needed.

PR #360, opened 2026-08-19
Tighten mailbox transition discipline and regression coverage

### LINKED ISSUE #333: Pin the mutation-invisible mailbox cell behaviors: cross-incarnation receive isolation, latest-value displacement, bind's observe_all evidence
**Priority: P2.** From the 2026-08-19 full review. Three behaviors in `crates/shelterwood-mailbox/src/cell.rs` are mutation-invisible: hostile mutation of the code leaves the entire suite green.

- [ ] **Cross-incarnation receive isolation (~line 1078).** `receive()`'s eligibility check (`BindingStatus::Bound(current) => current == incarnation`, and the incarnation half of the Frozen-drain arm) has no test anywhere: every test constructs its `MailboxReceiver` with the currently bound incarnation. Mutating the Bound arm to `true` (or dropping the `current == incarnation` conjunct in the Frozen arm) passes the suite, silently allowing a stale receiver from a prior incarnation to drain a rebound mailbox's messages. The mode half of the Frozen arm IS pinned (`receive_modes_pin_live_cutoffs_and_frozen_drain`), which makes the missing incarnation half easy to overlook. Test: bind incarnation A, close, rebind incarnation B with queued messages, assert a receiver constructed for A gets `None` in both Drain and LiveThrough modes (and symmetrically for a frozen mailbox).
- [ ] **Live latest-value displacement lock discipline (~line 1496).** `accept_locked`'s displacement of the previous latest envelope (`state.latest.replace` -> `effects.displaced.push`) is the most common displacement path, yet no test observes where or when that displaced message is destroyed (existing displacement tests cover only the bind-time promotion path). A mutation that drops `displaced` inline under the mailbox mutex — the exact violation the effects machinery exists to prevent — passes the suite. The `LockCheckingMessage` harness already in this file's tests is the right probe: bind a Latest mailbox, submit a `LockCheckingMessage`, submit a second message displacing it, assert the first message's destructor observed the mutex unlocked and ran after the second acceptance was visible.
- [ ] **bind's `observe_all` evidence (~line 1310).** `bind()` calls `waiters.observe_all(incarnation)` so FIFO-overflow waiters that stay parked through the bind report the new incarnation as withdrawal evidence — the comment argues the withdrawal-race consistency in detail, but deleting the call passes the suite (existing timeout tests get their evidence from submit-time `observe()` on an already-bound mailbox, never from the bind edge over previously unbound waiters). Test: park more sends than queue capacity on an unbound mailbox, bind, withdraw an unpromoted waiter, assert `WithdrawalOutcome::Withdrawn { observed: Some(bound_incarnation), .. }`.

### LINKED ISSUE #338: Consolidate the triplicated MailboxRuntime test doubles and the duplicated erased receiver
**Priority: P2 (simplification, ~100 lines of copy-paste test doubles).** From the 2026-08-19 full review.

Three hand-rolled delegating `MailboxRuntime` doubles exist in the mailbox crate's tests: `AdapterRuntime` (`capability.rs:230`), `ControlledClockRuntime` (`futures.rs:942`, overrides only `now()`), and `StagedReplyRuntime` (`reply.rs:212`, overrides only `oneshot()`). Each restates all five trait methods. In addition, `StagedReceiver` (`reply.rs:181-210`) duplicates `AdapterOneShotReceiver` (`capability.rs:274-303`) verbatim, including the four-arm `OneShotClose -> ErasedOneShotClose` mapping.

That copy-paste must be kept in sync whenever `MailboxRuntime` or the erased one-shot surface changes; a divergence in one copy would silently weaken the test it backs.

**Remedy.** In `capability::tests`, expose the erased adapter wrappers (`AdapterOneShotReceiver` at least) and add one delegating test runtime with overridable hooks (e.g. optional clock closure and optional oneshot factory falling back to the inner adapter). `ControlledClockRuntime`, `StagedReplyRuntime`, and `StagedReceiver` then collapse to a few lines each, and the trait-delegation boilerplate lives in exactly one place. This stays within the "test doubles delegate, never restate" rule since every hook still bottoms out in the real adapter primitives.

### LINKED ISSUE #339: Withdrawn-arm panics fire while holding operation/mailbox guards, deviating from the verdict-release-panic convention
**Priority: P2/P3 (quality — convention deviation; unreachable today, so not a bug).** From the 2026-08-19 full review; flagged independently by the module reviewer and the lock-rule auditor.

Two arms in `crates/shelterwood-mailbox/src/cell.rs` panic while holding framework guards, deviating from the travelling convention (compute the verdict, release, then panic — `MailboxCell::bind` is the named pattern; `MailboxState::take_waiters` uses `debug_assert!` where releasing is impossible):

- [ ] `withdraw()`'s `OperationOutcome::Withdrawn => panic!("a send operation was withdrawn more than once")` (~line 1234) fires while holding both the operation guard and the live `MailboxTxn`; the unwind poisons the operation mutex AND the mailbox mutex, after which every sender/receiver hits `.expect("mailbox mutex poisoned")` — one violated invariant becomes a mailbox-wide cascade.
- [ ] `SendOperation::poll`'s `OperationOutcome::Withdrawn => panic!("a withdrawn send future was polled")` (~line 317) unwinds while the operation state guard is held. Unlike `take_waiters` it is not excused by an impossible release: the arm could set a flag, drop the guard, and panic. (No cascade into the mailbox mutex here — a withdrawn operation is already unlinked from every waiter queue.)

Both arms are unreachable via the current public API (`SendFuture` transitions to `Done` before withdrawing, and poll-after-withdraw panics earlier in futures.rs), and the file is otherwise scrupulously consistent about the convention.

**Remedy.** Carry a verdict out of the critical section and panic after `transaction.finish()` (mirroring `SubmitTransition`), or downgrade to `debug_assert!` per the `take_waiters` precedent. The release-mode `unreachable!()` arms in `park`/`accept_locked` are the same class but lower value.

### LINKED ISSUE #344: shelterwood-mailbox residuals: empty-batch flush overhead, small duplications, TimedSend, Inline-waker withdrawal twin test
**Priority: P3 batch.** From the 2026-08-19 full review of `crates/shelterwood-mailbox` (beyond the issues filed separately for the mutation-invisible cell behaviors, the withdrawn-arm panics, and the test-double consolidation).

- [ ] **Empty-batch flush overhead** (`cell.rs:636`): `MailboxEffects::drop` unconditionally constructs a `MailboxEffectBatch` — cloning the `changed` and `runtime` Arcs (two atomic RMW pairs) — and runs `flush()` even when nothing was queued. An empty `try_recv` poll (the terminating poll of every actor-loop drain) pays this on each call; the struct's own doc justifies borrowing the cell precisely to avoid "two atomic refcount pairs" per transition, which the unconditional Drop reintroduces (the deferred path likewise boxes a no-op closure). Remedy: early-return from the drop when the batch would be empty.
- [ ] **Small duplications in cell.rs** (~25 lines): (1) `submit()`'s Frozen and Unbound arms are identical except for the `observe()` call — one park path taking `Option<Incarnation>`; (2) `MailboxEffects::new` and `::deferred` restate the same seven-field initializer differing only in `external`; (3) the Bound/Frozen-to-`Some(incarnation)` match appears in both `current_observation()` (~1114) and `withdraw()` (~1162) — a `MailboxState` helper next to `status()`.
- [ ] **`TimedSend` newtype and duplicated Terminated-error block** (`futures.rs:407`): `TimedSend<M>` is a one-field pass-through whose only purpose is carrying the `DeadlineOperation` impl; implement the private trait directly on `SendFuture<M>` (both crate-internal, no coherence conflict) and delete the struct, its field hop, and the `withdraw_send` thin wrapper. Separately, `CallOperation::poll_deadlined` builds the identical `close_reply` + `dispose(error.message)` + `CallError{kind: Terminated}` block twice (~610-621 and ~654-670), differing only in the kind mapping — extract a `fail_send` helper.
- [ ] **Inline waker disposition on the send_timeout withdrawal path has no hostile-waker twin** (`futures.rs:452`): `withdraw_send` documents a load-bearing split — the `SendTimeout` timeout path uses `WithdrawalDisposition::Inline`, so the registered waker's destructor runs on the caller's task, its panic stays visible, and the recovered message must already be owned by the result so a panicking waker drop cannot divert it. Every Isolated counterpart has a dedicated hostile-waker test; nothing pins the Inline path. A refactor could silently flip it to Isolated (changing panic visibility) or reorder `take_outcome` after `finish` (re-opening the diversion hazard) without a test failing. Test: park a send_timeout behind a thread-recording, panicking-drop waker, advance past the deadline, poll, assert the destructor ran on the polling thread, the panic surfaced from the poll, and the message was not diverted.

ralexstokes:
## Summary

- pin mailbox incarnation isolation, latest-value displacement discipline, and bind-time waiter observation with mutation-sensitive tests
- consolidate mailbox test runtimes behind one hookable adapter and reuse the erased one-shot receiver
- move withdrawn and missing-terminal-message invariant panics past operation/mailbox guard release
- complete the mailbox residual cleanup batch: skip empty effect flushes, reduce duplicated transitions/initializers/observation logic, remove `TimedSend`, share call send-failure cleanup, and pin inline timeout-waker behavior
- remove the teardown double box and tighten sealed `MailboxTermination::finish` to return its mandatory disposal directly

## Design decisions

- Kept `replace_binding`'s release-mode fail-safe and corrected its contract to state that a declined waiter-carrying transition strands parked senders. A witness token would add transition choreography without eliminating the terminal-misuse case, so it would not actually make invalid replacement unconstructible.
- Tightened `MailboxTermination::finish` because the trait is sealed, `MailboxTeardown` is its only implementation, and every constructed teardown owns one payload.
- The hostile `send_timeout` test drives the private deadline operation after advancing paused time so the timer does not itself own the deliberately panicking raw waker; this isolates the inline withdrawal contract being tested.
- No checklist items were omitted.

## Adversarial review

The fresh review found and fixed three coverage defects in follow-up commit `602d133`:

- the bind-time `observe_all` regression had been shadowed by withdrawing while still bound; the test now closes before withdrawal and fails when `observe_all` is removed
- the inline-waker mutation could abort before cleanup; the fixture now owns the waker through `ManuallyDrop` from construction and fails normally under mutation
- guard-release coverage now includes both Waiting and Terminated missing-message invariant branches plus repeated withdrawal

## Validation

- `./tools/dev cargo test -p shelterwood-mailbox`
- `./tools/dev cargo test -p shelterwood-cells`
- `./tools/dev just ci` (772 tests plus formatting, clippy, docs, API reachability, packaged docs, and external-consumer checks)

Closes #333
Closes #338
Closes #339
Closes #344

Tracking: #348

### DIFF
--- a/crates/shelterwood-cells/src/cells/member.rs
+++ b/crates/shelterwood-cells/src/cells/member.rs
@@ -662,9 +662,7 @@ impl MemberCell {
         );
         if let Some(teardown) = teardown {
             txn.defer(move || {
-                if let Some(payload) = teardown.finish() {
-                    runtime::dispose_detached(payload);
-                }
+                runtime::dispose_detached(teardown.finish());
             });
         }
         // First terminalizer wins. A losing edge neither reclassifies startup
--- a/crates/shelterwood-mailbox/src/capability.rs
+++ b/crates/shelterwood-mailbox/src/capability.rs
@@ -227,15 +227,57 @@ pub(crate) mod tests {
         MailboxRuntime, MailboxSignal, MailboxSignalWatcher,
     };
 
-    struct AdapterRuntime;
+    type ErasedOneShot = (
+        Box<dyn ErasedOneShotSender>,
+        Pin<Box<dyn ErasedOneShotReceiver>>,
+    );
 
-    impl MailboxRuntime for AdapterRuntime {
+    type OneShotHook = dyn Fn() -> ErasedOneShot + Send + Sync;
+    type NowHook = dyn Fn() -> Instant + Send + Sync;
+
+    /// The one mailbox runtime used by this crate's tests. Optional hooks let
+    /// a focused test control one capability while every other method keeps
+    /// using the real runtime adapter primitives below.
+    pub(crate) struct TestRuntime {
+        oneshot: Option<Box<OneShotHook>>,
+        now: Option<Box<NowHook>>,
+    }
+
+    impl TestRuntime {
+        pub(crate) fn new() -> Self {
+            Self {
+                oneshot: None,
+                now: None,
+            }
+        }
+
+        pub(crate) fn with_oneshot(
+            mut self,
+            oneshot: impl Fn() -> ErasedOneShot + Send + Sync + 'static,
+        ) -> Self {
+            self.oneshot = Some(Box::new(oneshot));
+            self
+        }
+
+        pub(crate) fn with_now(
+            mut self,
+            now: impl Fn() -> Instant + Send + Sync + 'static,
+        ) -> Self {
+            self.now = Some(Box::new(now));
+            self
+        }
+    }
+
+    impl MailboxRuntime for TestRuntime {
         fn oneshot(
             &self,
         ) -> (
             Box<dyn ErasedOneShotSender>,
             Pin<Box<dyn ErasedOneShotReceiver>>,
         ) {
+            if let Some(oneshot) = &self.oneshot {
+                return oneshot();
+            }
             let (sender, receiver) = oneshot();
             (
                 Box::new(AdapterOneShotSender(sender)),
@@ -252,7 +294,7 @@ pub(crate) mod tests {
         }
 
         fn now(&self) -> Instant {
-            now()
+            self.now.as_ref().map_or_else(now, |now| now())
         }
 
         fn sleep_until(&self, deadline: Option<Instant>) -> BoxedSleep {
@@ -271,7 +313,13 @@ pub(crate) mod tests {
         }
     }
 
-    struct AdapterOneShotReceiver(OneShotReceiver<ErasedValue>);
+    pub(crate) struct AdapterOneShotReceiver(OneShotReceiver<ErasedValue>);
+
+    impl AdapterOneShotReceiver {
+        pub(crate) fn new(receiver: OneShotReceiver<ErasedValue>) -> Self {
+            Self(receiver)
+        }
+    }
 
     impl ErasedOneShotReceiver for AdapterOneShotReceiver {
         fn poll_receive(
@@ -323,6 +371,6 @@ pub(crate) mod tests {
     }
 
     pub(crate) fn runtime() -> Arc<dyn MailboxRuntime> {
-        Arc::new(AdapterRuntime)
+        Arc::new(TestRuntime::new())
     }
 }
--- a/crates/shelterwood-mailbox/src/cell.rs
+++ b/crates/shelterwood-mailbox/src/cell.rs
@@ -2485,6 +2485,9 @@ pub(super) mod tests {
             first.poll(None, Waker::noop()),
             super::OperationPoll::Accepted(bound) if bound == incarnation
         ));
+        let closed = close(&mailbox, incarnation).expect("the bound incarnation closes");
+        let (_token, payload) = closed.into_parts();
+        drop(payload);
 
         let mut withdrawal = mailbox.withdraw(&second, super::WithdrawalDisposition::Inline);
         assert!(matches!(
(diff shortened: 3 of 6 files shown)


ralexstokes:
Fresh adversarial review fixed three coverage defects in 602d133: an observe_all test shadowed by Bound-state evidence, an inline-waker mutation that could abort before cleanup, and incomplete missing-message guard-release branches. Each target mutation now fails normally; full local CI passes with 772 tests.

PR #361, opened 2026-08-19
Unify private definition dispatch

### LINKED ISSUE #353: Design question: generic add<D: Definition> under the eight nominal entry points (declaration-matrix ruling)

ralexstokes:
## Summary

- keep all eight nominal public `add_*` entry points and every existing slot method, but route them through one sealed private `Definition` dispatch with associated `Kind` and `Handles`
- centralize static/dynamic reservation in a private `SlotKind` path that owns actor-mailbox attachment and subtree-flavor selection
- make builder and dynamic additions thin concrete wrappers over one generic helper per receiver flavor, preserving definition isolation across eager id/readiness hooks
- record the declaration-matrix ruling in SPEC and keep external conformance checks explicit for every public slot signature

## Decision and tradeoffs

The broad matrix collapse remains rejected. `ActorDef`/`ActorOnceDef`, `RawDef`/`RawOnceDef`, `TaskDef`/`TaskOnceDef`, and `SubtreeDef`/`SubtreeOnceDef` remain nominal public types. Their construction-source bounds, missing one-shot `.restart()` setter, and `TaskOnceDef<T>`'s typed completion handle remain compile-time properties.

The narrower private trait is worthwhile when paired with `SlotKind`: `Definition` alone would only move the same choreography into eight trait impls. Together they reduce sixteen behavior-bearing add bodies to two helpers, six static/dynamic reservation constructors to two generic functions, and remove the endpoint-flavor x definition-mode cross-product from the slot cores. The remaining per-definition impls contain only the irreducible mapping to construction and handle set.

The trait is `pub(super)` and cannot be implemented or named by consumers. Public methods retain concrete definition parameters and concrete return types, preserving rustdoc grouping, autocomplete, inference, and per-kind compiler errors. The three `reserve_*` families also remain public and distinct: actor mailbox type and subtree flavor are fixed before a definition is available.

`api_trait_conformance` intentionally continues spelling out both static and dynamic slot signatures. Shrinking that public inventory would weaken the evidence that the internal generic path did not leak into the API; the test names and comments now state that boundary explicitly.

## Stack / merge order

This draft is stacked on #359 (`agent/review-348-facade`) because it relies on that PR's definition-isolation changes.

1. Merge #359 first.
2. Retarget this PR from `agent/review-348-facade` to `main` (and rebase if GitHub does not carry the stack forward cleanly).
3. Re-run CI, then review/merge this PR.

## Validation

- `./tools/dev cargo nextest run -p shelterwood --test api_trait_conformance --test one_shot_resource_ownership --test dynamic --test raw --test disposal` — 110/110 passed
- `./tools/dev cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings`
- `./tools/dev just ci`
  - formatting and warnings-denied Clippy
  - 770/770 nextest tests
  - doctests and rustdoc warnings
  - public API runtime reachability
  - packaged docs and external consumer checks
  - Nix formatting

Closes #353
Tracking #348

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1780,8 +1780,15 @@ API-shape rules:
 - The spec surface for one kind names operations identically across kinds —
   the one-shot operation is the `_once` twin, never a differently-named
   `spawn_once`.
-- Adding a new child kind or mode extends the shared record, not a
-  hand-maintained matrix.
+- Adding a new child kind or mode extends the shared record and the private
+  declaration dispatch, not duplicated reserve/add/define choreography. The
+  dispatch is deliberately sealed inside the façade: its associated handle
+  set and slot kind let the implementation share that choreography, but the
+  eight public add entry points and the nominal slot methods above it remain
+  concrete. Making the dispatch public would admit an oversized extension
+  surface and turn per-kind parameter errors into generic trait-bound errors;
+  collapsing the reserve methods would also be false because actor mailbox
+  type and subtree flavor are fixed before a definition exists.
 
 **Slots — the reserve-before-define surface.** §3.2's cell machinery has
 one public face, uniform across the three kinds and both scope flavors.
--- a/crates/shelterwood/src/tree/builders.rs
+++ b/crates/shelterwood/src/tree/builders.rs
@@ -3,7 +3,7 @@ use std::{fmt, sync::Arc};
 use crate::{
     ActorDef, ActorOnceDef, ActorRef, ChildId, Intensity, ScopeDefaults,
     admission::ReserveError,
-    plan::{BuilderCore, LowerError, SlotCell},
+    plan::{BuilderCore, LowerError},
     policy::{ResolvedDefaults, ScopeFlavor},
     raw::{RawDef, RawOnceDef},
     runtime,
@@ -13,10 +13,7 @@ use crate::{
 
 use super::{
     ActorSlot, Subtree, SubtreeDef, SubtreeOnceDef, SubtreeSlot, System, TaskSlot,
-    slots::{
-        ActorSlotCore, StaticSlotEndpoint, SubtreeSlotCore, TaskSlotCore, attach_actor_mailbox,
-    },
-    system::sealed,
+    slots::{ActorKind, Definition, SubtreeKind, TaskKind, reserve_static},
 };
 
 /// A root lowering error.
@@ -49,10 +46,15 @@ pub(super) fn dispose_rejected<D: Send + 'static>(
     error
 }
 
-fn attach_actor_slot<M: Send + 'static>(slot: Arc<SlotCell>) -> ActorSlot<M> {
-    let mailbox = attach_actor_mailbox(&slot);
-    ActorSlot {
-        core: ActorSlotCore::new(StaticSlotEndpoint(slot), mailbox),
+fn add_definition<D: Definition>(
+    core: &mut BuilderCore,
+    id: impl Into<ChildId>,
+    definition: D,
+) -> Result<D::Handles, ReserveError> {
+    let mut definition = runtime::Isolated::new(definition);
+    match reserve_static::<D::Kind>(core, id) {
+        Ok(slot) => Ok(slot.define(definition.take().expect("isolated definition is available"))),
+        Err(error) => Err(dispose_rejected(definition, error)),
     }
 }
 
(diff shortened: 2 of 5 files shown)


ralexstokes:
Adversarial architecture review completed at head 1cb11a7.

Finding addressed:

- The private Definition trait still received SlotCore by value, so each raw/task/subtree implementation owned the final endpoint.define transition. That duplicated the bind/admit choreography across semantic lowering implementations and gave the seam more authority than its stated purpose required.
- 1cb11a7 changes Definition::lower to borrow the slot and return only its typed handles plus ChildConstruction. SlotCore::define is now the sole owner of consuming and binding/admitting the endpoint. A new definition mode can supply semantic lowering without being able to reimplement endpoint lifecycle behavior.

Adversarial checks:

- The stack remains cleanly scoped to base PR #359; this review commit changes only tree/slots.rs.
- Rustdoc JSON exposes exactly three public copies of every add_* and reserve_* method (Tree, DynamicTree, DynamicScopeRef), while Definition, SlotKind, and their markers are not publicly reachable.
- A downstream negative compiler probe still reports concrete TaskDef versus RawDef parameter mismatches at the nominal methods; the private trait does not leak into diagnostics.
- Focused nextest: api_trait_conformance, one_shot_resource_ownership, dynamic, raw, and disposal — 110/110 passed. This covers all kind/mode admissions, typed TaskOnce handles, split/fused ownership, hostile id/readiness hooks, rejected-definition disposal, and one-shot resource ownership.
- Full ./tools/dev just ci passed: warnings-denied Clippy, 770/770 tests, doctests, rustdoc warnings, public runtime reachability, packaged docs, external consumer, and Nix formatting.

Decisions retained: all eight nominal add methods, all three reserve families, distinct Once/Def types and bounds, typed handles, concrete inference/error surfaces, eager evaluation order, and isolated rejected-definition disposal. The narrowed path adds no lock acquisition; raw readiness still runs with its definition isolated, and endpoint binding remains on the existing post-reservation path. No unresolved review findings remain.

Issue #366, opened 2026-08-20
Cancellation/unwind containment: confirmed fixes from the #363 adjudication


ralexstokes:
Successor to #363. Every finding in that tracker was re-verified against the code at `96c078ca0d25cc9f00a4c55f120ab36a744e9cd7` (current `main` differs only by spec-reference renumbering in comments). This issue carries the confirmed, actionable work with corrected severities and remedies; the full adjudication — including refuted findings and remedy corrections — is recorded in the closing comment on #363. The waker-proxy work split out to #367 and the crate-merge design round to #368; the construction-after-latched-stop semantics question stays closed as refuted (the resolved dispositions of the `Handler`/raw contract and the phase-authority kernel are Batch 4 below).

Ordering note (**corrected 2026-08-20**): the abort-class items in Batch 1 do **not** need subprocess regressions. An in-process assertion is indeed inert where the failure mode is a double-panic abort — but the assertion is not what reports the failure. `cargo nextest` runs every test in its own process, so the abort surfaces as SIGABRT on that single test rather than poisoning the run (confirmed by execution during the #370 review). `justfile:25-26` makes nextest the only runner for `--lib --bins --tests --examples`; `cargo test` is used solely for `--doc`, which never sees these. Both CI jobs go through `nix develop --command just ci`.

Write abort-class regressions as ordinary `#[test]`s, keeping the assertions that pin *which* panic survives and what was destroyed. Reach for a subprocess harness only when a test must observe state past the abort point, or genuinely depends on the child's single-threaded environment — and say so in a comment when you do.

The original rule sent three PRs (#369, #370, #372) to a hand-rolled harness that re-implemented isolation the runner already provides, and the harness carried its own defect: libtest exits 0 when `--exact` matches nothing, so a hardcoded `TEST_NAME` going stale makes the whole regression pass while running zero tests. That vacuity was demonstrated by mutation on two of those PRs (revert the fix **and** rename the module ⇒ test still passed). The idiom's original site, `crates/shelterwood/tests/reply_waker_containment.rs`, has the same hazard. All four sites are now ordinary `#[test]`s (#369, #370, #372, and the idiom's original in #385); the accepted trade is that a plain `cargo test --lib` run degrades from one failing test to a whole-run abort — still a loud failure, not a silent pass.

## Batch 1 — small, high-value fixes

[...]

PR #369, opened 2026-08-20
Contain obligation fallback panics during unwind

ralexstokes:
Addresses the highest-priority Batch 1 item from #366.

Obligation fallback discharge now runs through PanicAccumulator. An ordinary fallback panic still reaches the caller, while a fallback reached during an existing unwind is contained so it cannot double-panic and abort the process.

The new subprocess regression drives the nested-unwind case and proves the original panic reaches its boundary.

Verification: ./tools/dev just ci (804 tests plus all repository checks).

### DIFF
--- a/crates/shelterwood/src/driver/storage.rs
+++ b/crates/shelterwood/src/driver/storage.rs
@@ -40,7 +40,8 @@ impl<T> Obligation<T> {
 
 impl<T> Drop for Obligation<T> {
     fn drop(&mut self) {
-        self.discharge();
+        let mut panics = crate::runtime::PanicAccumulator::default();
+        panics.run(|| self.discharge());
     }
 }
 
@@ -48,6 +49,7 @@ impl<T> Drop for Obligation<T> {
 mod tests {
     use std::{
         panic::{AssertUnwindSafe, catch_unwind},
+        process::Command,
         sync::{
             Arc,
             atomic::{AtomicUsize, Ordering},
@@ -62,6 +64,10 @@ mod tests {
 
     fn panic_after_counting(fallbacks: Arc<AtomicUsize>) {
         fallbacks.fetch_add(1, Ordering::SeqCst);
+        panic!("injected fallback panic");
+    }
+
+    fn panic_fallback((): ()) {
         panic!("injected fallback panic");
     }
 
@@ -92,6 +98,39 @@ mod tests {
             fallbacks.load(Ordering::SeqCst),
             1,
             "drop cannot repeat a fallback that panicked"
+        );
+    }
+
+    #[test]
+    fn obligation_fallback_panic_is_contained_during_an_existing_unwind() {
+        const CHILD_ENV: &str = "SHELTERWOOD_OBLIGATION_UNWIND_CHILD";
+        const TEST_NAME: &str = "driver::storage::tests::obligation_fallback_panic_is_contained_during_an_existing_unwind";
+
+        if std::env::var_os(CHILD_ENV).is_some() {
+            let panic = catch_unwind(|| {
+                let _obligation = Obligation::new((), panic_fallback);
+                panic!("outer panic");
+            })
+            .expect_err("the original unwind reaches its boundary");
+            assert_eq!(panic.downcast_ref::<&str>(), Some(&"outer panic"));
+            return;
+        }
+
+        let output = Command::new(std::env::current_exe().expect("unit-test executable"))
+            .arg("--exact")
+            .arg(TEST_NAME)
+            .arg("--nocapture")
+            .arg("--test-threads=1")
+            .env(CHILD_ENV, "1")
+            .output()
+            .expect("nested-unwind subprocess starts");
+
+        assert!(
+            output.status.success(),
+            "nested-unwind subprocess must preserve the original panic instead of aborting\nstatus: {}\nstdout:\n{}\nstderr:\n{}",
+            output.status,
+            String::from_utf8_lossy(&output.stdout),
+            String::from_utf8_lossy(&output.stderr),
         );
     }
 }
--- a/crates/shelterwood/src/driver/storage.rs
+++ b/crates/shelterwood/src/driver/storage.rs
@@ -61,7 +61,6 @@ impl<T> Drop for Obligation<T> {
 mod tests {
     use std::{
         panic::{AssertUnwindSafe, catch_unwind},
-        process::Command,
         sync::{
             Arc,
             atomic::{AtomicUsize, Ordering},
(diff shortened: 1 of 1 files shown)


crates/shelterwood/src/driver/storage.rs
```diff
@@ -94,4 +100,37 @@ mod tests {
             "drop cannot repeat a fallback that panicked"
         );
     }
+
+    #[test]
+    fn obligation_fallback_panic_is_contained_during_an_existing_unwind() {
+        const CHILD_ENV: &str = "SHELTERWOOD_OBLIGATION_UNWIND_CHILD";
+        const TEST_NAME: &str = "driver::storage::tests::obligation_fallback_panic_is_contained_during_an_existing_unwind";
```
ralexstokes: **Nit (verified empirically): this `--exact` filter is a fail-open point.**

`TEST_NAME` hardcodes the full module path. libtest exits **0** when `--exact <name>` matches nothing ("0 passed; N filtered out"), so a stale name makes the parent's `status.success()` assertion at line 128 pass while running nothing.

Evidence: with the production fix **still reverted**, I additionally changed this constant to `driver::renamed_module::tests::…` and the test **PASSES**. So the regression's continued validity rests on a string nobody is forced to update — and #366's own dispositions contemplate opportunistic file splits, which is exactly the drift that would trigger it.

One-line remedy: also assert the child actually ran.

```rust
let stdout = String::from_utf8_lossy(&output.stdout);
assert!(stdout.contains("1 passed"), "…{stdout}");
```

(Optionally derive the module half with `module_path!().split_once("::").expect(...).1`; the stdout assertion is the belt that also catches a fn rename.)

**Not a defect this PR introduces** — it is inherited verbatim from `crates/shelterwood/tests/reply_waker_containment.rs:119-127`, so it is a codebase-wide trap. It is *sharper* here only because that precedent filters on a crate-root integration-test fn name while this encodes a three-segment module path. The same issue was independently confirmed on #370. Since #366 Batch 1 calls for several more subprocess regressions, the better fix is one shared helper asserting both exit status and "1 passed", landed centrally, rather than a patch per PR.

crates/shelterwood/src/driver/storage.rs
```diff
@@ -40,14 +40,16 @@ impl<T> Obligation<T> {
 
 impl<T> Drop for Obligation<T> {
     fn drop(&mut self) {
-        self.discharge();
+        let mut panics = crate::runtime::PanicAccumulator::default();
```
ralexstokes: Two nits on this hunk, neither blocking:

1. **Undocumented containment split.** The `Obligation` doc block (lines 5-9) still says only that dropping executes the fail-closed fallback and that fallbacks must never await or join. Worth stating the new asymmetry there: a fallback panic resumes from an ordinary drop, and is discarded when the obligation is destroyed during an existing unwind (losing the diagnostic being the lesser outcome). Without it, the three-way split between this `Drop` (self-contained), the direct `discharge()` at `driver.rs:424` (contained by the caller's accumulator), and the existing test at line 87 (which asserts the panic *reaches* the caller) reads as accidental rather than decided.

2. **Path spelling.** `crate::runtime::PanicAccumulator` inline; every sibling driver module does `use crate::runtime;` and writes `runtime::PanicAccumulator`. Purely cosmetic.

ralexstokes (commented):
**Verdict: merge-with-nits.** Nothing here is blocking. The fix is correct, minimal, and at the right seam, and it implements the #366 Batch 1 bullet as specified: the `Drop` seam covers all four `Obligation` sites uniformly, and both the explicit `discharge()` in `ScopeRuntime::drop` (`crates/shelterwood/src/driver.rs:424`, already inside that destructor's own accumulator) and the existing `storage.rs` fallback test are untouched.

### Verified by execution (own worktree off the PR head, since removed)

- **The regression is not vacuous.** Committed the PR head, injected a mutation reverting `Obligation::drop` to bare `self.discharge()`, committed, re-ran: the new test **FAILS**, and on the right signal — the parent assertion reports `status: signal: 6 (SIGABRT) (core dumped)` with the child's `panic in a destructor during cleanup / thread caused non-unwinding panic. aborting.` in captured stderr. The test discriminates abort vs. no-abort through child exit status, not a side channel. The child-side `assert_eq!(panic.downcast_ref::<&str>(), Some(&"outer panic"))` (`storage.rs:115`) sits outside the `catch_unwind`, so it is not an inert assertion, and it pins that the *original* panic survives rather than the fallback's.
- `cargo nextest run -p shelterwood --lib`: 212/212 pass at PR head.
- `cargo +nightly fmt --all -- --check`: clean (the long `TEST_NAME` string literal does not trip the gate). `cargo clippy -p shelterwood --lib --tests --all-features`: clean.
- Not run: `just ci` / `just ci-nix` (shared machine) — relying on the green PR check. Non-Linux behavior unchecked; CI is Linux-only and the pattern is portable in principle.

### Checked and refuted — recording so a later review does not re-file

- **`PanicAccumulator` semantics.** `Drop` -> `resume_preferred_panic { primary: None, cleanup: first }` -> discard both when `thread::panicking()`, else resume (`crates/shelterwood-core/src/panic.rs:60-124`). Exactly the claimed split. `catch_panic` supplies `AssertUnwindSafe`, so capturing `&mut self` is fine, and `discard_panic` already tolerates a hostile payload destructor.
- **Seam placement.** `discharge()` takes the payload before invoking the fallback, so a contained fallback panic cannot leave a re-dischargeable payload; `self.payload` is `None` on every post-`discharge` path, and the accumulator local drops before the (trivial) field glue.
- **All four sites.** Exactly four `Obligation` fields exist — `AdmissionRequest::response` (`admission_control.rs:584`, ctor at `:305`), `DynamicEntry::removal` (`:78`, ctors at `:102/:114/:123`), `ReportToken::completion` (`child.rs:17/42`), `ChildStart::terminality` (`child.rs:162/189`) — and all four reach their fallback only through `Drop`.
- **The named abort scenario is genuinely closed.** `Pending::Admission(request)` (`driver.rs:1237`) is destroyed by the `pending.drain(..)` guard at `driver.rs:1208` when the loop body unwinds; that path now runs contained.
- **No bare `complete()` on an unwind path.** `Obligation::complete`'s closure still runs bare, but no `Drop` impl in `driver.rs` / `child.rs` / `admission_control.rs` reaches it; every call site is an orderly path.
- **Lock rule.** No new lock, none held across the new code, `PanicAccumulator` takes none. The change strictly narrows what escapes a destructor.
- **SPEC.** No guarantee text needs changing. SPEC §15.4 already names "discharges it, panic-contained, from `Drop`" as *the* mechanism (`specs/SPEC.md:2776`), so this makes an already-stated rule true at this site. SPEC's double-panic carve-out (`specs/SPEC.md:1386-1405`) and `docs/shutdown-and-resources.md:160-162` scope the abort exclusion to *user* destructors inside the user's own unwind — unaffected. `Obligation` is named nowhere in `specs/` or `docs/`.

PR #370, opened 2026-08-20
Contain completed output destruction in panic boundary

ralexstokes:
Addresses the `CatchUnwindFuture` Batch 1 item from #366.

When a completed inner future panics during destruction, the already-produced output is now destroyed inside containment before the future-destructor panic is returned.

**Corrected justification (the original "abort class" framing does not hold).** Reverting the fix on a committed tree and driving the hostile-output/hostile-payload combination in a subprocess produces **exit 101, not SIGABRT**: when the output destructor panics, the retained payload has already moved into the return place, and Rust never drops the return place of an unwinding function, so the payload is *leaked* rather than double-dropped. The real pre-fix defects, both fixed here, are:

1. **Precedence inversion with an escape.** The output-destructor panic unwound out of `poll` instead of the documented future-destructor panic becoming the error — contradicting the type's own rustdoc. At `raw/definition.rs:253` that escape skips mailbox freeze, resource join and the ordered epilogue; at `driver/child.rs:580` it skips `report.record(...)`.
2. **A leaked panic payload** on that same escape.

The hostile-destructor pair is now a plain in-process test. Nextest runs every test in its own process and every workspace lane reaches the unit tests through it, so a double-panic abort already fails exactly one test — the subprocess harness was re-implementing that isolation, and its stale-name guard existed only to keep the harness honest. Collapsing it also lets the test identify the *surviving payload*, which it could not do across a process boundary, so it now discriminates an inverted precedence too.

**Coverage.** Both poll arms are pinned for containment *and* precedence; the doc's ordering promise (a clean inner future destroyed exactly once before the output is released) and the drop-glue accumulator (resumes outside an unwind, contains during one) are pinned too.

**Verification.** `./tools/dev just ci` on the first commit (806 tests plus all repository checks). On the follow-up commits: 559/559 `-p shelterwood`, `just fmt`, clippy clean, plus a mutation battery in which each mutation is caught by the intended test — revert the fix; invert the Ok-arm precedence; invert the Err-arm precedence; bare drop in the Err arm; skip the Ok-arm future destruction; drop glue discarding instead of resuming; drop glue dropping bare (SIGABRT).

### DIFF
--- a/crates/shelterwood/src/raw/disposal.rs
+++ b/crates/shelterwood/src/raw/disposal.rs
@@ -43,7 +43,10 @@ impl<F: Future> Future for CatchUnwindFuture<F> {
                 let future = this.future.take();
                 match catch_panic(|| drop(future)) {
                     Ok(()) => Poll::Ready(Ok(value)),
-                    Err(payload) => Poll::Ready(Err(payload)),
+                    Err(payload) => {
+                        discard_panic(catch_panic(|| drop(value)).err());
+                        Poll::Ready(Err(payload))
+                    }
                 }
             }
             Ok(Poll::Pending) => Poll::Pending,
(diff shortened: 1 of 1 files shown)


crates/shelterwood/src/raw/disposal.rs
```diff
@@ -43,7 +43,10 @@ impl<F: Future> Future for CatchUnwindFuture<F> {
                 let future = this.future.take();
                 match catch_panic(|| drop(future)) {
                     Ok(()) => Poll::Ready(Ok(value)),
-                    Err(payload) => Poll::Ready(Err(payload)),
+                    Err(payload) => {
+                        discard_panic(catch_panic(|| drop(value)).err());
```
ralexstokes, line 49: **Finding: the abort-class framing this change is justified by does not reproduce.** The fix is correct and worth landing — what changes is the *reason*.

Evidence: with exactly this line reverted on a committed tree, the hostile-output child (recursively-panicking payload) exits **101, not SIGABRT**. Observed directly, twice: the child prints the `panic_any` diagnostic, then `injected output destructor panic`, then exits 101.

Why: when `value`'s destructor panics, `payload` has *already* been moved into the return place `_0` by the tail expression, and Rust does not drop the return place when a function unwinds. The hostile payload is therefore **leaked**, not double-dropped. No second panic, no abort.

The real pre-fix defects, both of which this line does fix:

1. **Precedence inversion with an escape.** The output-destructor panic unwinds out of `poll` instead of the documented future-destructor panic becoming the error. At `raw/definition.rs:253` that escape skips mailbox freeze, resource join, and `resume_preferred_panic_outside_unwind` — the whole ordered epilogue; at `driver/child.rs:580` it skips `report.record(...)` and falls back to the obligation's drop path. That is the severity here, and it is user-reachable.
2. **A leaked panic payload** on that same escape.

Also worth stating in the PR narrative: the pre-fix code contradicted the type's *own* rustdoc (disposal.rs:14-16, unchanged by this PR), which already promised that the output is discarded and the destructor panic becomes the error. This is a code-catches-up-to-doc fix, which is a stronger and more checkable justification than the abort claim.

Ask: correct the #366 bullet text when it is closed. The tracker wording is what the next reviewer will trust, and "abort class" there will send someone hunting a double panic that this shape cannot produce.

crates/shelterwood/src/raw/disposal.rs
```diff
@@ -181,3 +184,191 @@ impl<T> Drop for Contained<T> {
...
+                .as_mut()
+                .poll(&mut Context::from_waker(Waker::noop())),
+        );
+        assert_eq!(panic_message(&payload), Some("injected poll panic"));
+        discard_panic(Some(payload));
+        assert_eq!(future_drops.load(Ordering::SeqCst), 1);
+    }
+
+    #[test]
+    fn hostile_output_cannot_abort_while_a_hostile_future_panic_is_retained() {
```
ralexstokes: **Nit, paired with the finding above: retitle.** `hostile_output_cannot_abort_...` and the assert message at line 368 ("must return the retained panic instead of aborting") both assert an abort that the reverted code does not produce — the child exits 101. Something like `hostile_output_destructor_cannot_escape_the_boundary` and "must return the retained panic instead of escaping the boundary" would describe what the test actually discriminates.

Keeping the subprocess isolation is fine — `status.success()` is a superset check that would still catch a real abort. Note for calibration though: under the *inverted-precedence* mutation this test passes; only the in-process `completed_future_drop_panic_wins_...` catches that, because this child asserts drop counts and not the payload identity. Worth a sentence in the test comment so a future reader does not over-trust the subprocess as the precedence pin.

crates/shelterwood/src/raw/disposal.rs
```diff
@@ -181,3 +184,191 @@ impl<T> Drop for Contained<T> {
...
+        const CHILD_ENV: &str = "SHELTERWOOD_CATCH_UNWIND_OUTPUT_CHILD";
+        const TEST_NAME: &str = "raw::disposal::tests::hostile_output_cannot_abort_while_a_hostile_future_panic_is_retained";
```
ralexstokes: **Nit: this subprocess filter fails open on a rename.** `TEST_NAME` is a hardcoded module path. Verified against the built test binary: `<test-bin> --exact <nonexistent-name>` prints `running 0 tests ... test result: ok. 0 passed` and exits **0**. So if this test or its module is ever renamed or moved, the child runs zero tests, exits 0, and the parent's `assert!(output.status.success())` at line 366 passes vacuously — the regression silently stops being covered, with no signal anywhere.

This is inherited, not introduced: the same shape is at `crates/shelterwood/tests/reply_waker_containment.rs:119`, and it was independently confirmed on #369. Three sites now share one weakness, so the fix is one shared test helper rather than three per-site patches — spawn by name, assert the exit status **and** that stdout reports `1 passed` (the count assertion as an *additional* guard, not the primary one; a grep-only probe would fail open in the other direction).

ralexstokes (commented):
**Verdict: merge.** Nothing below blocks. The production change is the remedy #366 specifies, in the right place, and the tests are non-vacuous — I verified that by mutation on a committed tree in a throwaway worktree at `c4400e5`, not by reading them.

### Mutation table (executed)

| Mutation | Detected by |
| --- | --- |
| Revert the fix (`Err(payload) => Poll::Ready(Err(payload))`) | `completed_future_drop_panic_wins_and_output_drop_is_contained` **fails** (the output-destructor panic escapes `poll`) + the subprocess test **fails** |
| Keep containment, return the *output* payload (inverted Ok-arm precedence) | `completed_future_drop_panic_wins_...` **fails** on the message assertion (`left: Some("injected output destructor panic")`) |
| Invert Err-arm precedence (return the future-drop payload) | `poll_panic_wins_and_future_drop_panic_is_contained` **fails** on the message assertion |
| Bare `drop(future)` in the Err arm | `poll_panic_wins_...` **fails** (escaping panic) |

So both arms are pinned for containment **and** for precedence — not merely against deletion of the `discard_panic` call, which is the failure mode this kind of test usually has. Assertions all sit in the test body or are observed through a child exit status, so none are inert.

Other checks: `nextest -p shelterwood` 556/556 pass (214/214 lib), `just fmt` clean, `cargo clippy -p shelterwood --lib --tests --all-features` clean. I did not run full `just ci`.

Containment reasoning also holds on the paths I could not mutate: `discard_panic` is itself contained (`catch_panic` + `mem::forget`, shelterwood-core/src/panic.rs:28-37), so a recursively hostile output payload cannot unwind into the live `payload`; no lock is held anywhere in `poll`, so the lock rule is not engaged; and the fix reaches all four instantiations (`raw/definition.rs:253`, `:275`, `driver/child.rs:580`, `raw/context.rs:1280`) because it lives in the type.

### Nits (clearly optional, none blocking)

- **No coverage for `Drop for CatchUnwindFuture`** (disposal.rs:59-66). That path resumes a hostile inner-future destructor panic *out of drop glue* via `PanicAccumulator`, on a non-unwinding caller. The new module is the natural home for it and it is the one remaining unpinned panic path on this type. (Not anchorable inline — outside the diff.)
- **No happy-path assertion.** Nothing pins the rustdoc's first sentence: a clean future is destroyed *before* the output is released, exactly once. A two-line `Ok(value)` test with a drop-order probe would close it.
- **Rustdoc could be strengthened** (disposal.rs:14-16). It says the output "is discarded"; after this change that discarding is itself contained — the output destructor's own panic is swallowed. Worth saying, since that is the property the new line establishes.
- **Venue, out of scope and not a regression.** The discarded output is still destroyed inline on the incarnation/offload task, so a blocking user destructor stalls the executor. Same class as #366's separate "destroy opaque panic payloads off the executor" bullet; pre-fix it dropped inline too.

ralexstokes:

- **Abort-class refutation** (`disposal.rs`): the fix stands; the *justification* was corrected. Test renamed to `hostile_output_destructor_cannot_escape_while_a_hostile_future_panic_is_retained` and the assert message now says "instead of escaping the boundary" (`6b46c1c`); PR body rewritten with the exit-101 evidence; issue #366's bullet corrected to name the precedence inversion plus the leaked return place instead of a double panic. The conversion in `cd3ea3c` then re-asserted the refuted claim in a doc comment — fixed in `424f8a1`, which moves the abort sentence to the one test that genuinely aborts and has each comment name its own measured failure mode.
- **Retitle nit**: done in `6b46c1c`.
- **Fail-open `--exact` filter**: superseded rather than patched. `cd3ea3c` removes the subprocess harness entirely — `cargo nextest` already gives every test its own process, so there is no `TEST_NAME` left to go stale. Issue #366's ordering note has been corrected accordingly, including the per-bullet instructions in Batch 1 and Batch 2.

Also added in this branch: drop-glue coverage (resume outside an unwind, containment during one) and a happy-path assertion that a clean future is destroyed exactly once before its output is released — the rustdoc's first sentence was previously unpinned. Nine mutations, each caught by the intended test.

Issue #367, opened 2026-08-20
Isolate cancelled-timer waker retirement behind a framework-owned proxy (timer path only)


ralexstokes:
Split out of #366's "deferred pending a ruling" section, with the ruling now made: **in scope, timer path only.** Follow-up to the #363 adjudication.

## Why this reopens a decided rule

The `WithdrawalDisposition::Inline` adjudication (`shelterwood/src/futures.rs`, "the registered waker destructor stays inline, so its panic remains visible") held that a caller-side waker drop stays confined to the caller's own frame. That premise is broken for the timer path by a fact that was not on the table when it was decided: Tokio's timer cancel path ([`Handle::clear_entry`, tokio-1.53.1 `runtime/time/mod.rs:380-390`](https://docs.rs/tokio/1.53.1/src/tokio/runtime/time/mod.rs.html)) discards the registered waker **while holding the global time-driver mutex**. The firing path deliberately releases the lock before waking; the cancel path does not — and current tokio master has the same shape, so a pin bump will not dissolve it. A blocking hostile-waker destructor therefore stalls every timer registration and cancellation in the entire runtime (measured ~450 ms cross-thread against a 600 ms destructor in a standalone probe). The blast radius escapes the hostile caller, which makes this a framework liveness property — correctness, not performance.

The oneshot cancel path is different and the inline ruling **stands** there: tokio's oneshot close is atomics-only, so the damage from a hostile waker destructor stays in the hostile caller's own frame. #366 Batch 1's containment drops close the abort class on both paths; this issue is only about the blocking case that containment cannot address.

## Scope

- [ ] **Proxy the caller waker in `Deadlined`** (`shelterwood-mailbox`), so the waker Tokio's wheel stores — and destroys under its driver lock — is framework-owned and trivial to drop. The stored caller waker retires through `runtime.dispose` on cancellation. Covers `send_timeout`, `call`, `recv`, and both `CallFuture` phases, since all deadline through the same wrapper. A stable proxy also means Tokio's `will_wake` check stops re-registering on every poll — a small offset against the cost.
- [ ] **Placement: extend the `WakerSlot` pattern** (`shelterwood-mailbox/src/cell.rs:111-196` — private storage, effects-sink-mediated dispositions, `will_wake` short-circuit), promoted from its `cell.rs`-internal home to a crate-internal utility. **Not** below the `MailboxRuntime` seam: a proxy in the Tokio adapters would make every future runtime adapter reimplement containment.
[...]

PR #372, opened 2026-08-20
Contain mailbox deadline and receiver retirement

ralexstokes:
Addresses the Deadlined and DisposingReceiver Batch 1 item from #366, including the later timer-retirement addition.

Deadlined now retires its timer on drop through PanicAccumulator, and retires it synchronously inside a non-resuming boundary when the operation completes. Retirement at completion cannot be deferred to a disposal worker: the wheel entry would stay armed until that worker reached it, which is the spurious wake the clear exists to prevent. It cannot resume either — the completed result is a live local across the waker destructor, so a resuming boundary would destroy the delivered output instead of returning it, and a hostile output destructor would make that a double panic and a process abort. A delivered result therefore outranks a waker-cleanup diagnostic, the same precedence CatchUnwindFuture's completed-output arm applies.

DisposingReceiver now owns its channel through an explicit take path and accumulates panics across close, value recovery, disposal submission and final receiver retirement.

Regressions drive the complete public send_timeout, call, and recv futures with a hostile registered waker during an existing unwind, and drive recv through the ready-after-parking arrangement — the only shape in which a completing future still holds an armed timer — with a hostile timer-waker destructor and a hostile reply destructor. These are plain in-process tests: `just test` runs the integration targets under `cargo nextest`, which gives every test its own process, so a double-panic abort fails exactly the test that caused it. A unit regression parks, completes before the deadline, then advances a paused clock past that deadline and requires the retained future not to wake its caller.

Verification: ./tools/dev cargo nextest run --workspace (808 tests), fmt and clippy clean. Each production change is mutation-checked against the test that pins it, and each abort-class regression was confirmed to report SIGABRT against its mutation.

### DIFF
--- a/crates/shelterwood-mailbox/src/capability.rs
+++ b/crates/shelterwood-mailbox/src/capability.rs
@@ -158,41 +158,58 @@ pub(crate) fn dispose<T: Send + 'static>(runtime: &Arc<dyn MailboxRuntime>, valu
 
 /// Receive state that keeps an unclaimed user value out of holder drop glue.
 pub(crate) struct DisposingReceiver<T> {
-    inner: OneShotReceiver<T>,
+    inner: Option<OneShotReceiver<T>>,
     runtime: Arc<dyn MailboxRuntime>,
 }
 
 impl<T: Send + 'static> DisposingReceiver<T> {
     pub(crate) fn new(inner: OneShotReceiver<T>, runtime: Arc<dyn MailboxRuntime>) -> Self {
-        Self { inner, runtime }
+        Self {
+            inner: Some(inner),
+            runtime,
+        }
     }
 }
 
 impl<T> DisposingReceiver<T> {
+    fn inner_mut(&mut self) -> &mut OneShotReceiver<T> {
+        self.inner
+            .as_mut()
+            .expect("a live disposing receiver retains its channel")
+    }
+
     pub(crate) fn runtime(&self) -> Arc<dyn MailboxRuntime> {
         Arc::clone(&self.runtime)
     }
 
     pub(crate) fn close(&mut self) {
-        self.inner.close();
+        self.inner_mut().close();
     }
 }
 
 impl<T: Send + 'static> DisposingReceiver<T> {
     pub(crate) fn poll_receive(&mut self, context: &mut Context<'_>) -> Poll<Option<T>> {
-        self.inner.poll_receive(context)
+        self.inner_mut().poll_receive(context)
     }
 
     pub(crate) fn close_and_poll_receive(&mut self, context: &mut Context<'_>) -> OneShotClose<T> {
-        self.inner.close_and_poll_receive(context)
+        self.inner_mut().close_and_poll_receive(context)
     }
 }
 
 impl<T> Drop for DisposingReceiver<T> {
     fn drop(&mut self) {
-        if let Some(value) = self.inner.close_and_take_erased() {
+        let mut inner = self
+            .inner
+            .take()
+            .expect("a live disposing receiver retains its channel");
+        let mut value = None;
+        let mut panics = crate::panic::PanicAccumulator::default();
+        panics.run(|| value = inner.close_and_take_erased());
+        if let Some(value) = value {
             self.runtime.dispose(value);
         }
+        panics.run(|| drop(inner));
     }
 }
 
--- a/crates/shelterwood-mailbox/src/capability.rs
+++ b/crates/shelterwood-mailbox/src/capability.rs
@@ -205,10 +205,23 @@ impl<T> Drop for DisposingReceiver<T> {
             .expect("a live disposing receiver retains its channel");
         let mut value = None;
         let mut panics = crate::panic::PanicAccumulator::default();
+        // Recovery runs first so an unclaimed value reaches isolated disposal
+        // before the receiver -- and therefore before the waker clone it
+        // registered -- is retired; a hostile waker destructor can neither
+        // divert nor destroy it. If recovery itself unwinds, the value stays
+        // in the channel and `inner`'s own drop glue destroys it inline
+        // rather than through the isolated lane: accepted, because reaching
+        // it requires a destructor that has already panicked, and the
+        // alternative is retrying a step that just failed.
         panics.run(|| value = inner.close_and_take_erased());
-        if let Some(value) = value {
-            self.runtime.dispose(value);
-        }
+        // `dispose` can fall back to destroying the value on this thread when
+        // task and native-thread creation are exhausted, so submission belongs
+        // inside the boundary too.
+        panics.run(|| {
+            if let Some(value) = value {
+                self.runtime.dispose(value);
+            }
+        });
         panics.run(|| drop(inner));
     }
 }
--- a/crates/shelterwood-mailbox/src/deadline.rs
+++ b/crates/shelterwood-mailbox/src/deadline.rs
@@ -111,7 +111,24 @@ impl<F: DeadlineOperation + Unpin> Future for Deadlined<F> {
             this.operation
                 .poll_deadlined(context, budget, DeadlinePhase::InitialAttempt)
         {
-            this.timer = None;
+            // Completion retires the wheel entry here, and only here: a
+            // future held past its result would otherwise keep a timer armed
+            // with a clone of the caller's waker and deliver that expiry to a
+            // caller who already has its answer. Retirement must therefore be
+            // synchronous -- handing the timer to isolated disposal leaves the
+            // entry armed until a worker thread reaches it, which is the
+            // spurious wake this exists to prevent.
+            //
+            // A `RawWaker` vtable is caller code, and `result` -- which may
+            // own a user value -- is a live local across that destructor. A
+            // resuming boundary would destroy the completed output instead of
+            // returning it, and a hostile output destructor would turn that
+            // into a double panic and a process abort, so the panic is
+            // contained rather than resumed: a delivered result outranks a
+            // waker-cleanup diagnostic. This is the precedence
+            // `CatchUnwindFuture`'s completed-output arm already applies.
+            let timer = this.timer.take();
+            crate::panic::discard_panic(crate::panic::catch_panic(|| drop(timer)).err());
             return Poll::Ready(result);
         }
         if this.phase == DeadlinePhase::InitialAttempt {
@@ -137,6 +154,14 @@ impl<F: DeadlineOperation + Unpin> Future for Deadlined<F> {
 }
 
 impl<F> Drop for Deadlined<F> {
+    /// Retires an unelapsed timer inside a boundary.
+    ///
+    /// Cancelling an armed timer destroys the clone of the caller's waker it
+    /// registered, and a `RawWaker` vtable is caller code. There is no caller
+    /// left to surface a panic to here, and during an existing unwind a bare
+    /// destructor panic is a double panic and a process abort, so the
+    /// retirement runs through an accumulator: it resumes on an ordinary drop
+    /// and is contained while another unwind is already in progress.
     fn drop(&mut self) {
         let timer = self.timer.take();
         let mut panics = crate::panic::PanicAccumulator::default();
@@ -148,7 +173,11 @@ impl<F> Drop for Deadlined<F> {
 mod tests {
     use std::{
         future::Future,
-        task::{Context, Poll, Waker},
+        sync::{
+            Arc,
+            atomic::{AtomicUsize, Ordering},
+        },
+        task::{Context, Poll, Wake, Waker},
     };
 
     /// Stays pending on its first elapsed poll, modelling an atomic
@@ -158,9 +187,15 @@ mod tests {
         elapsed_polls: usize,
     }
 
-    struct ReadyImmediately;
+    /// Parks once, then completes strictly before its deadline -- the only
+    /// arrangement in which a retained future can still hold an armed wheel
+    /// entry carrying a clone of the caller's waker.
+    #[derive(Default)]
+    struct ReadyAfterParking {
+        polls: usize,
+    }
 
-    impl super::DeadlineOperation for ReadyImmediately {
+    impl super::DeadlineOperation for ReadyAfterParking {
         type Output = ();
 
         fn poll_deadlined(
@@ -169,10 +204,30 @@ mod tests {
             _budget: crate::deadline::Deadline,
             _phase: super::DeadlinePhase,
         ) -> Poll<()> {
-            Poll::Ready(())
+            self.polls += 1;
+            if self.polls < 2 {
+                Poll::Pending
+            } else {
+                Poll::Ready(())
+            }
         }
 
         fn short_circuit(&mut self) {}
+    }
+
+    /// Records wakes so an armed wheel entry is observable as the spurious
+    /// wake it would deliver rather than as a cleared field.
+    #[derive(Default)]
+    struct WakeRecorder(AtomicUsize);
+
+    impl Wake for WakeRecorder {
+        fn wake(self: Arc<Self>) {
+            self.wake_by_ref();
+        }
+
+        fn wake_by_ref(self: &Arc<Self>) {
+            self.0.fetch_add(1, Ordering::SeqCst);
+        }
     }
 
     impl super::DeadlineOperation for PendingOnFirstExpiry {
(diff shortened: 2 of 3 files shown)


crates/shelterwood-mailbox/src/deadline.rs
```diff
@@ -111,6 +111,7 @@ impl<F: DeadlineOperation + Unpin> Future for Deadlined<F> {
             this.operation
                 .poll_deadlined(context, budget, DeadlinePhase::InitialAttempt)
         {
+            this.timer = None;
```
ralexstokes: ### P1 confirmed — the armed timer is destroyed bare while the output is live

**Reachability is plain ready-after-pending**, no exotic scheduling: poll 1 leaves the operation `Pending`, so `Deadlined` polls the timer and the `BoxedSleep` registers a clone of the caller's waker; poll 2's `poll_deadlined(..., InitialAttempt)` returns `Ready`. `this.timer = None` then destroys the still-armed timer **synchronously and bare**, invoking `RawWaker::drop` — caller code — while `result` is a live local.

**Reproduced, not reasoned.** I injected a probe into this file's test module in a scratch worktree: a `PendingThenReady` operation plus a waker whose `drop` panics, over the real timer from `capability::tests::runtime()`.

```
SCRATCH second_poll_unwound=true output_destroyed=true
```

The waker-drop panic escapes `Deadlined::poll` and the completed output is destroyed during the resulting unwind. Adding a panicking destructor to that output escalates exactly as described:

```
thread ... panicked at library/core/src/panicking.rs:233:5:
panic in a destructor during cleanup
thread caused non-unwinding panic. aborting.
...
process didn't exit successfully: ... (signal: 6, SIGABRT: process abort signal)
```

**Severity, split from the original framing.** The abort needs *two* hostile destructors — the waker's and the user value's. But the milder outcome needs only the hostile waker, and it is bad on its own: a **successfully completed** operation silently loses its user value and surfaces a cleanup panic as its outcome. That reaches all three public futures — `recv` and `call` lose `T`, and `send_timeout` loses the recovered `M` carried inside `SendError` on the `Terminated` arm. So this is not solely an abort-class item; it is also a value-loss and outcome-fidelity item, in exactly the class this PR exists to close. Before this PR the drop site did not exist at all: the timer was retired in drop glue.

**`PanicAccumulator` alone does not fix this.** Its `Drop` *resumes* on a non-unwinding path, and it runs while the returned value is still in the return slot — so wrapping the clear in an accumulator reproduces the same abort. The fix has to be a path that cannot unwind here at all, or one that discards deliberately.

*Preferred:* route the timer to the disposal lane, which is what `SendFuture::drop` already does for a caller-waker clone ("A `RawWaker` vtable is caller code and there is no caller left to surface a panic to"). `BoxedSleep` is `Pin<Box<dyn Future<Output = ()> + Send + 'static>>`, so `crate::capability::dispose(&this.runtime, timer)` type-checks as-is. It also moves Tokio's time-driver-lock deregistration off the caller's task — a free step in #367's direction, without claiming #367.

*Minimal alternative:* the shape #366 itself prescribes for `CatchUnwindFuture`'s Ok-arm — `discard_panic(catch_panic(|| drop(timer)).err())` before the return — with the precedence rationale in a comment, since discarding on a normal return path deviates from `resume_preferred_panic`'s own doc ("containment is only correct where losing the diagnostic is the lesser outcome, which is true in a destructor and false on a normal return path"). Here the completed output outranks a waker-destructor diagnostic; that judgement should be written down rather than implied.

**Why this slipped past the new test.** `a_completed_operation_retires_its_timer_immediately` uses `ReadyImmediately`, which is `Ready` on the **first** poll. The timer is constructed but never polled, so no waker is ever registered inside it, and this line is provably harmless in that shape. The one arrangement where the line is dangerous — ready *after* a pending poll — is the arrangement the test does not build. Whichever remedy you pick, it needs a ready-after-pending regression with a hostile waker destructor (and a hostile output destructor for the abort half).

crates/shelterwood-mailbox/src/deadline.rs
```diff
@@ -135,6 +136,14 @@ impl<F: DeadlineOperation + Unpin> Future for Deadlined<F> {
     }
 }
 
+impl<F> Drop for Deadlined<F> {
```
ralexstokes, line 156: **Nit.** No comment. Every neighbouring containment drop in this codebase explains why it exists — `SendFuture::drop`, `CallOperation::drop`, `MailboxEffectQueue::drop`. Four bare lines here leave the next reader to rediscover that `BoxedSleep`'s drop runs a caller-supplied `RawWaker` vtable and that Tokio's cancel path destroys the registered clone inline.

crates/shelterwood-mailbox/src/deadline.rs
```diff
@@ -194,6 +220,25 @@ mod tests {
...
+        let mut context = Context::from_waker(Waker::noop());
+
+        assert!(matches!(
+            future.as_mut().poll(&mut context),
+            Poll::Ready(())
+        ));
+        assert!(
+            future.timer.is_none(),
+            "a completed but retained deadline future leaves no armed timer"
+        );
```
ralexstokes: **Nit (test quality).** This pins the assignment by restating it. `future.timer.is_none()` cannot distinguish "the field was cleared" from "the wheel entry was disarmed", and it does not exercise #366's stated motivation for the change — "can spuriously wake the caller if the deadline elapses first". The PR body's "proves a completed-but-retained deadline future leaves no armed timer" overstates what is shown: with `ReadyImmediately` the timer is never polled, so there was never an armed wheel entry holding a waker in the first place.

The artifact-level probe would be: paused clock, an operation that is `Pending` on poll 1 and `Ready` on poll 2, a counting waker registered by that first poll, then `advance` past the deadline and assert no wake arrives. That shape tests the real claim — and it is the same shape that would have caught the `deadline.rs:114` defect.

crates/shelterwood-mailbox/src/capability.rs
```diff
@@ -158,41 +158,58 @@ pub(crate) fn dispose<T: Send + 'static>(runtime: &Arc<dyn MailboxRuntime>, valu
...
 impl<T> Drop for DisposingReceiver<T> {
     fn drop(&mut self) {
-        if let Some(value) = self.inner.close_and_take_erased() {
+        let mut inner = self
+            .inner
+            .take()
+            .expect("a live disposing receiver retains its channel");
+        let mut value = None;
+        let mut panics = crate::panic::PanicAccumulator::default();
+        panics.run(|| value = inner.close_and_take_erased());
```
ralexstokes, line 216: **Nit.** If `close_and_take_erased()` panics — a hostile waker destructor is the reachable trigger — `value` stays `None` and a present user value falls to `inner`'s drop glue on the next line. That is still single-drop, contained, and leak-free (I checked), but it silently downgrades the disposal venue from the isolated lane to inline-on-this-thread, which is the thing `DisposingReceiver` exists to prevent. Worth a comment recording it as accepted, if not a fix.

crates/shelterwood-mailbox/src/capability.rs
```diff
@@ -158,41 +158,58 @@ pub(crate) fn dispose<T: Send + 'static>(runtime: &Arc<dyn MailboxRuntime>, valu
...
+        if let Some(value) = value {
             self.runtime.dispose(value);
         }
```
ralexstokes: **Nit.** This is the one step of the restructured drop left outside the accumulator. `dispose_detached` has a documented submitting-thread fallback under exhausted task/native-thread creation, so a panicking user-value destructor taken on that fallback escapes this drop glue bare and aborts during an unwind. Wrapping the whole `if let` in `panics.run(...)` is uniform and free. Pre-existing in kind, but this PR is the containment pass for this exact function, so it may as well close it.

ralexstokes (commented):
## Review: changes-needed — one confirmed defect **introduced by this PR**; everything else is correct and well-pinned

The containment work here is right, and the test design is better than the issue asked for. Both production containment changes are pinned by mutations that fail exactly the right test, the subprocess regressions observe **exit status** rather than a side channel, `#367` is correctly *not* over-claimed anywhere, and the new drop ordering is a genuine improvement over the field order it replaces. The blocking item below is scoped to one line — `deadline.rs:114`, the opportunistic timer-clear — not to the containment drops.

### Blocking

**`crates/shelterwood-mailbox/src/deadline.rs:114`** — the new `this.timer = None;` destroys an armed timer bare while the completed `result` is a live local. Codex's P1 is **confirmed and reproduced as `SIGABRT`**; details and both remedies inline, and adjudicated in-thread on the original comment.

### Nits (non-blocking)

- `deadline.rs:236-239` — the timer-clear regression restates the assignment it pins and does not test #366's stated spurious-wake motivation.
- `capability.rs:209-211` — `self.runtime.dispose(value)` is the one step of the restructured drop left outside the accumulator.
- `capability.rs:208` — if `close_and_take_erased()` panics, a present value is destroyed inline instead of routed to disposal.
- `deadline.rs:139` — `Drop for Deadlined` carries no explanatory comment, unlike every neighbouring containment drop.

### What held up under verification

Worktree at `adc381a`. `cargo +nightly fmt --all --check` and `cargo clippy -p shelterwood-mailbox -p shelterwood --all-targets` clean. All three new tests pass under both `cargo test` and `cargo nextest`. Existing suites unaffected: `shelterwood-mailbox --lib` (51), `--test deadlines`, `--test reply_waker_containment`, `--test one_shot_resource_ownership`.

Three mutations on the committed tree, each failing exactly the right test:

| Mutation | Result |
| --- | --- |
| A: delete `Drop for Deadlined` | `send_timeout_drop_contains_timer_waker_during_unwind` → SIGABRT/FAIL; `call` + `recv` still pass |
| B: revert `DisposingReceiver::drop` to bare | `call` + `recv` → SIGABRT/FAIL; `send_timeout` still passes |
| C: revert `this.timer = None` | `a_completed_operation_retires_its_timer_immediately` fails |

A and B are **complementary rather than redundant**, and the reason is worth recording: `call` and `recv` are driven with `Duration::MAX`, which `Deadline::after` maps to `Deadline(None)`, so `sleep_until(None)` yields `future::pending()` and **no timer waker is ever registered** in those two tests. The timer path is therefore pinned only by `send_timeout`, and the receiver path only by `call`/`recv`. Adequate for the production changes as written; it does mean no test covers `call`/`recv` with a *finite* deadline, where both waker clones exist at once.

Two concerns I raised and then refuted, so they are not re-litigated:

- **Re-poll after `Ready` does not break.** After a `Ready` from the `InitialAttempt` arm, `timer` is `None` while `phase` is still `InitialAttempt`, so a re-poll whose operation returned `Pending` would hit `expect("an unelapsed deadline future retains its timer")`. No operation can produce that: `SendFuture` re-polls to `Ready` (`Sent`) or panics (`Done`), and `ReplyOperation`/`CallOperation` panic inside the consumed one-shot first. Not reachable.
- **Disposal-before-waker-drop ordering is preserved**, and in fact strengthened. Tokio's `Receiver::close()` does not release the rx waker (it drops in `drop_rx`), so `dispose(value)` strictly precedes the waker drop in the new shape. I also traced the explicit take path for double-drop and use-after-take: `inner` is taken exactly once, `inner_mut()`'s `expect` is unreachable while alive, and a panic mid-`close_and_take_erased` leaves the value to `inner`'s drop glue — single drop, no leak.

Containment ordering is now deterministic in the fixed direction: `Deadlined::drop` runs before the `operation` field's drop glue, so the timer's waker clone is retired before `SendFuture::drop` hands its own clone to isolated disposal. The pre-fix field order raced those two across threads.

Not checked: full `just ci` / `just ci-nix` (shared machine — scoped targets only), and whether any SPEC/doc prose wants a sentence about the new completion-time timer retirement.

crates/shelterwood-mailbox/src/deadline.rs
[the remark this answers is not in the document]
ralexstokes: **Confirmed.** Good catch — the mechanism is exactly as described, and I reproduced it as a process abort rather than reasoning about it.

Probe: a scratch test in this file's `mod tests` with a `PendingThenReady` operation (`Pending` on poll 1, `Ready` on poll 2) and a waker whose `drop` panics, over the real timer from `capability::tests::runtime()`.

```
SCRATCH second_poll_unwound=true output_destroyed=true
```

The waker-drop panic escapes `Deadlined::poll` and the completed output is destroyed during the resulting unwind. Giving that output a panicking destructor escalates it:

```
thread ... panicked at library/core/src/panicking.rs:233:5:
panic in a destructor during cleanup
thread caused non-unwinding panic. aborting.
...
process didn't exit successfully: ... (signal: 6, SIGABRT: process abort signal)
```

Two refinements to the framing.

**Severity is broader than the abort.** The abort needs *two* hostile destructors — the waker's and the user value's. The milder outcome needs only the hostile waker, and it is a defect on its own: a **successfully completed** operation silently loses its user value and surfaces a cleanup panic as its outcome. That reaches all three public futures — `recv` and `call` lose `T`, and `send_timeout` loses the recovered `M` carried inside `SendError` on the `Terminated` arm. So this is a value-loss and outcome-fidelity item as well as an abort-class one.

**`PanicAccumulator` is not the fix here**, which is worth stating explicitly because it is the idiom this PR reaches for everywhere else. Its `Drop` *resumes* on a non-unwinding path, and it runs while the returned value is still in the return slot — so wrapping the clear in an accumulator reproduces the same abort. "A path that cannot unwind while the output is locally owned" has to mean either moving the destruction off this thread or discarding deliberately:

- *Preferred:* route the timer to the disposal lane, as `SendFuture::drop` already does for a caller-waker clone. `BoxedSleep` is `Pin<Box<dyn Future<Output = ()> + Send + 'static>>`, so `crate::capability::dispose(&this.runtime, timer)` type-checks as-is, and it incidentally moves Tokio's time-driver-lock deregistration off the caller's task — a free step toward #367 without claiming it.
- *Minimal:* `discard_panic(catch_panic(|| drop(timer)).err())` before the return — the shape #366 prescribes for `CatchUnwindFuture`'s Ok-arm — with the precedence rationale in a comment, since discarding on a normal return path deviates from `resume_preferred_panic`'s own doc.

One note on the AGENTS.md citation: the reasoning stands on the codebase's lock-rule and containment discipline in `CLAUDE.md` and on the reproduction above, not on that reference.

ralexstokes:
Fixes pushed in `399ee05`. Two notes for the record, one of which corrects my own review.

### The remedy is *not* the disposal lane I recommended

My inline comment preferred routing the retired timer to isolated disposal. I implemented that first and it **fails the artifact-level probe**, so I abandoned it:

```
assertion `left == right` failed: a completed but retained deadline future
must not wake its caller when the deadline elapses
  left: 1
 right: 0
```

Disposal is asynchronous. The field is cleared immediately but the wheel entry stays armed until a worker thread reaches the `BoxedSleep`, so a deadline that elapses in that window still delivers its expiry to a caller who already has its answer — precisely the spurious wake #366 asked for this clear. The two requirements at this site are in tension: the destructor must not unwind while `result` is live, *and* the disarm must be synchronous. Only a synchronous drop inside a non-resuming boundary satisfies both, so the shipped fix is `discard_panic(catch_panic(|| drop(timer)).err())` — the shape #366 prescribes for `CatchUnwindFuture`'s completed-output arm — with the precedence written into the comment: a delivered result outranks a waker-cleanup diagnostic. My "free step toward #367" remark does not survive that, and I have not claimed any #367 progress anywhere in the change.

Verification that the reproduced abort is closed, same probe as before, now with the hostile output destructor still in place:

```
SCRATCH second_poll_unwound=false output_destroyed=false
```

Mutation-checked, each failing exactly one test: bare timer clear ⇒ the new ready-after-parking regression aborts (SIGABRT); no timer clear ⇒ the paused-clock probe fails; no `Drop for Deadlined` ⇒ `send_timeout` aborts; bare `DisposingReceiver::drop` ⇒ `call` + `recv` abort. The subprocess helper now also asserts the child reports `1 passed`; I verified that guard by mistyping the filter name, which previously passed vacuously. Full workspace suite: 808/808.

### Out of scope, but confirmed by execution: a second abort site on this path, pre-existing

While targeting the timer's waker clone I found that the **one-shot's** registered clone is destroyed bare as the reply value is delivered — inside `poll_receive`, below `ReplyOperation::poll_deadlined`, with the value live. Making that clone (rather than the timer's) hostile aborts the process **on this branch with the fix applied**:

```
injected timer waker drop panic
injected reply destructor panic
panic in a destructor during cleanup
thread caused non-unwinding panic. aborting.
```

This is not a regression from this PR — nothing here touches that path, and it reproduces the same way on `main`. It is the same class as the Batch 1 item, at a site #366 does not name: the issue's "the value-present case is safe" sentence is about `DisposingReceiver::drop`, not about the delivery path. Flagging rather than folding it in, per the usual rule that structural fixes get filed instead of merged into an open PR. Happy to file it if you want it tracked.

ralexstokes:

- **P1 — armed timer destroyed bare while the output is live** (Codex `r3825717230` and the review thread): fixed in `399ee05`. Confirmed by execution before and after — the reproduced `signal: 6, SIGABRT` is gone, and the probe now reports `second_poll_unwound=false output_destroyed=false` with the hostile output destructor still in place.

  **Remedy correction, worth recording:** the disposal lane — recommended in the original review and the obvious sibling of `SendFuture::drop` — was implemented first and **fails**. Disposal is asynchronous, so the field clears while the wheel entry stays armed until a worker reaches the `BoxedSleep`; a deadline elapsing in that window still wakes a caller who already has its answer, which is exactly the spurious wake this branch's timer clear exists to prevent. Containment and synchronous disarm are in tension here, and only a synchronous drop inside a *non-resuming* boundary satisfies both. Shipped `discard_panic(catch_panic(|| drop(timer)).err())` — the shape #366 prescribes for `CatchUnwindFuture`'s completed-output arm — with the precedence rationale in the comment. `PanicAccumulator` is **not** a substitute: its `Drop` resumes on a normal return path and runs while the value is still in the return slot.

- **Test gap that let it through**: the existing regression used `ReadyImmediately`, ready on the *first* poll, so the timer was never polled and no waker was ever registered. `399ee05` adds `recv_ready_after_parking_contains_the_retired_timer_waker`, which targets the timer's clone **by ordinal** (`Deadlined::poll` mints the operation's clone first, the timer's second).
- **Tautological timer-clear test**: replaced with a paused-clock probe asserting a caller parked on the first poll is never woken after the deadline elapses — #366's actual stated motivation, rather than restating the assignment.
- **`dispose` outside the accumulator**: fixed in `399ee05`.
- **Missing comment on `Drop for Deadlined`**: added in `399ee05`.
- **`close_and_take_erased()` panicking**: comment, not fix — reaching it requires a destructor that has already panicked, and the only available "fix" is retrying the step that just failed. Inline destruction is recorded as accepted.

`972bd7b` then removes the subprocess harness from all four tests (nextest already isolates per test); the full mutation battery still discriminates, now reported as SIGABRT against the named test. Workspace 808/808.

**Separately filed: #383** — the one-shot's registered waker clone is retired bare on the reply *delivery* path, with the value live. Confirmed by execution, reproduces on `main`, and **not** fixed here: it is a different clone at a different seam, and it is not a destructor, so no Batch 1 containment drop reaches it.

PR #373, opened 2026-08-20
Detach opaque panic payload disposal

ralexstokes:
Part of #366.

Moves opaque user panic-payload destruction onto the detached disposal lane after extracting any string diagnostic. This prevents a blocking or panicking payload destructor from running on the executor that publishes the child exit.

Detaching alone is not sufficient, because the disposal lane classifies a destructor panic by calling back into `contain_panic_payload`: a bare submission would let a payload whose `Drop` `panic_any`s a fresh copy of itself requeue forever. `DiscardedPanicPayload`'s destructor runs the existing terminal discard (`discard_panic`, which `mem::forget`s the replacement by construction), so the venue guarantee holds while the disposal job never observes a panic to classify. The doc comment records why the wrapper cannot be inlined away.

SPEC §5.5 is scoped to what the code delivers: the guarantee covers an opaque payload *carried by a child exit*, states that the destruction is terminal, cross-references §6.5's thread-exhaustion exception, and records that a *displaced* payload (a rejected panic-slot record, a losing cleanup panic) is still discarded inline.

Two regressions in `crates/shelterwood/tests/disposal.rs`, both plain `#[test]`/`#[tokio::test]` that CI runs unconditionally — "gate" here refers to the `DestructorGate` fixture, not a `cfg` or feature gate:

- `blocking_panic_payload_does_not_stall_current_thread_exit_publication` — a current-thread runtime proving exit publication completes while the payload destructor remains blocked, and that the destructor runs off the driver thread.
- `a_self_regenerating_panic_payload_is_destroyed_a_bounded_number_of_times` — a self-regenerating payload driven through the public actor API, pinning terminality on the path a user reaches (the isolated pin lives beside `shelterwood_core::panic`'s `discarding_a_recursively_hostile_panic_payload_is_contained`).

Measured destruction counts for a self-regenerating payload, before and after the terminal discard:

| path | without the terminal step | with it |
|---|---|---|
| direct `dispose_detached` | 109,618 @300ms → 219,592 @600ms | 2, then stops |
| public actor API (`panic_any` from a `TaskOnceDef`) | 63,356 @300ms → 125,151 @600ms | 1, then stops |

Validation: `./tools/dev just ci`; both regressions mutation-checked (removing the newtype fails the terminality test in 5 ms; reverting the venue change fails the venue test in 5.0 s, bounded rather than hanging).

Stack: independent Batch 1 slice; targets `main`.

### DIFF
--- a/crates/shelterwood-runtime/src/spawn.rs
+++ b/crates/shelterwood-runtime/src/spawn.rs
@@ -446,8 +446,38 @@ pub(super) fn contain_panic_payload(payload: PanicPayload) -> Option<String> {
     // A custom panic payload is user-owned too. Its destructor may panic or
     // block, so retire it on the detached disposal lane before publishing
     // completion.
-    dispose_detached(payload);
+    dispose_detached(DiscardedPanicPayload(Some(payload)));
     message
+}
+
+/// Terminal, off-executor destruction for an opaque user panic payload.
+///
+/// The wrapper is load-bearing, not ceremony: submitting a bare payload to
+/// [`dispose_detached`] closes a cycle. `DisposalJob::finish` classifies a
+/// destructor panic by calling [`contain_panic_payload`], which submits the
+/// replacement payload for detached disposal, whose destructor panics again.
+/// A self-regenerating payload — one whose `Drop` `panic_any`s a fresh copy
+/// of itself — then spins the disposal lane for the life of the process,
+/// firing the panic hook every iteration.
+///
+/// [`discard_panic`] is what breaks the cycle, and does so by construction:
+/// it catches the replacement and `mem::forget`s it precisely because
+/// dropping the replacement would recurse outside the boundary
+/// (`shelterwood_core::panic`, pinned by that module's
+/// `discarding_a_recursively_hostile_panic_payload_is_contained`). Running it
+/// from this destructor keeps the venue guarantee — the payload is still
+/// destroyed off the exit-publishing executor — while the disposal job
+/// itself never observes a panic to classify.
+///
+/// Removing this wrapper reintroduces the unbounded spin; the regression is
+/// pinned by `blocking_panic_payload_does_not_stall_current_thread_exit_publication`'s
+/// sibling `a_self_regenerating_panic_payload_is_destroyed_a_bounded_number_of_times`.
+struct DiscardedPanicPayload(Option<PanicPayload>);
+
+impl Drop for DiscardedPanicPayload {
+    fn drop(&mut self) {
+        discard_panic(self.0.take());
+    }
 }
 
 fn panic_message(payload: &(dyn Any + Send + 'static)) -> Option<String> {
--- a/crates/shelterwood-runtime/src/spawn.rs
+++ b/crates/shelterwood-runtime/src/spawn.rs
@@ -443,9 +443,10 @@ pub(super) fn contain_panic_payload(payload: PanicPayload) -> Option<String> {
             None
         }
     };
-    // A custom panic payload is user-owned too. Its destructor may panic, so
-    // discard it under a fresh unwind boundary before publishing completion.
-    discard_panic(Some(payload));
+    // A custom panic payload is user-owned too. Its destructor may panic or
+    // block, so retire it on the detached disposal lane before publishing
+    // completion.
+    dispose_detached(payload);
     message
 }
 
--- a/crates/shelterwood/tests/disposal.rs
+++ b/crates/shelterwood/tests/disposal.rs
@@ -16,7 +16,7 @@ use std::{
 
 use crate::common::{
     DestructorBlocker, DestructorGate, POLL_TIMEOUT, PanicOnDrop, ReleaseGate, assert_eventually,
-    next_event, policy::never, poll_once,
+    assert_quiet, next_event, policy::never, poll_once,
 };
 use shelterwood::{
     Backoff, CallErrorKind, Cancellation, ChildId, ChildState, DynamicTree, ExitError, ExitKind,
@@ -158,6 +158,20 @@ struct PanickingPanicPayload;
 impl Drop for PanickingPanicPayload {
     fn drop(&mut self) {
         panic!("panic payload destructor");
+    }
+}
+
+/// A panic payload whose own destructor panics with a fresh copy of itself.
+///
+/// The counter records how many copies the framework destroys, which is the
+/// property under test: the terminal discard must stop the chain rather than
+/// let each replacement queue the next one.
+struct SelfRegeneratingPanicPayload(Arc<AtomicUsize>);
+
+impl Drop for SelfRegeneratingPanicPayload {
+    fn drop(&mut self) {
+        self.0.fetch_add(1, Ordering::SeqCst);
+        std::panic::panic_any(SelfRegeneratingPanicPayload(Arc::clone(&self.0)));
     }
 }
 
@@ -507,7 +521,7 @@ fn blocking_panic_payload_does_not_stall_current_thread_exit_publication() {
         let destructor_thread = started_rx
             .recv_timeout(POLL_TIMEOUT)
             .expect("panic payload destruction starts");
-        let released_after_publication = release_rx.recv_timeout(Duration::from_secs(1)).is_ok();
+        let released_after_publication = release_rx.recv_timeout(POLL_TIMEOUT).is_ok();
         release_gate.release();
         (destructor_thread, released_after_publication)
     });
@@ -546,6 +560,47 @@ fn blocking_panic_payload_does_not_stall_current_thread_exit_publication() {
     );
     assert_ne!(destructor_thread, driver_thread);
     assert!(matches!(exit.kind(), ExitKind::Panicked { message: None }));
+}
+
+/// Detaching the payload must not cost the chain's terminal step.
+///
+/// The disposal lane classifies a destructor panic by calling back into
+/// `contain_panic_payload`, so a payload submitted bare would queue its own
+/// replacement forever. `shelterwood_core::panic`'s
+/// `discarding_a_recursively_hostile_panic_payload_is_contained` pins the
+/// terminal discard in isolation; this pins it on the path a user reaches,
+/// where the payload leaves an actor through `panic_any`.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn a_self_regenerating_panic_payload_is_destroyed_a_bounded_number_of_times() {
+    let destructions = Arc::new(AtomicUsize::new(0));
+    let payload = SelfRegeneratingPanicPayload(Arc::clone(&destructions));
+    let mut tree = Tree::new();
+    let (task, _completion) = tree
+        .add_task_once(
+            "regenerating",
+            TaskOnceDef::new(move |_| async move {
+                std::panic::panic_any(payload);
+                #[allow(unreachable_code)]
+                Ok::<(), ExitError>(())
+            }),
+        )
+        .expect("valid task");
+    let system = tree.spawn().expect("runtime is available");
+    assert!(matches!(
+        task.wait().await.kind(),
+        ExitKind::Panicked { message: None }
+    ));
+    assert_eq!(system.wait().await, shelterwood::StopReason::Finished);
+
+    // Exactly one copy is destroyed: the replacement its destructor raises is
+    // the already-panicking diagnostic the terminal discard leaks on purpose.
+    // Without that step the count climbs by six figures per second, so the
+    // quiet window fails on its first sample instead of spinning.
+    assert_eventually!(|| destructions.load(Ordering::SeqCst) >= 1).await;
+    assert_quiet(Duration::from_millis(200), || {
+        destructions.load(Ordering::SeqCst) > 1
+    })
+    .await;
 }
 
 #[test]
--- a/crates/shelterwood/tests/disposal.rs
+++ b/crates/shelterwood/tests/disposal.rs
@@ -7,6 +7,7 @@ use std::{
     sync::{
         Arc, Mutex,
         atomic::{AtomicUsize, Ordering},
+        mpsc,
     },
     task::{Context, Poll, Wake, Waker},
     thread::{self, ThreadId},
@@ -60,6 +61,26 @@ impl Drop for DropProbe {
 struct BlockingDropProbe {
     dropped: tokio::sync::mpsc::UnboundedSender<ThreadId>,
     _blocker: DestructorBlocker,
+}
+
+struct BlockingPanicPayload {
+    started: mpsc::Sender<ThreadId>,
+    _blocker: DestructorBlocker,
+}
+
+impl BlockingPanicPayload {
+    fn new(gate: &DestructorGate, started: mpsc::Sender<ThreadId>) -> Self {
+        Self {
+            started,
+            _blocker: gate.blocker(),
+        }
+    }
+}
+
+impl Drop for BlockingPanicPayload {
+    fn drop(&mut self) {
+        let _ = self.started.send(thread::current().id());
+    }
 }
 
 impl BlockingDropProbe {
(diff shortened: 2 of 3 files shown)


crates/shelterwood-runtime/src/spawn.rs
```diff
@@ -443,9 +443,10 @@ pub(super) fn contain_panic_payload(payload: PanicPayload) -> Option<String> {
             None
         }
     };
-    // A custom panic payload is user-owned too. Its destructor may panic, so
-    // discard it under a fresh unwind boundary before publishing completion.
-    discard_panic(Some(payload));
+    // A custom panic payload is user-owned too. Its destructor may panic or
+    // block, so retire it on the detached disposal lane before publishing
+    // completion.
+    dispose_detached(payload);
```
ralexstokes: **Blocking — this reopens a recursion the codebase closed by construction.**

`dispose_detached` is not a terminal sink for a *panic payload*, because the disposal machinery classifies destructor panics by calling right back into this function:

```
contain_panic_payload (spawn.rs:449)
  -> dispose_detached(payload)
  -> dispatch_disposal
  -> DisposalJob::finish            (disposal.rs:105-110)
  -> catch_panic(|| drop(value))
  -> the destructor panic_any()s a replacement payload
  -> contain_panic_payload(payload) (disposal.rs:110)
  -> dispose_detached(payload)      -- cycle closed
```

Pre-PR this terminated **by construction**, not by luck. `discard_panic` (`crates/shelterwood-core/src/panic.rs:28-38`) catches the replacement and `std::mem::forget`s it, with a comment saying exactly why:

> A payload whose own destructor panics cannot be dropped safely: dropping the replacement payload would merely recurse outside the boundary. Leak only this already-panicking diagnostic.

and `crates/shelterwood-core/src/panic.rs:157` (`discarding_a_recursively_hostile_panic_payload_is_contained`) already pins it. This line removes the only terminal step from the path.

### Measured (temp probes, worktree restored clean)

Payload: `struct Forever(Arc<AtomicUsize>)` whose `Drop` bumps the counter and `panic_any(Forever(...))`s a fresh one.

| path | pre-PR (`discard_panic`) | this PR (`dispose_detached`) |
|---|---|---|
| direct `dispose_detached` of the payload | 2 destructions, then stops | 109,618 @300ms → 219,592 @600ms |
| **public actor API**: `panic_any(Forever)` from a `TaskOnceDef` | **1** destruction, then stops | **63,356 @300ms → 125,151 @600ms** |

The second row is the one that matters: it is reached through the supported public surface, not a synthetic probe. The exit still publishes correctly (`Panicked { message: None }`) — the damage is behind it.

### Characterization

Not a memory leak: each job submits exactly one successor and is freed when processed, so queue depth stays ~1. It is a **CPU-and-log livelock** — ~350k iterations/sec for the life of the process on the `shelterwood-disposal` thread, **firing the panic hook every iteration**, i.e. an unbounded stderr flood on a library-named thread. It also survives runtime shutdown: once Tokio starts rejecting blocking submissions, `submit_blocking_job` returns `false` and the chain lands permanently on the shared fallback thread.

### Blast radius is wider than the Codex comment frames it

The cycle is entered from `DisposalJob::finish`, which classifies the destructor panic of **any** detached-disposal value. So the trigger is not only `panic_any` users: any mailbox payload, actor state, or factory capture whose destructor panics with a self-regenerating payload enters the same loop.

### Remedy (verified)

Keep the venue fix; restore terminality with a newtype whose `Drop` is the existing terminal discard, so the disposal job never sees a panic at all:

```rust
/// Terminal off-executor destruction for an opaque user panic payload.
struct DiscardedPanicPayload(Option<PanicPayload>);

impl Drop for DiscardedPanicPayload {
    fn drop(&mut self) {
        discard_panic(self.0.take());
    }
}

// contain_panic_payload:
dispose_detached(DiscardedPanicPayload(Some(payload)));
```

I applied exactly this and measured: destructions bounded at **1**, and all **39** tests in `crates/shelterwood/tests/disposal.rs` pass — including this PR's own `blocking_panic_payload_does_not_stall_current_thread_exit_publication`, so the venue guarantee is preserved. `Box<dyn Any + Send>` keeps the wrapper `Send + 'static`.

Worth a regression beside `crates/shelterwood-core/src/panic.rs:157`'s: assert a bounded destruction count after a self-regenerating payload leaves an actor, so the terminality property is pinned on the disposal path too and not only on `discard_panic` in isolation.

ralexstokes (commented):
## Verdict: changes-needed — one confirmed, execution-verified regression

The **fix direction is right and the diff is 90% there**: the venue move is the correct remedy for #366's "Destroy opaque panic payloads off the executor", the §5.5 sentence lands in the correct section, and the new regression is a genuine venue pin that survives a mutation check. What the diff loses is **terminality**, not correctness of direction — `discard_panic` was doing two jobs (get the payload off the stack *and* stop a recursively hostile payload from regenerating), and only the first was replaced. This is a small, local fix on top of what is already here, not a restart.

**Blocking:** `crates/shelterwood-runtime/src/spawn.rs:449` re-enters `contain_panic_payload` through `DisposalJob::finish`, turning a self-regenerating panic payload into an unbounded CPU-and-log livelock. Confirmed by execution through the public actor API (63,356 destructions @300ms → 125,151 @600ms, versus 1 pre-PR). Details and a verified one-type remedy inline.

**Codex P2 (`dispose_detached`'s exhaustion fallback):** mechanism real, but severity overstated and remedy declined — SPEC already carries that caveat globally at `specs/SPEC.md:1042-1044` and it predates this PR. Adjudication posted in-thread; **no change requested**.

### What held up (recording so it is not re-litigated)

- **Diagnostic extraction unchanged and complete.** `panic_message` (`spawn.rs:452-459`) still handles both `&str` and `String`; extraction runs before the move, so ordering is preserved. The `Err(inspection_panic)` arm is effectively unreachable — `downcast_ref` runs no user code.
- **The `PanicPayload` bound is legitimate.** `PanicPayload = Box<dyn Any + Send + 'static>` (`crates/shelterwood-core/src/panic.rs:13`) satisfies `dispose_detached`'s `T: Send + 'static` with no lifetime or `Send` hazard.
- **Venue claim matches the code.** Traced the publisher: `crates/shelterwood/src/driver/child.rs:582-598` — the child task `resume_unwind`s, `runtime::join(handle)` on the exit-publisher task calls `contain_panic_payload` (`spawn.rs:417`). That is exactly the executor §5.5's new sentence names.
- **§5.5 is correct post-#364.** `### 5.5 Destruction venue` at `specs/SPEC.md:744`; all five cross-references (`843`, `1313`, `2061`, `2806`, `3182`) still resolve.
- **No packaged-docs sync needed.** `tools/sync-packaged-docs.sh` mirrors only `README.md` and `docs/{embedding,observation,retry-and-ordering,shutdown-and-resources}.md` into `crates/shelterwood/doctests/`. `specs/` is not mirrored and this PR touches none of the mirrored sources, so `packaged-docs-check` is unaffected.
- **Mutation check passed on a committed tree.** Reverting only `spawn.rs:449` to `discard_panic(Some(payload))` fails `blocking_panic_payload_does_not_stall_current_thread_exit_publication` in **1.005s** — it does *not* hang. The failure mode is bounded by the helper's `recv_timeout`, after which the dropped receiver makes `release.send(...).expect(...)` panic. The test pins the **venue**, not a proxy: `assert_ne!(destructor_thread, driver_thread)` on a `new_current_thread` runtime where the driver thread *is* the test thread, with every assertion in the test body rather than inside a caught-panic region.
- **Suite green at HEAD:** all 39 `crates/shelterwood --test disposal` tests, and 151 `shelterwood-runtime` + `shelterwood-core` unit tests.

Verified by execution unless noted; the P2 exhaustion path was traced by reading rather than by forcing `thread::Builder::spawn` to fail, and I did not re-run the full `just ci`.

specs/SPEC.md
```diff
@@ -750,10 +750,12 @@ surfaces on that task even though the replacement remains accepted.
 Framework-initiated disposal of externally submitted mailbox or
 reply-bearing payloads — including mailbox teardown, timeout/withdrawal
 cleanup, and accepted-prefix batch disposal — runs detached from the
-initiating task with per-element panic containment. Incarnation-owned
-continuations, timer messages, and offload state instead follow §6.5 and
-§8's incarnation teardown and verdict rules. No single disposal-thread
-identity is promised.
+initiating task with per-element panic containment. After extracting any
+string diagnostic, the framework likewise destroys an opaque user panic
+payload on the detached disposal lane rather than on the executor publishing
```
ralexstokes: **Nit (SPEC/code scope).** As written this sentence is unqualified about "an opaque user panic payload", but the framework still destroys *secondary* panic payloads inline on the actor's own task — `PanicSlot::record`'s rejected payload (`crates/shelterwood/src/raw/disposal.rs:83`) and `keep_first_panic`'s losing candidate (`crates/shelterwood-core/src/panic.rs:45-50`) both go straight to `discard_panic`.

The trailing clause ("rather than on the executor publishing the exit") makes the intended scope inferable, but scoping it explicitly — e.g. "an opaque user panic payload **carried by a child exit**" — removes the ambiguity and keeps a future review from filing the displaced-payload sites as a §5.5 violation.

Otherwise the placement is right: `### 5.5 Destruction venue` is the correct section post-#364, and all five `§5.5` cross-references (`843`, `1313`, `2061`, `2806`, `3182`) still resolve.

crates/shelterwood/tests/disposal.rs
```diff
@@ -475,6 +496,58 @@ fn final_factory_capture_is_destroyed_outside_the_current_thread_driver() {
...
+    let driver_thread = thread::current().id();
+    let gate = DestructorGate::default();
+    let (started, started_rx) = mpsc::channel();
+    let (release, release_rx) = mpsc::channel();
+    let release_gate = gate.clone();
+    let helper = thread::spawn(move || {
+        let destructor_thread = started_rx
+            .recv_timeout(POLL_TIMEOUT)
+            .expect("panic payload destruction starts");
+        let released_after_publication = release_rx.recv_timeout(Duration::from_secs(1)).is_ok();
```
ralexstokes: **Nits on the regression — the test itself is good.** It survives a mutation check (revert `spawn.rs:449` and it fails in 1.005s rather than hanging) and it pins the **venue** rather than a proxy, via `assert_ne!(destructor_thread, driver_thread)` on a `new_current_thread` runtime where the driver thread *is* the test thread, with every assertion in the test body. Two wording/robustness points:

1. **"gated" in the PR body means the `DestructorGate` fixture, not a `cfg`/feature gate.** This is a plain `#[test]` in the default `--test disposal` target, so CI runs it unconditionally — which is the right outcome, but the PR description reads as though it were feature-gated and might invite someone to go looking for the gate.

2. **This line is the flake-prone knob.** The positive wait above uses `POLL_TIMEOUT` (5s) while the negative window — "publication must arrive within" — is a hardcoded `Duration::from_secs(1)`. Under load a slow publication turns into a spurious `released_after_publication == false`. It matches neighbouring tests in this file, so this is style-consistent; flagging it only as the first thing to suspect if this test ever goes intermittent.

crates/shelterwood-runtime/src/spawn.rs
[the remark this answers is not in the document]
ralexstokes: **Confirmed — and this is the blocking finding on the PR. Good catch.**

Mechanism verified exactly as described, severity correct, and the remedy direction ("leak the replacement") is right — the codebase already says so in a comment. `discard_panic` (`crates/shelterwood-core/src/panic.rs:28-38`) catches the replacement and `std::mem::forget`s it because *"dropping the replacement payload would merely recurse outside the boundary"*, and `panic.rs:157` (`discarding_a_recursively_hostile_panic_payload_is_contained`) pins it. Swapping in `dispose_detached` removes the only terminal step from this path.

Measured with a `struct Forever(Arc<AtomicUsize>)` whose `Drop` bumps a counter and `panic_any`s a fresh `Forever`:

| path | pre-PR | this PR |
|---|---|---|
| direct `dispose_detached` | 2 destructions, then stops | 109,618 @300ms → 219,592 @600ms |
| **public actor API** (`panic_any(Forever)` from a `TaskOnceDef`) | **1**, then stops | **63,356 @300ms → 125,151 @600ms** |

Two refinements to your framing:

1. **It is a CPU-and-log livelock, not unbounded memory.** Each job submits exactly one successor and is freed when processed, so queue depth stays ~1. What is unbounded is ~350k iterations/sec for the life of the process on the `shelterwood-disposal` thread, with the panic hook firing every iteration — an unbounded stderr flood on a library-named thread. It also survives runtime shutdown: once Tokio rejects blocking submissions, the chain lands permanently on the shared fallback thread.

2. **Blast radius is wider than "a hostile `panic_any` payload".** The cycle is entered from `DisposalJob::finish` (`crates/shelterwood-runtime/src/disposal.rs:110`), which classifies the destructor panic of *any* detached-disposal value — so any mailbox payload, actor state, or factory capture with a self-regenerating hostile destructor triggers it too, not only payloads that arrived via `panic_any`.

The one-type fix I verified (bounded to 1 destruction, all 39 `crates/shelterwood/tests/disposal.rs` tests green including the PR's own new regression) is spelled out in the review comment on `crates/shelterwood-runtime/src/spawn.rs:449`: a `DiscardedPanicPayload(Option<PanicPayload>)` newtype whose `Drop` calls `discard_panic`, submitted through `dispose_detached`. That keeps the venue guarantee this PR is buying and restores terminality.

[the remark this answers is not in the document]
ralexstokes: **Mechanism confirmed; severity overstated and remedy declined — recommending no change here.** Taking the observation seriously, disagreeing on framing.

The path is real: `dispatch_disposal_with` (`crates/shelterwood-runtime/src/disposal.rs:277-309`) ends in `job.finish()` on the submitting thread when both `submit_blocking_job` and native thread creation fail, and the repo's own test comment at `disposal.rs:718-721` names it ("on the regression path `dispose_detached` runs the destructor inline").

The decisive point is that **SPEC already carries this caveat globally, and it predates this PR** — `specs/SPEC.md:1042-1044`:

> If the system cannot create the disposal worker either, disposal's final no-loss fallback destroys the state synchronously; **thread exhaustion is the sole exception to the isolation guarantee.**

Every other claim in §5.5 is equally unconditional on its face — "runs detached from the initiating task with per-element panic containment" has exactly the same fallback — so the newly added sentence is no more absolute than the paragraph it joins, and it inherits the same documented exception. Rewriting only the new sentence would make §5.5 internally inconsistent, implying the surrounding guarantees are somehow stronger.

`dispose_critical` is also the wrong instrument here. Per `CLAUDE.md`, it exists for values retired **inside a framework critical section** — where the fail-safe is a job left unreclaimed for the life of the process rather than user destruction under a held lock. The exit publisher (`crates/shelterwood/src/driver/child.rs:582-598`) holds no lock; it is an executor stall we are avoiding, not a lock-rule violation. #366 prescribes `dispose_detached` for this site for that reason. Trading "inline destruction under total native-thread exhaustion" for "payload memory held for process lifetime" is the worse deal on a non-critical path — and under total thread exhaustion the process has larger problems than one inline destructor.

Optional nit if the author wants belt-and-braces: append a cross-reference such as "(§6.5's thread-exhaustion exception applies)" to the new sentence. That documents the edge without weakening §5.5's other guarantees.

ralexstokes:

- **P1 — recursively panicking payloads requeue forever** (Codex `r3825732356` and the blocking review thread): fixed in `6b0d473`. `dispose_detached` now receives a `DiscardedPanicPayload` newtype whose `Drop` runs `discard_panic`, restoring the terminal step while keeping the venue fix. Measured, through the public actor API (`panic_any` from a `TaskOnceDef`):

  | path | before | after |
  |---|---|---|
  | direct `dispose_detached` | 109,618 @300ms → 219,592 @600ms | **2**, then stops |
  | public actor API | 63,356 @300ms → 125,151 @600ms | **1**, then stops |

  The newtype's doc comment names the `DisposalJob::finish → contain_panic_payload → dispose_detached` cycle and `discard_panic`'s `mem::forget` as the step that breaks it, so a future refactor cannot delete it the way this regression arose. New regression `a_self_regenerating_panic_payload_is_destroyed_a_bounded_number_of_times` fails in 0.005s without the wrapper — bounded, not a hang.

- **P2 — inline fallback under thread exhaustion** (Codex `r3825732363`): **declined**, rationale in-thread. The mechanism is real, but SPEC already carries this caveat globally and predates this PR ("thread exhaustion is the sole exception to the isolation guarantee"), so the new sentence is no more unconditional than the text it joins. `dispose_critical` is the wrong instrument on a path that holds no lock — it would trade inline destruction under total thread exhaustion for payload memory held for the process lifetime.

- **SPEC scope nit**: fixed in `6b0d473` — §5.5 is now scoped to "an opaque user panic payload carried by a child exit", records that the destruction is terminal, cross-references the thread-exhaustion exception, and states that a *displaced* payload is still discarded inline.
- **Test nits**: the PR body no longer says "gated" ambiguously (it is the `DestructorGate` fixture, not a cfg gate — CI runs both regressions unconditionally), and the venue test's negative window now uses the shared `POLL_TIMEOUT` instead of a hardcoded 1s.

Mutation checks: deleting the newtype fails the new test in 0.005s; reverting the venue fix fails the original test in 5.004s, bounded by `POLL_TIMEOUT`, while the terminality test correctly passes — the two pin independent properties. 40 `--test disposal` tests and 151 runtime/core unit tests green.

PR #374, opened 2026-08-20
Finish issue 366 Batch 1 contract fixes

ralexstokes:
Part of #366.

Completes the remaining small Batch 1 items:

- clamps equal jitter to its exact integer half-delay lower bound and covers the smallest positive sample
- makes `SystemRun::drop` inert after its driver has already been joined
- documents the immediate `Context::stop` intake/resource freeze and pins every local-work facade result after a same-callback stop
- records the stop-before-first-settlement construction behavior in SPEC §11

Validation: `./tools/dev just ci` (805 tests).

Stack: independent Batch 1 slice; targets `main`.

### DIFF
--- a/crates/shelterwood-core/src/policy.rs
+++ b/crates/shelterwood-core/src/policy.rs
@@ -469,6 +469,7 @@ impl Backoff {
                     half_duration(delay)
                 } else {
                     duration_from_nanos(duration_nanos(delay) * (0.5 + sample * 0.5))
+                        .max(half_duration(delay))
                 }
             }
         };
@@ -1363,6 +1364,18 @@ mod tests {
         assert_eq!(
             even.next_delay(RestartAttempt(1), JitterSample::new(0.0)),
             Duration::from_nanos(1 << 59)
+        );
+    }
+
+    #[test]
+    fn smallest_positive_sample_equal_jitter_respects_the_exact_half_delay() {
+        let odd = Duration::from_nanos((1 << 60) + 1);
+        let fixed = Backoff::fixed(odd, Jitter::Equal).expect("valid backoff");
+        let smallest_positive = JitterSample::new(f64::from_bits(1));
+        assert_ne!(smallest_positive, JitterSample::new(0.0));
+        assert_eq!(
+            fixed.next_delay(RestartAttempt(1), smallest_positive),
+            Duration::from_nanos((1 << 59) + 1)
         );
     }
 
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -184,7 +184,12 @@ impl<'a, A: Actor> Context<'a, A> {
         }
     }
 
-    /// Requests a clean local stop after the current callback.
+    /// Requests a clean local stop.
+    ///
+    /// External intake and incarnation-owned continuations, timers, and
+    /// offloads freeze at this call. The current callback may finish, but any
+    /// later attempt from it to queue such work is rejected. Terminal
+    /// publication follows callback completion.
     ///
     /// During a successful initializer with effective [`Readiness::AfterInit`],
     /// the automatic readiness edge is published before this request becomes
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -134,6 +134,9 @@ fn classify_retained_root_driver_join(
 
 impl Drop for SystemRun {
     fn drop(&mut self) {
+        if self.driver.is_none() {
+            return;
+        }
         // After a clean shutdown the root epochs are `Idle`, not `Exhausted`,
         // so this writes a real `ScopeRequest` — targeting the pending next
         // incarnation — into dead control state and pulses the member record.
--- a/crates/shelterwood/tests/actor.rs
+++ b/crates/shelterwood/tests/actor.rs
@@ -133,6 +133,137 @@ async fn handler_actor_runs_init_handle_and_stop_through_the_raw_wrapper() {
     );
 }
 
+#[derive(Clone, Copy, Debug)]
+enum StopSurfaceMessage {
+    Stop,
+    Continuation,
+    Timer,
+    Offload,
+}
+
+const STOP_CONTINUATION_REJECTED: usize = 1 << 0;
+const STOP_TIMEOUT_REJECTED: usize = 1 << 1;
+const STOP_INTERVAL_REJECTED: usize = 1 << 2;
+const STOP_CLEAR_REPORTS_EMPTY: usize = 1 << 3;
+const STOP_OFFLOAD_REJECTED: usize = 1 << 4;
+const STOP_SCOPED_OFFLOAD_REJECTED: usize = 1 << 5;
+const ALL_STOP_RESULTS: usize = STOP_CONTINUATION_REJECTED
+    | STOP_TIMEOUT_REJECTED
+    | STOP_INTERVAL_REJECTED
+    | STOP_CLEAR_REPORTS_EMPTY
+    | STOP_OFFLOAD_REJECTED
+    | STOP_SCOPED_OFFLOAD_REJECTED;
+
+struct StopSurfaceActor(Arc<AtomicUsize>);
+
+impl Actor for StopSurfaceActor {
+    type Msg = StopSurfaceMessage;
+    type Args = Arc<AtomicUsize>;
+
+    async fn init(observed: Self::Args, _: &mut Context<'_, Self>) -> Result<Self, ExitError> {
+        Ok(Self(observed))
+    }
+
+    async fn handle(&mut self, message: Self::Msg, context: &mut Context<'_, Self>) -> ExitResult {
+        if !matches!(message, StopSurfaceMessage::Stop) {
+            return Ok(());
+        }
+
+        context.stop();
+        let mut observed = 0;
+        if context
+            .continue_with(StopSurfaceMessage::Continuation)
+            .is_err_and(|rejected| {
+                matches!(rejected.into_inner(), StopSurfaceMessage::Continuation)
+            })
+        {
+            observed |= STOP_CONTINUATION_REJECTED;
+        }
+        if context
+            .set_timeout("timeout", StopSurfaceMessage::Timer, Duration::ZERO)
+            .is_err_and(|rejected| {
+                matches!(
+                    rejected.into_inner(),
+                    ("timeout", StopSurfaceMessage::Timer)
+                )
+            })
+        {
+            observed |= STOP_TIMEOUT_REJECTED;
+        }
+        if context
+            .set_interval(
+                "interval",
+                StopSurfaceMessage::Timer,
+                Duration::from_secs(1),
+            )
+            .is_err_and(|rejected| {
+                matches!(
+                    rejected.into_inner(),
+                    ("interval", StopSurfaceMessage::Timer)
+                )
+            })
+        {
+            observed |= STOP_INTERVAL_REJECTED;
+        }
+        // Clearing does not queue work. The stop already emptied the timer
+        // store, so the live callback facade reports that there is nothing
+        // left to retract.
+        if matches!(context.clear_timer(&"timer"), Ok(false)) {
+            observed |= STOP_CLEAR_REPORTS_EMPTY;
+        }
+        if context
+            .offload(
+                async {},
+                |_| StopSurfaceMessage::Offload,
+                Duration::from_secs(1),
+            )
+            .is_err_and(|rejected| {
+                let (work, continuation) = rejected.into_inner();
+                drop(work);
+                matches!(continuation(Ok(())), StopSurfaceMessage::Offload)
+            })
+        {
+            observed |= STOP_OFFLOAD_REJECTED;
+        }
+        if context
+            .offload_scoped(
+                async {},
+                |_| StopSurfaceMessage::Offload,
+                Duration::from_secs(1),
+            )
+            .is_err_and(|rejected| {
+                let (work, continuation) = rejected.into_inner();
+                drop(work);
+                matches!(continuation(Ok(())), StopSurfaceMessage::Offload)
+            })
+        {
+            observed |= STOP_SCOPED_OFFLOAD_REJECTED;
+        }
+        self.0.store(observed, Ordering::SeqCst);
+        Ok(())
+    }
+}
+
+#[tokio::test]
+async fn stop_immediately_rejects_every_public_local_work_submission() {
+    let observed = Arc::new(AtomicUsize::new(0));
+    let mut tree = Tree::new();
+    let actor = tree
+        .add_actor_once(
+            "stop-surface",
+            ActorOnceDef::<StopSurfaceActor>::new(Arc::clone(&observed)),
+        )
+        .expect("valid actor");
+    let system = tree.spawn().expect("runtime is available");
+    system.wait_started().await.expect("actor starts");
+    actor
+        .send(StopSurfaceMessage::Stop)
+        .await
+        .expect("stop message is accepted");
+    assert_eq!(system.wait().await, shelterwood::StopReason::Finished);
+    assert_eq!(observed.load(Ordering::SeqCst), ALL_STOP_RESULTS);
+}
+
 #[tokio::test]
 async fn handler_decorator_reenters_the_inner_actor_context_across_callbacks() {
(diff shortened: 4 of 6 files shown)


crates/shelterwood/tests/actor.rs
```diff
@@ -133,6 +133,137 @@ async fn handler_actor_runs_init_handle_and_stop_through_the_raw_wrapper() {
...
+                )
+            })
+        {
+            observed |= STOP_INTERVAL_REJECTED;
+        }
+        // Clearing does not queue work. The stop already emptied the timer
+        // store, so the live callback facade reports that there is nothing
+        // left to retract.
+        if matches!(context.clear_timer(&"timer"), Ok(false)) {
+            observed |= STOP_CLEAR_REPORTS_EMPTY;
```
ralexstokes, line 228: **Nit — item 4 (test quality): `STOP_CLEAR_REPORTS_EMPTY` is vacuous.**

`clear_timer(&"timer")` names a key that was never armed. The two `set_*` calls above used `"timeout"` and `"interval"`, and *both were rejected*, so no timer by any name exists — `Ok(false)` comes back with or without the preceding `context.stop()`.

Evidence, not inference: under the mutation that removes the stop from the guards (drop `local_stop` from `is_stopping()`, stop clearing `accepting`), this was the **only surviving bit** — the test failed with `observed == 8`, i.e. exactly `1 << 3`. Every other bit detected the mutation; this one cannot detect anything.

The comment two lines up ("The stop already emptied the timer store, so the live callback facade reports that there is nothing left to retract") describes a behavior the assertion does not test. Fix is small: arm a timer *before* `context.stop()`, then assert `clear_timer` on that key reports `Ok(false)` afterwards — that actually pins the emptied-store claim.

crates/shelterwood/src/actor.rs
```diff
@@ -184,7 +184,12 @@ impl<'a, A: Actor> Context<'a, A> {
         }
     }
 
-    /// Requests a clean local stop after the current callback.
+    /// Requests a clean local stop.
+    ///
+    /// External intake and incarnation-owned continuations, timers, and
+    /// offloads freeze at this call. The current callback may finish, but any
```
ralexstokes, line 190: **Nit — item 4 (naming/scope, not a hole): the new test pins the rejection half of this sentence, not the freeze half.**

Executed: replacing `RawContext::stop`'s body with a bare `self.local_stop.fire()` — no `receiver.freeze()`, no `resources.freeze()` — leaves `stop_immediately_rejects_every_public_local_work_submission` **passing**. `is_stopping()` still reads the fired latch, so all six guards still reject; nothing in the new test observes that continuations, timers and offloads were actually released.

To be explicit: this is **not** a repo-wide gap. `tests/offloads.rs::cancellation_destructor_panics_do_not_poison_or_short_circuit_cleanup` does catch that mutation. The note is only that the test's name and this doc sentence promise "freeze at this call", while the test's reach is the rejection consequence. No action needed if that split is intentional.

ralexstokes (commented):
## Verdict: merge-with-nits — no correctness defects

All four items do what #366 asked and the code is right. Everything below is a docs-consistency miss (one) and test-quality gaps (four), organized by item.

**Ran:** `cargo +nightly fmt --all --check`; `clippy -D warnings` on both changed crates (`--all-targets --all-features`); `cargo doc` with `RUSTDOCFLAGS=-D warnings`; `tools/sync-packaged-docs.sh --check`; 644 tests (`-p shelterwood -p shelterwood-core --all-features --lib --tests`) — all green. Plus three mutation experiments and one scratch integration probe (described below, then removed).

**Did not run:** the rest of `just ci` — `runtime-api-check`, `external-consumer-check`, `nixfmt-check`, doctests — and `ci-nix`.

### What held up under adversarial checking

- **Item 2 (jitter).** Re-derived the math independently rather than trusting the issue text: for `d = 2^60+1`, `duration_nanos(d)` rounds to `2^60`, and `f64::from_bits(1) * 0.5` ties-to-even to exactly `0.0`, so `0.5 + 0.0 = 0.5` and the product is `2^59` — precisely 1 ns below `half_duration(d) = 2^59+1`. Swept the post-clamp formula over `{1, 2, 3, 2^53-1, 2^53, 2^53+1, 2^60, 2^60+1, 2^62+7, 10^9+1}` x 11 samples (including both boundaries and `1 - 2^-53`): zero violations of the documented `[d/2, d]` range and zero monotonicity inversions. The integer and float branches agree at the boundary — the float branch is now `>= half_duration`, equal as `sample -> 0+`. Upper bound survives because `half_duration(d) <= d` for every `d >= 1 ns` and the trailing `.min(maximum)` is untouched. **Mutation check executed on a committed tree:** reverting `.max(half_duration(delay))` fails the new test with `left: ...423488s / right: ...423489s`. Non-vacuous, and it sits directly beside the zero-sample twin.
- **Item 4 (facade assertions).** Enumerated the public `Rejected` surface myself rather than trusting the test's claim to be exhaustive: `Context` has exactly six (`actor.rs:220, 234, 251, 264, 283, 304`). All six are exercised — **no gaps**. Part II surfaces named in the B.1 table (`send_after_to`, `watch`, `project`) do not exist yet. Evidence is published as an `AtomicUsize` bitset and judged in the test body at `tests/actor.rs:264`, not inside the handler, so the swallowed-assertion trap is correctly avoided. **Mutation check executed:** dropping `local_stop` from `is_stopping()` and no longer clearing `accepting` fails the test `left: 8, right: 63`.
- **Item 5 (SPEC §11).** Confirmed by running code, not by reading it. Scratch integration test on a current-thread runtime: `tree.spawn()` then a synchronous `system.scope().request_shutdown()` before any `await`, so the latch is set before the driver task's first poll. Result: `init` ran exactly once and `system.wait()` returned `ShutdownRequested`. The code path agrees — `run_scope_incarnation` runs `publish_initial_children()` + `settle_supervisor()` at `driver.rs:1080-1082`, before the loop's first `root.take_shutdown_request(scope.epoch)` at `:1087`. Section numbering is right: the sentence lands under `## 11. Shutdown`, inside the pending-incarnation-stop-latch bullet.
- **No packaged-docs sync needed.** `tools/sync-packaged-docs.sh` mirrors only `README.md` and `docs/{embedding,observation,retry-and-ordering,shutdown-and-resources}.md`. `specs/` is not in the list and `--check` passes.
- **The `cargo fmt` blind spot no longer applies.** Probed it directly: appending deliberately misformatted code to `crates/shelterwood/tests/actor.rs` *is* caught by `cargo +nightly fmt --all --check`. `register_test_modules!` is gone from the tree and `tests/*.rs` are ordinary auto-discovered integration targets, so the new test code is format-checked like any other.
- **Item 3 (`SystemRun::drop`) is inert-after-join, not a skipped cleanup.** `self.driver` is taken only in `join_driver` (`driver.rs:106`), so `None <=> joined`. The Drop body is a single statement; the early return skips only that, and field drop glue for `root`/`driver` still runs. The unclean path (`driver == Some`) is unchanged, so `System`'s `#[must_use = "dropping the sole system owner requests graceful shutdown"]` contract is intact.
- **B.1 row now agrees with SPEC's own sentences** at §7:1065 ("intake freezes at the request") and B item 15:3109 ("External intake freezes at stop"). Live staying `✓` alongside `continue_with` = `✓` is consistent with the legend ("✓ = available"), and the row's prose carries the after-stop caveat.

### Findings

One real docs miss (4a, inline at `specs/SPEC.md:431`) and five nits, all inline below and attributed to their item.

---

### Finding 4a — `specs/SPEC.md:431` (the one pre-merge ask)

_Posted in the summary rather than inline: line 431 is outside the diff, so GitHub's review API cannot anchor there._

**Finding 4a (item 4 — highest-severity item in this PR; docs).**

This line still reads "effective after the current callback". That is verbatim the phrasing #366 flagged as wrong at `actor.rs:187` — and it cross-references **B.1**, the exact row this PR corrects two thousand lines below.

Failure scenario: a reader in §4 follows the `(B.1 — ...)` pointer into the capability matrix and gets two contradictory statements of one rule — here, publication-timing framing ("after the current callback"); there, "freezes intake and incarnation-owned resources at the call, so later same-callback work is rejected". A future review reading §4 first re-files precisely the finding this PR exists to close.

Suggested wording, matching the amended row:

> so the return channel carries errors only and stop has one mechanism (§1 principle 2). Clean self-stop is `ctx.stop()` alone (B.1 — intake and incarnation-owned resources freeze at the call, terminal publication follows the callback, `Err` outcome wins, idempotent).

This is the only finding I would ask to land before merge; everything else is a nit.

specs/SPEC.md
```diff
@@ -3280,7 +3282,7 @@ in `on_stop`. ✓ = available; **R** = MUST be `Rejected`-or-absent (§6.4);
...
-| `stop()` (clean self-stop; `Err` outcome wins; idempotent — during drain the already-stopping no-op; arms the child's configured §11 ladder as the stop bound. Live/Drain: effective after the current callback; in a successful effective-`AfterInit` initializer, after its automatic readiness edge. Raw: freezes intake at the call — drain the frozen prefix via `try_recv` after `recv` yields `None`; §1 principle 5's public primitive for the blanket loop's `stop()`) | ✓ | ✓ | ✓ | — |
+| `stop()` (clean self-stop; `Err` outcome wins; idempotent — during drain the already-stopping no-op; arms the child's configured §11 ladder as the stop bound. Live: freezes intake and incarnation-owned resources at the call, so later same-callback work is rejected; the callback may finish before terminal publication. In a successful effective-`AfterInit` initializer, stop publication follows its automatic readiness edge, while the freeze remains immediate. Drain is already frozen. Raw: freezes intake at the call — drain the frozen prefix via `try_recv` after `recv` yields `None`; §1 principle 5's public primitive for the blanket loop's `stop()`) | ✓ | ✓ | ✓ | — |
```
ralexstokes, line 3341: **Nit — item 4 (coverage): the new effective-`AfterInit` clause is true in code but has no facade assertion.**

"stop publication follows its automatic readiness edge, while the freeze remains immediate" checks out — `Context::stop` routes the `Initializing` + `AfterInit` case to `defer_stop_until_after_init`, which calls `receiver.freeze()` + `freeze_and_report()` at the call, and `is_stopping()` includes `deferred_init_stop`. So a same-`init` `continue_with` after `stop()` *is* rejected.

Nothing pins it. `tests/readiness.rs` has four stop-in-init actors (`:53`, `:72`, `:88`, `:104`) and none attempts a submission afterwards; the new test drives only the `Live` stage. Adding one init-stage bit to `StopSurfaceActor::init` would close it cheaply — otherwise the "freeze remains immediate" half of this row is a spec claim with no fence.

(The row itself reads correctly and now agrees with §7:1065 and B item 15:3109.)

crates/shelterwood/src/driver.rs
```diff
@@ -134,6 +134,9 @@ fn classify_retained_root_driver_join(
 
 impl Drop for SystemRun {
     fn drop(&mut self) {
+        if self.driver.is_none() {
```
ralexstokes, line 137: **Nit — item 3: correct, but the guard has no regression test.**

The change itself is sound: `self.driver` is taken only in `join_driver` (`:106`), so `None <=> joined`; the Drop body is a single statement and the early return skips only that, with field drop glue for `root`/`driver` still running; the unclean path is unchanged.

Being candid about *why* there is no test: by construction. Removing the guard restores `main`'s behavior, and `main` is green — so nothing in the suite can distinguish the two, and no amount of searching would find a fence. (Reasoned, not executed; I did not run the full suite with the guard removed.)

Worth flagging because this was the one item whose stated goal was "converting a documented-fragile invariant into an enforced one". The enforcement now exists and the comment below still explains why it matters, but the invariant is exactly as unfenced as before. A crate-internal probe is feasible if you think it earns its keep: hold the root cell, run `System::wait()`, then assert `root.take_shutdown_request(epoch)` is `false` after the `SystemRun` drops — that fails on `main` and passes here.

specs/SPEC.md
```diff
@@ -2095,8 +2095,10 @@ cooperative cancel → grace expiry → tidy-abort beat → hard abort
   accepted while the membership has *no live incarnation* (a restart
   window — B.9's `shutdown_and_wait` landing between incarnations) is
   held on the membership and armed onto the next incarnation at its
-  start, which then starts and immediately begins teardown. The two
-  rules partition by target and never conflict: fresh-restart-starts-
+  start, which then starts and immediately begins teardown. A stop latched
+  before the driver's first settlement likewise still constructs the first
+  incarnation and then immediately tears it down. The two rules partition by
+  target and never conflict: fresh-restart-starts-
```
ralexstokes: **Cosmetic — item 5 (two small things, both in this paragraph).**

1. The insertion left this line ragged mid-paragraph ("target and never conflict: fresh-restart-starts-"), short against the file's otherwise consistent wrap. No CI lint enforces markdown width, so purely style — but a reflow of the bullet would tidy it.
2. "The **two** rules partition by target and never conflict" now follows a paragraph that states three (pending-incarnation latch, fresh-restart-starts-clear, and the newly added pre-first-settlement window). The count and the antecedent both read slightly stale after the insertion. If the new sentence is meant as an instance of the pending-latch rule rather than a third rule, saying so explicitly would keep the "two rules" framing intact.

The sentence itself is accurate — I confirmed it by running code, not reading it (details in the summary).

crates/shelterwood-core/src/policy.rs
```diff
@@ -469,6 +469,7 @@ impl Backoff {
                     half_duration(delay)
                 } else {
                     duration_from_nanos(duration_nanos(delay) * (0.5 + sample * 0.5))
+                        .max(half_duration(delay))
```
ralexstokes, line 472: **Item 2 — confirmed correct, no action.** Noting the positive result inline so it is not mistaken for unreviewed.

Re-derived independently: for `d = 2^60+1`, `duration_nanos(d)` rounds to `2^60`; `f64::from_bits(1) * 0.5` ties-to-even to exactly `0.0`, so `0.5 + 0.0 = 0.5` and the product is `2^59` — exactly 1 ns below `half_duration(d) = 2^59+1`. The clamp restores it.

Range and monotonicity swept over `{1, 2, 3, 2^53-1, 2^53, 2^53+1, 2^60, 2^60+1, 2^62+7, 10^9+1}` x 11 samples: zero violations. Branches agree at the boundary (float branch is now `>= half_duration`, equal as `sample -> 0+`), and the `d` upper bound holds because `half_duration(d) <= d` for every `d >= 1 ns` with `.min(maximum)` untouched.

Mutation check on a committed tree: reverting this line fails `smallest_positive_sample_equal_jitter_respects_the_exact_half_delay` with `left: ...423488s / right: ...423489s` — the exact 1 ns the issue predicted.

ralexstokes:
Review findings closed — all six in `baefde0`, none a behavior change:

- **§4 / B.1 contradiction (the one pre-merge ask)**: SPEC §4 still described `ctx.stop()` as "effective after the current callback" while pointing at the B.1 row this branch corrects. Reworded to match; `git grep "after the current callback"` across `specs/`, `docs/` and `crates/` now returns nothing, so no third site survives.
- **`STOP_CLEAR_REPORTS_EMPTY` was vacuous**: it retracted a key that was never armed, so `Ok(false)` came back with or without the stop — it was the only surviving bit under mutation. Now arms a timer first and asserts the pre-arm succeeded, so the check cannot silently go empty again.
- **Freeze half unpinned locally**: a `RawContext::stop` reduced to a bare `local_stop.fire()` left the stop-surface test passing. The emptied timer store plus a `NotRunning` self-send now pin the intake and resource freeze in this file rather than only in the distant `tests/offloads.rs`.
- **Effective-`AfterInit` clause unfenced**: added a facade assertion that stops in `init` and observes the same frozen surface.
- **`SystemRun::drop` guard had no regression**: pinned through the root member-record pulse, which fires only when the request the guard suppresses is actually published. This was the item whose stated goal was converting a documented-fragile invariant into an enforced one, so it should not have shipped without a fence.
- **Cosmetics**: §11 paragraph reflowed, and the pre-first-settlement window is now named as the scope latch's own version of that window so "the two carry rules" keeps its antecedent.

The jitter clamp thread was a confirmed-correct note, no action: the math was re-derived independently (for `d = 2^60+1`, `f64::from_bits(1)*0.5` ties-to-even to 0.0, landing exactly 1 ns below `half_duration`), a property sweep over boundary delays × 11 samples found zero range or monotonicity violations, and the new smallest-positive-sample test fails without the clamp.

Issue #363, opened 2026-08-20
Tracking: 2026-08-20 review — cancellation/unwind bugs and simplification backlog


ralexstokes:
Follow-up to #348. This review covered the current five-crate workspace file by file, then traced cross-module ownership, cancellation, lock, and panic/disposal paths.

Unlike #348, this pass found several reachable correctness defects. Most are variations of one theme: normal transaction paths isolate user-owned values and wakers correctly, but cancellation and nested-unwind paths still have places where caller code can run inline, be destroyed on a framework task, or overwrite an earlier failure.

Reviewed commit: `96c078ca0d25cc9f00a4c55f120ab36a744e9cd7`.

A second independent review was checked against the same control flow. Its confirmed lifecycle, raw/handler-contract, and last-owner-disposal findings are folded into this tracker. Claims without a supported-path producer (notably a stale root finisher closing a newer root incarnation) and behaviors with an explicit alternative API or documented exhaustion exception are intentionally omitted.

## P0 — address before expanding the public API

- [ ] **Suppress construction before the first supervisor settlement when a stop source is already latched.** [`run_scope_incarnation`](https://github.com/ralexstokes/shelterwood/blob/96c078ca0d25cc9f00a4c55f120ab36a744e9cd7/crates/shelterwood/src/driver.rs#L1077-L1085) publishes residents and runs `settle_supervisor()` before sampling the root shutdown request. The resulting `StartChild` effects reach [`spawn_child`](https://github.com/ralexstokes/shelterwood/blob/96c078ca0d25cc9f00a4c55f120ab36a744e9cd7/crates/shelterwood/src/driver/child.rs#L672-L691), which rechecks removal and reducer draining state but not an already-latched shutdown, force, or ancestor stop. Dropping the sole `System` owner immediately after `spawn()` can therefore invoke first-generation user construction before teardown, contrary to shutdown-before-start ownership semantics. Sample stop sources before initial settlement and make `spawn_child` the single construction funnel by consulting the complete [`construction_is_suppressed`](https://github.com/ralexstokes/shelterwood/blob/96c078ca0d25cc9f00a4c55f120ab36a744e9cd7/crates/shelterwood/src/driver/events.rs#L210-L225) predicate. Add an already-fired-stop regression for each one-shot construction kind.

[...]

PR #375, opened 2026-08-20
Retain provisional failed outcomes

ralexstokes:
Part of #366.

Introduces `RetainedRecordedOutcome` beside `RetainedExit` so failed provisional outcomes retire through critical disposal when framework event or driver state discards them. The child task's report cell carries the retained form from the moment it is filled, and child-exit events carry it across failed forwarding, receiver teardown, pending-batch suffixes, and defensive early returns.

The core reconciliation/classification folds now return their discarded value to the runtime-owning caller. Framework callers reach them only through `classify_exit_retaining`, `reconcile_recorded_outcomes_retaining` and `classify_disposal_panic_retaining` in `shelterwood-cells`, which make the discard's venue isolated disposal; the raw two-value folds are left for core's own unit tests. This also keeps pending terminal exits retained while folding construction-disposal panics. Selected winners retain ordinary drop timing.

Coverage: carrier conversion/retirement tests, plus one destruction-venue regression per fold, each mutation-checked against deleting its retention —

- `reconcile_recorded_outcomes_retaining`: `disposal.rs::readiness_timeout_disposes_a_losing_application_error_off_the_driver`
- `classify_exit_retaining`: `retained_outcomes.rs::a_join_panic_disposes_the_recorded_application_failure_off_the_driver`
- `classify_disposal_panic_retaining`: `shutdown.rs::an_arrived_disposal_panic_disposes_its_losing_application_error_off_the_driver`
- the report cell's claim window: `reports.rs::an_unclaimed_report_disposes_its_failed_outcome_off_the_dropping_thread`

plus an ordinary nested-unwind regression for `pending.drain(..)` suffix destruction, which relies on nextest's per-test process isolation and pins both the surviving primary panic and disposal of the displaced user error.

`AGENTS.md`'s lock-rule section names the new carrier and the retaining wrappers.

Validation: rebased onto `main` (clean); `cargo nextest run -p shelterwood -p shelterwood-cells -p shelterwood-core` (709 tests), `cargo +nightly clippy --locked --workspace --all-targets --all-features -- -D warnings`, `just fmt`, `cargo doc --no-deps --workspace`.

Stack: independent Batch 2 slice; targets `main`.


ralexstokes:
**Verdict: no correctness defects found in the change itself. Four findings — one mutation-invisible coverage gap on the site #366 specifically called out, one residual raw-outcome destruction site the carrier does not reach, and two cleanup/altitude items.**

I traced every site the Batch 2 bullet enumerates and each is genuinely covered: the ordinary-path loser fold (`child.rs:1024-1029`), the `pending.drain(..)` suffix and driver-teardown receiver drops (via the type change at `events.rs:22`), the failed forward on the exit-joiner task (`child.rs:600`), `handle_exit`'s three defensive early returns (`child.rs:997-1006` — `recorded` is still the retained carrier there), and the pending-terminal un-retain gap (`child.rs:1236-1245`). Two things I checked and confirm are *not* problems: `ActiveChild::forced_outcome` is only ever `readiness_timed_out` (`startup.rs:79`) or `aborted` (`child.rs:914`), never `Failed`, so leaving it a raw `RecordedOutcome` is correct; and SPEC §"item 18" (`specs/SPEC.md:3143-3161`) explicitly carves the venue question out of the lock-rule item, so nothing here contradicts the spec and no spec edit is owed.

---

### 1. (Medium) The `classify_disposal_panic` loser retention is mutation-invisible — no test fails if you delete it

`crates/shelterwood/src/driver/child.rs:1243-1245`

```rust
let (selected, discarded) = classify_disposal_panic(exit.into_exit(), message);
exit = RetainedExit::new(selected);
drop(RetainedExit::new(discarded));
```

This is the one line in the PR that closes the gap the issue calls "the adjacent gap #363 missed … escaping the coverage CLAUDE.md claims for pending terminal disposal". `rank(Failed) < rank(Panicked)`, so whenever a construction-disposal panic arrives for an exited incarnation carrying an application failure, the *loser is always the user error*, and before this PR it was destroyed inline on the driver task. That is the most user-reachable half of the change — and nothing pins it.

The three tests that drive this path (`driver/tests/shutdown.rs:246`, `:271`, `:299`) all run through the `scope_with_arrived_factory_disposal_panic` fixture, whose recorded outcome is `ExitError::message("application failure")` (`shutdown.rs:68-70`) — a framework-owned `String`, so its destruction venue is unobservable. `assert_arrived_disposal_panic_published` only checks the published verdict. Mutating `drop(RetainedExit::new(discarded))` to `drop(discarded)` leaves the whole suite green.

Concrete failure the missing test would catch: a task returning `Err(user_error_with_blocking_destructor)` whose retained construction *also* panics on disposal ⇒ the user destructor runs on the driver task and stalls supervision.

Remedy: parameterise the fixture (or add one case) so the recorded error is a thread-reporting `ExitError`, and assert the disposal thread is not the driver thread — the same shape as the new `readiness.rs` test. This is a few lines given the fixture already exists.

### 2. (Low–Medium, PLAUSIBLE) `RecordedReport.outcome` is still a raw `RecordedOutcome` across the claim window

`crates/shelterwood/src/driver/child.rs:3-6` (`RecordedReport`), `:34-52` (`report_slot`), `:57-70` (`ReportCompletion::fill`), `:595-606` (the joiner).

The carrier is applied at the *send* site (`report.outcome.map(RetainedRecordedOutcome::new)`), which leaves one window uncovered: the child task fills the shared `OnceLock<RecordedReport>` with a `Failed` outcome when it ends, and the joiner task's `report_claim` is the sole remaining owner from that moment until it is next polled. If the joiner is dropped in that window — runtime teardown drops un-polled tasks, which is exactly the hard-shutdown shape — the `Arc<OnceLock<RecordedReport>>` drop glue destroys the type-erased user error inline on whatever thread tears the task down (tokio's runtime-shutdown thread). Same class as everything else in this PR; a blocking destructor there stalls runtime teardown.

Remedy: move the wrap one step earlier — have `ReportCompletion::fill` store `Option<RetainedRecordedOutcome>` in `RecordedReport`. That costs nothing (it runs on the child task at its own end), deletes the `.map(RetainedRecordedOutcome::new)` at `:600`, and closes the window by construction rather than by "the joiner always gets polled".

I have not reproduced this; the window is real by ownership but narrow, so PLAUSIBLE rather than CONFIRMED.

### 3. (Low) Two of the four new discard-handling sites are provably dead, and the tuple return enforces nothing

`crates/shelterwood/src/driver.rs:133-134` and `crates/shelterwood/src/driver/child.rs:126-133` both call `classify_exit(None, …)`. With `recorded == None`, `classify_exit`'s match can only take the `(None, Some(join))` or `(None, None)` arms, both of which return `discarded == None`. So `drop(discarded.map(RetainedExit::new))` at those two sites is unreachable, and reads to a future reviewer as live containment.

Related: `let (exit, _) = classify_exit(...)` compiles and silently restores the pre-PR behaviour. Nothing in the type system stops it — which is the shape CLAUDE.md asks to avoid ("makes an out-of-transaction mutation unavailable by construction").

Both are addressed by the same move: one helper beside `RetainedExit` in `shelterwood-cells`, e.g.

```rust
pub fn classify_exit_retaining(recorded, join, hard_abort_phase, cancellation) -> Exit {
    let (exit, discarded) = classify_exit(recorded, join, hard_abort_phase, cancellation);
    drop(discarded.map(RetainedExit::new));
    exit
}
```

That collapses the duplicated two-line idiom at all three real sites (plus the two dead ones disappear), and leaves the raw core fold reached only from core's own unit tests. `reconcile_recorded_outcomes` and `classify_disposal_panic` take the same treatment.

Minor, same area: the nested `{ … }` inside the `else` arm at `child.rs:125-134` is redundant; and `RetainedRecordedOutcome`'s `Debug` impl (`retained.rs:159-163`) has no caller — `DriverEvent`/`ChildEvent` derive no `Debug` and `ObservedExit` doesn't either.

### 4. (Low) Venue test is in the wrong file and restates a probe that has an established home

`crates/shelterwood/tests/readiness.rs:30-49, 777-831`

`crates/shelterwood/tests/disposal.rs` is the destruction-venue file: it already owns `DropProbe`, `assert_disposed_off_current` / `assert_disposed_off` (`disposal.rs:195-215`), and the `#[test]` + `Builder::new_current_thread` + `assert_ne!(dropped, driver_thread)` idiom this new test hand-rolls with `spawn_blocking` (`disposal.rs:449-470`). `tests/common/ownership.rs` is where shared probes live (it already exports `PanicOnDrop`). A fresh `ThreadProbe` in `readiness.rs` is the third restatement of that type in the tree (`shelterwood-cells/src/cells/mod.rs:77` has the other). Suggest moving the test to `disposal.rs` and putting the `Error`-implementing probe in `common/ownership.rs`.

I did verify the test is **not** vacuous: `#[tokio::test(start_paused = true)]` forces the current-thread flavor and `runtime::spawn` is `tokio::task::spawn`, so the driver really does run on the test thread; and `assert!(matches!(task.wait().await.kind(), ExitKind::ReadinessTimedOut { .. }))` pins that the application error is the *loser*, which is what makes `assert_ne!` meaningful. Reverting the retention makes it fail.

---

### Notes (not findings)

- **Branch is 12 commits behind `main`** (merge base `e808e68`, so #369–#374 and #385 are absent). `git merge-tree` reports a clean merge, but #374 edited `driver.rs` three lines below this PR's `classify_retained_root_driver_join` hunk, and the green CI here is not a combined-tree result. Update before merging, per the repo's stacked-merge rule.
- `pending_drain_suffix_contains_failed_outcome_destruction_during_unwind` builds its own `Vec<(ArbitrationClass, Pending)>` rather than driving `run_scope`'s loop, so it pins the carrier's behaviour inside an unwind but not the driver's actual suffix guard. That's the right trade given the fix is a type change on `ChildEvent::Exited`, but worth a one-line comment saying so. The abort-class signal itself checks out: pre-fix, the drain suffix destroys `PanickingDrop` during the unwind ⇒ double panic ⇒ SIGABRT, which nextest reports per-test.
- `CLAUDE.md`'s lock-rule section enumerates the retained-carrier family and names only `RetainedExit` ("…protect failed exits through `RetainedExit`"). Adding the provisional-outcome sibling there would keep the one document a reviewer is told to read first accurate.
- No prior review comments on this PR to adjudicate.

ralexstokes:
Thanks — all four findings addressed. Branch rebased onto `main` first (clean, no conflicts with #374's `driver.rs` hunk); everything below was verified on the combined tree. Pushed as `06599a2` + `a716573`.

### 1. (Medium) `classify_disposal_panic` loser retention is mutation-invisible — **fixed**

Confirmed and fixed. `scope_with_arrived_factory_disposal_panic` is now a thin wrapper over `scope_with_arrived_factory_disposal_panic_over(recorded: ExitError)` (`driver/tests/shutdown.rs`), so the three existing cases keep their `ExitError::message` fixture and one new case supplies a payload that reports its drop thread:

`shutdown.rs::an_arrived_disposal_panic_disposes_its_losing_application_error_off_the_driver`

The probe itself is `ThreadReportingError` / `thread_reporting_error()` in `driver/tests/support.rs`, next to the existing `GateProbeError`, rather than a fourth local restatement.

**Mutation check** (committed tree, `drop(RetainedExit::new(discarded))` → `drop(discarded)` inside `classify_disposal_panic_retaining`): the test fails with `left: ThreadId(2), right: ThreadId(2)`.

While pinning that site I checked the other two folds the same way and found the coverage claim was weaker than the review states:

- The venue test you validated (`readiness_timeout_disposes_…`) does **not** pin `classify_exit`'s discard — on the readiness-timeout path the loser is discarded by `reconcile_recorded_outcomes`, because `active.forced_outcome` is `readiness_timed_out`. Mutating `classify_exit_retaining` alone leaves it green; mutating `reconcile_recorded_outcomes_retaining` is what fails it.
- So `classify_exit`'s own discard — reachable as recorded `Failed` + `JoinOutcome::Panic`, i.e. a task that returned an application error and then panicked on its way out of the runtime — had nothing pinning it either. Added `retained_outcomes.rs::a_join_panic_disposes_the_recorded_application_failure_off_the_driver`, mutation-checked the same way.

All four retention sites are now individually mutation-pinned (see the updated PR description for the map).

### 2. (Low–Medium) `RecordedReport.outcome` raw across the claim window — **fixed, as proposed**

Agreed on the ownership analysis; took your remedy. `RecordedReport.outcome` is `Option<RetainedRecordedOutcome>` and the wrap happens in `ReportCompletion::fill` (on the child task, at its own end), which deletes the `.map(RetainedRecordedOutcome::new)` at the send site.

It also turned out to be directly pinnable rather than only closed-by-construction: `reports.rs::an_unclaimed_report_disposes_its_failed_outcome_off_the_dropping_thread` records a failing outcome and then drops the `ReportClaim` without receiving — the exact shape of a joiner dropped un-polled. **Mutation check** (revert to the send-site wrap): the test fails.

### 3. (Low) Dead discard sites + unenforcing tuple return — **fixed, with a scope adjustment**

Took the helper, in `shelterwood-cells/src/cells/retained.rs` beside `RetainedExit`, for all three folds:

- `classify_exit_retaining`
- `reconcile_recorded_outcomes_retaining` (takes `Option<RetainedRecordedOutcome>`, so `handle_exit` no longer unwraps the carrier by hand)
- `classify_disposal_panic_retaining` (takes and returns `RetainedExit`, so the carrier survives the fold rather than being rebuilt around it)

Every framework call site now routes through them; `handle_exit`'s six lines collapse to two, `driver.rs:133` back to one expression, and the redundant `{ … }` block at `child.rs:125` is gone with it. The raw two-value folds in `shelterwood-core` are reached only from core's own unit tests, and their rustdoc now points at the wrappers.

I did **not** make the raw folds `#[must_use]`-enforcing: that stops `let _ = classify_exit(..)` but not `let (exit, _) = classify_exit(..)`, which is the shape you flagged, and a newtype that would stop it costs core a dependency on a concept only cells can honour. Routing framework code through the wrapper is the enforcement.

`RetainedRecordedOutcome::Debug`: **accepted as-is.** It is now reachable in the sense that matters — `RecordedReport` holds the carrier and `as_outcome` is exercised — but you are right that no caller formats it today. Keeping it is three lines of parity with `RetainedExit`, and dropping `Debug` from a `pub` carrier makes any future `derive(Debug)` on `ChildEvent` fail confusingly. Not worth churning either way.

### 4. (Low) Venue test in the wrong file — **fixed**

Moved to `tests/disposal.rs`, immediately before `duplicate_id_rejection_disposes_blocking_definition_off_the_caller`. It now uses the file's own `DropProbe::reporting` and `assert_disposed_off_current` instead of a hand-rolled `spawn_blocking` + `assert_ne!`, so the third `ThreadProbe` restatement is gone. `DropProbe` gains `Debug`/`Display`/`Error` impls so it can stand in as an application-error payload — that felt better than adding a separate error probe to `common/ownership.rs`, since `disposal.rs` is the only consumer and `DropProbe` is already that file's venue probe. `readiness.rs` is back to its prior imports.

**Mutation check** after the move (as above, `reconcile_recorded_outcomes_retaining`): fails.

### Notes

- **Branch staleness** — rebased onto `main` (now includes #369–#374, #385). No conflicts; `driver.rs:133` and #374's neighbouring hunk merged cleanly and the result was rebuilt and retested before anything else.
- **`pending_drain_suffix_…` scope** — added the one-line comment saying it drives the drain directly rather than `run_scope`'s loop, and why that's the right trade.
- **`AGENTS.md`** (`CLAUDE.md` is a symlink to it) — the lock-rule `Arc`-traffic bullet now names `RetainedRecordedOutcome` as the one-step-earlier sibling of `RetainedExit` and names the three retaining wrappers as the supported entry points.

### Verification

`cargo nextest run -p shelterwood -p shelterwood-cells -p shelterwood-core` → 709 passed. `cargo +nightly clippy --locked --workspace --all-targets --all-features -- -D warnings` → clean. `just fmt` → clean. `cargo doc --no-deps --workspace` → clean. Every added/changed test was mutation-checked on a committed tree and the mutation reverted with `git checkout`.

PR #377, opened 2026-08-20
Close never-started nested scope bodies

ralexstokes:
Part of #366.

This closes the observation surface for nested scope bodies that are spawned but never begin their first incarnation:

- attaches a `NestedScopeStart` obligation to the spawned body before runtime submission
- publishes a final `Stopped { NeverStarted }` snapshot/lifecycle state if the body is dropped before first poll — gated on the scope never having begun an incarnation (`ScopeState::Unstarted`, SPEC B.6's "no incarnation has ever spawned" in the scope plane)
- preserves parent-driver ownership of the shared membership's terminal exit and delays stream closure until that terminal parent projection is visible
- leaves a body dropped before its *restart* incarnation alone: that scope already published a real, terminal prior-incarnation reason, so the fallback only closes its stream. (An earlier revision replaced that reason with `NeverStarted`; that was unsound — it depended on which of the two terminality owners ran first, and the winning branch contradicted the shared membership's `Aborted` exit. See the review thread.)
- keeps a structured startup result installed without an incarnation (identity exhaustion) rather than refilling it with `ShutdownRequested`
- verifies public `SnapshotClosed` behavior after terminal root shutdown

Regression coverage includes immediate abort on a current-thread runtime, submission through an already-shut-down runtime handle, both arrival orders of the never-polled-restart case, structured-startup preservation, pre-terminal stream ordering, and the public final-snapshot-before-closure contract.

Validation: `./tools/dev cargo nextest run --workspace` (813 tests on the #375 stack), workspace clippy `-D warnings`, `cargo +nightly fmt --all --check`.

Stack: Batch 2 follow-up on #375 to avoid their shared test-support conflict.


ralexstokes:
## Review

**Verdict: request changes.** The first-spawn never-started case is correct and both new driver regressions genuinely discriminate the fix (mutation-verified below). The *restart* half of the change is unsound: it makes a previously deterministic outcome race-dependent, and the branch it introduces publishes a `Stopped { NeverStarted }` that disagrees with the membership exit and with SPEC B.5.

---

### P1 — Restarted never-polled bodies: nondeterministic terminal projection, and the winning branch contradicts SPEC B.5

`crates/shelterwood-cells/src/cells/scope.rs:1270-1290` (`close_never_started_body`) resets and republishes unconditionally, but `crates/shelterwood/src/driver/child.rs` (`discharge_child_terminality`) and `scope.rs:660-672` (`terminalize_child`'s trailing close) can reach the same scope first. For a *restarted* nested body the two orders now produce different terminal projections.

`MemberRecord::last_incarnation` is set at `MemberTransition::Starting` (`member.rs:135`), i.e. at spawn — not at exit. So a never-polled restart body leaves `last_incarnation = Some(n)` and the nested record reading `Stopped { previous }`, which is exactly the pair `terminalize_child` treats as "the nested epilogue already published its final state, close now".

Probes (run on this head, current-thread runtime, driver-level fixture with a restartable `SubtreeDef` whose prior incarnation finished `Stopped { Finished }`):

* **Parent wins** — `ScopeRuntime::drop` calls `abort_handle.abort()` and then `child.terminality.discharge()` a few instructions later, while the aborted task's future can only be dropped by a scheduler tick. On a current-thread runtime this is deterministic:

  ```
  PROBE immediately after driver drop: closed=true state=Stopped { reason: Finished }
  PROBE after yields:                  closed=true state=Stopped { reason: Finished }
  ```

  The stream closes on the **stale prior reason** and `close_never_started_body` then returns at its `observation.closed` guard. The PR description's claim — *"replaces a restarted body's stale prior stop reason and incarnation-local startup/restart state with the never-started result"* — does not hold for the reachable producer named in #366 (`ScopeRuntime::drop` aborting a subtree spawned in the driver's last poll). `wait_started()` likewise still returns `Ok(())` on that path.

* **Body wins** — force the task drop before terminalization (abort + yields, then drop the driver):

  ```
  PROBE body-wins ordering: closed=false state=Stopped { reason: NeverStarted } total_restarts=TotalRestarts(0)
  PROBE member stage=Starting last_incarnation=Some(Incarnation { .. })
  PROBE after driver drop:  closed=true state=Stopped { reason: NeverStarted }
  PROBE member terminal record=Terminal(Exit { kind: Aborted { phase: WithinGrace }, cancellation: Observed })
  ```

  The nested scope's final projection is `Stopped { NeverStarted }` while the shared membership terminalizes `Aborted { WithinGrace } / Observed` with `last_incarnation: Some(..)`. That pair is what SPEC forbids:

  * SPEC B.5: `NeverStarted` is "membership terminal with **no incarnation ever spawned**".
  * SPEC B.4 (~line 3568): the `Stopped { NeverStarted }` terminal event is for "a scope membership **that never spawns**".
  * `shelterwood-core/src/exit.rs:82-86`: `NeverStarted` is the lattice top precisely because "whenever a membership terminalizes **without ever spawning**, the scope-state projection must agree with the membership exit, in either arrival order." Here it disagrees.
  * `child.rs`'s own comment above this branch still says "A restarting scope already published its real prior-incarnation reason" — this PR contradicts it without updating it.

  On a multi-thread runtime the worker can win this race, so which of the two projections a user observes is genuinely nondeterministic. Before this PR the restart case was deterministic (stale reason, closed) — imperfect, but stable and consistent with the membership exit.

**Remedy.** Gate the reset + publication on the same predicate `discharge_child_terminality` uses — `self.member.record().last_incarnation.is_none()` — so the fallback only ever fires for a membership that truly never spawned, and record the restarted-body stale reason as accepted (#366 explicitly offers that disposition). If the restart correction is actually wanted, it has to be made deterministic on *both* sides (a "body spawned, incarnation not begun" flag on the cell, cleared in `begin_incarnation`, that `terminalize_child` also consults) **and** the membership exit for that case has to become `Exit::never_started()` so the pair still agrees — that is a larger semantic change than this PR is scoped for. Either way the PR description's fourth bullet should be corrected or dropped.

### P3 — Epoch-exhausted nested start now loses its structured startup failure

`driver.rs:895/911`: `begin_nested_incarnation(&scope)?` early-returns *before* `start.complete(drop)`, so the `?` drops the obligation. `begin_nested_incarnation` has just installed `Err(StartupError::StartupFailed(IdentityExhausted))` via `set_startup`; the fallback then clears `record.startup = None` (`scope.rs:1281`) and `publish_stopped_locked` refills it with `Err(ShutdownRequested)`. `wait_started()` on that scope degrades from the structured cause to a generic shutdown, and the state becomes `Stopped { NeverStarted }` rather than the startup-failure projection.

Only reachable through `ScopeEpochs::Exhausted` (2^64 incarnations) or a double-`begin` that cannot happen, so operationally this is a note rather than a defect — but it is a behaviour change on that path that no test covers. `nested_membership_exhaustion_is_structured_and_fail_closed` does **not** exercise it: its exhaustion is at *lowering* (child-id domain), after `start.complete(drop)` has already run. Cheapest fix is the same `last_incarnation`/started-guard as P1, or discharging the obligation explicitly before the `?`.

### Nit — the branch predates #369; the explicit `PanicAccumulator` becomes redundant

This stack is based on `e808e68` (pre-#369). On `main`, `Obligation::drop` already runs `discharge()` inside a `PanicAccumulator` (`storage.rs:53-57`), and every sibling fallback (`admission_control.rs:305`, `child.rs:42`, `child.rs:188`) is written bare. After merging/rebasing onto `main`, `close_never_started_scope_body` (`driver.rs:838-841`) is the only fallback with its own accumulator. Harmless (nested accumulators compose), but worth dropping so the containment seam stays in one place.

### Verified, no finding

* Both new driver regressions discriminate. Stubbing `close_never_started_body` to an early `return` fails both: `nested_body_aborted_before_first_poll_...` on the pre-terminal `Stopped { NeverStarted }` assertion, and `nested_body_spawned_after_runtime_shutdown_...` with `state: Unstarted`, `closed=false`. (My initial guess that the second was covered by the parent path was wrong — `last_incarnation` being set at spawn is exactly why `terminalize_child` cannot reach it.)
* No lock-rule violation. `close_never_started_body` runs entirely on framework-owned record data; the raw `record.startup` is released while `retained_exits` still owns the failed `Exit` (refcount traffic), and `refresh_retained_exits`'s guard swap can only submit to `dispose_critical`, which is legal under the gate. It is the same shape `begin_incarnation` (`scope.rs:773-779`) already uses. The obligation never fires with the gate held: `ObservationTxn::commit` releases the guard before flushing effects, so a hostile-waker resume out of `mailbox.bind`/`transition_child_stage` unwinds `spawn_child`'s `body` with the gate free.
* The obligation cannot fire against a live incarnation: `start.complete(drop)` is the statement after a successful `begin_nested_incarnation`, with no await between, and Tokio cancellation cannot interrupt an in-progress poll. It also strictly precedes the `JoinHandle` completion, so a restart decision cannot overtake it.
* `snapshot_subscription_yields_snapshot_closed_after_terminal_shutdown` is a real behavioural pin (it would fail if closure preceded the final snapshot) and closes #366's Batch 2 coverage item.

---

### Adjudication of existing comments

Both Codex comments were filed against `1f2b4025b2` and are **already addressed** by the current head (`5248ae8`); no action needed on either, but the framing is worth recording.

1. **"Delay closure until the parent terminal publication" (P2) — correct, and fixed.** The `if self.membership_terminal()` gate plus `terminalize_child`'s existing trailing close is the right shape, and the new pre-terminal assertion in `nested_body_aborted_before_first_poll_...` pins it. Confirmed by mutation.
2. **"Clear stale incarnation counters before the final snapshot" (P2) — the observation is right, the remedy is what produced P1 above.** `total_restarts`/`startup` are indeed incarnation-local and `begin_incarnation` does reset them, so *if* the never-started publication is going to stand in for a fresh incarnation then the reset belongs with it. But applying the remedy to the **restarted** case is what creates the `Stopped { NeverStarted }` / `Aborted{..} + last_incarnation: Some` mismatch, and it only lands in one of two possible orders. The reset is correct scoped to a membership that never spawned; it should not have been extended to a restart.

No fabricated spec citations this round — neither comment quoted the spec.

ralexstokes:
## Review response — `0fabe75`

All three findings addressed. The P1 is fixed on this branch with a variant of
your first remedy; the P3 falls out of the same change; the nit is confirmed
but is not yet true on this base.

---

### P1 — restarted never-polled bodies: **fixed** (not with the literal predicate)

Confirmed the diagnosis, including the arrival-order dependence and the
`Stopped { NeverStarted }` / `Aborted { WithinGrace } + last_incarnation: Some(..)`
pair. The remedy needed one adjustment.

**Why not `last_incarnation.is_none()`.** That predicate is false for the
*first* spawn too — `MemberTransition::Starting` sets `last_incarnation` at
spawn for every incarnation, which is the same fact that keeps
`terminalize_child`'s own `NeverStarted` branch out of reach here (your
"Verified, no finding" bullet 1). Gating on it would make
`close_never_started_body` dead code and re-open #366's actual bug; both new
driver regressions fail under it.

**What landed instead** (`shelterwood-cells/src/cells/scope.rs`): gate the
publication on the *scope plane* having never begun an incarnation —

```rust
if matches!(self.record().state, ScopeState::Unstarted) {
    self.publish_stopped_locked(txn, StopReason::NeverStarted, None, None);
}
if self.membership_terminal()
    && matches!(self.record().state, ScopeState::Stopped { .. })
{
    self.close_observation_locked(txn);
}
```

`ScopeState::Unstarted` is SPEC B.6's "membership exists, no incarnation has
ever spawned" verbatim, and `Stopped { NeverStarted }` is its terminal twin, so
this is the same distinction `begin_incarnation` draws — `Unstarted` holds iff
no scope epoch was ever minted. A restarted body's scope is `Stopped { prior }`,
so the fallback now supplies a *missing* terminal projection and never replaces
a published one. Closure is the only effect it owns unconditionally, and its
trailing close picked up the same `Stopped { .. }` guard `terminalize_child`
already applies, so no stream can end on a `Starting`/`Running` payload.

Consequences: the restart case is back to its pre-PR outcome (stale-but-real
prior reason, closed) in **both** orders, `child.rs`'s "a restarting scope
already published its real prior-incarnation reason" comment is true again
without editing it, and the reset that motivated Codex's second comment is gone
— under this gate `total_restarts` is provably `ZERO` (a scope only charges
restarts inside an incarnation) and `startup` is either `None` or a cause worth
keeping (see P3).

**Evidence.** Rather than probe the two orders through the scheduler, the new
regression drives them explicitly, so it is deterministic on every runtime
rather than only on current-thread:

`cells::scope::tests::never_polled_restart_body_keeps_its_prior_reason_in_either_arrival_order`
builds a nested scope sharing the child's `MemberCell`, runs and finishes
incarnation 1 (`Stopped { Finished }`), transitions
`RestartScheduled` → `Starting { 2 }` without a `begin_incarnation`, then runs
`terminalize_child` and `close_never_started_body` in each order.

Mutation check — the fix reverted to this PR's previous body, tests unchanged:

```
running 3 tests
test never_started_body_publishes_before_membership_closes_observation ... ok
test never_polled_restart_body_keeps_its_prior_reason_in_either_arrival_order ... FAILED
  assertion `left == right` failed: a never-polled restart keeps its prior
  incarnation's reason (parent_first=false)
    left: Stopped { reason: NeverStarted }
   right: Stopped { reason: Finished }
test never_started_body_preserves_a_structured_startup_failure ... FAILED
  a structured startup cause survives the never-started fallback: Some(Err(ShutdownRequested))
```

`parent_first=true` passes on the old code and `parent_first=false` fails: that
*is* the nondeterminism, reproduced as two deterministic orders. With the fix
restored all three pass, and both of your driver regressions still pass
unchanged.

### P3 — epoch-exhausted nested start: **fixed** by the same change

Dropping `record.startup = None` is sufficient. `publish_stopped_locked` only
installs `Err(ShutdownRequested)` when `record.startup.is_none()`, so the
`Err(StartupFailed(IdentityExhausted))` that `begin_nested_incarnation` installs
before its `?` drops the obligation now survives, and `wait_started()` keeps
reporting the structured cause.

Pinned by `never_started_body_preserves_a_structured_startup_failure` (see the
mutation output above), which installs the cause directly rather than burning
2^64 incarnations.

I did **not** take the "discharge the obligation before the `?`" alternative:
on that path the parent's `terminalize_child` cannot publish `NeverStarted`
(`last_incarnation` is `Some`) and its trailing close is guarded on a `Stopped`
projection, so suppressing the fallback would leave the nested stream open
forever — #366's bug on the exhaustion path.

The projection stays `Stopped { NeverStarted }` rather than a startup-failure
projection: the scope genuinely never began an incarnation, `NeverStarted`
outranks `StartupFailed` in B.6's lattice, and the structured cause is
preserved where `wait_started` reads it. **Accepted** with that rationale.

### Nit — `PanicAccumulator`: confirmed, **deferred to the rebase**

Verified against `origin/main`: `Obligation::drop` there wraps `discharge()` in
a `PanicAccumulator`. It does **not** on this branch's base
(`driver/storage.rs` here is a bare `self.discharge()`), so
`close_never_started_scope_body`'s accumulator is currently the only
containment on the path and removing it now would be a regression until the
rebase. #375 is being edited concurrently, so I deliberately did not rebase.
The accumulator comes out in the same commit that lands this stack on
post-#369 `main`.

---

### Also

* **PR description corrected** — the fourth bullet no longer claims the
  restarted-body replacement; it now states what the fallback does and does not
  do for a restart, with a pointer to this thread.
* **Filed #391** for the residual semantic question your B.6 citation raises.
  The `Stopped { NeverStarted }` / `Aborted + last_incarnation: Some(..)` pair
  survives for the *first*-spawn never-polled case, which is what this PR is
  for. It is not new: `last_incarnation` is set at spawn while
  `begin_incarnation` runs on first poll (a window SPEC ~L3855 sanctions), and
  `wait_stopped` has always mapped `ScopeState::Unstarted` to `NeverStarted`,
  so that answer predates this PR — the PR only makes the record, the final
  snapshot and the last event agree with it instead of leaving the projection
  non-terminal and the stream open. Whether `NeverStarted` describes the scope
  plane or the membership plane is a spec adjudication with three options, all
  larger than this PR; #391 lays them out. No correctness exposure either way:
  the case is terminal, closed and order-independent.

### Verification

`cargo nextest run --workspace` — 813 passed, 0 failed.
`cargo clippy --workspace --all-targets -- -D warnings` — clean.
`cargo +nightly fmt --all` then `--check` — clean.
Mutation check performed on the committed tree, then reverted.

ralexstokes:
Superseded by #394. GitHub closed this permanently when its base branch (#375) merged and was deleted, and the rebase force-push made it unreopenable.

The work is unchanged and now targets `main` directly: same three commits rebased onto post-#375 `main`, plus the deferred rebase nit from the review — `close_never_started_scope_body`'s own `PanicAccumulator` is now bare, since `Obligation::drop` contains its fallback and this obligation has no explicit discharge path. Verified 839/839 on the combined tree.

The review findings and their resolutions stay recorded here.

PR #378, opened 2026-08-20
Contain offload completion wakes

ralexstokes:
Part of #366.

This contains caller-owned `Guard::finished()` wake panics on every ordinary `SharedOffloadFuture` completion path. Instead of escaping through the framework-owned offload task and reaching the incarnation only as the join result's stringified stand-in, the caller's original payload is retained in the incarnation panic slot and the actor signal is pulsed.

Completion notification and its publication are now separate edges. `fire_finished` claims the completion latch's *own* transition with `Latch::fire_silently`, runs the waker-visible `notify` under `catch_panic`, and only then release-stores a `finished_published` flag that ledger reclamation consults instead of the latch's earlier fired bit. Because the claim is the latch's own, the thread that runs the wake is by construction the thread that publishes: a concurrent cancellation can neither retire the work mid-wake nor win a second, divergent claim and run the wake beside the publisher.

**This changes when an incarnation fails**, and the change is now written down. A `Guard::finished()` waiter's waker panic is incarnation-owned disposal on the *ordinary* completion path, not only at the intake freeze, so a third party awaiting `finished()` with a panicking waker fails a live incarnation at its next receive boundary and suppresses `on_stop`. SPEC §6.2's funnel enumeration, §6.5's `Guard` bullet, `Guard::finished`'s own rustdoc, and the `recv`/`try_recv` retention lists all say so. The accepted trade in the other direction is recorded with it: pinning the entry until publication means a completion waker that *blocks* rather than panicking holds the work in the ledger, so teardown joins it and caller code can delay exit publication.

Two deterministic regressions, both with timeout-bounded handshakes so a suppressed fire reports instead of wedging the runner, and both panicking with an opaque payload so a retained original is distinguishable from the join's `String` stand-in. The first blocks inside a hostile finished waiter and checks that reclamation cannot retire the work until the exact caller panic is retained. The second uses a real `ActorWork` on a multi-threaded runtime to exercise the chain the fix protects in production: entry retained mid-wake → `join_offloads` awaits the task → the original payload reaches the post-join resource take.

Validation: `./tools/dev just ci` (805 tests), after rebasing onto current `main`.

### DIFF
--- a/crates/shelterwood/src/raw/context.rs
+++ b/crates/shelterwood/src/raw/context.rs
@@ -2036,6 +2036,60 @@ mod tests {
         assert!(!resources.offloads[0].finished.is_fired());
     }
 
+    #[crate::runtime::test]
+    async fn ordinary_offload_completion_retains_a_finished_waiter_wake_panic() {
+        let mut resources = RawResources::<()>::default();
+        let mut panic_signal = resources.disposal.signal.watcher();
+        let finished = Latch::default();
+        let mut waiter = Box::pin(finished.fired());
+        let hostile = Waker::from(Arc::new(PanickingWake("ordinary finished wake panic")));
+        assert!(
+            waiter
+                .as_mut()
+                .poll(&mut Context::from_waker(&hostile))
+                .is_pending()
+        );
+
+        let state = SharedOffloadState::new(
+            Box::pin(async {}),
+            resources.disposal.clone(),
+            finished.clone(),
+        );
+        resources.offloads.push(OffloadResource {
+            cancellation: Latch::default(),
+            finished: finished.clone(),
+            state: Some(Arc::clone(&state)),
+            task: None,
+        });
+        let mut work = SharedOffloadFuture(state);
+        assert!(
+            Pin::new(&mut work)
+                .poll(&mut Context::from_waker(Waker::noop()))
+                .is_ready(),
+            "a caller wake panic does not escape through the offload task"
+        );
+
+        assert!(finished.is_fired(), "all completion waiters were released");
+        resources.reclaim_finished();
+        assert!(
+            resources.offloads.is_empty(),
+            "ledger reclamation cannot discard the retained diagnostic"
+        );
+        assert!(matches!(
+            crate::runtime::timeout(Duration::from_secs(1), panic_signal.changed()).await,
+            crate::runtime::Timeout::Completed(())
+        ));
+        let payload = resources
+            .disposal
+            .panic
+            .take()
+            .expect("the caller wake panic survives ledger reclamation");
+        assert_eq!(
+            panic_message(&payload),
+            Some("ordinary finished wake panic")
+        );
+    }
+
     /// `freeze` is the drain that normally empties the continuation queue, but
     /// it is not a guaranteed one: `Drop for RawResources` skips it once the
     /// incarnation is already frozen, and runs it under `catch_panic` so an
--- a/crates/shelterwood/src/raw/context.rs
+++ b/crates/shelterwood/src/raw/context.rs
@@ -740,7 +740,13 @@ impl<M> RawResources<M> {
     /// count, so a per-turn walk of it is proportional to work the actor
     /// itself has outstanding.
     fn reclaim_finished(&mut self) {
-        self.offloads.retain(|offload| !offload.finished.is_fired());
+        self.offloads.retain(|offload| {
+            if let Some(state) = &offload.state {
+                !state.finished_published()
+            } else {
+                !offload.finished.is_fired()
+            }
+        });
     }
 
     fn resume_pending_panic(&self) {
@@ -1736,6 +1742,12 @@ mod tests {
 
     struct PanickingWake(&'static str);
 
+    struct BlockingPanickingWake {
+        entered: Arc<Barrier>,
+        release: Arc<Barrier>,
+        message: &'static str,
+    }
+
     impl Wake for PanickingWake {
         fn wake(self: Arc<Self>) {
             panic!("{}", self.0);
@@ -1743,6 +1755,24 @@ mod tests {
 
         fn wake_by_ref(self: &Arc<Self>) {
             panic!("{}", self.0);
+        }
+    }
+
+    impl BlockingPanickingWake {
+        fn block_then_panic(&self) {
+            self.entered.wait();
+            self.release.wait();
+            panic!("{}", self.message);
+        }
+    }
+
+    impl Wake for BlockingPanickingWake {
+        fn wake(self: Arc<Self>) {
+            self.block_then_panic();
+        }
+
+        fn wake_by_ref(self: &Arc<Self>) {
+            self.block_then_panic();
         }
     }
 
@@ -2036,13 +2066,19 @@ mod tests {
         assert!(!resources.offloads[0].finished.is_fired());
     }
 
-    #[crate::runtime::test]
-    async fn ordinary_offload_completion_retains_a_finished_waiter_wake_panic() {
+    #[test]
+    fn ordinary_offload_completion_retains_a_finished_waiter_wake_panic() {
         let mut resources = RawResources::<()>::default();
-        let mut panic_signal = resources.disposal.signal.watcher();
+        let panic = Arc::clone(&resources.disposal.panic);
         let finished = Latch::default();
         let mut waiter = Box::pin(finished.fired());
-        let hostile = Waker::from(Arc::new(PanickingWake("ordinary finished wake panic")));
+        let entered = Arc::new(Barrier::new(2));
+        let release = Arc::new(Barrier::new(2));
+        let hostile = Waker::from(Arc::new(BlockingPanickingWake {
+            entered: Arc::clone(&entered),
+            release: Arc::clone(&release),
+            message: "ordinary finished wake panic",
+        }));
         assert!(
             waiter
                 .as_mut()
@@ -2061,27 +2097,36 @@ mod tests {
             state: Some(Arc::clone(&state)),
             task: None,
         });
-        let mut work = SharedOffloadFuture(state);
-        assert!(
-            Pin::new(&mut work)
-                .poll(&mut Context::from_waker(Waker::noop()))
-                .is_ready(),
-            "a caller wake panic does not escape through the offload task"
-        );
+        let poller = thread::spawn(move || {
+            let mut work = SharedOffloadFuture(state);
+            assert!(
+                Pin::new(&mut work)
+                    .poll(&mut Context::from_waker(Waker::noop()))
+                    .is_ready(),
+                "a caller wake panic does not escape through the offload task"
+            );
+        });
+        entered.wait();
 
         assert!(finished.is_fired(), "all completion waiters were released");
         resources.reclaim_finished();
         assert!(
-            resources.offloads.is_empty(),
-            "ledger reclamation cannot discard the retained diagnostic"
+            !resources.offloads.is_empty(),
+            "the fired bit cannot retire work before its hostile wake is contained"
         );
-        assert!(matches!(
-            crate::runtime::timeout(Duration::from_secs(1), panic_signal.changed()).await,
-            crate::runtime::Timeout::Completed(())
-        ));
-        let payload = resources
-            .disposal
-            .panic
+        assert!(
+            panic.take().is_none(),
+            "an actor turn can precede the blocked wake's panic"
+        );
+
+        release.wait();
+        poller.join().expect("the caller wake panic is contained");
+        resources.reclaim_finished();
+        assert!(
+            resources.offloads.is_empty(),
+            "the ledger retires work after notification publication"
+        );
+        let payload = panic
             .take()
             .expect("the caller wake panic survives ledger reclamation");
         assert_eq!(
--- a/crates/shelterwood/src/raw/context.rs
+++ b/crates/shelterwood/src/raw/context.rs
@@ -1787,7 +1787,7 @@ mod tests {
     /// Generous enough to survive a loaded shared machine, finite so that a
     /// mutation which suppresses the fire under test reports instead of
     /// wedging the runner.
-    const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(30);
+    const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(10);
 
     /// A single-use, timeout-bounded signal between two threads.
     ///
@@ -2207,18 +2207,25 @@ mod tests {
         });
         entered.wait("the completion wake to reach the hostile waiter");
 
-        assert!(finished.is_fired(), "all completion waiters were released");
+        // Sample the mid-wake state, then release before asserting on it: an
+        // assertion that fires while the wake is still parked would strand the
+        // peer on its timeout and report the expiry instead of the mismatch.
+        let fired_mid_wake = finished.is_fired();
         resources.reclaim_finished();
-        assert!(
-            !resources.offloads.is_empty(),
+        let pinned_mid_wake = resources.offloads.len();
+        let recorded_mid_wake = panic.take();
+        release.open();
+
+        assert!(fired_mid_wake, "all completion waiters were released");
+        assert_eq!(
+            pinned_mid_wake, 1,
             "the fired bit cannot retire work before its hostile wake is contained"
         );
         assert!(
-            panic.take().is_none(),
+            recorded_mid_wake.is_none(),
             "an actor turn can precede the blocked wake's panic"
         );
 
-        release.open();
         poller.join().expect("the caller wake panic is contained");
         resources.reclaim_finished();
         assert!(
@@ -2275,15 +2282,17 @@ mod tests {
         });
 
         entered.wait("the completion wake to reach the hostile waiter");
-        assert!(finished.is_fired(), "all completion waiters were released");
+        let fired_mid_wake = finished.is_fired();
         resources.reclaim_finished();
+        let pinned_mid_wake = resources.offloads.len();
+        release.open();
+
+        assert!(fired_mid_wake, "all completion waiters were released");
         assert_eq!(
-            resources.offloads.len(),
-            1,
+            pinned_mid_wake, 1,
             "the ledger keeps the task handle teardown must join while the wake runs"
         );
 
-        release.open();
         resources.join_offloads().await;
         assert!(
             resources.offloads.is_empty(),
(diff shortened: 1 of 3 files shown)


ralexstokes:
**Verdict: the publication edge is sound — I found no lost wake, strand, ABA, double-publication, or insufficient ordering — but the PR ships a new user-observable retention class with no doc or SPEC text, and one of its stated invariants is falsifiable by a real interleaving.**

Protocol audit (for the record, so the next reader does not have to redo it). I enumerated: concurrent task-completion vs. `freeze`-driven `OffloadResource::cancel`; loser-returns-early; second `fire` as a no-op; `take_for_poll` early-return after `cancelled`; reclamation observing the edge; panic-slot read racing the wake; `Guard::drop`/`Blocking::drop` cancellation; deadline expiry; abort-mid-poll. The orderings are adequate: the `Release` store at `offload.rs:234-235` is the only thing that must be released, `record()`'s slot mutex and signal pulse precede it, and `finished_published()`'s `Acquire` load at `offload.rs:239` is the matching consumer. The state byte is monotone `0 → 1 → 2`, so there is no ABA. `WaiterRegistry::wake_all` already runs each waker behind a `PanicAccumulator`, so a hostile waiter cannot strand its siblings and `catch_panic(|| finished.fire())` sees exactly the accumulator's first payload. Ledger retention does not perturb arbitration (`ReadyBatch::steady` takes `events.watermark()`, not `offloads.len()`). `reclaim_finished` is unreachable between `freeze()` and `join_offloads()` on every path (`recv`/`try_recv` stop branches and the `definition.rs:270-277` epilogue), so the retained entry always reaches the join that fences the late `record`.

---

### 1. [Medium] The new ordinary-path retention class is undocumented, and contradicts the `recv`/`try_recv` rustdoc

`crates/shelterwood/src/raw/context.rs:1111-1116` (and the twin at `1170-1178`) enumerate what may be retained at a receive boundary and scope the waker class explicitly to teardown:

> an offload-work panic, and — **because the stop branches freeze first** — … a waker panic from the `Guard::finished()` waiters and cancellation latches that cancelling those offloads wakes

After this PR that attribution is wrong. A `Guard::finished()` waiter's wake panic is now recorded on the *ordinary* completion path (`fire_finished` from all three `SharedOffloadFuture::poll` arms), with no freeze involved, and `next_ready` resumes it at `context.rs:1365` (`reclaim_finished` → `resume_pending_panic`) on a fully live incarnation.

The consequence is user-visible and, as far as I can tell, new: **an incarnation now fails because a third party's waker panicked while awaiting `Guard::finished()`**, and per SPEC §6.2 that failure suppresses `on_stop`. SPEC §6.2's definition of the funnel — "the offloaded future and its continuation closure, plus the queued continuations, armed timers and queued offload completions released at the intake freeze" — does not contain a completion-waiter waker, and SPEC §6.5's `Guard::is_finished`/`finished()` bullet only orders the *work* payload against the notification, saying nothing about the notification's own wake. §6.5 also currently says panics that reach the actor are "in the offloaded future or the continuation".

This is the same shape as the batch-1 items that each carried a SPEC sentence (#373 → §5.5). Remedy: reword the em-dash clause in both rustdocs so the finished-waiter waker class is listed for the ordinary path too, add the class to SPEC §6.2's funnel enumeration, and extend §6.5's `Guard` bullet to say the notification's own waker panic is retained as incarnation-owned disposal. Worth adding a line to `Guard::finished`'s own rustdoc (`offload.rs:66-69`) as well — a caller who hands `Guard` to a task with a panicking waker currently has no way to know it kills the actor. CONFIRMED (doc/spec contradiction; the code behavior is the one #366 asked for).

### 2. [Low–Medium] Teardown now unconditionally joins a completion whose wake is in flight

Before this PR, `reclaim_finished` retired an entry as soon as the fired bit was visible — which `Latch::fire` sets *before* `notify()` — so an offload blocked inside a caller waker could have its `ActorWork` handle dropped and never be joined. After the fix the entry is pinned until `FINISHED_PUBLISHED`, so `join_offloads` (`context.rs:756-770`) always awaits it.

Interleaving: a caller awaits `Guard::finished()` with a waker whose `wake()` blocks (does not panic) — exactly the `BlockingPanickingWake` fixture minus the panic. The offload task sits inside `wake_all`; the actor reaches teardown; `state.cancel()` loses the CAS and returns, `task.abort()` cannot preempt a running poll, and `join_offloads` awaits it. Incarnation teardown, and therefore exit publication and supervision, are now stalled by caller code for as long as the waker blocks. Previously teardown could sail past.

I do not think this should change the design — retiring early is precisely the bug being fixed, and on a current-thread runtime the executor was already blocked. But SPEC §6.2 says in terms "Retention is the guarantee, **not a join**", and this quietly moves one case from not-a-join to always-joined. Record it as an accepted trade in the same SPEC/doc edit as finding 1, rather than leaving the next reviewer to rediscover it. CONFIRMED as a behavior change; PLAUSIBLE as something anyone will hit.

### 3. [Low] The publication invariant stated in the comment and the PR body is falsifiable

`offload.rs:231-235`:

> Release-publish only after a hostile wake has either returned or had its panic installed in the incarnation slot.

and the PR body's "The winning notifier publishes completion only after every hostile wake has returned or its panic has been recorded."

Counter-interleaving, using the second fire site at `offload.rs:332-334`:

1. Offload task: `fire_finished` wins the CAS → `PUBLISHING`, then is descheduled *before* `self.finished.fire()`.
2. Actor (`freeze` → `OffloadResource::cancel`): `state.cancel()` → `fire_finished` → CAS fails → returns without firing. `dispose`, `task.abort()`.
3. Actor: trailing `panics.run(|| self.finished.fire())` — `fire_silently` still wins here, so **the actor thread** runs `wake_all` and enters the hostile wake.
4. Offload task resumes: `finished.fire()` returns `false` (already fired), no notify, no panic, and it stores `FINISHED_PUBLISHED` while the hostile wake is still executing on the actor thread.

So the publication-CAS winner can publish with a wake in progress. It is benign *today* for one non-obvious reason: the in-progress wake is on the actor thread itself, and `reclaim_finished` is unreachable from there (it runs on that same thread, and never between `freeze` and `join_offloads` anyway). That is a much narrower safety argument than the comment asserts, and it depends on `OffloadResource::cancel` keeping its trailing fire on the actor thread. Remedy: either weaken the comment to state the actual invariant ("the ledger cannot retire the entry while the publishing *thread* is inside the wake"), or remove the now-redundant trailing `self.finished.fire()` in the `state.is_some()` case (see finding 5) so the edge really is the only notifier. CONFIRMED.

### 4. [Low] Stale comment and a now-dead capture in `OffloadResource::cancel`

`offload.rs:316-321`:

```rust
// `state.cancel` disposes the future before firing `finished`.
// Pull that contained destructor failure into the accumulator
// before recording a later wake panic escaping the call.
let state_panic = catch_panic(|| state.cancel()).err();
```

No wake panic escapes `state.cancel()` any more — `fire_finished` contains it and routes it to the slot, where `PanicSlot::record`'s first-wins policy discards it if the disposal panic got there first. The net precedence is unchanged (good), but the comment now describes a path that cannot happen, and `state_panic` can only ever be a poisoned-mutex `.expect`. Update the comment to say the wake panic now arrives via `retained.take()`, and note that first-wins discard is where the loser goes.

### 5. [Low] The three-state byte partly duplicates a primitive the crate already has, and the fork it forces is avoidable

`Latch::fire_silently` exists for exactly this ("linearize the fire … defer the waker-visible `notify`", `shelterwood-runtime/src/sync.rs:217-230`) and already returns a unique winner. `fire_finished` adds a second claim CAS over the same transition, and the two claims can disagree (finding 3 is precisely that divergence). The minimal shape is `if !self.finished.fire_silently() { return; }` + `catch_panic(|| self.finished.notify())` + one `AtomicBool` `Release`-stored by the winner — one atomic instead of two claims, and no possibility of divergence. The wrinkle is `offload.rs:332-334`, which would then be able to win the claim without publishing; but that trailing fire is already dead whenever `state.is_some()` (`state.cancel()` always calls `fire_finished` first), so it can be moved under `if self.state.is_none()`.

Collapsing the claim would also let `reclaim_finished` (`context.rs:742-750`) drop its two-branch fork. That fork exists only because the zero-deadline short-circuit at `context.rs:1264` builds an `OffloadResource` with `state: None`, and it is the one remaining uncontained `finished.fire()` in the file. It happens to be safe — the `Latch` is local and the `Guard` has not been returned, so the registry is provably empty — but that is worth a one-line comment, since nothing structural enforces it.

Also: the CAS at `offload.rs:214-221` uses `AcqRel`/`Acquire`, but the winner reads nothing published by another thread and the loser does nothing at all, so `Relaxed`/`Relaxed` is sufficient. Not a bug, just noise on a line where a reader will stop to reason.

### 6. [Low] Test notes — non-vacuous, but it can hang instead of fail, and it skips the join interaction

I mutation-checked `ordinary_offload_completion_retains_a_finished_waiter_wake_panic` and it is genuinely load-bearing, which is worth stating explicitly given this repo's history: reverting `reclaim_finished` to `!offload.finished.is_fired()` fails the `!resources.offloads.is_empty()` assertion (the fired bit is already set at that point), and removing the `catch_panic` in `fire_finished` fails `poller.join().expect(...)`. All assertions are in the test thread, none inside a handler or `catch_unwind`, so nothing is swallowed. Three caveats:

- Mutations that stop the latch firing at all (e.g. dropping the `finished.fire()` call) make the main thread block forever on `entered.wait()` — the test **hangs** rather than fails. A `Barrier` with no timeout is a fail-slow oracle. Consider a bounded handshake so the regression reports rather than wedges the runner.
- The fixture uses `task: None`, so it does not exercise the interaction the fix actually protects in production: entry retained → `join_offloads` awaits the task → `take_resource_panic` after the join. Given finding 2, a companion test with a real `ActorWork` would be more valuable than the `reclaim_finished` assertion alone.
- Nothing here can fail if `Ordering::Release` at `offload.rs:235` is weakened to `Relaxed` — `poller.join()` is a full fence. That is expected for an ordering, but it means the comment is the only thing pinning it; say so.

---

**chatgpt-codex-connector, P1 "Retain the offload until finished notification completes" (`offload.rs:229`) — CONFIRMED against the commit it reviewed, and already fixed in this PR.** The comment was filed against `418f0820` (`original_line: 204`), i.e. the first commit, where `fire_finished` was bare `catch_panic` with no publication edge. The described race is real for that commit: `Latch::fire` sets the fired bit in `fire_silently` before `notify()`, so `reclaim_finished`'s `!offload.finished.is_fired()` predicate could retire the entry — dropping the `ActorWork` handle that `join_offloads` needs — while the hostile wake was still running, after which `record()` would land in a slot nobody reads again. Commit `62e4c576` is exactly the remedy it prescribed ("keep a separate completion/publication state"), and I verified above that the resulting edge closes it: the entry survives to `join_offloads`, and `take_resource_panic` at `definition.rs:277` runs after that join. Severity and remedy were both framed correctly this time. The one residue is finding 3 — the edge is slightly weaker than its own comment claims — and the hard-abort path (`Drop for RawResources` runs `freeze` without a join), which SPEC §6.2 already carves out as explicitly not-a-join, so it is out of scope here.

No other review comments on the PR. No fabricated spec citations to correct — Codex quoted no spec text.

ralexstokes:
All six findings addressed in `d0e5a2b`, `18811d8`, `3ef71cb`. Finding 5 taken, in a corrected form — one part of its proposed remedy is refuted below. Thank you for the interleaving enumeration; it is what made taking 5 tractable.

### 1. [Medium] Undocumented ordinary-path retention class — **fixed**

Confirmed and the attribution was wrong exactly as described. Four places now carry it:

- `raw/context.rs` — `recv`'s retention list moves the `Guard::finished()` waiter class out of the "because the stop branches freeze first" clause and states the consequence in terms: *"a third party awaiting `Guard::finished()` with a panicking waker fails a live incarnation here, and per SPEC §6.2 a failed incarnation skips `on_stop`."* The cancellation-latch and teardown-notification wakes stay in the freeze-only clause, where they belong. The `try_recv` twin gets the same split across its outside-drain / during-drain halves.
- SPEC §6.2 — the funnel enumeration now ends *"…plus the wakers of `Guard::finished()`'s own completion waiters, which are caller-owned but run by the incarnation on the ordinary completion path as well as at the freeze"*.
- SPEC §6.5 — the `Guard` bullet is extended: *"The notification's own wake belongs to §6.2's funnel too: a completion waiter whose waker panics has that payload retained as incarnation-owned disposal — on the ordinary completion path, not only at the freeze — so a third party awaiting `finished()` can fail the incarnation and suppress its `on_stop` … Callers are expected to await `finished()` from a task whose waker neither panics nor blocks."*
- `Guard::finished`'s own rustdoc, which is the surface a caller actually reads before handing the guard to a task.

### 2. [Low–Medium] Teardown now unconditionally joins an in-flight completion — **recorded as an accepted trade**

Confirmed; design unchanged, as you suggest. Recorded in three places rather than left for the next reviewer:

- SPEC §6.2, immediately after the "retention is the guarantee, not a join" sentence it qualifies: *"The completion notification of §6.5 is the one carve-out in the other direction: because its wake panic MUST be retained before the framework may forget the work, a completion waiter whose waker *blocks* rather than panicking holds that work in the incarnation's resource ledger, and teardown joins it — so caller code can delay teardown and exit publication there."*
- SPEC §6.5 and `Guard::finished`, in the same breath as the panic class.
- A new doc comment on `join_offloads` itself, where the stall materializes.

### 3. [Low] The publication invariant was falsifiable — **fixed by removing the divergence, not by weakening the claim**

Your counter-interleaving is right, and the second claim CAS was its whole cause. Rather than restate the narrower argument, the divergence is gone: `fire_finished` now claims the **latch's own** transition.

```rust
fn fire_finished(&self) {
    if !self.finished.fire_silently() {
        return;
    }
    if let Err(payload) = catch_panic(|| self.finished.notify()) {
        self.record(payload);
    }
    self.finished_published.store(true, Ordering::Release);
}
```

There is now exactly one claim, so the wake-running thread and the publishing thread are the same thread by construction. Your step 3 cannot occur: the trailing `self.finished.fire()` in `OffloadResource::cancel` finds the latch already claimed and does nothing, so the hostile wake never migrates to the actor thread. The comment's invariant is now literally true and does not depend on `reclaim_finished`'s reachability.

### 4. [Low] Stale comment / dead capture in `OffloadResource::cancel` — **fixed**

Rewritten to describe the path that exists: *"`state.cancel` disposes the future and then contains its own completion wake, so neither a destructor panic nor a caller waker panic escapes the call: both arrive through the retained slot, where `PanicSlot`'s first-wins policy discards the loser. `state_panic` can therefore only be a poisoned-mutex `expect`."*

### 5. [Low] Simplification — **taken, minus the trailing-fire gate, which I believe is wrong**

Taken: `AtomicU8` + three constants + a second CAS → `fire_silently` + one `AtomicBool`. The `AcqRel`/`Acquire` noise dissolves with the CAS. Net −8 lines and one fewer atomic on the completion path.

I re-derived correctness rather than inheriting yours, since this touches the protocol you had just certified. Enumerated: task completion vs. `freeze`-driven `OffloadResource::cancel` in both orders; loser-returns-without-publishing; the `Pending`-arm cancel-observed-on-return path; `take_for_poll`'s early return after `cancelled`; guard-drop and `Blocking`-drop cancellation (both fire `cancellation`, and the guard-drop case still reaches `fire_finished` through the operation's `select_two` Left arm); deadline expiry; `task.abort()` mid-wake (Tokio cannot preempt a running poll, so the wake completes); and the panic-slot read racing the wake (`PanicSlot`'s mutex plus the `Release`/`Acquire` pair on the flag). Publication is stored by one thread, after that thread's `notify` returned or had its payload recorded, and `finished_published` is loaded `Acquire` before any retirement.

**Refuted: moving the trailing fire under `if self.state.is_none()`.** It is not dead when `state.is_some()`. `SharedOffloadState::cancel` reaches its own `fire_finished` only after `self.state.lock().expect(…)` and `self.dispose(future)`, and `dispose` records through `PanicSlot::record`, which also `.expect()`s. `catch_panic(|| state.cancel())` returning `Err` therefore means `fire_finished` never ran — and gating the trailing fire would leave every `Guard::finished()` waiter unwoken forever on that path. Only a poisoned mutex gets there, which is close to unreachable given the lock rule, but the fallback is free and the gate buys nothing now that the single claim has already removed finding 3. The unconditional fire now carries a comment saying both halves of that: no-op in the common case (which is what keeps the wake off the actor thread), sole notifier in the poisoned case.

**Also declined: collapsing `reclaim_finished`'s two-branch fork.** It exists because the zero-deadline short-circuit builds `state: None`, and no amount of claim collapsing changes that — there is no `SharedOffloadState` to hold the flag. Both branches now carry the one-line rationale you asked for, including why the zero-deadline `finished.fire()` at `context.rs` is safe (it fires while `guard` is still on its way back to the caller, so the waiter registry is provably empty) and a note to keep that fire ahead of the `return Ok(guard)`.

### 6. [Low] Test notes — **all three addressed**

- **Fail-slow oracle.** `Barrier` is gone. A small `Gate` (`sync_channel(1)` + `recv_timeout`, 10s) replaces it, and both tests now *sample* the mid-wake state, release the peer, and only then assert — so an assertion failure cannot strand the peer and turn a mismatch report into a timeout report. Mutation-checked: deleting the `notify` call makes both tests **fail in 10.00s** with `timed out waiting for the completion wake to reach the hostile waiter`, where before they would have hung.
- **`task: None`.** New companion `a_pinned_completion_wake_panic_survives_the_offload_task_join`, `#[crate::runtime::test(flavor = "multi_thread", worker_threads = 2)]`, with a real `spawn_actor_work(SharedOffloadFuture(state))`. It blocks in the hostile waker, asserts the ledger still holds the entry (`len == 1`) while the wake is in flight, releases, `join_offloads().await`s, and takes the payload from the slot afterwards — the exact retain → join → `take_resource_panic` chain.
- **Distinguishing the payload.** Both wakers now `panic_any(OpaqueWakePanic(..))` instead of `panic!("{}", ..)`. That matters: `panic!("{}", x)` produces a `String`, which is bit-for-bit what `join_offloads` would have recorded from `JoinOutcome::Panic { message }` — so the old `panic_message(&payload) == Some("…")` assertion could not have told a retained original from the join's stand-in. With an opaque payload it can, and does.
- **The `Release` ordering.** Still unpinned by any test — `poller.join()` and the runtime join are both full fences. Said so explicitly, in the comment at the store (*"No test pins that pairing — only this comment does"*) and in the first test's doc comment.

Mutation checks, on the committed tree, reverted after each:

| mutation | result |
|---|---|
| `reclaim_finished` also retires on `finished.is_fired()` | both tests fail in 0.004s (`left: 0, right: 1`) |
| `notify()` outside `catch_panic` | both fail; companion reports `left: None, right: Some(OpaqueWakePanic(..))` |
| `notify()` + containment deleted entirely | both fail in 10.00s with the handshake timeout, not a hang |

Validation: `./tools/dev just ci` green, 805 tests. PR body updated — the old "every hostile wake has returned or its panic has been recorded" claim was the overstated one you flagged, and it is now true rather than reworded.

PR #380, opened 2026-08-20
Harden observation transaction failure paths

ralexstokes:
Part of #366: the consolidated Batch 3 hardening pass.

This closes the three internal-invariant gaps without changing supported-path behavior:

- creates `ObservationTxn`'s `PanicAccumulator` before snapshot installation, contains each install independently, transfers every attempted publication to post-unlock disposal, and drains later cuts/effects before resuming the first panic
- retains user-bearing by-value inputs before fallible bookkeeping at restart scheduling, supervised child terminalization, scope finish, and resident admission
- routes pruned and cleared resident graphs through `dispose_detached`, making final member/mailbox teardown explicitly asynchronous, with per-element panic containment on `ResidentChild::drop` so SPEC §5.5's containment clause holds at every depth rather than once around the displaced `Vec`

Fault-injection regressions cover a panicking staged install, illegal restart and residency transitions, a poisoned scope-finish mutex, a rejected last-owner admission, and two hostile resident payload destructors in one displaced set. A blocking mailbox-payload destructor regression proves resident clearing returns before the destructor is released and that destruction runs on the detached lane. Existing immediate-release coverage was updated for the asynchronous contract, and `AGENTS.md` now records every detached-disposal route including `admit_child_locked`'s.

The three regressions that depend on a `debug_assert!` firing are gated on `#[cfg(debug_assertions)]`; without the gate they fail (not merely go inert) under `--release`. See #393.

Deferred to issues: #389 (SPEC §15.5's structural `Removed`-on-drop is unreachable from a disposal worker), #390 (let residency own the projection so `ResidentAdmission` can be deleted), #393 (no release-profile test lane).

Validation: `./tools/dev just ci`, plus `cargo nextest run --release --workspace` (803/803).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,10 +60,19 @@ rests on:
   traffic and only the last clone submits disposal. Scope state and startup
   results need that protection too: a structured startup failure recursively
   owns the triggering child's `Exit`.
-  `ScopeCell::clear_residents_locked` and `prune_child_locked` route displaced
-  `Arc<MemberCell>`s through `runtime::dispose_detached` after unlock: the last
-  member owner can also be the last owner of a mailbox containing unread user
-  messages, which `RetainedExit` does not cover.
+  `ScopeCell::clear_residents_locked`, `prune_child_locked` and
+  `admit_child_locked` route `Arc<MemberCell>`s through
+  `runtime::dispose_detached` after unlock: the last member owner can also be
+  the last owner of a mailbox containing unread user messages, which
+  `RetainedExit` does not cover. The first two displace an existing resident;
+  the third holds the *incoming* projection in a `ResidentAdmission` guard, so
+  a bookkeeping panic before installation retires the same graph the same way.
+  Containment lives on `ResidentChild::drop` rather than on the displaced
+  `Vec`, because `Vec`'s slice drop glue keeps going after an element panics
+  and a second hostile mailbox destructor would then panic inside the first
+  one's unwind. Per-element containment therefore holds at every depth of a
+  nested residency and on `ScopeCell`'s own drop glue, which is what SPEC §5.5
+  asks of this lane.
 - **Framework `dyn` seams.** `MailboxControl`, `MailboxTermination`,
   `MailboxRuntime`, `MailboxEffectSink`, `ActorIdentity` and `DynamicRoute` are
   implementation seams
@@ -111,8 +120,8 @@ caller, so compute the verdict, release, *then* panic (`MailboxCell::bind` is
 the pattern; where releasing is impossible, `debug_assert!` instead, as
 `MailboxState::take_waiters` does). And a value that may block on destruction
 goes to `runtime::dispose_detached`, not merely past the unlock. That second
-convention is currently met for mailbox payloads, construction closures and
-offloads. Framework-retained `Exit` copies meet it through `RetainedExit`,
+convention is currently met for mailbox payloads, construction closures,
+offloads, displaced resident graphs and rejected admission projections. Framework-retained `Exit` copies meet it through `RetainedExit`,
 including driver completions and pending terminal disposal; exits handed to
 users keep ordinary drop timing. Its fail-safe under exhausted thread
 creation is an unreclaimed queued job — memory held for the life of the
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,10 +60,10 @@ rests on:
   traffic and only the last clone submits disposal. Scope state and startup
   results need that protection too: a structured startup failure recursively
   owns the triggering child's `Exit`.
-  `ScopeCell::clear_residents_locked` and `prune_child_locked` still defer
-  displaced `Arc<MemberCell>`s: the last member owner can also be the last
-  owner of a mailbox containing unread user messages, which `RetainedExit`
-  does not cover.
+  `ScopeCell::clear_residents_locked` and `prune_child_locked` route displaced
+  `Arc<MemberCell>`s through `runtime::dispose_detached` after unlock: the last
+  member owner can also be the last owner of a mailbox containing unread user
+  messages, which `RetainedExit` does not cover.
 - **Framework `dyn` seams.** `MailboxControl`, `MailboxTermination`,
   `MailboxRuntime`, `MailboxEffectSink`, `ActorIdentity` and `DynamicRoute` are
   implementation seams
--- a/crates/shelterwood-cells/src/cells/gate.rs
+++ b/crates/shelterwood-cells/src/cells/gate.rs
@@ -117,21 +117,24 @@ impl<'a> ObservationTxn<'a> {
     }
 
     fn commit(&mut self) {
+        let mut panics = runtime::PanicAccumulator::default();
         // Installation precedes the unlock, and the order is load-bearing:
         // released first, two transactions on one gate could interleave as
         // "T1 unlocks, T2 stages and installs a newer cut, T1 installs its
         // stale one", leaving every ungated borrow behind the tree until some
         // later publication corrected it. SPEC §14 promises this ordering.
-        // This install segment must remain panic-free apart from unreachable
-        // invariant assertions. A panic raised here runs from `Drop`, escapes
-        // before the post-unlock accumulator, and would strand later snapshot
-        // installations plus every queued effect; the current path invokes no
-        // user code and defers generation exhaustion into `effects`.
-        for publication in self.snapshots.drain(..) {
-            publication.install(&mut self.effects);
+        // Installations invoke no user code, but keep even their unreachable
+        // invariant assertions inside the accumulator. A broken installation
+        // must not strand the remaining committed cuts or their queued effects,
+        // and its panic resumes only after the guard is gone.
+        for mut publication in self.snapshots.drain(..) {
+            panics.run(|| publication.install(&mut self.effects));
+            // The projection captures its publishing scope. Transfer the
+            // whole attempted publication to the post-unlock list whether
+            // installation succeeded or panicked.
+            self.effects.push(Box::new(move || drop(publication)));
         }
         drop(self.guard.take());
-        let mut panics = runtime::PanicAccumulator::default();
         for effect in self.effects.drain(..) {
             // One hostile waker must not prevent the remaining committed
             // observation edges from notifying their waiters.
--- a/crates/shelterwood-cells/src/cells/member.rs
+++ b/crates/shelterwood-cells/src/cells/member.rs
@@ -751,6 +751,11 @@ mod tests {
     use super::*;
     use crate::cells::test_support::{ThreadProbe, isolated_scope};
 
+    // The retention this pins only has an observable effect when the
+    // framework invariant it guards is checked, and that check is a
+    // `debug_assert!`. Release builds decline the panic entirely, so the test
+    // is gated on the profile it can hold in rather than left to fail there.
+    #[cfg(debug_assertions)]
     #[test]
     fn illegal_restart_transition_retires_its_user_error_off_thread() {
         let scope = isolated_scope("root", ScopeFlavor::Ordered);
--- a/crates/shelterwood-cells/src/cells/member.rs
+++ b/crates/shelterwood-cells/src/cells/member.rs
@@ -528,7 +528,23 @@ impl MemberCell {
     }
 
     pub fn transition_locked(&self, txn: &mut ObservationTxn<'_>, transition: MemberTransition) {
+        // `RestartScheduled` carries a user error by value. Retain it before
+        // the watch lock and reducer assertions so an invariant panic can
+        // unwind the raw transition as refcount traffic only.
+        let retained_exit = match &transition {
+            MemberTransition::RestartScheduled { exit, .. } => {
+                Some(RetainedExit::new(exit.clone()))
+            }
+            MemberTransition::Admitted
+            | MemberTransition::Starting { .. }
+            | MemberTransition::Running
+            | MemberTransition::Stopping => None,
+        };
         self.update_locked(txn, |record| record.apply_transition(transition));
+        if let Some(retained_exit) = retained_exit {
+            // The record now owns an equivalent retained copy.
+            drop(retained_exit.into_exit());
+        }
     }
 
     pub fn set_options(&self, options: ResolvedCommonOptions) {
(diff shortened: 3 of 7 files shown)


ralexstokes:
**Verdict: no correctness defects found. The three claims hold; findings below are hardening/doc gaps, none blocking.**

## Findings

### 1. `clear_residents_locked` / `prune_child_locked` submit the whole graph as one disposal job — no per-element containment (Low, CONFIRMED mechanism / PLAUSIBLE reachability)

`crates/shelterwood-cells/src/cells/scope.rs:1187` (`dispose_detached(residents)` for a whole `Vec<ResidentChild>`) and `:1157`.

`Vec`'s slice drop glue keeps dropping the remaining elements when one element's destructor panics, so a *second* hostile mailbox-payload destructor in the same displaced set panics during that unwind → process abort. `DisposalJob::finish` wraps the whole `Vec` in a single `catch_panic`, which does not help: the second panic happens inside the first one's unwind, below the boundary.

This is the same shape as on `main` (`drop(residents)` inside `panics.run`), so it is **not a regression** — but the PR is the moment this line is being re-authored, the batch item is explicitly about closing the `Arc`-graph tail, and SPEC §5.5 says framework-initiated disposal of externally submitted mailbox payloads runs "detached from the initiating task **with per-element panic containment**". One job for N residents does not deliver that.

Remedy: either `runtime::dispose_all(residents)` (already exists; each value gets its own unwind boundary — costs N jobs), or, to keep one job, a newtype whose `Drop` loops `discard_panic(catch_panic(|| drop(child)).err())` over the vec. The same applies to `prune_child_locked`, though there N = 1 and only the nested subtree can multiply it.

### 2. `AGENTS.md` under-describes what the code now does (Low, CONFIRMED)

Two gaps in the recorded contract:

- `AGENTS.md:118-122` (the closing convention paragraph) still reads "a value that may block on destruction goes to `runtime::dispose_detached` … That second convention is currently met for mailbox payloads, construction closures and offloads." After this PR it is also met for displaced resident graphs — the exact enumeration the PR just changed one bullet above.
- The PR adds a **third** `dispose_detached` route, `ResidentAdmission` in `admit_child_locked` (`scope.rs:66-72`), which no line of `AGENTS.md` mentions. The updated bullet names only `clear_residents_locked` and `prune_child_locked`.

### 3. Detached residency teardown forecloses SPEC §15.5's structural `Removed` (Low, CONFIRMED as a doc/design tension)

SPEC §15.5 states: "Residency in a scope's observed child set is itself such a value — **its drop emits `Removed`**, making §3.2's exact pairing structural rather than remembered." Today `ResidentChild` has no `Drop` impl at all and `Removed` is emitted explicitly by `emit_locked`, so nothing breaks *now*. But routing the displaced set to a disposal worker means the value's drop can never again reach an `ObservationTxn` — the SPEC-prescribed shape becomes unimplementable for this graph rather than merely unimplemented.

This is the concrete answer to "what property does the detached lane weaken": not a live property, but the option to make §3.2's pairing structural. Worth one sentence somewhere (SPEC §15.5 exception, or the `clear_residents_locked` comment) so the next reader does not try to implement it and discover the conflict.

### 4. `ResidentAdmission` is heavier than the problem (Low, simplification)

`scope.rs:58-88`. `Rc<RefCell<Option<ResidentProjection>>>` + `Ref::map` + two `expect`s + one boxed no-op effect, paid on **every** admission — including the `set_admitted_children` startup batch, which loops `admit_child_locked` once per declared child (`scope.rs:1124-1131`). The `Ref` guard also has to be manually `drop`ped before `install()` or the `borrow_mut` panics, which is a live footgun for the next editor of this function.

Two cheaper shapes:
- `Rc<Cell<Option<_>>>` with take/put — removes the `Ref` lifetime juggling and the double-borrow hazard entirely.
- Better: push `ResidentChild::new(child)` into `current_children()` **first**, then do gate adoption, parent wiring and `transition_locked`. Residency then owns the projection across every fallible step, and `clear_residents_locked` already disposes it correctly; nothing can observe the half-wired resident because the gate is held for the whole window. That deletes the guard type outright. (Needs care: the commit-time snapshot producer walks residency, so the `Admitted` transition must still land before commit — it does.)

### 5. Three of the five new regressions are `debug_assert`-gated (Note)

`illegal_restart_transition_retires_its_user_error_off_thread`, `nonresident_terminal_exit_is_retained_before_the_residency_assertion` and `rejected_resident_admission_detaches_its_last_mailbox_owner` all depend on a `debug_assert!` firing. `justfile:25` has no release lane, so they are inert under `--release`. Consistent with #366 Batch 4's own note about not adding a release lane — but the "structural rather than commented" claim is only pinned in the dev profile, and the poisoned-mutex test (`.expect()`) is the only one of the five that holds in both.

## Adjudication (things checked and found correct)

**`commit`'s unwind path is sound; no abort, no stranding.** Traced `gate.rs:119-143`: `panics` is the first local so it drops last; each `install` and each effect is contained independently; effects pushed by a *panicking* install are still drained because `install` mutates `self.effects` in place before it unwinds; `drop(self.guard.take())` is a `MutexGuard` drop and cannot panic; `PanicAccumulator::drop` → `resume_preferred_panic`, which *discards* both payloads when `thread::panicking()`, so an `ObservationTxn` dropped during an unwind cannot double-panic. A panic during the drain cannot strand later wakes — `keep_first_panic` retains the first and `discard_panic` has a `mem::forget` terminator for a hostile payload's own hostile destructor. The watch guards are `PoisonError::into_inner`-tolerant (`sync.rs:688`), so a contained install panic cannot wedge a hub for later readers either.

**Effect-order change in `install` is benign.** Retiring the spent publication *after* the pulse (rather than the projection *before* everything) means the publishing scope's `Arc` is now held across the pulse instead of released before it — strictly safer for a hostile waker that re-enters. The extra `WatchSender` clone under the gate is refcount traffic on framework data.

**Every new under-gate `drop(x.into_exit())` / `drop(reason.into_public())` has a provable surviving owner** — i.e. is refcount traffic, not destruction, so it is legal under the "`Arc` traffic that cannot reach zero" exemption:
- `member.rs:530-548`: `apply_transition` stores the exit in `last_exit` and `update_locked` calls `refresh_retained_exits`; `RetainedExit::install` replaces a non-equal guard set (equality is `Arc::ptr_eq` for `Failed`), so the record owns it before the transient guard is surrendered.
- `scope.rs:679-682`: both `terminalize_locked` branches keep an owner — the winner in `MemberMailbox::Terminal`/`record.stage`, the loser via `txn.defer(drop(RetainedExit))`.
- `scope.rs:906-910`: `publish_stopped_locked` keeps an owner on **both** arms — `published` installs `state` into the record *and* moves it into the `ScopeState` lifecycle event; `!published` defers `drop(state); drop(transient_retained)` in raw-then-guards order.

**`publish_child_restart` looked like an uncovered fifth site of the same class and is not.** `scope.rs:620-637` holds `transition`, `exited` and `scheduled` raw across `modify_silently`, `pulse` and `emit_locked(exited)`. But the driver frame (`driver/child.rs:1056-1095`) keeps its own `exit` binding live across the whole call and `LifecycleEventKind::RestartScheduled` carries no exit, so every raw copy inside the gate is refcount traffic. No change needed — though the argument rests on a caller-side invariant that nothing states.

**Detached disposal cannot leak the resident graph.** Non-critical `DisposalJob::drop` calls `finish()` (`disposal.rs:120-127`), so a Tokio runtime torn down after an accepted `spawn_blocking` still destroys the value on the thread that drops the closure. And `run_fallback_disposals` releases the queue mutex before running a job, so the `RetainedExit::drop` → `dispose_critical` re-submission that happens *inside* a member's teardown cannot self-deadlock the shared worker.

**Mutation-checked all five new tests; none is vacuous.** Every assertion is outside its `catch_unwind`, and each fails on the corresponding revert:
- `panicking_snapshot_install_does_not_strand_later_cuts_or_effects` — reverting `commit` leaves `succeeding_receiver` at `Unstarted` and `effects == 0`.
- `illegal_restart_transition_…` / `nonresident_terminal_exit_…` / `rejected_resident_admission_…` — reverting the retention makes the raw value the last owner, so the probe reports the test thread and `assert_ne!` fires.
- `poisoned_finish_bookkeeping_…` — on `main` the raw `reason`/`exit` unwind on the test thread; note this one also depends on `RetainedStopReason`'s field order (raw reason before guards), which is what routes the probe to `dispose_critical`.

**`clearing_residents_detaches_the_last_mailbox_owner_after_unlock` is not flaky.** The 100 ms window opens only *after* the destructor's `entered` message arrives, and `clear_residents` sends `cleared` at submission time — strictly before the worker can reach the destructor. The `!gate.is_held()` probe is now cross-thread, but it is still discriminating: a regression to an inline deferred drop makes `drop_thread == clear_thread` *and* makes the same-thread `try_lock` report held.

ralexstokes:
Review findings addressed in 0fb4a5b (plain push on top of `c743736` — **no force-push**, no rebase; #382 is stacked on this branch and will need to pick up the new commit).

Two fixed here, one refuted in part, three deferred to filed issues. Walking them in order.

---

### 1. Per-element containment on the displaced set — **fixed**, and first **reproduced**

The mechanism is real and I got a deterministic repro before touching anything. Two residents, each the last owner of a bound mailbox holding one payload whose destructor panics, cleared in one transaction:

```
DROP first  panicking=false
DROP second panicking=true
thread 'shelterwood-disposal' panicked at core/src/panicking.rs:233:
panic in a destructor during cleanup
thread caused non-unwinding panic. aborting.
```

The backtrace is exactly the shape you described: `Envelope` → `MailboxState` drop glue (an un-closed mailbox never builds a `MailboxTeardown`, so the payload is destroyed inline rather than re-routed) → `Arc<MemberCell>` → `ResidentProjection` → `ResidentChild` → `[ResidentChild]` slice glue → `Vec::drop` → `DisposalJob::finish`'s single `catch_panic`, which is above the second panic and cannot see it.

One thing worth flagging, because it nearly cost me the finding: **the first version of that test passed.** The abort happens on the disposal worker *after* both destructors have reported, so the test body had already returned and nextest printed `PASS` — with `thread caused non-unwinding panic. aborting.` sitting in the captured output and an overall exit code of 0. A regression written the obvious way is a coin flip. The committed test parks behind a sentinel `dispose_detached` queued *after* the displaced set: the fallback disposal queue is FIFO on one thread, so a worker that aborts can never reach the sentinel, and the failure precedes the test's return instead of racing it. Mutation-checked — reverting the containment gives `SIGABRT [ ] (1/1) ... clearing_residents_contains_every_hostile_payload_destructor`, deterministically.

On the remedy, I took neither of the two you proposed, for a reason I think is worth the extra paragraph. Both `dispose_all` (N jobs) and a containing newtype around the `Vec` (1 job) put the boundary on the *collection*, and that only covers the displaced root. A `ResidentChild` for a nested scope owns an `Arc<ScopeCell>` whose own `current_children` is another `Vec<ResidentChild>` destroyed by plain slice glue — so two hostile payloads one level down still abort, and `prune_child_locked`'s N=1 case is exactly where that matters ("only the nested subtree can multiply it" — it does, and a collection-level boundary doesn't reach it).

So the boundary went onto the element instead. `ResidentChild::projection` is now an `Option`, and `ResidentChild::drop` takes it and runs `discard_panic(catch_panic(|| drop(projection)).err())`. That gives per-element containment for *every* `Vec<ResidentChild>` at any depth, and on `ScopeCell`'s own drop glue too — and it means neither `clear_residents_locked` nor `prune_child_locked` changed at all. The diagnostic is discarded rather than reported because every venue that destroys a resident already discards it: `dispose_detached` passes an empty completion, and plain drop glue has nobody to report to. Cost is one `expect` on the ~10 field reads, which are now `resident.projection()`.

`crates/shelterwood-cells/src/cells/scope.rs` (`ResidentChild`), test `clearing_residents_contains_every_hostile_payload_destructor`.

### 2. `AGENTS.md` under-describes the code — **fixed**

Both gaps closed. The lock-rule bullet now names `admit_child_locked` alongside the two displacement routes and says what distinguishes it (it protects the *incoming* projection with `ResidentAdmission`), and it records that containment lives on `ResidentChild::drop` rather than on the displaced `Vec`, with the reason. The closing convention list now reads "mailbox payloads, construction closures, offloads, displaced resident graphs and rejected admission projections".

On your question about the two files staying in sync: they cannot drift — `CLAUDE.md` is a symlink to `AGENTS.md`. One edit covers both.

### 3. Detached teardown forecloses SPEC §15.5's structural `Removed` — **deferred to #389**, note added

Confirmed, and it is a genuine SPEC-versus-code contradiction rather than a stale comment: §15.5 states the corollary normatively ("its drop emits `Removed`"), and a disposal worker has neither an `ObservationTxn` nor the gate. Picking between amending §15.5 and splitting residency into an observation-side token plus an ownership-side payload is a spec adjudication, which by this repo's convention gets filed rather than merged into an open PR — so #389.

Your requested sentence is in the tree regardless: `clear_residents_locked`'s comment now says the resident's own destructor can never reach an `ObservationTxn`, that §15.5's shape is therefore unavailable on this lane, and that the edge is emitted explicitly instead, pointing at #389.

### 4. `ResidentAdmission` is heavier than the problem — **footgun refuted; simplification deferred to #390**

The double-borrow hazard is not one. `projection()` returns `Ref<'_, ResidentProjection>` borrowed from `&self`, and `install()` takes `self` by value, so holding the projection across the install is a **compile** error, not a `RefCell` panic. Verified by moving the `drop(projection)` after the push:

```
error[E0505]: cannot move out of `child` because it is borrowed
    --> crates/shelterwood-cells/src/cells/scope.rs:1204:38
```

The next editor gets stopped by rustc, not by a runtime panic in production.

The weight is real, but the `Rc`/`RefCell` is buying something specific, which was not visible in the code and is why the finding reads as unmotivated. A bare `Option<ResidentProjection>` guard disposing from its own `Drop` would submit the disposal job **with the observation gate still held**. That is legal under the lock rule — a submission runs no user code, the `RetainedExit::drop` exemption — but `AGENTS.md` states the preference the other way: "a caller that already owns an effects sink should still flush it", because a submission can cost a native thread start. `admit_child_locked` does own a sink. So `Rc<Cell<Option<_>>>` would not help either; it removes the `Ref`, not the second owner. I've added that rationale to `ResidentAdmission`'s doc so the shape stops looking arbitrary.

On cost: `set_admitted_children` is startup-only and the decision layer is a control plane (§15.6), so one `Rc` plus one boxed effect per declared child is not a hot-path argument.

Your *second* suggestion is the right end state and I agree with it — push `ResidentChild::new(child)` into `current_children()` first and let residency own the projection across gate adoption, parent wiring and `transition_locked`, deleting the guard type outright. That is a restructure of a lock-rule-critical function with at least three things to check first (the commit-time snapshot producer walking residency, the `contains_member`/`resident_projection` lookups that would briefly see an untransitioned member, and the unwind path now *leaving* a half-wired resident in residency instead of disposing it at commit — a behavioural change needing its own test). Filed as #390 with that checklist rather than merged here.

### 5. Three regressions are `debug_assert`-gated — **fixed here, lane question deferred to #393**

Correcting the framing, because the disposition changes with it: they are not inert under `--release`, they **fail**. Each drives the code into an illegal transition and asserts on the resulting panic, so with the `debug_assert!` compiled out the `expect_err` has nothing to catch:

```
$ cargo nextest run --release -p shelterwood-cells --no-fail-fast
Summary: 49 tests run: 46 passed, 3 failed
  FAIL illegal_restart_transition_retires_its_user_error_off_thread
  FAIL rejected_resident_admission_detaches_its_last_mailbox_owner
  FAIL nonresident_terminal_exit_is_retained_before_the_residency_assertion
```

Those three were the *only* failures in the crate, i.e. this PR was introducing release-lane breakage into a lane nobody runs. They are now `#[cfg(debug_assertions)]`, which makes the profile dependency explicit and restores a release-clean workspace: `cargo nextest run --release --workspace` is 803/803. Whether to actually *run* a release lane is a repo-wide CI question — filed as #393 with the three options (add the lane, add a structural check, or declare dev the contract), noting that #366 Batch 4 raised the same question independently.

---

**Adjudication section**: read it all, nothing to contest. The `publish_child_restart` analysis in particular saved me a re-derivation, and I agree the argument rests on a caller-side invariant nothing states — that is worth a comment on `driver/child.rs` at some point, but not in this diff.

**Verification**: `cargo +nightly fmt --all` (written, not checked), workspace `nextest` 809/809 debug and 803/803 release, workspace clippy `--all-targets --all-features -D warnings` clean, doctests clean. Mutation-checked the one new test as described above; the five existing ones were already mutation-checked in your pass and are unchanged apart from the three `cfg` gates.

PR #381, opened 2026-08-20
Document Handler error teardown asymmetry

ralexstokes:
Addresses the Handler/raw decorator contract item from #366.

Pins where a callback error's teardown stops and where mailbox acceptance actually closes.

- `Handler` errors return to their immediate decorator with the incarnation-resource half already frozen and joined. That half-teardown touches **only** the resource half, so it leaves the mailbox binding exactly as the callback loop left it: still accepting when the error preceded any intake freeze, already frozen when the error was raised during frozen-prefix drain or after a stop request — `RawContext::stop` and the `recv`/`try_recv` shutdown boundary freeze intake themselves. The guarantee is "left as found", not "left live".
- Continuing to drive that raw loop is unsupported, and the docs now say the enforcement is partial: a second `Handler::run` panics, but receives keep delivering while `continue_with`/`set_timeout`/`set_interval`/`offload` silently return `Rejected`.
- The framework performs neither teardown step on a plain `RawActor`'s behalf; its decorator observes whatever that loop itself froze.
- SPEC §5.4 is reconciled with this. Its second clause named exit publication as the close point for an incarnation ending without a stop phase; the raw incarnation boundary freezes the mailbox as soon as the body returns or unwinds, before `join_resources` and both user destructors, so that freeze is the close point. A hard abort is reachable only as a grace-ladder escalation of a stop request that already froze intake, so its close point is that stop's freeze; abrupt driver teardown freezes and closes together. Acceptance is closed no later than exit publication on every path.

Façade regressions pin both halves. The decorator probe records the resource capabilities *and* a self-`try_send`, and all verdicts are compared in the test body. `handler_error_freezes_resources_while_leaving_an_unfrozen_mailbox_alone` pins the still-accepting direction for a `Handler` init error and a plain `RawActor` error; the new `handler_drain_error_returns_with_intake_already_frozen` pins the already-frozen direction on the normative default `MailboxShutdown::Drain` path.

Validation: `cargo nextest run -p shelterwood` (555), clippy `-D warnings`, doctests, `cargo doc --no-deps`, packaged-docs check. Mutation checks for each new assertion are recorded in the review-response comment.

### DIFF
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -484,12 +484,22 @@ impl<A: Actor> RawActor for Handler<A> {
     /// decorator — not the raw runner — may resume after this future
     /// returns, and it must not observe still-live incarnation resources;
     /// every error exit therefore freezes and joins them here first. That
-    /// half-teardown deliberately does not freeze the mailbox, so a decorator
-    /// may inspect the returned error with the mailbox still live and the
-    /// resource capabilities now rejected. Resuming the raw loop past this
-    /// point is unsupported; once the decorated raw stack returns, the raw
-    /// incarnation boundary freezes the mailbox before destroying the context
-    /// and actor.
+    /// half-teardown touches only the resource half: it leaves the mailbox
+    /// binding exactly as this loop left it — still accepting when the error
+    /// preceded any intake freeze, already frozen when it was raised during
+    /// frozen-prefix drain or after a `Context::stop` request, since
+    /// [`RawContext::stop`] and the [`recv`](RawContext::recv) /
+    /// [`try_recv`](RawContext::try_recv) shutdown boundary freeze intake
+    /// themselves. Either way the decorator finds the resource capabilities
+    /// already rejected. Resuming the raw loop past this point is
+    /// unsupported, and unenforced beyond a second call to this method
+    /// panicking: receives keep delivering whatever the mailbox state still
+    /// allows, while `continue_with`, `set_timeout`, `set_interval` and
+    /// `offload` silently return `Rejected` instead of diagnosing. Once the
+    /// decorated raw stack returns, the raw incarnation boundary freezes the
+    /// mailbox before joining resources and destroying the context and
+    /// actor, which is where acceptance closes for an incarnation that ends
+    /// without a stop phase (§5.4).
     async fn run(&mut self, raw: &mut RawContext<Self::Msg>) -> ExitResult {
         let HandlerState::Uninit(args) = std::mem::replace(&mut self.state, HandlerState::Spent)
         else {
@@ -551,10 +561,12 @@ impl<A: Actor> RawActor for Handler<A> {
 /// Propagates a callback error after §6.5's orderly teardown: incarnation-owned
 /// work is frozen, cancelled, and joined before control returns to the caller,
 /// which at the advertised composition point may be a raw decorator rather
-/// than the raw incarnation boundary itself. This helper leaves the mailbox
-/// live for that decorator; the outer raw incarnation boundary freezes it when
-/// the decorated stack returns. Returning to the loop after this resource-only
-/// teardown is unsupported.
+/// than the raw incarnation boundary itself. This helper touches only the
+/// resource half, leaving the mailbox binding as it found it: still accepting
+/// when the error preceded any intake freeze, already frozen when it was
+/// raised during frozen-prefix drain or after a stop request. The outer raw
+/// incarnation boundary freezes the mailbox when the decorated stack returns.
+/// Returning to the loop after this resource-only teardown is unsupported.
 async fn fail_after_teardown<M: Send + 'static>(
     raw: &mut RawContext<M>,
     error: ExitError,
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -483,7 +483,12 @@ impl<A: Actor> RawActor for Handler<A> {
     /// this wrapper is the advertised raw-decorator composition point, so a
     /// decorator — not the raw runner — may resume after this future
     /// returns, and it must not observe still-live incarnation resources;
-    /// every error exit therefore freezes and joins them here first.
+    /// every error exit therefore freezes and joins them here first. That
+    /// half-teardown deliberately does not freeze the mailbox: an incarnation
+    /// ending without a stop phase remains accepting until exit publication
+    /// (SPEC §5.4). A decorator may inspect the returned error and the now-
+    /// rejected resource capabilities, but resuming the raw loop past this
+    /// point is unsupported.
     async fn run(&mut self, raw: &mut RawContext<Self::Msg>) -> ExitResult {
         let HandlerState::Uninit(args) = std::mem::replace(&mut self.state, HandlerState::Spent)
         else {
@@ -545,7 +550,9 @@ impl<A: Actor> RawActor for Handler<A> {
 /// Propagates a callback error after §6.5's orderly teardown: incarnation-owned
 /// work is frozen, cancelled, and joined before control returns to the caller,
 /// which at the advertised composition point may be a raw decorator rather
-/// than the raw incarnation boundary itself.
+/// than the raw incarnation boundary itself. The mailbox stays bound until
+/// the raw incarnation boundary publishes the exit, as SPEC §5.4 requires;
+/// returning to the loop after this resource-only teardown is unsupported.
 async fn fail_after_teardown<M: Send + 'static>(
     raw: &mut RawContext<M>,
     error: ExitError,
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -484,11 +484,12 @@ impl<A: Actor> RawActor for Handler<A> {
     /// decorator — not the raw runner — may resume after this future
     /// returns, and it must not observe still-live incarnation resources;
     /// every error exit therefore freezes and joins them here first. That
-    /// half-teardown deliberately does not freeze the mailbox: an incarnation
-    /// ending without a stop phase remains accepting until exit publication
-    /// (SPEC §5.4). A decorator may inspect the returned error and the now-
-    /// rejected resource capabilities, but resuming the raw loop past this
-    /// point is unsupported.
+    /// half-teardown deliberately does not freeze the mailbox, so a decorator
+    /// may inspect the returned error with the mailbox still live and the
+    /// resource capabilities now rejected. Resuming the raw loop past this
+    /// point is unsupported; once the decorated raw stack returns, the raw
+    /// incarnation boundary freezes the mailbox before destroying the context
+    /// and actor.
     async fn run(&mut self, raw: &mut RawContext<Self::Msg>) -> ExitResult {
         let HandlerState::Uninit(args) = std::mem::replace(&mut self.state, HandlerState::Spent)
         else {
@@ -550,9 +551,10 @@ impl<A: Actor> RawActor for Handler<A> {
 /// Propagates a callback error after §6.5's orderly teardown: incarnation-owned
 /// work is frozen, cancelled, and joined before control returns to the caller,
 /// which at the advertised composition point may be a raw decorator rather
-/// than the raw incarnation boundary itself. The mailbox stays bound until
-/// the raw incarnation boundary publishes the exit, as SPEC §5.4 requires;
-/// returning to the loop after this resource-only teardown is unsupported.
+/// than the raw incarnation boundary itself. This helper leaves the mailbox
+/// live for that decorator; the outer raw incarnation boundary freezes it when
+/// the decorated stack returns. Returning to the loop after this resource-only
+/// teardown is unsupported.
 async fn fail_after_teardown<M: Send + 'static>(
     raw: &mut RawContext<M>,
     error: ExitError,
--- a/crates/shelterwood/tests/actor.rs
+++ b/crates/shelterwood/tests/actor.rs
@@ -404,6 +404,123 @@ impl<R: RawActor> RawActor for ResumeProbeDecorator<R> {
     }
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+struct ErrorReturnCapabilities {
+    continuation_rejected: bool,
+    timer_rejected: bool,
+    offload_rejected: bool,
+}
+
+struct ErrorReturnProbe<R> {
+    inner: R,
+    observed: Arc<Mutex<Option<ErrorReturnCapabilities>>>,
+}
+
+impl<R: RawActor<Msg = ()>> RawActor for ErrorReturnProbe<R> {
+    type Msg = ();
+
+    fn readiness() -> Readiness {
+        R::readiness()
+    }
+
+    async fn run(&mut self, context: &mut RawContext<Self::Msg>) -> ExitResult {
+        let result = self.inner.run(context).await;
+        assert!(result.is_err(), "the probe requires an inner error return");
+        let capabilities = ErrorReturnCapabilities {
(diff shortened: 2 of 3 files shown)


ralexstokes:
Review at `c856b24`. The mechanism the PR documents is real and the code matches the *positive* claims (`fail_after_teardown` freezes only `self.resources`; the boundary freezes the mailbox at `raw/definition.rs:265-267`, then `drop_raw` / `drop_actor` at `279`/`282`, then returns for publication). What does not hold up is the **consequent** attached to the mailbox half in three places, plus one spec section left unreconciled. Two of those are checkable on the default configuration.

## P2 — new §6.5 text contradicts §5.4, which the PR leaves unamended

`specs/SPEC.md:1056-1058` (new): "the raw incarnation boundary freezes the mailbox before destroying the context and actor; this precedes exit publication."

`specs/SPEC.md:726-727` (untouched): acceptance "closes ... or, for an incarnation that ends without a stop phase (panic, hard abort, **plain return**), at its exit publication (§8)".

The code agrees with the new text and therefore contradicts §5.4. `raw/definition.rs:265-267` freezes the mailbox as the first thing after the decorated raw stack returns; `shelterwood-mailbox/src/cell.rs:1388-1400` moves the binding `Bound → Frozen`, and `cell.rs:1067` makes `try_send` on `Frozen` return `NotRunning`. The window is not a formality: `join_resources().await` at `definition.rs:275` sits inside it and is unbounded (it joins every live offload), and `drop_raw`/`drop_actor` run arbitrary user destructors after that. So a sender observes `NotRunning` long before the exit is published, on the plain-return and error-return paths alike.

Note that `RawIncarnationOwner::drop` (`definition.rs:229-240`) performs *no* mailbox freeze, so §5.4's "at its exit publication" is still correct for hard abort — it is only the return and panic paths that close early.

This PR is the one whose job is to pin the close point, and issue #366's Batch 4 item explicitly asked the spec note to *cite* §5.4. It now neither cites nor reconciles it. Remedy: amend §5.4's second clause to name the raw-boundary mailbox freeze as the close point for return/panic endings, keeping exit publication for hard abort. CONFIRMED.

## P2 — "the decorator observes that half still live" is false on the default drain path

`specs/SPEC.md:1054-1055`; `crates/shelterwood/src/actor.rs:487-489` ("so a decorator may inspect the returned error with the mailbox still live"); `crates/shelterwood/src/actor.rs:554-556` ("This helper leaves the mailbox live for that decorator").

`Handler::run` reaches its drain loop only after `raw.recv().await` returned `None`, and `RawContext::recv` freezes the receiver *and* the resources before returning `None` (`raw/context.rs:1129-1150`); `try_recv` does the same at `raw/context.rs:1184-1191`. So:

```
while let Some(message) = raw.recv().await { ... }   // None here => receiver already Frozen
match raw.mailbox_shutdown() {
    MailboxShutdown::Drain => {
        while let Some(message) = raw.try_recv() {
            ...
            if let Err(error) = handled { return fail_after_teardown(raw, error).await; }
```

A handler error raised while draining the frozen prefix returns to the decorator with the mailbox **already frozen**. `MailboxShutdown::Drain` is the normative default (Appendix A, `SPEC.md:3238`), so no unusual configuration is needed. Two more counterexamples: a `Live` callback that calls `Context::stop()` (`actor.rs:202-209` → `RawContext::stop` → `receiver.freeze()`) and then returns `Err` in the same callback; and an `AfterInit` initializer that calls `Context::stop()` (`defer_stop_until_after_init`, `raw/context.rs:939-947`, which also calls `receiver.freeze()`) and then returns `Err`.

The first clause of each sentence is accurate — the half-teardown does not *itself* freeze the mailbox. The consequent is the over-reach: what the helper actually guarantees is that it leaves the mailbox binding **as it found it**. Remedy: "leaves the mailbox binding untouched — still live when the error precedes any intake freeze, already frozen when it was raised during drain or after a stop request". The new test only exercises the init-error case, so it cannot catch this. CONFIRMED.

## P3 — "A plain `RawActor` returning `Err` has not performed either teardown step"

`specs/SPEC.md:1058-1060`. As written this is a universal claim about plain raw actors, and it is false for the most common exit shape: a raw loop that ends because `recv()` returned `None` has had both halves frozen *by `recv` itself* (`raw/context.rs:1129-1150`) before it returns anything, `Err` included. Same for any raw actor that called `RawContext::stop()`.

The intended claim is that the *framework* performs neither step on a plain raw actor's behalf. `PlainRawError` in the new test returns `Err` on the first statement of `run`, so the test pins only the case where the universal statement happens to hold. Remedy: scope the sentence to framework action — "the framework performs neither teardown step on a plain `RawActor`'s behalf; its decorator observes whatever the loop itself froze, via `stop()` or via the `recv`/`try_recv` shutdown boundary."

## P3 — "unsupported" is unenforced, and the doc doesn't say what a violation does

`actor.rs:489-490`, `actor.rs:556-557`, `SPEC.md:1053-1054`. Enforcement is partial and incidental: a second `Handler::run` panics (`actor.rs:493-497` — the state is `Running` or `Spent`, never `Uninit`). Nothing stops a decorator from continuing to drive the shared `RawContext`: `recv()`/`try_recv()` keep delivering mailbox messages while `continue_with`/`set_timeout`/`offload` silently return `Rejected` (`raw/context.rs:968-971`, `977-992`) — a silently degraded loop, not a diagnostic. Since the paragraph exists to make the convention checkable, say that it is unenforced and name the observable consequence.

## Test notes (not blocking)

Not vacuous, and it does pin what it claims. The inner `assert!(result.is_err(), ...)` sits in the decorator's `run` and would be eaten by the incarnation boundary, but it degrades into a *test-body* failure through `.expect("the decorator observed its returned context")` on the unset observation, so it cannot pass silently; the verdicts themselves are compared in the test body. Mutation check: deleting `raw.freeze_resources()` from `fail_after_teardown` leaves `is_stopping()` false and `resources.accepting` true at the probe (no `.await` between the inner return and the probe, so no shutdown-cancellation race), flipping all three Handler flags to `false` and failing the first `assert_eq!`.

Two coverage gaps worth closing while here:

- The tests pin only the **resource** half. The mailbox half — the asymmetry the PR is named for — is unpinned in both directions. A `context.myself().try_send(())` probe added to `ErrorReturnCapabilities` would pin it, and would have surfaced the P2 above.
- Both cases exercise only a pre-readiness error. The drain-path handler error has no coverage at all.

## Checks that came back clean

- Packaged docs: `tools/sync-packaged-docs.sh` covers only `README.md` and `docs/{embedding,observation,retry-and-ordering,shutdown-and-resources}.md`. This PR touches none of them, so there is no `doctests/` divergence. `docs/shutdown-and-resources.md`'s raw-actor drain section makes no conflicting claim.
- No other `specs/` statement conflicts: §4.3 (`SPEC.md:610-613`), §15.1 L3 (`SPEC.md:2580-2583`) and obligation 15 (`SPEC.md:3165`) all describe the stop-path close point, which is unaffected.
- The positive ordering claims all hold: `fail_after_teardown` touches only `self.resources` (`raw/context.rs:1564-1570`), and the boundary order is mailbox freeze → resource freeze → join → context drop → actor drop → return/publish.
- Lock rule: no new lock sites; the new spec text does not restate §15.4.

**chatgpt-codex-connector, P2 "Align the binding claim with the raw epilogue" (`actor.rs:554`, reviewed commit `b958185`) — CONFIRMED against that commit, and already fixed by `c856b24` on the current head.** The text it quoted ("The mailbox stays bound until the raw incarnation boundary publishes the exit, as SPEC §5.4 requires") was factually wrong for exactly the reason given, and it is gone.

Two framing corrections:

- Its remedy menu was half wrong. "Move the freeze to the actual publication path" would be a behavior change that reopens the acceptance window across the whole unbounded `join_resources` await and both user destructors, buying nothing. "Describe the close point as return to the raw boundary" was the right half, and is what `c856b24` did.
- It did not name the residual, which is the one thing still broken: §5.4 itself (`SPEC.md:726-727`) still states the wrong close point. That is the P2 at the top of this review. Fixing only the PR's own prose left the document internally contradictory in the opposite direction.

Its `AGENTS.md:L9-L15` citation is the lock-rule preamble and has nothing to do with the finding — spurious, though not a fabricated quote.

ralexstokes:
All four findings addressed in `ae5a3e6`. Each claim below was re-derived from the code, not taken from the review on trust.

## P2 (1) — §5.4 contradicted the new §6.5 text — FIXED

`specs/SPEC.md` §5.4 second clause, was:

> or, for an incarnation that ends without a stop phase (panic, hard abort, plain return), at its exit publication (§8)

now:

> or, for an incarnation that ends without a stop phase (plain return, error return, or panic), at the raw incarnation boundary's mailbox freeze — taken the moment the incarnation body returns or unwinds, before its resources are joined and before the context and actor are destroyed (§6.5), and therefore strictly before exit publication (§8). A hard abort is reachable only as a grace-ladder escalation (§11) of a stop request that already froze intake, so its close point is that stop's freeze; abrupt driver teardown freezes and closes together. On every path acceptance is closed no later than exit publication.

Why it is true of the code:

- `raw/definition.rs:264-268` runs `mailbox.freeze(incarnation, …)` unconditionally right after `CatchUnwindFuture::new(actor.run(raw)).await`, so it covers `Ok`, `Err` and caught-panic alike. `MailboxCell::freeze` (`shelterwood-mailbox/src/cell.rs:1388`) moves `Bound → Frozen`, and `try_send` on `Frozen` is `NotRunning` (`cell.rs:1067`). The rest of the boundary — `freeze_resources`, `join_resources().await` (unbounded), `drop_raw`, `drop_actor` — is *after* that freeze, and the incarnation future only then resolves, so publication is strictly later.
- The hard-abort sentence is checked, not assumed. `StopAction::HardAbort` is produced only by `ladder.advance` (`driver/child.rs:918-924`); `active.ladder` is assigned only in `begin_stop_child` (`child.rs:855`), nine lines after that function's own `mailbox.freeze` (`child.rs:844-847`). The only other `abort_handle.abort()` is `driver.rs:416`, in driver drop, which freezes *and* closes at `driver.rs:402-403`. `RawIncarnationOwner::drop` itself still performs no freeze — that part of the review is right — but it is never reached without one of those two prior freezes, so I did **not** keep "at exit publication" as the hard-abort close point; the accurate statement is the §6.2 freeze, with "no later than exit publication" as the path-independent backstop.

Cross-checks: §15.1 obligation 15 (`SPEC.md:3178`) and the identity table (`SPEC.md:3432`) both describe the intake freeze and remain consistent; `SPEC.md:1058`'s "the mailbox binding outlives actor destruction" is about *binding* lifetime (`Frozen` is still bound — `close` is the driver's, at publication) and is unaffected.

## P2 (2) — "the decorator observes that half still live" — FIXED

Confirmed reachable on the default configuration. `Handler::run`'s drain loop is entered only after `recv()` returned `None`, and both `None` branches of `recv` call `self.receiver.freeze()` first (`raw/context.rs:1130-1131`, `1147-1148`); `try_recv` does the same (`context.rs:1189-1190`); `RawContext::stop` (`context.rs:928-932`) and `defer_stop_until_after_init` (`context.rs:938-947`) freeze intake directly. `MailboxReceiver::freeze` (`shelterwood-mailbox/src/cell.rs:1684`) is the same binding freeze. So the drain-path error, the `stop()`-then-`Err` live handler, and the stop-in-init-then-`Err` initializer all return with the mailbox already frozen.

Replaced the consequent in all three places with the guarantee that is actually provable — the helper touches only the resource half, so it leaves the binding as it found it:

- `SPEC.md` §6.5: "It touches **only** that half: the mailbox binding is left exactly as the callback loop left it — still accepting when the error preceded any intake freeze (an `init` or live-handler error with no stop request), already frozen when the error was raised while draining the frozen prefix or after a `ctx.stop()`, because the stop call and the `recv`/`try_recv` shutdown boundary freeze intake themselves (§6.2)."
- `actor.rs:486-493` (`Handler::run`): "That half-teardown touches only the resource half: it leaves the mailbox binding exactly as this loop left it — still accepting when the error preceded any intake freeze, already frozen when it was raised during frozen-prefix drain or after a `Context::stop` request, since `RawContext::stop` and the `recv`/`try_recv` shutdown boundary freeze intake themselves."
- `actor.rs:564-567` (`fail_after_teardown`): "This helper touches only the resource half, leaving the mailbox binding as it found it: still accepting when the error preceded any intake freeze, already frozen when it was raised during frozen-prefix drain or after a stop request."

## P3 (3) — the plain-`RawActor` universal claim — FIXED

Was: "A plain `RawActor` returning `Err` has not performed either teardown step, so its decorator observes both halves live until it returns to that boundary." Scoped to framework action, as the review proposed:

> The framework performs neither teardown step on a plain `RawActor`'s behalf, so its decorator observes whatever that loop itself froze.

## P3 (4) — "unsupported" is unenforced — FIXED

Enforcement re-checked: `Handler::run` panics on re-entry because the state is `Running`/`Spent`, never `Uninit` (`actor.rs:509-512`). `freeze_resources` sets `resources.accepting = false`, which makes `continue_with`/`set_timeout`/`set_interval`/`offload` return `Rejected` (`raw/context.rs:967-971`, `976-992`), while `next_ready()` still reads the mailbox, so receives keep working. Both docs now say so:

> Resuming the raw loop past this point is unsupported, and unenforced beyond a second call to this method panicking: receives keep delivering whatever the mailbox state still allows, while `continue_with`, `set_timeout`, `set_interval` and `offload` silently return `Rejected` instead of diagnosing.

## Test coverage

`ErrorReturnCapabilities` gains `mailbox_send: Result<(), SendErrorKind>`, recorded in the decorator by `context.myself().try_send(())` — the exact probe the review named. Recording happens in the decorator; every verdict is compared in the test body, so nothing depends on an assertion inside actor code.

- `handler_error_freezes_resources_while_leaving_an_unfrozen_mailbox_alone` (renamed from `…_while_plain_raw_error_leaves_them_live`) now pins `mailbox_send: Ok(())` for both the `Handler` init-error case and `PlainRawError`.
- New `handler_drain_error_returns_with_intake_already_frozen` covers the gap: `DrainErrorActor` self-sends two messages during `init`, calls `ctx.stop()` on the first live delivery so the second becomes the frozen accepted prefix, then returns `Err` while `is_draining()`. It pins `mailbox_send: Err(SendErrorKind::NotRunning)` — the other direction.
- The harness no longer keys on a startup verdict (the drain case reaches readiness); it waits on the observation itself and then shuts the tree down.

Mutation checks (committed tree, mutation injected, reverted with `git checkout`):

| mutation | result |
| --- | --- |
| `fail_after_teardown` also calls `raw.stop()` (i.e. starts touching the mailbox half) | `…_leaving_an_unfrozen_mailbox_alone` FAILS: `mailbox_send` `Ok(())` → `Err(NotRunning)` |
| drop `receiver.freeze()` from `RawContext::stop`, both `recv` shutdown branches and `try_recv` | drain test still passes — the driver's `begin_stop_child` freeze (`child.rs:844-847`) independently establishes it |
| …plus dropping that driver freeze | `handler_drain_error_returns_with_intake_already_frozen` FAILS: `Err(NotRunning)` → `Ok(())` |
| drop `raw.freeze_resources()` from `fail_after_teardown` | `…_leaving_an_unfrozen_mailbox_alone` and `handler_error_joins_offloads_before_returning_to_a_raw_decorator` both FAIL |

The second row is worth recording: the frozen-intake property is redundantly implemented, so the drain test pins the *property* the doc states rather than any single call site. The last row also shows the drain test does not discriminate the resource half (`is_stopping()` already rejects those capabilities there) — that half stays pinned by the pre-readiness cases.

## Verification

`cargo +nightly fmt --all`; `cargo nextest run -p shelterwood` (555 passed); `cargo clippy -p shelterwood --all-targets -D warnings`; `cargo test -p shelterwood --doc`; `cargo doc -p shelterwood --no-deps` clean (the new intra-doc links resolve); `tools/sync-packaged-docs.sh --check` clean — `specs/SPEC.md` is not a packaged source, matching the review's own check.

PR #382, opened 2026-08-20
Reject illegal member projection transitions in release

ralexstokes:
Addresses the reducer-legality item from #366 and is stacked on #380.

Converts member projection transitions from debug-only source-stage assertions to a fallible reducer result, and makes residency, lifecycle, snapshot, and restart publication conditional on acceptance. Rejected user-bearing inputs leave through post-unlock detached disposal.

Admission probes legality, performs the recursive gate handoff under the child subtree's current observation gate, and applies the transition only after that handoff succeeds. A rejected repeated admission therefore cannot rehome a subtree that remains resident under its original parent, and an unwind out of `install_observation_gate_locked`'s `assert!` cannot publish `Admitted` for a half-re-homed subtree. The regression pins original residency, parent identity, and descendant gate identity; the formerly debug-only live-route handoff test now asserts fail-closed behavior in release too.

**Scope of the "rejection is a no-op" claim.** A refusal is a total no-op *inside the cell layer*: no record field changes, no watch version advances, no lifecycle event is emitted, no residency is pushed. It says nothing about state the caller committed before presenting the transition. Three driver sites have such a prefix — `publish_child_restart` after `schedule_restart` has charged the child and the intensity window, the spawn `Starting` after the mailbox bind and body dispatch, and `admit_child_locked` after `entry.promote` — so a refusal there is a half-applied operation, not silence. Those are filed as #392; each site carries a comment pointing at it.

What keeps the class detectable in the meantime:

- every function returning the verdict is `#[must_use]` (`MemberCell::transition{,_locked}`, `ScopeCell::transition_child_stage`, `publish_child_restart`, `admit_child{,_locked}`, `set_admitted_children`), so it cannot be discarded by writing `;`;
- each of the seven driver sites binds the verdict and `debug_assert!`s it, naming the source stages it may present. That restores the debug-lane abort the deleted reducer assertions provided, and does it outside every framework lock — the old ones panicked under the watch mutex and the observation gate, poisoning both.

Also in this PR after review: one gate-handoff retry loop (`MemberCell::with_handoff_gate`) instead of two; AGENTS.md's gate-to-gate exemption updated to name it and both of its users; the live-dynamic-route `debug_assert!` dropped from the admission path (the legality probe makes it unconstructible there) with its regression restored on the reservation-time adoption path that still carries the assert.

Validation:

- `./tools/dev just ci` (811 tests)
- `cargo nextest run --workspace --release` (807 pass; `nonresident_terminal_exit_is_retained_before_the_residency_assertion` is a base-branch artifact — #380's tip already gates it on `debug_assertions`, this PR is based on an older #380 commit)
- mutation-checked: accept-everything matrix, handoff-before-probe ordering, discarded per-child verdict in `set_admitted_children`, and the removed live-route assert each fail their pinning test

Stack: keep based on #380 until that Batch 3 foundation lands, then retarget to `main`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,11 +116,19 @@ rests on:
   gate is outermost: everything else is taken under it or standalone, never
   around it. Inside, the documented orders are cells' member mailbox →
   `MailboxCell::state`, and `MailboxCell::state` → `SendOperation::state`.
-  Gate-to-gate is the one exception to "outermost": `adopt_observation_gate`
-  takes the adoptee's *current* gate while the adopting parent's is held, in
-  the parent-to-child direction only, and re-reads the installed pointer
-  under the acquired guard so a concurrent handoff retries rather than
-  deadlocks.
+  Gate-to-gate is the one exception to "outermost":
+  `MemberCell::with_handoff_gate` takes the adoptee's *current* gate while the
+  adopting parent's is held, in the parent-to-child direction only, and
+  re-reads the installed pointer under the acquired guard so a concurrent
+  handoff retries rather than deadlocks. Both of its users — plain adoption
+  (`ScopeCell::adopt_observation_gate`) and admission
+  (`ScopeCell::admit_observation_gate`, `ScopeCell::admit_child_locked`) —
+  hold the exemption. Admission widens what runs inside the doubled section:
+  the member record's watch-value mutex is taken there, and admission's
+  `MemberTransition` reaches `RetainedExit` clone/drop under both gates. That
+  stays inside the rule — the record is framework-owned data and the retained
+  clone is provably non-last — but it is the deepest the exemption goes, so a
+  new operation added to the doubled section needs the same accounting.
 
 Two conventions that are not the rule itself but travel with it: panicking
 while holding a mutex the codebase `.expect()`s poisons it for every later
--- a/crates/shelterwood-cells/src/cells/member.rs
+++ b/crates/shelterwood-cells/src/cells/member.rs
@@ -29,20 +29,6 @@ pub enum MemberStage {
     Terminal(Exit),
 }
 
-impl MemberStage {
-    fn tag(&self) -> &'static str {
-        match self {
-            Self::Reserved => "Reserved",
-            Self::Admitted => "Admitted",
-            Self::Starting => "Starting",
-            Self::Running => "Running",
-            Self::Restarting => "Restarting",
-            Self::Stopping => "Stopping",
-            Self::Terminal(_) => "Terminal",
-        }
-    }
-}
-
 /// One non-terminal member-record transition owned by the cell layer.
 pub enum MemberTransition {
     Admitted,
@@ -108,48 +94,52 @@ impl MemberRecord {
     /// Applies one driver-requested transition.
     ///
     /// Every watch-channel writer routes stage changes through here (see
-    /// [`MemberCell::transition`] for the wake-bus contract), so each arm
-    /// asserts the source stages its driver call sites can actually present.
+    /// [`MemberCell::transition`] for the wake-bus contract). The reducer
+    /// rejects an event whose source stage is not one its driver call sites
+    /// can present, including in release builds.
     ///
     /// Exits are safe to retire inside the watch mutation: the record's guard
     /// set still covers the displaced value, and [`Self::refresh_retained_exits`]
     /// re-establishes that cover before the mutation returns.
-    fn apply_transition(&mut self, transition: MemberTransition) {
+    fn apply_transition(&mut self, transition: MemberTransition) -> Result<(), MemberTransition> {
+        let legal = matches!(
+            (&self.stage, &transition),
+            (MemberStage::Reserved, MemberTransition::Admitted)
+                | (
+                    MemberStage::Admitted | MemberStage::Restarting,
+                    MemberTransition::Starting { .. }
+                )
+                | (
+                    MemberStage::Starting | MemberStage::Reserved,
+                    MemberTransition::Running
+                )
+                | (
+                    MemberStage::Starting | MemberStage::Running,
+                    MemberTransition::Stopping
+                )
+                | (
+                    MemberStage::Starting | MemberStage::Running | MemberStage::Stopping,
+                    MemberTransition::RestartScheduled { .. }
+                )
+        );
+        if !legal {
+            return Err(transition);
+        }
+
         match transition {
             MemberTransition::Admitted => {
-                debug_assert!(
-                    matches!(self.stage, MemberStage::Reserved),
-                    "admission must consume a fresh reservation, not {}",
-                    self.stage.tag()
-                );
                 self.stage = MemberStage::Admitted;
             }
             MemberTransition::Starting { incarnation } => {
-                debug_assert!(
-                    matches!(self.stage, MemberStage::Admitted | MemberStage::Restarting),
-                    "a spawn must start an admitted or restarting member, not {}",
-                    self.stage.tag()
-                );
                 self.stage = MemberStage::Starting;
                 self.incarnation = Some(incarnation);
                 self.last_incarnation = Some(incarnation);
                 self.restart_at = None;
             }
             MemberTransition::Running => {
-                debug_assert!(
-                    matches!(self.stage, MemberStage::Starting | MemberStage::Reserved),
-                    "readiness must promote a starting member (or the root scope's own \
-                     never-admitted reservation), not {}",
-                    self.stage.tag()
-                );
                 self.stage = MemberStage::Running;
             }
             MemberTransition::Stopping => {
-                debug_assert!(
-                    matches!(self.stage, MemberStage::Starting | MemberStage::Running),
-                    "a stop ladder must begin on a starting or running member, not {}",
-                    self.stage.tag()
-                );
                 self.stage = MemberStage::Stopping;
             }
             MemberTransition::RestartScheduled {
@@ -157,14 +147,6 @@ impl MemberRecord {
                 restart_count,
                 restart_at,
             } => {
-                debug_assert!(
-                    matches!(
-                        self.stage,
-                        MemberStage::Starting | MemberStage::Running | MemberStage::Stopping
-                    ),
-                    "a restart must be scheduled from an active incarnation's exit, not {}",
-                    self.stage.tag()
-                );
                 self.stage = MemberStage::Restarting;
                 self.incarnation = None;
                 self.last_exit = Some(exit);
@@ -172,6 +154,7 @@ impl MemberRecord {
                 self.restart_at = restart_at;
             }
         }
+        Ok(())
     }
 }
 
@@ -407,8 +390,8 @@ impl MemberCell {
     /// field read by a loop precondition must be changed through a pulsing path
     /// like this one, never by a silent write outside an observation gate.
     #[cfg(any(test, feature = "test-util"))]
-    pub fn transition(&self, transition: MemberTransition) {
-        self.with_observation_txn(|txn| self.transition_locked(txn, transition));
+    pub fn transition(&self, transition: MemberTransition) -> bool {
+        self.with_observation_txn(|txn| self.transition_locked(txn, transition))
     }
 
     fn with_observation_txn<R>(&self, operation: impl FnOnce(&mut ObservationTxn<'_>) -> R) -> R {
@@ -527,9 +510,13 @@ impl MemberCell {
         txn.pulse(&self.record);
     }
 
-    pub fn transition_locked(&self, txn: &mut ObservationTxn<'_>, transition: MemberTransition) {
+    pub fn transition_locked(
+        &self,
+        txn: &mut ObservationTxn<'_>,
+        transition: MemberTransition,
+    ) -> bool {
         // `RestartScheduled` carries a user error by value. Retain it before
-        // the watch lock and reducer assertions so an invariant panic can
+        // the watch lock and reducer validation so an unrelated panic can
         // unwind the raw transition as refcount traffic only.
         let retained_exit = match &transition {
             MemberTransition::RestartScheduled { exit, .. } => {
@@ -540,11 +527,29 @@ impl MemberCell {
             | MemberTransition::Running
             | MemberTransition::Stopping => None,
         };
-        self.update_locked(txn, |record| record.apply_transition(transition));
+        let mut rejected = None;
+        self.record
+            .modify_silently(|record| match record.apply_transition(transition) {
+                Ok(()) => record.refresh_retained_exits(),
+                Err(transition) => rejected = Some(transition),
+            });
+        if let Some(rejected) = rejected {
+            // Rejection is a total no-op. The rejected event may own a failed
+            // exit, so retire it with the transaction rather than under the
+            // observation gate or watch lock.
+            txn.defer(move || runtime::dispose_detached(rejected));
+            if let Some(retained_exit) = retained_exit {
+                // The deferred transition proves this clone cannot be last.
+                drop(retained_exit.into_exit());
+            }
+            return false;
+        }
+        txn.pulse(&self.record);
         if let Some(retained_exit) = retained_exit {
             // The record now owns an equivalent retained copy.
             drop(retained_exit.into_exit());
         }
+        true
     }
 
     pub fn set_options(&self, options: ResolvedCommonOptions) {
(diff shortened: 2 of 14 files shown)


ralexstokes:
Reviewed `cb36a4e` on top of `agent/issue-366-batch3-hardening`; the legality matrix reproduces the five deleted `debug_assert!`s exactly, the admission reordering preserves the documented parent-to-child gate discipline, and the three new regressions are non-vacuous. No reachable bug found. Findings below are about what the conversion gives up and one unwind-window ordering regression.

## Medium — every production call site discards the rejection, and nothing else reports it

`crates/shelterwood-cells/src/cells/member.rs:551`, `:393`, `crates/shelterwood-cells/src/cells/scope.rs:642`, `:667`, `:1184`, `:1188` all return `bool`; none is `#[must_use]`, and the workspace enables only `dead_code`/`unreachable_pub`. All six production callers throw the result away:

- `crates/shelterwood/src/driver.rs:785` — `root.admit_child_locked(...)`, then unconditionally `AdmissionInstall::Admitted(key)`
- `crates/shelterwood/src/driver.rs:1013` — root `MemberTransition::Running`
- `crates/shelterwood/src/driver/child.rs:739` — `Starting` + `Started`
- `crates/shelterwood/src/driver/child.rs:843` — `Stopping`
- `crates/shelterwood/src/driver/child.rs:1072` — `publish_child_restart`
- `crates/shelterwood/src/driver/startup.rs:67` — `Running` + `Ready`

`MemberStage::tag()` was deleted with the asserts, tracing is gone since #376, and no `debug_assert!` guards the `!legal` branch. Net effect: an illegal transition that previously aborted the whole debug test suite now returns `false` into `;` in every build, in silence. That is a real loss — the debug lane was the only detector this class had, and #366's item asks for a check "consumed by the publication sites", which the cell layer does but the driver does not.

The consequence is caller-side half-update, which is what makes silence expensive rather than merely quiet:

- `child.rs:1072`: `schedule_restart` has already advanced `child.restarts` and `self.intensity`, and the driver goes on to arm the restart timer from `decision`. A rejection emits neither `Exited` nor `RestartScheduled`, and skips the `total_restarts` write — the driver restarts the child while the projection and lifecycle stream never record the exit.
- `child.rs:739`: `mailbox.bind(token, incarnation, ...)` and `dispatch_child_construction` have already run. A rejection means the incarnation runs with the record still reading `Admitted`/`Reserved` and no `Started` event. The later `terminalize_child` tolerates the missing residency with a `debug_assert!` only (`scope.rs:726`), so the nested scope's observation never closes — the exact stall class Batch 2 is filed against.
- `driver.rs:785`: `entry.promote(...)` already ran; the dynamic entry is promoted with no residency behind it.

So "rejection is a total no-op" is true of the cell layer and false of the system. Suggested remedy (cheap, and matches the `Acceptance` `#[must_use]` precedent AGENTS.md already cites): put `#[must_use]` on all six, and make each driver site state its handling explicitly — even `let _ = ...;` with a one-line comment naming why the transition is unreachable is better than an invisible discard. If a loud debug detector is wanted back without breaking the new tests, gate it on a flag the three legality tests set, or add `MemberRecord::would_accept` and `debug_assert!` at the *call sites* rather than in the reducer. **CONFIRMED** (mechanical: the returns are discarded; the reachability of any individual rejection is presumed nil, per Batch 3's premise).

## Low — the reorder moves the record mutation ahead of the recursive handoff, so a failed handoff now publishes a stage change

`crates/shelterwood-cells/src/cells/scope.rs:445-482`. The attempt closure applies `MemberTransition::Admitted` first, then `install_observation_gate_locked` + `adopt_descendant_observation_gates_locked`. `install_observation_gate_locked` ends in a real `assert!` ("a resident member must share its tree observation gate"), live in release. Before this PR the transition ran after adoption, so an unwind out of that assert left the record untouched. Now the record already reads `Admitted`, and `ObservationTxn::drop` → `commit()` flushes the queued pulse during the unwind (pinned by `gate.rs`'s `observation_txn_flushes_during_unwind_without_replacing_the_primary_panic`), publishing `Admitted` for a subtree that is only half re-homed.

Failure scenario is unwind-only and the assert is believed unreachable, hence Low. The clean shape recovers both properties at once: factor the matrix into `MemberTransition::is_legal_from(&MemberStage)` (or `MemberRecord::would_accept`), *probe* it under the child gate before touching anything, and apply the transition after the handoff succeeds. That is also what Codex asked for and it removes the need for the two-phase attempt closure entirely. **PLAUSIBLE**.

## Low — `try_adopt_observation_gate_with` duplicates the handoff-retry loop

`member.rs:468` and `member.rs:505` are the same capture/lock/re-read/retry loop with different early-return semantics. The old one is expressible as the new one: `try_adopt_observation_gate_with(gate, report, |current| { if !current.shares_gate(gate) { install(current) } true })`. This loop is the single trickiest concurrency shape in the crate and the one AGENTS.md singles out by name; two copies means the next fix has to find both. **CONFIRMED** (simplification, not a bug).

## Low — AGENTS.md/CLAUDE.md's gate-to-gate exemption still names only one function

AGENTS.md:98-106 (and the identical CLAUDE.md paragraph) states the exemption as "`adopt_observation_gate` takes the adoptee's *current* gate while the adopting parent's is held". This PR adds a second holder of that exemption (`try_adopt_observation_gate_with` / `ScopeCell::admit_observation_gate`) and, unlike the old one, it now runs a watch-value mutex acquisition (`record.modify_silently`) and a `RetainedExit` clone/drop under **both** gates. The nesting is still parent-to-child-only with the installed-pointer re-read, so the rule holds — but the normative paragraph should name the new function and record that a record mutation now happens inside the doubled section. Also worth carrying: `admit_observation_gate` drops the explanatory sentence its sibling at `scope.rs:411` carries ("Re-homing the parent would first have to acquire that same gate, so rereading its installed pointer here cannot race a parent handoff"), which is the actual justification for the `debug_assert!` above it. **CONFIRMED** (doc).

## Low — the live-dynamic-route `debug_assert!` lost its only test and is now unreachable by construction

`scope.rs:475`. `gate_handoff_rejects_a_scope_with_an_unadmitted_dynamic_reservation_without_rehoming` used to be the regression for that assert; after the retarget the rejection short-circuits before the assert is ever evaluated, so the assert has no coverage. Worse, it looks unreachable now: re-homing requires the `Admitted` transition to be legal, which requires stage `Reserved`, while a live dynamic route requires a started driver (stage `Starting`/`Running`). Either delete it as subsumed by the legality check, or keep a test that actually reaches it — an assert with no reachable trigger and no test is a comment. **PLAUSIBLE**.

## Checks that came back clean

- Legality matrix is byte-equivalent to the five deleted asserts (`Reserved→Admitted`; `Admitted|Restarting→Starting`; `Starting|Reserved→Running`; `Starting|Running→Stopping`; `Starting|Running|Stopping→RestartScheduled`). No silent tightening or loosening.
- Lock rule on the rejection paths: `member.rs:556-568` retains the incoming `Exit` before the watch lock; the rejected `MemberTransition` leaves via `txn.defer(dispose_detached(..))`, and the transient `RetainedExit` clone dropped under the gate is provably non-last because the deferred transition owns the sibling `Arc<ExitErrorInner>`. Same for the deferred `LifecycleEventKind` at `scope.rs:654` and the `(exited, scheduled)` pair at `scope.rs:677`. In the admission path the defer lands on the *destination* txn while the child's old gate guard is released inside `try_adopt_observation_gate_with`, so both guards are gone before the flush.
- Rejected `ResidentProjection`: `ResidentAdmission::new` defers disposal at the top of `admit_child_locked`, and the `Rc<RefCell<..>>` is still `Some` on the `return false`, so the whole graph goes to `dispose_detached`. The `Ref` borrow is released before commit (declaration order puts `projection`'s drop ahead of `child`'s), so no `already borrowed` panic.
- No new ABBA. `try_adopt_observation_gate_with` reproduces the capture → lock → re-read-installed-pointer → retry structure exactly; the already-on-destination-gate case runs `attempt` **without** re-acquiring, so no reentrant self-deadlock. `report_capture` still fires once per differing-gate iteration, so the existing `adoption_retries_when_the_captured_child_gate_has_already_been_replaced` probe is unaffected.
- Observable cut: `transition_child_stage` and `publish_child_restart` rejections publish nothing at all (no event, no snapshot chain, no `total_restarts`), and `admit_child_locked` rejects before `set_parent`, before the residency push and before `Added`. No split cut within the cell layer. Reordering the `total_restarts` write to after the transition is safe — same txn, and pulse order between two already-final records is not observable.
- Test non-vacuity, checked by mutating the implementation mentally: reverting `cb36a4e` (transition after handoff) breaks `rejected_nested_admission_preserves_original_parent_and_gate` on both the parent-identity and the descendant-gate assertions; accepting the illegal transition breaks the `transition(..) == false` and `MemberStage::Reserved` assertions; removing `txn.defer(dispose_detached(..))` breaks the `assert_ne!` on the retiring thread. `modify_silently` does not advance the watch version (`runtime/src/sync.rs:776`), so the `changed()`-stays-`Pending` assertion in `illegal_restart_transition_is_rejected_in_every_build` genuinely pins the absent pulse. All assertions are in test bodies, none inside an actor handler or `catch_unwind`.
- PR claims verified true: the new test does pin original residency (`has_resident_child`), parent identity (`Arc::ptr_eq` on `parent()`), and descendant gate identity; the live-route test did drop `#[cfg(debug_assertions)]` and `#[should_panic]`.
- No `specs/SPEC.md` contradiction found — the spec does not describe the member-stage reducer matrix.
- Nothing in this diff depends on #380's changes beyond the `ObservationTxn` commit hardening it introduces.

**chatgpt-codex-connector, P1 "Validate admission before rehoming the child" (`scope.rs:1171`) — CONFIRMED against the reviewed commit, ALREADY FIXED by `cb36a4e`.** The observation was correct at `a51e5ce`: `adopt_observation_gate` + `set_parent` ran unconditionally and the rejection came afterwards, so a rejected re-admission left the subtree on the destination gate and parent. Its AGENTS.md:98-106 citation is accurate this time (verified against the file) — no fabricated spec text. Two framing notes: the remedy it proposed ("validating before handoff, or rolling back") is the better of the two shapes and is still worth taking (see the Low finding above) — the PR instead chose transition-then-handoff, which fixes the rejection case but moves a record mutation ahead of the recursive install; and its "cross-tree lock acquisition can violate the required parent-to-child gate ordering" consequence overstates the pre-fix damage, since the stranded child was left on the *destination* gate with its old parent, which makes subsequent operations gate-inconsistent but does not by itself construct a child-to-parent acquisition. Severity P1 was right; the mechanism sentence was half right.

ralexstokes:
Addressed in `caccd58`. Finding by finding.

## Medium — discarded rejections — **fixed (cheap half) + filed (#392)**

The framing was right: the conversion made an illegal transition *silent in both* builds when it was supposed to make it fail closed in release *in addition to* aborting in debug. Both halves of the suggested remedy are in, plus a third piece.

**`#[must_use]` on every returning entry point.** `MemberCell::transition`, `MemberCell::transition_locked`, `ScopeCell::transition_child_stage`, `publish_child_restart`, `admit_child`, `admit_child_locked` — and a seventh the finding did not list: `ScopeCell::set_admitted_children` swallowed `admit_child_locked`'s verdict for every child, and it is production-reachable through `ScopeRuntime::publish_initial_children`. It now returns whether every child was admitted. Test call sites became `assert!(...)`, which is a small improvement in its own right: ~50 test setups now assert that their setup actually happened.

**A loud debug detector, at the call sites rather than in the reducer.** Each of the seven driver sites binds the verdict and `debug_assert!`s it with a comment naming the source stages it may present. This is not merely equivalent to the deleted reducer assertions — it is strictly better placed. The old ones panicked *inside* `record.modify_silently`, i.e. under the watch-value mutex and the observation gate, which is the shape AGENTS.md tells you not to write ("compute the verdict, release, *then* panic"). The new ones run outside every framework lock. The three cell-layer rejection tests are unaffected, since none of them goes through a driver site, so no flag or feature gate was needed.

**The partial-effect half is filed as #392**, and you are right that it is the expensive part. I did not fix it here, and I want to be explicit about why rather than hand-wave it:

- `child.rs` `publish_child_restart` — cannot be reordered at all. `publish_child_restart` consumes `decision`, which only exists because `schedule_restart` already charged `child.restarts` and `self.intensity`. Probing first would mean splitting `schedule_restart` into a derive half and a charge half.
- `child.rs` spawn `Starting` — reorderable in principle (`MemberCell::would_accept` now exists), but moving the transition ahead of `dispatch_child_construction` changes the publication order of `Started` relative to the body's first poll. That is observable and needs adjudication, not a mechanical reorder.
- `driver.rs` `admit_child_locked` after `entry.promote` — reorderable, but `AdmissionInstall` has no variant for "the reservation was refused by the projection reducer", and inventing one is inventing recovery semantics for an invariant violation.

So #392 records the three sites, the two candidate closures (structural unreachability via driver type state, or per-site reordering), and the fact that all three are presumed unreachable. Each site carries a comment pointing at it. **The PR description has been corrected** — the "total no-op" claim now says explicitly that it is layer-local and names the three sites.

## Low — record mutation ahead of the recursive handoff — **fixed, and taken in the shape you suggested**

`MemberTransition::is_legal_from` now carries the matrix; `MemberCell::would_accept` probes it. `ScopeCell::admit_observation_gate` and the member arm of `admit_child_locked` are now probe → handoff → apply. An unwind out of `install_observation_gate_locked`'s `assert!` leaves the record reading its pre-admission stage, which is the pre-PR property, while the rejection case still never touches the subtree, which is `cb36a4e`'s property. Both at once, as you predicted.

The reason the probe/apply split is not a TOCTOU: the child gate is held across both (by the caller when the gates already match, by `with_handoff_gate` otherwise), and every stage writer needs an `ObservationTxn` on that gate. The `debug_assert!` on the apply says so. This is written down on `would_accept` and on `admit_observation_gate`.

One thing this does *not* have is a test for the unwind window itself — reaching `install_observation_gate_locked`'s `assert!` requires an inconsistent installed-gate pointer that no constructible sequence produces. The ordering is pinned indirectly: moving the probe back after the handoff fails both `rejected_nested_admission_preserves_original_parent_and_gate` and `gate_handoff_rejects_a_scope_with_an_unadmitted_dynamic_reservation_without_rehoming` (mutation-checked).

## Low — duplicated handoff-retry loop — **fixed**

One loop now: `MemberCell::with_handoff_gate`. `adopt_observation_gate_with` is expressed on top of it exactly as you sketched, with a `debug_assert!` that unconditional adoption never refuses. `report_capture` still fires once per differing-gate iteration, so `adoption_retries_when_the_captured_child_gate_has_already_been_replaced` still probes what it probed.

## Low — AGENTS.md gate-to-gate exemption — **fixed**

(`CLAUDE.md` is a symlink to `AGENTS.md`, so one edit covers both.) The paragraph now names `MemberCell::with_handoff_gate` as the exempt primitive, names both of its users, carries the justifying sentence its sibling had (`admit_observation_gate` has it back in the code too), and records that admission widens the doubled section: the record's watch-value mutex is taken inside it and the transition reaches `RetainedExit` clone/drop under both gates. It also says why that still satisfies the rule — framework-owned data, provably non-last clone — and that a new operation added to the doubled section needs the same accounting.

## Low — the live-dynamic-route `debug_assert!` — **fixed, both halves**

Your reachability argument is correct for the *admission* path, and stronger than "unreached": re-homing requires the `Admitted` transition to be legal, which requires `Reserved`, while a live route requires a started driver. So the assert is removed from `admit_observation_gate` and replaced by a comment stating that derivation.

But the assert did not disappear — it survives on `ScopeCell::adopt_observation_gate`, the reservation-time path, which has no legality probe. That copy is the one this PR's retarget actually left uncovered, and it now has a regression: `plain_gate_adoption_rejects_a_scope_with_a_live_dynamic_route` drives `adopt_child_observation_gate` on a scope carrying a route. Mutation-checked (delete the assert ⇒ test fails). Separately, `gate_handoff_rejects_a_scope_with_an_unadmitted_dynamic_reservation_without_rehoming` now pins the refusal at the API boundary too, since `set_admitted_children` reports it: `assert!(!root.set_admitted_children(...))`.

## Adjudication note

Your reading of the Codex P1 is the one I acted on — its remedy (validate before handoff) was the better shape and is now taken, and its "cross-tree lock acquisition" consequence sentence does overstate the pre-fix damage.

## Verification

- `./tools/dev just ci` — clean (811 tests)
- `cargo nextest run --workspace --release` — 807 pass. One failure, `nonresident_terminal_exit_is_retained_before_the_residency_assertion`, is a **base-branch artifact and not from this PR**: it depends on a `debug_assert!` firing, and #380's tip already gates it on `debug_assertions`; this branch is based on an older #380 commit and is deliberately not rebased while #380 is being edited concurrently. It reproduces on `cb36a4e` unmodified.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean
- Mutation checks (commit-then-break-then-revert): accept-everything `is_legal_from` fails all five rejection tests; handoff-before-probe fails the two subtree tests; `let _ =` on the per-child verdict in `set_admitted_children` fails the driver test; deleting the surviving live-route assert fails its new regression.

PR #384, opened 2026-08-20
Split mailbox cell support modules

ralexstokes:
Addresses the behavior-neutral mailbox-cell structural follow-up from #366, intended to land after the four behavioral batches.

Mechanically moves the embedded mailbox suite and the waker storage module out of the 3,294-line cell file, preserving test names and visibility. It also removes `MailboxTxn`'s broad mutable dereference in favour of named binding, waiter, payload, and bind-token transitions, without changing protocol behavior. `parts()` still hands the state out for the four sites that mutate it beside their own effects, so the invariant the removal buys is narrower than "every mutation is named": there is no longer any way to mutate the state without the effects sink in scope.

Review follow-up (46067ac) corrects that doc claim, drops the two redundant `#[path]` attributes, and routes `take_payload` through `MailboxPayload<M>` so panic containment for unread user messages lives in the type rather than in both call sites' convention.

Validation: `./tools/dev just ci` (803 tests).

### DIFF
--- a/crates/shelterwood-mailbox/src/cell.rs
+++ b/crates/shelterwood-mailbox/src/cell.rs
@@ -2,7 +2,7 @@ use std::{
     collections::{BTreeMap, VecDeque},
     fmt,
     num::NonZeroUsize,
-    ops::{Deref, DerefMut},
+    ops::Deref,
     sync::{
         Arc, Mutex, MutexGuard,
         atomic::{AtomicBool, Ordering},
@@ -108,91 +108,8 @@ pub(super) struct OperationState<M> {
     registration: Option<WaiterId>,
 }
 
-mod waker_slot {
-    use std::{sync::Arc, task::Waker};
-
-    use crate::{MailboxRuntime, capability::dispose, panic::PanicAccumulator};
-
-    /// The only storage surface for a caller-owned waker.
-    ///
-    /// Its value is private even from the parent mailbox module, and every
-    /// mutating operation requires an effects sink, so replacing or taking an
-    /// `Option<Waker>` and accidentally dropping it beside a guard does not
-    /// type-check.
-    #[derive(Default)]
-    pub(super) struct WakerSlot(Option<Waker>);
-
-    pub(super) enum WakerAction {
-        Wake,
-        DropInline,
-        Dispose(Arc<dyn MailboxRuntime>),
-    }
-
-    enum WakerEffect {
-        Wake(Waker),
-        DropInline(Waker),
-        Dispose(Arc<dyn MailboxRuntime>, Waker),
-    }
-
-    #[derive(Default)]
-    pub(super) struct WakerEffects(Vec<WakerEffect>);
-
-    impl WakerSlot {
-        pub(super) fn will_wake(&self, waker: &Waker) -> bool {
-            self.0
-                .as_ref()
-                .is_some_and(|registered| registered.will_wake(waker))
-        }
-
-        pub(super) fn replace(&mut self, waker: Waker, effects: &mut WakerEffects) {
-            if let Some(displaced) = self.0.replace(waker) {
-                effects.push(displaced, WakerAction::DropInline);
-            }
-        }
-
-        pub(super) fn take(&mut self, action: WakerAction, effects: &mut WakerEffects) {
-            if let Some(waker) = self.0.take() {
-                effects.push(waker, action);
-            }
-        }
-
-        pub(super) fn is_empty(&self) -> bool {
-            self.0.is_none()
-        }
-    }
-
-    impl WakerEffects {
-        pub(super) fn is_empty(&self) -> bool {
-            self.0.is_empty()
-        }
-
-        fn push(&mut self, waker: Waker, action: WakerAction) {
-            self.0.push(match action {
-                WakerAction::Wake => WakerEffect::Wake(waker),
-                WakerAction::DropInline => WakerEffect::DropInline(waker),
-                WakerAction::Dispose(runtime) => WakerEffect::Dispose(runtime, waker),
-            });
-        }
-
-        pub(super) fn flush(&mut self, panics: &mut PanicAccumulator) {
-            for effect in self.0.drain(..) {
-                match effect {
-                    WakerEffect::Wake(waker) => panics.run(|| waker.wake()),
-                    WakerEffect::DropInline(waker) => panics.run(|| drop(waker)),
-                    WakerEffect::Dispose(runtime, waker) => {
-                        panics.run(|| dispose(&runtime, waker));
-                    }
-                }
-            }
-        }
-    }
-
-    impl Drop for WakerEffects {
-        fn drop(&mut self) {
-            self.flush(&mut PanicAccumulator::default());
-        }
-    }
-}
+#[path = "cell/waker_slot.rs"]
+mod waker_slot;
 
 use waker_slot::{WakerAction, WakerEffects, WakerSlot};
 
@@ -765,6 +682,11 @@ impl<M: Send + 'static> MailboxEffectBatch<M> {
 }
 
 /// A mailbox transition guard paired with its mandatory post-unlock effects.
+///
+/// It exposes immutable state through `Deref`, but every mutation is a named
+/// operation on the transaction. This keeps binding, waiter-domain, payload,
+/// and bind-token transitions coupled to the effects sink instead of exposing
+/// a broad mutable state escape hatch.
 struct MailboxTxn<'a, 's, M: Send + 'static> {
     state: Option<MutexGuard<'a, MailboxState<M>>>,
     effects: MailboxEffects<'a, 's, M>,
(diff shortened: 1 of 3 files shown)


ralexstokes:
**Verdict: behavior-neutral as claimed — no visibility erosion, no lost test, no lock-rule regression. Three low-severity hygiene findings, none blocking.**

## Verification performed (the claims this PR has to earn)

- **`waker_slot` move is byte-identical.** Dedenting the old inline `mod waker_slot { … }` (`main:cell.rs:111-196`) and diffing against `cell/waker_slot.rs` yields exactly one difference: the removed closing brace. Every visibility is preserved verbatim — `pub(super) struct WakerSlot`, `pub(super) enum WakerAction`, private `enum WakerEffect`, `pub(super) struct WakerEffects`, and critically the tuple field `WakerSlot(Option<Waker>)` stays **unannotated**, i.e. private to `waker_slot` and unreachable from `cell`. The module's parent is still `cell` in both arrangements, so `pub(super)` denotes the same set. CLAUDE.md's "its storage is private even from the parent module" still holds structurally, and no operation returns or replaces a `Waker` without a `WakerEffects` sink. Nothing outside `cell.rs` names `WakerSlot`.
- **No test lost or renamed.** Attribute multiset is identical old vs. new (29 × `#[test]`, 1 × `#[crate::runtime::test(start_paused = true)]`); the `fn` name set is identical (65 names, `diff` clean); the entire tests diff after dedent is pure rustfmt reflow, nothing else. `cargo nextest list -p shelterwood-mailbox --lib` on this branch collects **30** `cell::tests::*` tests, matching the attribute count. `mod tests` keeps `pub(super)`, so `futures.rs:704` and `reply.rs:167` still resolve their helper imports.
- **The fmt gate does cover the new files.** Probed empirically (repo has history here): injected malformed `fn`s into both `cell/tests.rs` and `cell/waker_slot.rs`, and `cargo +nightly fmt --all --check` reported a diff in each. `#[path]` + `#[cfg(test)]` is not a rustfmt blind spot here.
- **`MailboxTxn` field order and `Drop` are untouched** — `state: Option<MutexGuard<…>>` still declared before `effects`, and `Drop` still `drop(self.state.take())` before Rust drops the sink. The unwind flush still runs with no mailbox mutex held.
- **The named transitions match the old inline mutations statement-for-statement**, including `freeze_after_exhaustion`'s deliberate bypass of `replace_binding`'s empty-domain check and `take_next`'s eligibility/`.flatten()` shape. The only reordering is in `close`: the bind token is now minted before `take_payload`/`pulse` instead of after. Inert — everything is under one guard and `pulse()` only queues into the deferred sink.
- **No accidental `Deref` shadowing.** `MailboxState`'s method set is `status`, `current_observation`, `replace_binding`, `park`, `take_waiters`, `remove_waiter`, `waiters`; the only overlaps with the new inherent methods (`park`, `take_waiters`, `remove_waiter`) are straight forwarders, and no `&self` method is shadowed by a differently-behaving `&mut self` inherent one.

## Findings

### 1. LOW — doc comment overstates what the removal bought (`cell.rs:684-689`) — CONFIRMED

"every mutation is a named operation on the transaction … instead of exposing a broad mutable state escape hatch" is not what the type does. `MailboxTxn::parts()` (`cell.rs:716`) still hands out `&mut MailboxState<M>`, and four call sites use it (`cell.rs:1052`, `1123`, `1167`, `1365`); `bind` even destructures it (`let MailboxState { queue, latest, .. } = state;`). Any of those could call `replace_binding` or write `queue`/`latest` directly.

The invariant that *does* hold after this PR — and it is a real tightening — is narrower: **no mutation without the effects sink in scope**. `DerefMut` allowed sinkless mutation from anywhere in `cell`; `parts()` cannot. Failure scenario is documentary, not runtime: a future reader trusts the comment and assumes the transition list is exhaustive when auditing what can mutate under the guard — exactly the doc-vs-code drift #366 has been spending PRs correcting. Remedy: say that mutation is either a named transition or a `parts()` pairing that hands the state beside its sink.

### 2. LOW — redundant `#[path]` attributes deviate from the repo's own convention (`cell.rs:111`, `cell.rs:1713`) — CONFIRMED

`src/cell.rs` already resolves `mod waker_slot;` / `mod tests;` to `src/cell/*.rs` with no attribute. Verified by deleting both attributes in a scratch worktree: `cargo check --locked -p shelterwood-mailbox --all-targets` passes. The established shape in this repo is exactly the bare form — `crates/shelterwood/src/driver.rs:1312-1313` is `#[cfg(test)] mod tests;` beside `driver/tests.rs`, and `raw`/`tree`/`cells` are the same; the only `#[path]` in the whole workspace is `crates/shelterwood/tests/common/main.rs`, where a test-target quirk forces it. Cost of keeping it: `#[path]` decouples module path from file path, so a future rename can silently leave a stale file, and it is one more construct for a tool to mis-resolve. Remedy: delete both attributes.

### 3. LOW — `take_payload` names the transition but hands back the least-typed carrier (`cell.rs:851-857`) — PLAUSIBLE (hygiene, not a regression)

It returns a bare `(VecDeque<Envelope<M>>, Option<Envelope<M>>)` of user-owned messages while the guard is held. Both call sites (`cell.rs:1437`, `cell.rs:1472`) immediately wrap it in the identical `MailboxPayload { queue: Some(queue), latest, retired: Vec::new() }`, whose `Drop` (`cell.rs:922`) is the thing that runs each user destructor through a `PanicAccumulator`. The raw tuple has no such protection and no `#[must_use]`: a call site that dropped the temporary — `transaction.take_payload();` compiles — would destroy user messages under the mailbox mutex with a bare, uncontained destructor panic. Not a regression (pre-PR the same values came out of inline `mem::take`), but this PR is precisely the one narrowing these transitions, and the codebase's "invalid states unconstructible" line argues for closing it here. Remedy: return `MailboxPayload<M>` directly (deduplicating both construction sites), or at minimum `#[must_use]`. Same argument, weaker, for `take_next`'s `Option<Envelope<M>>`.

### Nit — 14 copies of one `expect`

Each new transition re-derives `self.state.as_deref_mut().expect("a live mailbox transaction retains its guard")`; the string now appears 16 times in the file. `parts()` already computes exactly that as its first element, so `fn state_mut(&mut self) -> &mut MailboxState<M> { self.parts().0 }` (or the inverse factoring) collapses the repetition without widening anything — `state_mut` would be private to `cell` and reach no further than `parts()` already does. Related altitude nit: `take_next`'s eligibility match is pure `MailboxState` logic (it is the only place `MailboxTxn` knows about `queue`/`latest`/`kind` internals) and would sit more naturally as a `MailboxState::take_next` forwarder, like `park`/`take_waiters`/`remove_waiter`.

ralexstokes:
All three findings taken in 46067ac. Nothing deferred, nothing refuted.

### 1 — doc comment overstates the removal (`cell.rs:684-689`) — **fixed**

Confirmed: `parts()` mutates at four sites, so "every mutation is a named operation" was false. The comment now states the invariant that actually holds — mutation is either a named transition or a `parts()` pairing that hands the state out beside its sink, so no mutation happens without the effects sink in scope, which is exactly what `DerefMut` used to allow. The PR body repeated a softer version of the same claim; it is updated too, with the narrowing spelled out.

### 2 — redundant `#[path]` attributes — **fixed**

Both deleted; `mod waker_slot;` / `#[cfg(test)] pub(super) mod tests;` resolve to the same files, matching `driver.rs` + `driver/tests.rs`.

Re-probed the fmt gate afterwards, since the attribute was the only reason to suspect coverage in the first place and this repo has a history here. Appended a deliberately malformed `fn` to each file on the committed tree; `cargo +nightly fmt --all --check` reported `Diff in …/cell/tests.rs:1565` and `Diff in …/cell/waker_slot.rs:82`. Reverted with `git checkout`. Both files remain covered without the attributes — `#[cfg(test)]` does not hide the test module from rustfmt.

### 3 — `take_payload` hands back the least-typed carrier — **fixed**

`take_payload` now returns `MailboxPayload<M>` and carries `#[must_use]`. Added `MailboxPayload::unread(queue, latest)`, which collapses all three construction sites — the two `take_payload` callers (`close`, `prepare_termination`) plus the isolated-displaced construction in `MailboxEffectBatch::flush`. Panic containment for unread user messages is now a property of the return type instead of a convention both callers happen to follow.

Mutation-checked the fence rather than trusting it, on the committed tree: rewriting `close` to call `transaction.take_payload();` for effect produces `warning: unused return value of MailboxTxn::take_payload that must be used`, which CI's `-D warnings` makes an error; deleting the `#[must_use]` and rebuilding the same mutation produces zero warnings, so the attribute is load-bearing and not incidentally covered by the `Drop` impl. Both probes reverted.

On the weaker half of the finding: `Option` is **not** `#[must_use]` in std (verified with a scratch `fn f() -> Option<D>` where `D: Drop` — `f();` compiles silently), so `take_next` really was unfenced. Applied the rule uniformly to the transitions that hand a user-reachable value out from under the guard: `take_next`, `take_waiters` (a dropped `WaiterQueue` can release the last `Arc<SendOperation>` owning a user message), and `remove_waiter`. Not applied to `reset_bind_permit`, which returns a framework token.

No test added: the change is a type-level move with no new runtime behavior to pin — both call sites already built the identical `MailboxPayload`. The fence is pinned by the compile-time mutation check above instead.

### Nits — both applied, they shrink the diff

- `MailboxTxn::state_mut(&mut self) -> &mut MailboxState<M> { self.parts().0 }` collapses **14** copies of the guard `expect`; the string is now down to 1 occurrence (in `parts`) plus the `Deref` read. Nothing widens — `state_mut` is private and reaches no further than `parts()` already did.
- `take_next`'s eligibility match moved down to `MailboxState::take_next`, beside `park`/`take_waiters`/`remove_waiter`; `MailboxTxn::take_next` is now a forwarder like its siblings. `MailboxTxn` no longer names `queue`/`latest`/`kind` internals anywhere except `take_payload`'s `mem::take`.

Net: **69 insertions, 95 deletions** in `cell.rs`.

### Behavior neutrality, re-verified the way the review did

- **`fn` name set**: identical modulo exactly three intended additions — `state_mut`, `unread`, and the second `take_next` (the `MailboxState` method the `MailboxTxn` forwarder now calls). No removals, no renames.
- **`#[test]` attribute multiset**: byte-identical — 29 × `#[test]`, 1 × `#[crate::runtime::test(start_paused = true)]`.
- **`nextest list -p shelterwood-mailbox --lib`**: `cell::tests::*` collects the same **30** tests, `diff` clean against the pre-change listing.
- `cargo nextest run --workspace`: **803 passed, 0 skipped**. `cargo clippy --locked --workspace --all-targets -- -D warnings` clean. `cargo +nightly fmt --all --check` clean.
- Lock rule untouched: `MailboxTxn`'s field order and `Drop` are unchanged, `state_mut` routes through `parts()` rather than around it, `WakerSlot`'s storage stays private to `waker_slot`, and the only new drop-glue is `MailboxPayload`'s, which already existed on the same values.

PR #387, opened 2026-08-20
Split raw context timer storage

ralexstokes:
Addresses the behavior-neutral raw-context structural follow-up from #366, intended to land after the four behavioral batches.

Moves the keyed timer store and its dedicated tests from the actor context into a focused `raw::context::timers` module. The containment, indexing, deadline, interval-rearm, and detached-disposal behavior is unchanged.

The seam `raw::context` receives is four types — `TimerStore`, `TimerMessage`, `ArmingOrder`, `IntervalRearm` — and ten `TimerStore` methods. **No fields.** `TimerEntry`, `KeyHash`, `TimerLocation` and every struct field in the module are private to `timers`, so the parent can neither construct an arming order nor take ownership of a timer's user key. The store owns the arming-order counter it indexes with, and `take_due_once` retires a delivered timer's key through `RawDisposal` internally rather than handing an entry across the module boundary — which makes the detached-disposal invariant structural rather than a convention the parent has to remember.

Validation: `./tools/dev just ci` (804 tests).


ralexstokes:
**Verdict: no correctness findings — the move is genuinely byte-for-byte behavior-neutral; four low/nit-severity notes, one of which is a literal mismatch with the PR body's visibility claim.**

### Verification performed

Reconstructed the removed and added line sets from the diff and diffed them against each other. The *entire* delta between the block deleted from `context.rs` and the block added as `timers.rs` is: the new `use std::{...}` / `use super::super::disposal::{...}` header, nine `pub(super)` markers, one rustfmt signature reflow (`rearm_interval`), one added and one removed blank line, and the test-module rename. Nothing else — no reordered statements, no changed early returns, no `&mut`-to-clone, no borrow-split or drop-timing change. The `context.rs` side has exactly four hunks: the two import trims, `mod timers;` + `use timers::{…}`, and two pure deletions with zero `+` lines.

- **The five claimed behaviors** hold individually, by identity: containment (`replace` / `clear_and_dispose` still wrap key *and* message in `Contained` before any user `Hash`/`Eq` runs), indexing (`keyed`/`armings`/`deadlines` + `unlink`'s `debug_assert_eq!` pairing), deadline (`take_due` / `arm_deadline` / `next_deadline`), interval rearm (`rearm_interval` still clones before mutating `entry.deadline`, preserving the retry-the-arming property), detached disposal (`dispose_entry`, `Drop for TimerStore → clear`, and the parent's `disposal.dispose(key)` in `deliver_timer`).
- **Lock rule**: `TimerStore` takes no mutex and none is held across it — user `Hash`/`Eq`/`Clone` on timer keys run in `hash_key`, the `locate` predicate and `rearm_interval` exactly as before. Unchanged code, unchanged venue.
- **Test fidelity**: the sorted set of all `fn` names across `context.rs` on `main` vs `context.rs` + `timers.rs` on the branch is **identical**. `#[test]` count 28 → 20 + 8. No test renamed, dropped, or un-compiled: `mod timers;` is unconditional, `#[cfg(test)] mod tests` is inside it, and the moved `use super::{ArmingOrder, IntervalRearm, TimerMessage, TimerStore};` happens to name only timer types, so re-rooting `super` from `raw::context` to `raw::context::timers` resolves to the same items. The `timer_store_tests` → `tests` rename changes the nextest path `raw::context::timer_store_tests::*` → `raw::context::timers::tests::*`; grepped the tree — nothing references either path, so no repeat of the stale-`--exact`-filter vacuity from #366's ordering note.
- **fmt gate coverage** (checked empirically, given this repo's history): appended a deliberately misformatted `fn` to `timers.rs` in a scratch worktree and ran `cargo +nightly fmt --all --check` — it flagged the file. The new path is inside the gate.
- No `specs/` or `docs/` reference to `raw/context.rs` or the timer store, so nothing to re-point.

---

### 1. LOW — the seam exposes three fields, not just "narrow types and methods"

`crates/shelterwood/src/raw/context/timers.rs:16,25,31`

```rust
pub(super) struct ArmingOrder(pub(super) u64);
pub(super) struct TimerEntry<M> {
    pub(super) key: Box<dyn Any + Send>,
    ...
    pub(super) message: TimerMessage<M>,
```

The PR body says "the parent receives only the narrow timer types and methods it consumes". Taken literally that is not what shipped: three previously-private *fields* widened to `pub(super)` to make the split compile. Scope is correctly bounded — `mod timers` is private, `KeyHash`/`TimerLocation`/`deadline`/`arming_order`/`period`/`lookup_probes` stayed private, and `TimerEntry`'s remaining private fields mean the parent can destructure but never *construct* one — so nothing outside `raw::context` gains reach and there is no live defect.

What it costs is the discipline's enforceability. `TimerEntry` has no `Drop` impl; the only thing that routes a timer's user key and user message to `RawDisposal` is `TimerStore::dispose_entry`, which is private. `remove_arming` hands a bare `TimerEntry` across the module boundary, and today `deliver_timer` (`context.rs:1209-1217`) correctly does `disposal.dispose(key)` by hand. A future edit on the parent side that instead drops that entry — or takes an early return between `remove_arming` and the `dispose` call — destroys a user key and a user message bare on the actor task, which is the exact venue the detached-disposal convention exists to avoid. Before the split that was a same-file hazard the store could see; now the invariant is split across a module boundary with a field-level hole in it.

Remedy (behavior-preserving, and it would let all three fields go back to private):

- give the store the operation the parent actually wants — `pub(super) fn take_due_once(&mut self, arming: ArmingOrder) -> Option<M>` that unlinks, disposes the key, and returns the one-shot message, absorbing `deliver_timer`'s `unreachable!("a non-interval timer must own a one-shot message")`. That also lets `TimerEntry` and `TimerMessage`'s variants stay private, with `replace_once` / `replace_interval` replacing the constructed-by-the-parent `TimerMessage`.
- move `timer_orders: PoisonedCounter` (`context.rs:356`) into `TimerStore` and have `replace` mint and return its own `ArmingOrder`. It is the store's identity space; the only non-test parent use of the inner `u64` is the single construction at `context.rs:897`. Then `ArmingOrder`'s field can be private with a `#[cfg(test)]` constructor for the eight test sites.

CONFIRMED as a visibility widening and a claim mismatch; PLAUSIBLE as a design objection — reasonable to defer as follow-up rather than reopen this PR, but the PR body's sentence should be corrected either way.

### 2. NIT — `super::super::` in non-test source

`crates/shelterwood/src/raw/context/timers.rs:8` — `use super::super::disposal::{Contained, RawDisposal};`. All 47 occurrences of `super::super` in the workspace today are inside test modules; no non-test source file uses it. `raw/mod.rs:5` declares `mod disposal;`, so `use crate::raw::disposal::{Contained, RawDisposal};` is available and reads as an address rather than a hop count.

### 3. NIT — no module doc

`timers.rs` opens straight into `use std::{…}`. Every other file in `crates/shelterwood/src/raw/` carries a `//!` header (`mod.rs`, `context.rs`, `definition.rs`, `disposal.rs`, `offload.rs`). The `TimerStore` doc comment has the substance; one line at the top keeps the local convention.

### 4. NIT — `mod` declaration placement

`context.rs:34` puts `mod timers;` after the whole `use` block including a `#[cfg(test)] use`. The sibling `driver.rs` declares its submodules first (lines 3-8). Cosmetic.

ralexstokes:
All four addressed in `4408e22`. Taking the remedy for finding 1 rather than deferring it: this *is* the structural PR, and the field-level hole is squarely its subject matter.

### 1. LOW — the seam exposes three fields — **fixed, both halves of the proposed remedy**

Confirmed as written. `remove_arming` handing a bare `TimerEntry` across the module boundary was the substantive part: the only thing keeping a delivered timer's user key off the actor task was a hand-written `disposal.dispose(key)` at the parent call site, with nothing structural to stop a later edit from dropping the entry or returning early between the two.

**`take_due_once`** (`timers.rs:284`) replaces `remove_arming` at the seam. It unlinks, retires the key through disposal, and yields the one-shot message — absorbing `deliver_timer`'s destructure and its `unreachable!("a non-interval timer must own a one-shot message")`. `deliver_timer` (`context.rs:1194`) is now the single `.expect` on that call. `TimerEntry` and both its widened fields are private again; the type is no longer named outside `timers`.

**The arming-order counter moved into the store.** `timer_orders: PoisonedCounter` is gone from `RawResources`; `TimerStore` owns `arming_orders` and `replace` mints and returns its own `ArmingOrder`, so the inner `u64` is private. This came out cheaper than the sketch: it needed no `#[cfg(test)]` constructor. `mod tests` is a *descendant* of `timers`, so the eight in-module sites could just bind `replace`'s return, and the `order()` helper retired with them. The six `ArmingOrder(1)` sites in `context.rs`'s tests bind the return too.

The seam is now four types (`TimerStore`, `TimerMessage`, `ArmingOrder`, `IntervalRearm`) and ten methods, with **zero `pub(super)` fields** — `grep -n 'pub(super)' timers.rs` returns only item declarations. The parent can no longer fabricate, reuse, or install an arming order out of band either, which makes the `armings`-map uniqueness invariant structural alongside the disposal one. PR body corrected to describe the seam literally instead of by adjective.

I left `TimerMessage`'s variants reachable: the parent constructs `Once`/`Interval` in `set_timeout`/`set_interval` and consumes `IntervalRearm::Interval`. Splitting `replace` into `replace_once`/`replace_interval` would duplicate the shared `replace_timer` body to hide an enum that carries no ownership hazard — variants are a type's interface, not a field hole.

**Pinned, not just structural.** New test `one_shot_delivery_retires_its_key_through_disposal` (`timers.rs:732`) arms a one-shot under a key whose destructor panics, calls `take_due_once`, and asserts the panic landed in `disposal.panic` as cleanup evidence rather than unwinding into the caller. Mutation-checked on the committed tree: replacing `self.disposal.dispose(key)` with `drop(key)` fails it with `timer key destructor panic` escaping at the call site; reverted with `git checkout`.

### 2. NIT — `super::super::` in non-test source — **fixed**

`timers.rs:15` now `use crate::{identity::PoisonedCounter, raw::disposal::{Contained, RawDisposal}};`. (`super::disposal` isn't available from a grandchild, and `crate::raw::disposal` resolves since `mod disposal` is private-to-`raw` and `timers` is a descendant.)

### 3. NIT — no module doc — **fixed**

Six-line `//!` header at `timers.rs:1`, stating the ownership rule the new seam now enforces.

### 4. NIT — `mod` placement — **fixed**

`mod timers;` moved to `context.rs:3`, directly under the `//!` header, matching `driver.rs:3-8`.

---

### Behavior-neutrality re-verification (post-edit)

Re-ran your method against `main`:

- **`fn`-name set**, `context.rs` on `main` vs `context.rs` + `timers.rs` on the branch — delta is exactly six entries, all intended: `-order` / `+one_shot_delivery_retires_its_key_through_disposal`, `-remove_arming` / `+take_due_once`, and `+drop`/`+eq`/`+hash` from the new `PanickingDropKey` test fixture's trait impls. Nothing else moved.
- **Test names**: all 32 base `#[test]`/`#[crate::runtime::test]` names preserved verbatim, one added. 28 → 20 + 9 in the two files.
- **Statement order / early returns / drop timing**, walked on the two behavioral hunks:
  - `deliver_timer`: `locate_arming` → `unlink` → `dispose(key)` → `Once` match → `Some(message)`, in that order, same as `main`; the `.expect` fires on the same condition (`locate_arming` miss). `self.resources.disposal` and `self.resources.timers.disposal` are clones of the same `RawDisposal`, so the destination is unchanged.
  - `replace_timer`: the mint moved into `TimerStore::replace`, still ahead of `Contained::new` and therefore ahead of all user `Hash`/`Eq`. The only reordering is that `runtime::now()` / `Deadline::after` now precede it — both non-user, non-panicking. The exhaustion-panic path leaves key and message uncontained exactly as on `main`, so even that unreachable edge is unchanged.
- **Lock rule**: unchanged venue. `TimerStore` still takes no mutex and none is held across it; `take_due_once` adds no user code inside anything, and the key it destroys goes to `RawDisposal`, which is the whole point of the change.
- `./tools/dev just ci` green — 804 tests (803 + the one added). Clippy `-D warnings` clean on `--workspace --all-targets --all-features`; nothing new was silenced.

Nothing deferred, nothing filed.

PR #388, opened 2026-08-20
Terminalize never-started exit dispatch

ralexstokes:
Addresses the defensive `NeverStarted` exit-dispatch follow-up from #366.

Makes `ExitKind::NeverStarted` explicitly terminal before restart-policy evaluation, so a caller that accidentally routes that membership-only outcome through incarnation dispatch cannot have `RestartCondition::Always` schedule a restart for an incarnation that never ran. The `debug_assert!` that previously stood there is kept above the release return, so the misuse stays loud in debug and fails closed in release — #366 Batch 4's "fail closed in release *in addition to* aborting in debug" direction, not instead of it. The arm carries a comment citing SPEC §8.

The regression test is split into profile halves: a debug `#[should_panic]` twin and a release assertion that the same input returns `ExitDispatch::Terminal`. Both are mutation-checked — deleting the guard fails each in its own profile.

Validation: `./tools/dev just ci` (821 tests), workspace `cargo nextest` in debug, the focused regression plus `-p shelterwood-core` in both debug and release, and `cargo +nightly clippy --all-targets --all-features` on the touched crate in both profiles.

### DIFF
--- a/crates/shelterwood-core/src/engine.rs
+++ b/crates/shelterwood-core/src/engine.rs
@@ -251,6 +251,25 @@ pub fn dispatch_exit(
     scope: ScopeMode,
     membership: MembershipStatus,
 ) -> ExitDispatch {
+    // SPEC §8: `NeverStarted` is a membership fact, not an incarnation
+    // verdict — it "is never input to restart or intensity accounting".
+    // Every supported producer terminalizes the membership directly and never
+    // reaches dispatch, so arriving here is a framework bug: the assert makes
+    // it loud in debug, and the return fails closed in release rather than
+    // charging a restart for an incarnation that never ran.
+    //
+    // Failing closed is the lesser of two wrong outcomes, not a correct one:
+    // the sole dispatch caller holds a real incarnation, so a `Terminal`
+    // verdict here would publish a `NeverStarted` terminal paired with
+    // `last_incarnation: Some(..)`, which SPEC §B.3 excludes. `ExitDispatch`
+    // carries no incarnation, so the pairing is the caller's to choose and
+    // cannot be repaired from inside this function; an inconsistent terminal
+    // projection is still strictly better than an illegal restart, and the
+    // debug abort fires before any such projection can be published.
+    debug_assert!(
+        !matches!(exit.kind(), ExitKind::NeverStarted),
+        "NeverStarted is a membership outcome outside incarnation dispatch"
+    );
     if matches!(exit.kind(), ExitKind::NeverStarted) {
         return ExitDispatch::Terminal;
     }
@@ -1439,6 +1458,25 @@ mod tests {
         }
     }
 
+    /// Debug half of the two-profile guard: the misuse aborts loudly.
+    #[cfg(debug_assertions)]
+    #[test]
+    #[should_panic(expected = "NeverStarted is a membership outcome outside incarnation dispatch")]
+    fn funnel_dispatch_rejects_the_membership_level_never_started_outcome() {
+        let exit = Exit::never_started();
+        let _ = dispatch_exit(
+            &exit,
+            RestartPolicy::new(RestartCondition::Always, Backoff::Immediate),
+            ScopeMode::Running,
+            MembershipStatus::Active,
+        );
+    }
+
+    /// Release half: with the assert compiled out the same input fails closed
+    /// on `Terminal` instead of charging a restart. `RestartCondition::Always`
+    /// is the one condition whose `should_restart` returns `true`
+    /// unconditionally, so removing the guard flips this assertion.
+    #[cfg(not(debug_assertions))]
     #[test]
     fn funnel_dispatch_treats_the_membership_level_never_started_outcome_as_terminal() {
         let exit = Exit::never_started();
--- a/crates/shelterwood-core/src/engine.rs
+++ b/crates/shelterwood-core/src/engine.rs
@@ -251,10 +251,9 @@ pub fn dispatch_exit(
     scope: ScopeMode,
     membership: MembershipStatus,
 ) -> ExitDispatch {
-    debug_assert!(
-        !matches!(exit.kind(), ExitKind::NeverStarted),
-        "NeverStarted is a membership outcome outside incarnation dispatch"
-    );
+    if matches!(exit.kind(), ExitKind::NeverStarted) {
+        return ExitDispatch::Terminal;
+    }
     if scope == ScopeMode::Draining || membership == MembershipStatus::Removing {
         return ExitDispatch::Terminal;
     }
@@ -1440,16 +1439,17 @@ mod tests {
         }
     }
 
-    #[cfg(debug_assertions)]
     #[test]
-    #[should_panic(expected = "NeverStarted is a membership outcome outside incarnation dispatch")]
-    fn funnel_dispatch_rejects_the_membership_level_never_started_outcome() {
+    fn funnel_dispatch_treats_the_membership_level_never_started_outcome_as_terminal() {
         let exit = Exit::never_started();
-        let _ = dispatch_exit(
-            &exit,
-            RestartPolicy::default(),
-            ScopeMode::Running,
-            MembershipStatus::Active,
+        assert_eq!(
+            dispatch_exit(
+                &exit,
+                RestartPolicy::new(RestartCondition::Always, Backoff::Immediate),
+                ScopeMode::Running,
+                MembershipStatus::Active,
+            ),
+            ExitDispatch::Terminal,
         );
     }
 


ralexstokes:
Reviewed at `1ae4d05` (one file, 12+/12-, up to date with `origin/main`). **Verdict: approve.** The defensive framing checks out and the test is non-vacuous; one low-severity note below, then the things I checked and found clear.

## Low — the invariant loses its only in-code record, and the absorption is only partial

`crates/shelterwood-core/src/engine.rs:254-256`. The `debug_assert!` was the sole statement anywhere in the tree that a `NeverStarted` exit must never reach incarnation dispatch (`dispatch_exit` has no doc comment). It is replaced by a bare early return with no comment, so after this change a violation is invisible in both profiles and the rule survives only in the test name.

Two things sharpen this:

1. **Inconsistent with its sibling.** `ScopeLifecycle::begin_drain` (`engine.rs:886-889`) still `debug_assert!`s the same misuse class — `"NeverStarted is not a live-incarnation drain reason"`. The two `NeverStarted`-misuse guards in one file now disagree about whether the violation is loud.
2. **Terminal is not actually a consistent state for the hypothetical caller.** If the guard ever fires, it fires from `driver/child.rs:1056`, which is only reachable with `child.active` set for a real `incarnation` — so the `Terminal` arm calls `begin_terminal_disposal(key, exit, Some(incarnation), startup)` and publishes a `NeverStarted` terminal for a membership whose record has `last_incarnation: Some(..)`. SPEC §B.3 pins the opposite pairing ("`Some` iff any incarnation ever ran, `None` on a `NeverStarted` terminal", SPEC.md:3415-3417), and the observation projections read `last_incarnation.is_none()` as "never started" (`driver/child.rs:121`). So the guard trades an illegal restart for an illegal-but-quiet terminal projection. That is still the better of the two, but it is not "the state is now correct", and nothing reports it.

Counter-argument I weighed and accept partially: `engine` is `pub mod`, so `dispatch_exit` is callable by an unsupported direct dependent of `shelterwood-core` (same class as the `MailboxEffectSink` carve-out in CLAUDE.md), and a public pure function that panics on an input its signature admits is worse than a total one. That justifies making it total; it does not require deleting the debug-time signal for the one supported caller.

Remedy, cheapest first: (a) a one-line comment on the arm citing SPEC §8's "never input to restart or intensity accounting" (SPEC.md:1293-1296) — this is the minimum, since the rule is now undocumented at the site that enforces it; (b) optionally keep `debug_assert!` above the return and split the test into a debug `#[should_panic]` twin plus the release-profile assertion, which is what issue #366's Batch-4 direction ("fail closed in release" *in addition to* the debug check) implies; (c) the house "invalid states unconstructible" answer — an `IncarnationExit` newtype at the boundary — is real but not worth the cross-crate tax for a case with no producer.

Status: the deleted diagnostic is CONFIRMED; the downstream consequence in point 2 is on a structurally unreachable path.

## Checked and clear

- **Unreachability is real; the PR description is accurate.** The only non-test caller is `driver/child.rs:1056`, whose `exit` comes from `classify_exit(reconcile_recorded_outcomes(recorded, active.forced_outcome), join, ..)`. `classify_exit` can only produce `NeverStarted` via `recorded`, and `RecordedOutcome`'s field is private with exactly three constructors — `returned` (`Completed`/`Failed`), `readiness_timed_out`, `aborted` (`exit.rs:550-566`); `reconcile_recorded_outcomes` folds only those. `join` contributes `Panicked`/`Aborted`, the `(None, None)` arm is `Aborted`. The genuine `NeverStarted` producers (`discharge_child_terminality` at `child.rs:121-123`, `SlotCell::terminalize_never_started`, `ScopeCell::terminalize_never_started_locked`, `stop_reason_root_exit`) all call `terminalize_child`/`terminalize_locked` directly and never pass through dispatch. No reachable behavior changes.
- **No interaction with #377.** That PR's new producers publish scope-level `StopReason::NeverStarted` (`close_never_started_body` → `publish_stopped_locked`) via a `NestedScopeStart` obligation; it adds no `RecordedOutcome` and no `Exit::never_started()` on any path into `handle_exit`. The member exit for a body dropped before first poll is still `Aborted` from the cancelled join. Terminality here is correct for those producers because they never reach dispatch at all.
- **Placement is fine.** The two checks the guard now precedes both return `Terminal` themselves, so ordering against them is observationally neutral, and nothing between the old assert point and the new return point consumes the classification. Being ahead of `should_restart` is the point of the change.
- **Test is non-vacuous and profile-independent.** It asserts on the value `dispatch_exit` returns, not a side channel, with `RestartCondition::Always` — the one condition where `should_restart` returns unconditionally `true` (`policy.rs:544-549`). Delete or invert the guard and the assertion fails. No `catch_unwind`/handler swallowing, no `cfg`, and no `debug_assert!` remains on the path, so debug and release execute identical code; I did not need to build both profiles to establish that.
- **No SPEC change needed.** SPEC §8 already states the rule the code now enforces: `NeverStarted` "is a membership fact, not an incarnation verdict: it sits outside the precedence rule below, is never input to restart or intensity accounting" (SPEC.md:1293-1296). This PR moves code toward an existing spec sentence rather than establishing new semantics, so the absence of a spec edit is correct — only the code-side citation is missing (finding above).
- **Lock rule: N/A.** Pure function over a borrowed `&Exit`; the new return drops nothing, formats nothing, and invokes no user code.

ralexstokes:
Fixed in 58c650d — the LOW finding is confirmed and taken as written, with the optional remedy (b) rather than the minimum (a).

## What changed

`crates/shelterwood-core/src/engine.rs`: the `debug_assert!` is restored *above* the release early return, and the arm now carries the SPEC §8 citation the finding asked for.

```rust
// SPEC §8: `NeverStarted` is a membership fact, not an incarnation
// verdict — it "is never input to restart or intensity accounting".
// Every supported producer terminalizes the membership directly and never
// reaches dispatch, so arriving here is a framework bug: the assert makes
// it loud in debug, and the return fails closed in release rather than
// charging a restart for an incarnation that never ran.
```

The test splits into profile halves: the original debug `#[should_panic]` twin (restored under `#[cfg(debug_assertions)]`) plus a `#[cfg(not(debug_assertions))]` assertion that the same input returns `Terminal`. Both are mutation-checked against the committed tree — deleting the guard fails each in its own profile.

## Why (b) and not (a)

A comment alone would have left the release return as the only enforcement, which is exactly the shape being flagged on #382 right now: converting debug asserts into silent release rejections makes an illegal transition quiet in *both* profiles instead of loud in one and safe in the other. #366 Batch 4's phrasing is "fail closed in release **in addition to** aborting in debug"; the original PR read it as "instead of". Keeping both is the direction, and it also resolves sharpener 1 — `begin_drain` and `dispatch_exit` now agree that the misuse is loud. (`begin_drain` remains loud-only; adding a fail-closed half there is a different transition with a less obvious safe verdict, so I left it for a separate Batch-4 item rather than widening this PR.)

The counter-argument about `engine` being a `pub mod` survives intact: the release return is what makes the function total for an unsupported direct dependent of `shelterwood-core`, and it is still there. The assert binds only the framework's own callers, in a profile where an abort is the point.

## Sharpener 2 — accepted as-is, and now recorded in the code

You are right that `Terminal` is not a consistent state for the hypothetical caller: `driver/child.rs:1056` holds a real `incarnation`, so `begin_terminal_disposal(key, exit, Some(incarnation), startup)` would publish a `NeverStarted` terminal against a record with `last_incarnation: Some(..)`, which SPEC §B.3 excludes.

It cannot be repaired inside `dispatch_exit`. `ExitDispatch` is a two-variant restart-vs-terminal verdict carrying no incarnation; the pairing belongs to the caller, and the caller genuinely *has* an incarnation, so a consistent state would require it to also special-case the same predicate and pass `None` — duplicating the guard at the one site the guard exists to defend against. The house "invalid states unconstructible" answer (an incarnation-exit newtype at the boundary) is the real fix, and I agree with your (c) assessment that it is not worth the cross-crate tax for a case with no producer.

So the disposition is accepted-as-is, with two things changed from the state you reviewed: it is now written down at the site, and it is no longer silent —

```rust
// Failing closed is the lesser of two wrong outcomes, not a correct one:
// the sole dispatch caller holds a real incarnation, so a `Terminal`
// verdict here would publish a `NeverStarted` terminal paired with
// `last_incarnation: Some(..)`, which SPEC §B.3 excludes. `ExitDispatch`
// carries no incarnation, so the pairing is the caller's to choose and
// cannot be repaired from inside this function; an inconsistent terminal
// projection is still strictly better than an illegal restart, and the
// debug abort fires before any such projection can be published.
```

The debug abort now fires before the inconsistent projection can reach an observer, so the §B.3 violation is reachable only in a release build of a framework that is already violating §8 — and there it is still the better of the two wrong outcomes.

No SPEC edit; your reading that §8 already states the rule stands.

## Verification

- `cargo nextest run --workspace` — 821 passed.
- `cargo nextest run -p shelterwood-core` in **debug** (90 passed, debug half runs) and **`--release`** (90 passed, release half runs).
- Mutation check on the committed tree: guard deleted → `funnel_dispatch_rejects_…` FAILs in debug, `funnel_dispatch_treats_…_as_terminal` FAILs in release; reverted with `git checkout`.
- `cargo +nightly clippy --locked -p shelterwood-core --all-targets --all-features -- -D warnings`, debug and release. `just fmt` clean.

PR description updated — its "now an ordinary assertion and passes in both profiles" claim is false after this change.

Issue #407, opened 2026-08-21
Offload one-shot still hands the raw caller waker to Tokio: #383's abort geometry survives on receive_blocking


ralexstokes:
Follow-up from the #398 review round (PRs #401/#403/#405). The reply path is now proxied, but the same abort geometry #383 confirmed survives on the blocking-offload path.

## The residual

`receive_blocking` (`crates/shelterwood-runtime/src/spawn.rs`, the `poll_fn(|context| receiver.poll_receive(context)).await` at ~line 177) polls `shelterwood-runtime`'s own one-shot receiver with the **raw caller waker** — no `WakerProxy`. The one-shot delivers `BlockingOutcome<T>`: a user value (the offload closure's return, or a captured panic payload).

The Tokio 1.53.1 drop site verified for #383 applies verbatim: `Receiver::poll` binds the delivered value, then drops its `Arc<Inner>`; when that is the last owner, `Inner::drop` runs `rx_task.drop_task()` — a user waker destructor — while the frame owns the live delivered value. A caller awaiting an offload with a hostile-drop waker plus a panicking-destructor return value reproduces #383's double-panic abort on the ordinary success path. Confirmed by source inspection during the #405 review; not executed.

#383 as scoped named only `reply.rs`/`capability.rs`, so closing it there was correct — but the class is not closed process-wide.

## Remedy shape

Same as the reply seam (#398 ruling 3, as refined by the Option B venue split adjudicated on PR #403): register a framework-owned `WakerProxy` on every parked `poll_receive`, retire the stored caller waker with a synchronous contained drop at the delivery seam (poll path → contained inline; drop glue → disposal lane). `poll_one_shot_with_proxy` in `shelterwood-mailbox/src/capability.rs` is the template; whether the runtime crate can reuse it or needs its own copy depends on the #368 crate-merge layout, so this may be cheapest to do after (or during) #368.

## Regression

Public-path test driving the whole offload future: hostile waker destructor on successful delivery, plus the abort-class variant adding a panicking destructor on the delivered value (in-process; nextest isolation surfaces the abort as that test's SIGABRT). The #405 tests in `crates/shelterwood/tests/reply_delivery_waker_proxy.rs` are the pattern.

ralexstokes:
Audit follow-up (from the class audit this issue and #408 asked for, run against the merged #398 stack): **the same abort geometry survives at a second, warmer seam — `OneShotTaskRef::wait`.**

[...]

PR #410, opened 2026-08-21
Proxy public timer registrations

### LINKED ISSUE #408: wait_for_child/shutdown register raw caller wakers with the timer wheel: #367's stall class outside Deadlined
Follow-up from the #398 review round (PRs #401/#403/#405). #367 chartered the timer path as `Deadlined` only, and PR #403 closed it as scoped — but the *threat* #367 describes (a hostile blocking waker destructor discarded under the global time-driver mutex, stalling every timer registration and cancellation process-wide) survives at two public seams that hand the raw caller waker to Tokio's wheel directly:

1. **`ScopeRef::wait_for_child(..., timeout)`** — `crates/shelterwood/src/scope.rs` awaits `crate::runtime::sleep_until_std(expires)` (~line 164) inside a user-awaited future.
2. **`SystemHandle::shutdown`** — `crates/shelterwood/src/driver/shutdown.rs` (~line 66) wraps `wait_for_incarnation` in `runtime::timeout(...)`, another user-awaited wheel registration.

In both, the caller's waker is cloned into the wheel entry; on completion or drop the retained clone is discarded inside `Handle::clear_entry` under the driver lock — the exact venue #403 just removed for `Deadlined`. A user awaiting either with a blocking waker destructor stalls unrelated timer traffic runtime-wide.

Severity note: these are colder paths than `send_timeout`/`call`/`recv` (scope teardown and shutdown, not per-message), and the waker must be genuinely hostile — this is class-closure hygiene, not a P0. Confirmed by code reading during the #403 review; not reproduced.

## Remedy shape

Route both through the proxied `Deadlined`/timer machinery, or register a `WakerProxy` at these seams directly, with the Option B venue split adjudicated on #403 (poll path → synchronous contained retirement; drop glue → disposal lane). An audit for further `sleep_until_std`/`runtime::timeout` call sites reachable from user-awaited futures should ride along — these two were found by review, not by search.

## Regression

The bounded cross-thread stall shape from `crates/shelterwood/tests/timer_waker_proxy.rs`: hostile blocking waker destructor on one task, assert an unrelated task's timer arms and cancels promptly.

ralexstokes:
## Summary

- Extracts the timer half of `Deadlined` into a public-hidden, runtime-neutral `ProxiedSleep` above the `MailboxRuntime` seam. Runtime adapters supply only raw framework timers; the external wheel therefore retains the stable framework proxy and never the caller's `RawWaker`.
- Routes the runtime's `sleep_until`, `sleep_until_std`, `timeout`, and new absolute `timeout_at` selection through that boundary. Poll-path readiness and a sibling branch winning retire synchronously and contained; cancellation/drop glue submits the caller-waker destructor to the blocking disposal lane before synchronously clearing the framework-only wheel entry.
- Reuses `ProxiedSleep` from `Deadlined`, preserving #403's no-op first probe, stable identity, lost-wake handshake, unbounded no-allocation path, synchronous ready retirement, and drop-glue disposal venue.
- Closes all public user-polled timer seams found by the #408 audit: `ScopeRef::wait_for_child`, shared `shutdown_scope` (`ScopeRef::shutdown_and_wait`, `System::shutdown`, and rollback), and `RawContext::recv` while raw timers are armed. The warm raw-receive path keeps the timer outside the entire event selection so mailbox/event/control winners cancel it inline rather than paying the drop lane on every turn.

## Audit

The production sweep covered every `sleep_until_std`, `runtime::timeout`, and `runtime::sleep_until` site plus direct Tokio time calls.

- The only direct `tokio::time::sleep_until` registration remains the runtime adapter's public-hidden raw capability.
- Mailbox `Deadlined` wraps the raw `MailboxRuntime` capability directly.
- All façade/runtime `sleep_until*` and `timeout*` callers are proxied, including the three public seams above.
- The raw offload deadline and scope-driver wait also inherit the proxy even though they are polled only with framework wakers and are outside the hostile caller-waker class.

## Regression coverage

- Bounded cross-thread hostile-waker tests drive whole public futures for:
  - `ScopeRef::wait_for_child`
  - `ScopeRef::shutdown_and_wait`
  - `System::shutdown`
  - manually user-polled `RawContext::recv` with an armed timer
  - the existing `Deadlined` public mailbox path
- Each targets the timer's exact caller-clone ordinal, blocks that destructor, and proves another native thread can still arm and cancel an unrelated Tokio timer while the cancellation thread returns promptly.
- Focused `ProxiedSleep` tests pin immediate-ready no-entry/no-clone behavior, fused ready repolling, ready inline retirement, containing-branch inline cancellation, and drop-glue disposal on another thread. Runtime timeout tests pin zero-budget and exact-boundary operation precedence, bounded long-deadline slicing, and maximum/unarmable deadlines. Existing `Deadlined` tests continue to pin immediately-ready, already-elapsed, unbounded, synchronous cancellation, and no-spurious-wake behavior.

## Validation

- `./tools/dev just ci`
  - 897 nextest tests
  - doctests and rustdoc with warnings denied
  - public API reachability checks
  - packaged-doc and external-consumer checks
  - Nix formatting
- `./tools/dev cargo nextest run -p shelterwood --test timer_waker_proxy` (5/5)
- `./tools/dev cargo test -p shelterwood-mailbox -p shelterwood-runtime --lib` (74 mailbox + 66 runtime)

## Cross-PR coordination

#407 independently exposes a public-hidden post-unlock retirement callback on the same mailbox `WakerProxy` and adds a runtime-private venue wrapper for Tokio one-shots. This PR does not copy or move the primitive: `ProxiedSleep` uses that single mailbox implementation directly. The branches target `main` independently; they overlap additively in the mailbox module exports, so whichever merges second should rebase that small boundary edit.

Closes #408

### DIFF
--- a/crates/shelterwood-mailbox/src/capability.rs
+++ b/crates/shelterwood-mailbox/src/capability.rs
@@ -374,7 +374,7 @@ pub(crate) mod tests {
 
     use crate::runtime::{
         OneShotClose, OneShotReceiver, OneShotSender, Signal, SignalWatcher, dispose_detached, now,
-        oneshot, sleep_until,
+        oneshot, raw_sleep_until,
     };
 
     use super::{
@@ -455,7 +455,7 @@ pub(crate) mod tests {
         fn sleep_until(&self, deadline: Option<Instant>) -> BoxedSleep {
             deadline.map_or_else(
                 || Box::pin(std::future::pending()) as BoxedSleep,
-                sleep_until,
+                raw_sleep_until,
             )
         }
     }
--- a/crates/shelterwood-mailbox/src/deadline.rs
+++ b/crates/shelterwood-mailbox/src/deadline.rs
@@ -2,17 +2,12 @@ use std::{
     future::Future,
     pin::Pin,
     sync::Arc,
-    task::{Context, Poll, Waker},
+    task::{Context, Poll},
 };
 
 use shelterwood_core::DeadlineBudget;
 
-use crate::{
-    BoxedSleep, MailboxRuntime,
-    cell::waker_slot::{WakerAction, WakerEffects},
-    panic::PanicAccumulator,
-    waker_proxy::WakerProxy,
-};
+use crate::{MailboxRuntime, ProxiedSleep, panic::PanicAccumulator};
 
 pub(crate) use shelterwood_core::deadline::Deadline;
 
@@ -63,8 +58,7 @@ pub(super) struct Deadlined<F> {
     runtime: Arc<dyn MailboxRuntime>,
     budget_width: DeadlineBudget,
     budget: Option<crate::deadline::Deadline>,
-    timer: Option<BoxedSleep>,
-    timer_waker: Option<WakerProxy>,
+    timer: Option<ProxiedSleep>,
     pub(super) started: bool,
     phase: DeadlinePhase,
 }
@@ -83,7 +77,6 @@ impl<F> Deadlined<F> {
             budget_width: budget_width.into(),
             budget: None,
             timer: None,
-            timer_waker: None,
             started: false,
             phase: DeadlinePhase::InitialAttempt,
         }
@@ -116,32 +109,12 @@ impl<F> Deadlined<F> {
             // so it stays as free as it was before the proxy.
             return Poll::Pending;
         }
-        if self.timer_waker.is_none() {
-            let mut probe = Context::from_waker(Waker::noop());
-            if self
-                .timer
+        Pin::new(
+            self.timer
                 .as_mut()
-                .expect("an unelapsed deadline future retains its timer")
-                .as_mut()
-                .poll(&mut probe)
-                .is_ready()
-            {
-                return Poll::Ready(());
-            }
-            self.timer_waker = Some(WakerProxy::new());
-        }
-
-        let timer_waker = self
-            .timer_waker
-            .as_ref()
-            .expect("a parked timer retains its waker proxy");
-        timer_waker.register(context.waker());
-        let mut proxy_context = Context::from_waker(timer_waker.waker());
-        self.timer
-            .as_mut()
-            .expect("an unelapsed deadline future retains its timer")
-            .as_mut()
-            .poll(&mut proxy_context)
+                .expect("an unelapsed deadline future retains its timer"),
+        )
+        .poll(context)
     }
 
     /// Retires the caller waker on this thread, then synchronously removes the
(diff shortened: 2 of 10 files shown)


ralexstokes:
Review pass (Opus agent review, findings adjudicated and addressed on this branch):

**1. P2 (confirmed, fixed in 99a1dd9): `wait_scope`'s deadline arm paid the drop-glue disposal venue on every driver wakeup.** The deadline was a select arm, so whenever signal/parent-shutdown/message/control won, the parked `ProxiedSleep` dropped and submitted a blocking-lane job just to destroy the driver's own framework waker. Measured via an `eprintln` in `WakerEffects::flush`'s `Dispose` arm: 1007 submissions across `-p shelterwood --lib` vs 0 on main. Same class at the raw offload deadline (`select_two(work, sleep_until(..))`): 56 submissions in `--test offloads`, two blocking-pool jobs per deadlined offload, queued ahead of user offload work. Both sites now hoist the deadline through `timeout_at` — the treatment `wait_for_event` already used — so a non-timer winner cancels inline. After the fix the same instrumentation shows 6 (lib) / 2 (drain+deadlines+offloads) submissions, all genuine dropped-while-pending venues. Poll-order semantics are unchanged: `timeout_at` is operation-first, matching the old biased select where the deadline arm was last.

**2. P3 (confirmed, fixed in 2d27519): `cancel_inline` left `ProxiedSleep` unfused.** It emptied `timer`/`timer_waker` but not `completed`, so a poll after cancellation panicked on the emptied timer slot ("an incomplete proxied sleep retains its timer"). Unreachable from in-tree callers today, but a trap on a `pub` sibling-crate seam whose contract invites exactly that call pattern. Retirement now fuses at the single point that empties both slots; `containing_ready_path_cancels_the_losing_timer_inline` repolls to pin it.

**3. Test tautology (confirmed, fixed in b72e764): `maximum_timeout_stays_unbounded` passed with or without the overflow guard** — `Duration::MAX` short-circuits before any timer exists, so "still pending after advance" was guaranteed by `std::future::pending()` alone. The test now polls with a clone-counting caller waker and asserts zero clones: a parked `ProxiedSleep` clones the caller on its first pending poll, so the assertion pins that the guard constructs no timer and no proxy.

**Cleared by the review (no action):** lost-wake windows around the noop probe (level-triggered timers + the `woken` handshake, including the multi-slice re-arm), double retirement (`retire_timer_inline` takes the sleep out before drop), lock order (nothing taken under the proxy leaf; only the framework proxy drops under Tokio's driver lock), panic containment on all retirement paths, deadline edges (zero, exact-boundary, `MAX`, unarmable absolute, slicing), and public-seam equivalence vs main for `wait_for_child`, `shutdown_and_wait`, `System::shutdown`, and raw `recv`. The hostile-waker regressions were mutation-checked non-vacuous (reverting either public seam to `raw_sleep_until` fails them with the intended time-driver-mutex diagnostic).

`./tools/dev just ci` green at b72e764 (exit code checked, not tail-piped).

PR #411, opened 2026-08-21
Isolate runtime one-shot caller wakers behind proxy

### LINKED ISSUE #407: Offload one-shot still hands the raw caller waker to Tokio: #383's abort geometry survives on receive_blocking

ralexstokes:
## Summary

- reuse the mailbox WakerProxy through a runtime adapter that selects inline or detached caller-waker retirement
- keep DisposingReceiver caller wakers out of Tokio's result-delivery frame, contain ready-path destructor panics, and detach cancellation/drop cleanup
- cover public Context::run_blocking delivery plus the OneShotTaskRef::wait completion boundary with hostile waker and hostile value regressions

## Ordering note

The current one-shot task driver sends the typed value before it publishes Completed, so the public end-to-end OneShotTaskRef::wait path does not currently expose a pending completion receiver. The staged ONESHOT_SENDING unit test exercises that shared pending-to-ready boundary without claiming the current driver ordering reproduces it, and hardens the receiver against future or test-only producers.

## Validation

- ./tools/dev cargo test -p shelterwood-mailbox waker_proxy
- ./tools/dev cargo test -p shelterwood-runtime sync::tests::dropping_disposing_receiver_detaches_caller_waker_retirement
- ./tools/dev cargo nextest run -p shelterwood --test result_delivery_waker_proxy
- staged OneShotTaskRef::wait regression selected through nextest
- ./tools/dev just ci

Mutation checks against clean origin/main fail at the intended receiver seams: the run_blocking tests produce the hostile-waker panic and SIGABRT pairing, and the staged OneShotTaskRef test SIGABRTs when Tokio drops the raw caller waker while owning the hostile completion value.

Closes #407

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -95,6 +95,12 @@ rests on:
   sub-capabilities `MailboxRuntime` mints: `MailboxSignal`,
   `MailboxSignalWatcher` and the `ErasedOneShot*` family are public unsealed
   cross-crate traits for the same reason and ride under the same ruling.
+  `WakerProxy` and its `retire_with` seam ride with them: the proxy is a
+  public doc-hidden cross-crate type whose `retire_with` takes a
+  caller-supplied `fn(Waker)`, but the effect is queued under the proxy's
+  leaf mutex and invoked only after unlock, so no foreign code runs under
+  the lock, and the supported façade re-exports neither the type nor
+  anything that could install one.
   `MailboxEffectSink` is the sharpest case: the framework calls
   `defer_mailbox_effect` while holding both the resident-tree observation gate
   and `MemberCell::mailbox`, and its `MailboxEffectQueue` implementation is
--- a/crates/shelterwood-mailbox/src/cell/waker_slot.rs
+++ b/crates/shelterwood-mailbox/src/cell/waker_slot.rs
@@ -15,12 +15,14 @@ pub(crate) enum WakerAction {
     Wake,
     DropInline,
     Dispose(Arc<dyn MailboxRuntime>),
+    Run(fn(Waker)),
 }
 
 enum WakerEffect {
     Wake(Waker),
     DropInline(Waker),
     Dispose(Arc<dyn MailboxRuntime>, Waker),
+    Run(fn(Waker), Waker),
 }
 
 #[derive(Default)]
@@ -56,6 +58,7 @@ impl WakerEffects {
             WakerAction::Wake => WakerEffect::Wake(waker),
             WakerAction::DropInline => WakerEffect::DropInline(waker),
             WakerAction::Dispose(runtime) => WakerEffect::Dispose(runtime, waker),
+            WakerAction::Run(effect) => WakerEffect::Run(effect, waker),
         });
     }
 
@@ -67,6 +70,7 @@ impl WakerEffects {
                 WakerEffect::Dispose(runtime, waker) => {
                     panics.run(|| dispose(&runtime, waker));
                 }
+                WakerEffect::Run(effect, waker) => panics.run(|| effect(waker)),
             }
         }
     }
--- a/crates/shelterwood-mailbox/src/lib.rs
+++ b/crates/shelterwood-mailbox/src/lib.rs
@@ -45,6 +45,8 @@ pub use cell::*;
 pub use errors::*;
 pub use futures::*;
 pub use reply::*;
+#[doc(hidden)]
+pub use waker_proxy::WakerProxy;
 
 mod identity {
     pub(crate) use shelterwood_core::identity::*;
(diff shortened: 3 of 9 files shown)


ralexstokes:
Review pass (Opus agent review, findings adjudicated and addressed in 88d7f53). The reviewer traced every seam against the pinned Tokio 1.53.1 oneshot source and found **no lost wake, no missed or doubled retirement, no lock-rule violation, and no new abort-class path**; the PR's unreachability claim for the pending-completion receiver was confirmed (send happens-before the Completed verdict, so `wait_terminal` guarantees a first-poll ready probe), and the noop-probe/coop-budget/wake-conversion behavior is equivalent to main. Findings and dispositions:

**1. P2: the drop-glue venue comment contradicted the landed #405 reply-path ruling.** `retire_caller_waker`'s doc claimed "a holder's dropping thread is not an acceptable venue" universally, while `capability.rs`'s landed adjudication retires reply-receiver cancellation inline on exactly that thread. Kept the Detached venue — it's what the #412 stack builds on, cancellation is cold at these seams, and it keeps possibly-blocking drop glue off public holders' threads — but the comment now states it as this seam's adjudication and cites the reply receiver's inline ruling as the deliberate sibling choice, with a pointer to revisit if a hot cancellation path ever lands here.

**2. P3: the hostile-waker regressions never exercised the ready-path containment.** On the fixed code every sender-side completion consumes the installed clone through the proxy wake first, so `retire_caller_waker` always found an empty slot and `discard_panic` was mutation-invisible (verified: neither venue-flip nor discard→resume failed any existing test). A receiver-initiated close is the one ready edge no wake precedes; added `receiver_close_ready_edge_contains_the_installed_caller_waker`, which parks a panicking-drop caller clone, closes, and asserts the outcome comes back intact with the destructor run synchronously on the closing thread. Both mutations (Inline→Detached, discard→resume) now fail it.

**3. P3: `Drop`'s recovery comment overclaimed.** "Tokio's close/drop work below is framework-owned traffic" is true of the waker but not of a value stranded by a recovery unwind, which `inner`'s drop glue then destroys inline. Ported the reply receiver's accepted-residual sentence.

**4. P3: CLAUDE.md's seam inventory didn't cover the newly public `WakerProxy`.** Added it to the dyn-seam list: `retire_with`'s caller-supplied `fn(Waker)` is queued under the leaf mutex and invoked only after unlock, and the façade re-exports neither the type nor an installer.

**5. Cleanup: `DisposingReceiver::close` and `inner_mut` had zero callers** (including in the stacked #412) — deleted. `close_and_poll_receive` also had none in production; it now carries the staged-status doc and the containment test above as its pin, since it is the only route to the no-wake ready edge.

**Deferred (filed separately rather than churned into this PR):** the `poll_one_shot_with_proxy` duplication with `capability.rs`, the four copies of the hostile-waker test vtable, and the framework-enum-values-through-the-disposal-lane cost on `DisposingReceiver::Drop` — issue to follow, shared with #412's duplication findings.

`./tools/dev just ci` green at 88d7f53 (exit code checked).

PR #412, opened 2026-08-21
Proxy remaining public caller-waker handoffs

### LINKED ISSUE #409: Remaining raw caller-waker handoffs to Tokio without the abort or stall pairing: join seams and admission one-shots
Follow-up from the class audit #407/#408 asked for (run against the merged #398 stack). Two further families of public seams hand the raw caller waker to Tokio, but with weaker geometry than either parent issue: no framework or driver lock is involved (so no #367/#408-style process-wide stall), and the frame-owned value at the waker-drop venue is framework data on the common path (so the #383/#407 double-panic pairing needs an additional precondition or is absent). Filed together as class-closure hygiene with a severity assessment, not as P-anything.

## 1. `runtime::join` behind `System::wait` / `System::shutdown` / `start_or_shutdown`

`join_driver` (`crates/shelterwood/src/driver.rs:115`) awaits the root driver `JoinHandle` with the raw caller waker, reachable from three public futures: `System::wait`, `System::shutdown`, `start_or_shutdown` (`crates/shelterwood/src/tree/system.rs`).

Venue in pinned Tokio 1.53.1: `JoinHandle::poll` stores the caller waker in the task trailer; after completion wakes it, the stored (possibly hostile) waker survives until `drop_join_handle_slow` runs `trailer().set_waker(None)` — in the awaiting caller's frame, at the await point's handle drop, while the frame owns the bound `Result<RetainedStopReason, JoinError>`. The task state machine is lock-free, so there is no stall half.

The abort half needs the owned value to have a panicking destructor. `RetainedStopReason` retires through critical disposal (no panic), so the reachable pairing is `JoinError::Panic`, which owns the driver task's panic payload — a user payload only if an accumulator resume path re-raises one inside the driver task, itself already an error path. Narrow, but the raw-waker handoff is the same shape #398 closed elsewhere, and the payload half is not under our control.

The other production `join` sites are framework-task venues (Tokio's own waker) and out of class: `monitor_root_driver` (`driver.rs:170`), `join_offloads` (`raw/context.rs:467`), the driver child monitor's report claim (`driver/child.rs:606`).

## 2. `Admission` / `Removal` futures poll the raw runtime one-shot

`response.receive().await` (`crates/shelterwood/src/tree/admission.rs:57` and `:239`) polls `runtime::OneShotReceiver` — the unproxied Tokio one-shot (`shelterwood-runtime/src/sync.rs:543`) — from the public user-pollable `Admission<H>` and `Removal` futures.

Delivered values are plain framework enums (`Result<(), ReserveError>`, `RemoveOutcome`; `ReserveError`'s richest variant owns a `ChildId`), so there is no user value in the frame at the #383 drop site and no abort pairing. The residual is only that a hostile caller-waker destructor panic surfaces uncontained from the caller's own poll — blast radius limited to the awaiting task. Lowest severity in the class.

## Remedy shape

Both want the #398 pattern (framework-owned `WakerProxy` on park, synchronous contained retirement at delivery), but neither justifies it on its own urgency. The natural time is when #407's residuals move: the `OneShotTaskRef::wait` fix (see the audit comment on #407) puts a proxy field on `runtime::DisposingReceiver`, and the admission receivers can ride the same machinery; the join seam needs its own small proxy wrapper around `JoinHandle` polling, or acceptance-with-documentation given the precondition stack. All of this lands most cheaply during or after the #368 crate merge, which settles where the proxy type lives.

Confirmed by source inspection (including the Tokio 1.53.1 `drop_join_handle_slow` and oneshot `Inner::drop` venues); not executed.

ralexstokes:
Depends on #411 and is intentionally based on `agent/issue-407-offload-waker-proxy`. After #411 lands, this PR can be retargeted to `main` without changing its #409-only commits.

## Summary

- add a public-polling join wrapper that probes Tokio with a noop waker, parks only a stable framework proxy, and retires the caller waker before a ready join result crosses the boundary
- route the common `SystemRun::join_driver` seam used by `System::wait`, `System::shutdown`, and `start_or_shutdown` through that wrapper
- route public `Admission` and exact `Removal` one-shot receivers through #411's proxied `runtime::DisposingReceiver`
- add adversarial public-path regressions plus a scheduler-independent `SystemRun::wait` seam fixture for hostile caller-waker destruction, including a `JoinError::Panic` whose panic payload also panics on drop

## Handoff audit

The remaining supported public-user-polled sites are covered here:

- `SystemRun::join_driver`, shared by system wait, shutdown, and startup rollback
- `Admission`
- `Removal`

Related venues are accounted for as follows:

- `OneShotTaskRef::wait` and blocking one-shot receipt are covered by the #411 dependency
- mailbox reply futures are already covered by #405
- `monitor_root_driver`, child-driver monitoring, and raw offload joins are framework-task venues; they do not park caller-supplied public wakers and continue using the raw runtime join helper

The join ready path retires the caller waker first, then drops the Tokio handle while the returned result may own a panic payload. `PanicAccumulator` contains hostile retirement/destruction so the `JoinError::Panic` geometry cannot turn a second panic during unwind into an abort. Cancelling a still-pending public join instead transfers caller-waker destruction to detached disposal, keeping potentially blocking drop glue off the future holder's thread.

## Validation

- `./tools/dev cargo test -p shelterwood-runtime` — 68 passed
- `./tools/dev cargo test -p shelterwood --test result_delivery_waker_proxy --test admission_removal_waker_proxy --test system_join_waker_proxy` — 2 + 2 + 2 passed
- deterministic `SystemRun::wait` pending-join fixture — passed; terminal scope state is prepublished and a latch holds the synthetic driver incomplete, avoiding dependence on the real driver's terminal-publication/completion scheduler interval
- System join stress — 100/100 consecutive rounds passed (200 public shutdown/rollback executions plus 100 deterministic wait-seam executions)
- `./tools/dev cargo test -p shelterwood tree::admission::tests` — 5 passed
- mutation checks: removing the already-ready noop probe trips the caller-clone regression; removing synchronous ready retirement trips the narrow payload+waker ordering regression; making pending-drop retirement inline trips the detached-venue regression; reverting `join_driver` to raw `runtime::join` makes the deterministic `SystemRun::wait` seam fail at hostile waker drop; reverting Admission/Removal to raw receive makes both public one-shot regressions fail
- `./tools/dev just ci` — green, including 899/899 nextest tests, formatting, clippy, docs/doctests, public API reachability, packaging, and external-consumer checks

Closes #409

### DIFF
--- a/crates/shelterwood-runtime/src/spawn.rs
+++ b/crates/shelterwood-runtime/src/spawn.rs
@@ -684,13 +684,13 @@ mod tests {
         mem::ManuallyDrop,
         panic,
         sync::{
-            Arc,
+            Arc, Mutex,
             atomic::{AtomicBool, AtomicUsize, Ordering},
             mpsc,
         },
         task::{Context, Poll, RawWaker, RawWakerVTable, Wake, Waker},
         thread,
-        time::Duration,
+        time::{Duration, Instant},
     };
 
     use super::BlockingPoolJob;
@@ -721,32 +721,63 @@ mod tests {
         }
     }
 
-    #[derive(Default)]
-    struct JoinWakeState(AtomicBool);
+    struct JoinWakeState {
+        woken: AtomicBool,
+        dropped: mpsc::Sender<crate::test_support::ThreadDescription>,
+        clone_action: Mutex<Option<CloneJoinAction>>,
+    }
+
+    struct CloneJoinAction {
+        release: tokio::sync::oneshot::Sender<()>,
+        finished: super::AbortHandle,
+    }
 
     unsafe fn clone_join_waker(data: *const ()) -> RawWaker {
         // SAFETY: every pointer using this vtable came from an Arc of the
         // matching type. ManuallyDrop preserves the reference represented by
         // `data`; the returned raw waker owns only the new clone.
         let state = ManuallyDrop::new(unsafe { Arc::<JoinWakeState>::from_raw(data.cast()) });
+        let action = state
+            .clone_action
+            .lock()
+            .expect("join clone-action mutex is not poisoned")
+            .take();
+        if let Some(CloneJoinAction { release, finished }) = action {
+            release
+                .send(())
+                .expect("the join task still awaits its release");
+            let deadline = Instant::now() + Duration::from_secs(1);
+            while !finished.0.is_finished() {
+                assert!(
+                    Instant::now() < deadline,
+                    "the released join task completes while its no-op waker is parked"
+                );
+                thread::yield_now();
+            }
+        }
         RawWaker::new(Arc::into_raw(Arc::clone(&state)).cast(), &JOIN_WAKER_VTABLE)
     }
 
     unsafe fn wake_join_waker(data: *const ()) {
         // SAFETY: wake consumes the Arc reference represented by this waker.
         let state = unsafe { Arc::<JoinWakeState>::from_raw(data.cast()) };
-        state.0.store(true, Ordering::SeqCst);
+        state.woken.store(true, Ordering::SeqCst);
     }
 
     unsafe fn wake_by_ref_join_waker(data: *const ()) {
         // SAFETY: ManuallyDrop preserves the reference represented by `data`.
         let state = ManuallyDrop::new(unsafe { Arc::<JoinWakeState>::from_raw(data.cast()) });
-        state.0.store(true, Ordering::SeqCst);
+        state.woken.store(true, Ordering::SeqCst);
     }
 
     unsafe fn drop_join_waker(data: *const ()) {
         // SAFETY: drop consumes the Arc reference represented by this waker.
-        drop(unsafe { Arc::<JoinWakeState>::from_raw(data.cast()) });
+        let state = unsafe { Arc::<JoinWakeState>::from_raw(data.cast()) };
+        let current = thread::current();
+        let _ = state
+            .dropped
+            .send((current.id(), current.name().map(str::to_owned)));
+        drop(state);
         panic!("hostile public-join waker destructor");
     }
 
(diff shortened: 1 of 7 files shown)


ralexstokes:
Review pass (Opus agent review of the incremental diff vs `agent/issue-407-offload-waker-proxy`, findings adjudicated and addressed in e548289; the branch is also rebased onto #411's updated head, so the previous SHAs were force-replaced).

**Verdict from the review: no reachable correctness bug.** The join wrapper's lost-wake windows (probe→park and park→inner-poll), coop-budget interaction, retirement edges, the `JoinError::Panic` geometry (retirement under `PanicAccumulator` with `discard_panic`'s self-regenerating replacement while the payload is live in frame), lock-rule compliance, and `classify_join_result`'s byte-identical outcome mapping were all traced against pinned Tokio 1.53.1 internals and cleared. The handoff-audit claims were verified true at head (`monitor_root_driver`, child-driver monitoring, and raw offload joins all poll from inside framework-spawned tasks; the only other `join_resuming` site is test-util-gated). All five mutation checks in the PR description were re-confirmed, plus 12 consecutive green rounds of the two integration suites. Findings and dispositions:

**1. P3 (confirmed empirically, fixed): the four public fixtures were named for containment they never exercise.** Drop-counter instrumentation showed the hostile `drop` vtable fires zero times in all four on the fixed path — every completion consumes the installed clone through the proxy wake (`wake_by_ref` takes the slot; the next `register` immediately re-consumes via the `woken` handshake), so ready-path retirement always finds an empty slot. What they *do* pin is real and valuable: the raw caller waker is never parked in Tokio's trailer/one-shot (each fails at hostile waker destruction under the documented mutations). Renamed to `*_never_parks_its_caller_waker_*`, reworded the doc comments, and added file headers stating the division of labor: contained ready-path destruction and the detached cancellation venue are pinned by the runtime unit tests (`user_polled_join_separates_hostile_waker_and_panic_payload_destruction` and the detach-venue tests).

**2. P3 (confirmed, fixed): `runtime::join` carried no venue restriction while `join_user_polled` documented the split one-sidedly.** The #407/#409 class exists precisely because a public seam silently reached for the raw primitive. `join` now documents that it is for framework-task venues only and names `join_user_polled` as the public-seam requirement. (A newtype was considered and rejected per the repo's own ruling — it would tax call sites without enforcing anything.)

**3. Cleanup (deferred to #413):** the third restatement of the probe→register→poll scaffolding in `poll_join_user_waker`, the integration-test hostile-vtable copies, and the framework-enums-through-the-disposal-lane cost on dropped in-flight `Admission`/`Removal` (a #411-contract matter, noted there too).

`./tools/dev just ci` green at e548289 (exit code checked). After #411 lands this branch retargets to `main` unchanged.

Issue #398, opened 2026-08-21
Waker-proxy round: contain external-primitive waker retirement (implements #367, resolves #383)


ralexstokes:
Implementation round consolidating #367 (chartered: timer path) and #383 (needed a ruling: reply delivery). The two are halves of one mechanism — a framework-owned waker proxy registered with the external primitive, so the clone Tokio retires internally is trivial and the real caller waker retires through framework-controlled, contained paths. This issue tracks the round; the PRs close #367 and #383 individually.

## Rulings

1. **#383's amendment is accepted: the reply delivery path is proxied.** The confirmed abort on the ordinary success path of `call`/`recv` outranks the inline-visibility preference behind the original `WithdrawalDisposition::Inline` reasoning. Correctness-first.

2. **#367's "oneshot cancel stays inline" is superseded, not amended.** The proxy must be registered on every parked reply-receiver poll — #383's abort involves no deadline, so there is no narrower registration point. Once the registration is uniformly proxied, Tokio's one-shot never holds a user waker on *any* path, and the cancel path inherits containment as a side effect. The atomics-only-close reasoning that made inline acceptable there was sound and is recorded at `DisposingReceiver` as history explaining why cancel never *needed* the proxy — not as a live asymmetry, because none remains.

3. **Retirement venue splits by path, deliberately:**
   - **Timer cancel (#367):** the stored caller waker retires through the disposal lane (`runtime.dispose`), as chartered. The threat is a *blocking* destructor stalling every timer registration and cancellation under the global time-driver mutex; work that may block belongs on the lane.
   - **Reply delivery (#383):** the stored caller waker retires through a **synchronous contained drop** at the delivery seam — the shape #372 landed, for the reason #372 discovered: the disposal lane is not a drop-in substitute where a synchronous effect is required, and delivery is the hot path of every successful `call`/`recv`, where a per-delivery lane submission is real cost while a contained drop of a benign waker is nearly free. The containment must be ordered so a waker-destructor panic cannot unwind through a frame that owns the live user value — that combination is exactly the double-panic abort.

## Design constraints

[...]

PR #416, opened 2026-08-22
Fold cells and mailbox into the facade

### LINKED ISSUE #368: Design round: merge the implementation crates into the façade to seal the lock-rule seams
Design round, not an implementation go: adjudicate whether to merge the executor/cells/mailbox implementation crates into the `shelterwood` façade, keeping `shelterwood-core` as the separate leaf. This reopens the runtime-neutrality tradeoff recorded on #348 — deliberately, because its premises have changed, not because the conclusion was re-asserted (#363's P2 raised it again; this issue is the adjudication venue).

## Premises that changed since the tradeoff was recorded

- **Thread-per-core is deferred** (#37), and the standing rule is that TPC readiness must not justify structural looseness. Runtime swappability — the benefit the crate split buys — has no anticipated consumer.
- **The correctness-first directive**: no users, API breaks are free, and we do not build compat machinery. The cost side of merging (a breaking reshape of the crate graph) is therefore near zero right now, and only grows.
- **CLAUDE.md itself names the residual**: the split forces five-plus public unsealed seams that hold "by convention rather than by construction" — `MailboxRuntime`, `ActorIdentity`, `DynamicRoute`, and the minted sub-capabilities (`MailboxSignal`, `MailboxSignalWatcher`, the `ErasedOneShot*` family), with `MailboxEffectSink` as the sharpest case: the framework calls `defer_mailbox_effect` while holding both the observation gate and `MemberCell::mailbox`, and an unsupported direct dependent can construct and install its own sink. The façade's non-re-export is the only fence.

## What merging buys

- Every seam above becomes `pub(crate)`. The "unsupported direct dependent installs its own `MailboxEffectSink`" hazard becomes unrepresentable, and the private-supertrait sealing gymnastics for `MailboxControl`/`MailboxTermination` become unnecessary. The lock rule's weakest paragraph converts from a ruling held by convention to one held by construction — the same "invalid states unconstructible" principle the newtype program applied to values, applied to the crate graph.
- The erased adapters and downcasts at the seams disappear. (This is also the main devirtualization lever — the separate performance sprint would inherit it free — but the correctness case stands alone; perf is explicitly out of scope here.)
- `shelterwood-core` stays the leaf, preserving the #254 ruling (core is the leaf; disposal lives in runtime) and the reducer's isolation.

## Decisions the design pass must answer

- [x] **Prior-ruling check:** the #227 round ruled "layering needs a crate boundary, not core migration." Establish that ruling's scope — it was about keeping the cells layer out of core, which this proposal preserves — or supersede it explicitly if it reads broader.
- [x] **The dependency-inversion tests**: the `cargo tree -e normal` inversion checks (#251 round) enforce the current graph. Decide what the merged graph's enforceable invariant is (core-remains-leaf, façade-names-tokio) and rewrite the checks to pin that instead.
- [x] **The Tokio-naming rule** (#242: only `shelterwood-runtime` may name tokio, tests included) becomes module-level rather than crate-level. Decide the enforcement mechanism — the clippy/CI grep that currently keys on crate boundaries needs a module-path replacement, or the runtime module keeps its own crate after all (a smaller merge: cells + mailbox only). Evaluate both shapes.
- [x] **Test-support and cross-crate test surface**: `shelterwood-cells`' test utilities (`ThreadProbe` et al.) and any dev-dep uses by siblings need a home; recall the #251 rule that dev-deps can't share types across a would-be cycle — merging dissolves some of those constraints, list which.
- [x] **Packaged docs / README sync**: the byte-identical packaged-docs rule and `just packaged-docs-sync` cover per-crate READMEs; decide the merged layout's story.
- [x] **Module layout**: whether the current crate boundaries survive as top-level modules (they should — the internal ordering disciplines, e.g. the observation-gate-outermost rule, are documented against those boundaries) and whether any `#[doc(hidden)]`/`pub(crate)` shims (#233 pattern) are still needed at the façade's public boundary.
- [x] **Sequencing**: after the #366 batches land — the merge is mechanical but touches every file those fixes touch.

## Out of scope

Performance work (separate sprint), thread-per-core, and any behavior change: the merge must be observable-behavior-neutral, verified by the existing suite plus the API-reachability walk.

ralexstokes:
## Summary

- move the runtime-neutral mailbox capability interface and shared waker-proxy machinery into `shelterwood-core`, leaving `shelterwood-runtime` dependent only on core among workspace crates
- fold the cells and mailbox implementation crates into private façade modules and narrow their former cross-crate seams to `pub(crate)`
- update API enforcement, the external-consumer probe, manifests, lockfiles, `AGENTS.md`, and SPEC §§15.1/16.13 for the new crate topology

## Verification

- `./tools/dev just ci`
- `./tools/dev just ci-nix`
- `./tools/dev cargo tree -p shelterwood-runtime -e normal --depth 2`

Closes #368.


ralexstokes:
Review round (8 finder angles, verified findings only) found **no reachable correctness bug** — the fold is line-for-line faithful where it matters and every visibility narrowing is strictly stronger. Nine enforcement/residue findings were confirmed and are all addressed in 41d3f8b:

1. **ProxiedSleep white-box pins deleted, not relocated** — the move to core dropped `ProxiedPoll::is_parked` and the assertions that retirement empties both slots synchronously, leaving the #398 "entry is gone when retirement returns" contract mutation-invisible. Fixed: the three sync tests now live beside the type in `shelterwood-core/src/proxied_sleep.rs` with the white-box asserts restored (mutation-checked: deferring the wheel-entry drop to drop glue fails all three); the façade keeps only the disposal-lane test.
2. **Probe narrower than SPEC §16.13's MUST** — the installable-seams probe covered only `MailboxRuntime` of the capability family. Fixed: it now rejects façade reachability of `MailboxSignal`/`MailboxSignalWatcher`, the `ErasedOneShot*` family, and `WakerProxy`/`ProxiedPoll`/`ProxiedSleep`/`WakerSlot`/`WakerAction`/`WakerEffects` (mutation-checked with a leaked `pub use`). The grep now matches the E0432 header because rustc caps rendered span labels below the list length.
3. **"Runtime-free core manifest" was asserted nowhere** — new `tools/check-core-manifest.sh` pins core's direct normal deps to an allowlist (`thiserror`) in the `runtime-api-check` lane and the flake's api-enforcement derivation; SPEC §16.13 now states the check as a MUST.
4. **Stale `WakerSlot` doc + unlisted seam** — the "private even from the parent mailbox module" claim is corrected, the waker family carries an implementation-boundary ruling, and AGENTS.md's seam inventory names it. (`WakerEffects::is_empty` stays `pub`: the façade's `MailboxEffects` drop probes it cross-crate.)
5. **mailbox.rs fold residue** — the duplicate `panic` alias module, the scalar `ChildId`/`Incarnation`/`Membership` re-exports, and the `cfg(test)`-only re-export block are gone; panic helpers use the crate-wide `crate::runtime::*` idiom and tests import capability types from `shelterwood_core` directly.
6. **Dead E0365 shim files** — `observe.rs`/`admission.rs`/`cancellation.rs` deleted; `lib.rs` re-exports from `cells` directly and internal users import one spelling.
7. **Duplicated test constructors** — one `isolated_scope_with` in `cells::test_support` now feeds both the cells and driver suites, and the driver's `ThreadReportingError` delegates to the shared `ThreadProbe`.
8. **Inert `doc(hidden)` on `pub(crate)` items** removed (`LIFECYCLE_EVENT_CAPACITY`, `ParentCancellationToken`).
9. **Ghost directory skeletons** of the deleted crates removed from disk.

`just ci` is green locally including the new checks.

PR #442, opened 2026-08-22
Document admission lock ordering and exit disposal venue

### LINKED ISSUE #427: Lock-rule accounting and doc drift: dynamic-state mutex around gate acquisition; CLAUDE.md names deleted Acceptance; tokio-in-facade-tests policy
From the 2026-08-22 post-fold full review: lock-rule accounting and doc drift. The full cross-crate lock-order audit otherwise passed (capability seam minimal, core still leaf, effects-flush composition clean, specs/packaged doctests/justfile/Nix filter all in sync).

## 1. Two admission-lane paths hold the dynamic-state mutex *around* observation-gate acquisitions

The only sites contradicting CLAUDE.md's "gate is outermost … never around it" ordering:

- `crates/shelterwood/src/driver/admission_control.rs:283-296` — `DynamicControl::reserve` takes the dynamic-state mutex, then inside the critical section calls `mint_reserved_slot` (child-identity mutex) and `scope.adopt_child_observation_gate(...)`, which reaches `MemberCell::with_handoff_gate` (`cells/member.rs:515-544`) and **acquires the freshly minted child's gate while the state mutex is held**.
- `crates/shelterwood/src/driver.rs:772-820` — the admission-install transaction holds the state mutex across `root.admit_child_locked(...)`, which can enter the gate-to-gate doubled section.

Both are safe, but each rests on an unstated invariant:
1. Reserve path: the gate acquired under the state mutex belongs to a `MemberCell` created inside the same critical section (`plan.rs:31-45` mints a fresh gate) — unpublished, so uncontended by construction.
2. Admit path: by install time the slot's gate provably equals the root's held gate, so `with_handoff_gate` short-circuits without a second acquisition — because `reserve` adopted the slot onto the root gate and the root cannot have been re-homed since (`adopt_observation_gate` refuses any scope with a live dynamic route, `cells/scope.rs:448-451`, and reservations exist only while the route is live). The reserved slot is *not* in `current_children`, so if re-homing were ever possible it would strand the slot on the old gate — invariant (2) is load-bearing, not belt-and-braces.

Per CLAUDE.md's own rule ("a new operation added to the doubled section needs the same accounting"): add a comment at each site plus a sentence in the ordering paragraph acknowledging that the dynamic-state and child-identity mutexes ride above/beside gate acquisition on the admission lane and why.

## 2. CLAUDE.md names a deleted type: `Acceptance`

AGENTS.md:47 (CLAUDE.md symlinks to it). The "Moving a user value out" exemption ends with "…and `Acceptance`'s `#[must_use]` result all encode that." No `Acceptance` type exists; it was deleted in `02f16e8` ("Make mailbox effects structural and runtime-neutral", 2026-08-13) and the reference has been carried through every revision since, including PR #416's. The living `#[must_use]` moving-out encoders are `MailboxTxn::take_payload` (`mailbox/cell.rs:869-872`, whose doc now carries the argument), plus `remove_waiter`/`take_next`/`take_waiters` (:790-802) and `finish_returned` (:881). The sentence should name a living type. This was the only stale item in the lock-rule prose — every other named type/path verified present.

## 3. Policy re-adjudication: façade tests name tokio directly

`crates/shelterwood/src/driver/tests/construction.rs:242,314,325`, `system_join.rs:12`. Sanctioned mechanically by the façade's post-fold tokio dev-dependency, but the PR-242-era rule was "only shelterwood-runtime may name tokio, tests too." If that rule is still meant strictly, these are the sites to re-adjudicate (or the rule should be amended to dev-deps-allowed post-fold).

## 4. Micro doc items

- `flake.nix:16-17`: doubled "the" across the line break.
- README/`docs/*.md` repo-relative links (`[SPEC.md](specs/SPEC.md)`, acceptance-test paths) resolve on GitHub but not in the packaged rustdoc rendered via `include_str!` (`lib.rs:64-76`); rustdoc does not validate them, so `doc-check` passes. Likely an accepted trade — recording it.
- Untracked `pi-sketch.md` at the repo root (retired per prior ruling) — delete or ignore.

ralexstokes:
Closes #427.

## What changed

- accounts for both admission-lane lock-order exceptions at the exact call sites and in `AGENTS.md`: reservation briefly nests the published parent child-identity mutex through a non-escaping API, then acquires only an unpublished fresh child gate under dynamic state; install proves the reserved slot already shares the root gate and therefore performs no nested gate acquisition
- replaces the deleted `Acceptance` reference with the live must-use `MailboxTxn::take_payload` carrier
- distinguishes blocking-offload captured state on the detached lane from raw incarnation-owned async offload/timer/continuation state on SPEC §6.5's on-task funnel
- makes the retained failed-exit application-error disposal venue normative in SPEC §5.5 and removes §16's stale framework-thread aside
- explicitly re-adjudicates Tokio naming after the fold: production façade implementation stays behind runtime capabilities; façade tests, doctests, and examples may use the pinned dev dependency
- ignores the retired root-level `pi-sketch.md` review artifact without deleting a contributor's local copy

## Dispositions of recorded micro items

- The doubled “the” in `flake.nix` was already removed on current `main`.
- The old README/docs `include_str!` link concern was made stale by the completed documentation split: the README is no longer included into crate rustdoc, and the remaining included guide links use rustdoc paths.

## Fresh adversarial review

The post-open review found and fixed three accuracy problems in the first draft:

- the parent scope's child-identity mutex is already published, not owned by the newly minted child; the lock proof now accounts for its non-escaping API separately from the fresh child gate
- critical exit disposal guarantees no inline destruction on the retiring framework stack, but does not promise that an isolated worker starts only after the submitting guard is released
- the Tokio policy now distinguishes implementation imports/calls from prose and explicitly covers test-only synchronization

## Verification

- `./tools/dev just ci`
- `git diff --check`

This remains a draft until the required fresh adversarial review is reported complete.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ result
 
 # mdbook build output
 /book/book/
+
+# Retired local review sketch; the durable conclusions live in tracked issues.
+/pi-sketch.md
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -169,10 +169,11 @@ the pattern; where releasing is impossible, `debug_assert!` instead, as
 `MailboxState::take_waiters` does). And a value that may block on destruction
 goes to `runtime::dispose_detached`, not merely past the unlock. That second
 convention is currently met for mailbox payloads, construction closures,
-blocking-offload captured state, displaced resident graphs and rejected
-admission projections. Raw incarnation-owned values — async offload futures,
-continuations, timers and completions — instead follow SPEC §6.5's on-task
-contained funnel because its disposal verdict is part of exit classification.
+blocking-offload captured state, displaced resident graphs, and rejected
+lifecycle events and admission projections. Raw incarnation-owned values —
+async offload futures, continuations, timers and completions — instead
+follow SPEC §6.5's on-task contained funnel, because a destructor panic
+there is cleanup evidence that exit classification must fold in.
 Framework-retained `Exit` copies meet it through `RetainedExit`,
 including driver completions and pending terminal disposal; exits handed to
 users keep ordinary drop timing. Its fail-safe under exhausted thread
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -44,8 +44,8 @@ rests on:
   a drop. What must be outside the critical section is the *destination*:
   `MailboxTxn::finish`/`finish_returned` handing the value back only after
   releasing the guard, `withdraw`'s `Withdrawal` carrying its outcome and
-  post-unlock effects together, and `Acceptance`'s `#[must_use]` result all
-  encode that.
+  post-unlock effects together, and `MailboxTxn::take_payload`'s `#[must_use]`
+  carrier all encode that.
 - **`Arc` traffic that cannot reach zero.** Cloning an `Arc` under a lock is
   refcount work; dropping one is only refcount work while another owner is
   provable. Prefer restructuring over the proof — every violation the #235
@@ -142,6 +142,16 @@ rests on:
   stays inside the rule — the record is framework-owned data and the retained
   clone is provably non-last — but it is the deepest the exemption goes, so a
   new operation added to the doubled section needs the same accounting.
+  Dynamic admission has two further, narrowly accounted shapes around this
+  order. `DynamicControl::reserve` holds its state mutex while
+  `mint_reserved_slot` takes the child-identity mutex and adoption acquires the
+  new child's gate; both belong to a member minted inside that critical
+  section and are unpublished, hence uncontended. Admission install holds the
+  same state mutex inside the root gate while calling `admit_child_locked`, but
+  the reserved member was already adopted onto that root gate and the root
+  cannot be re-homed while its dynamic route is live, so the handoff check
+  short-circuits without a second gate acquisition. Changing either
+  publication or re-homing invariant requires re-deriving this exception.
   `WakerProxy`'s mutex (`crates/shelterwood-core/src/waker_proxy.rs`) sits at the
   other end of that order: it is a **leaf**. `wake_by_ref` acquires it from
   whatever thread drives the external primitive — the timer driver, a sender,
@@ -156,7 +166,11 @@ the pattern; where releasing is impossible, `debug_assert!` instead, as
 `MailboxState::take_waiters` does). And a value that may block on destruction
 goes to `runtime::dispose_detached`, not merely past the unlock. That second
 convention is currently met for mailbox payloads, construction closures,
-offloads, displaced resident graphs and rejected admission projections. Framework-retained `Exit` copies meet it through `RetainedExit`,
+blocking-offload captured state, displaced resident graphs and rejected
+admission projections. Raw incarnation-owned values — async offload futures,
+continuations, timers and completions — instead follow SPEC §6.5's on-task
+contained funnel because its disposal verdict is part of exit classification.
+Framework-retained `Exit` copies meet it through `RetainedExit`,
 including driver completions and pending terminal disposal; exits handed to
 users keep ordinary drop timing. Its fail-safe under exhausted thread
 creation is an unreclaimed queued job — memory held for the life of the
@@ -167,6 +181,15 @@ Reviewing a new `.lock()` is one pass: name every value the critical section
 can destroy, every callback it can invoke, and every panic it can raise. If
 any of those is user code, hand it to a transaction (`txn.defer`), an effects
 struct, or the caller.
+
+## Runtime naming
+
+Non-test façade code names only the runtime capability layer, never Tokio.
+After the crate fold, façade tests, doctests and examples may use the pinned
+Tokio dev-dependency when executor control or runnable example syntax is the
+subject; internal unit tests still prefer `crate::runtime` when it exposes the
+needed operation. This dev-only allowance does not widen the public API or the
+core crate's dependency boundary.
 
 ## Running anything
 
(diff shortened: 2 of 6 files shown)


ralexstokes:
## Review — PR #442

Reviewed against the lock rule, `specs/SPEC.md`, and the #427 findings. This is a
docs/comment-only change, so there is no behavior to mutation-test; instead I
re-derived every load-bearing claim against the code.

### Claims verified (all hold)

- **Reserve half.** `mint_membership` (`cells/scope.rs:238`) takes
  `ScopeCell::child_identity`, mutates only framework identity data, and releases
  the guard before returning; all three identity accessors (`mint_membership`,
  `adopt_or_mint_membership`, `evict_child_identity`) are take-and-release, so
  there is no `child_identity → state` reverse edge and no ABBA with the
  dynamic-state mutex.
- **Fresh gate is uncontended by construction.** `MemberCell::new`
  (`cells/member.rs:273`) installs `ObservationGate::new()` — a fresh
  `Arc<Mutex<()>>` per member — so the gate `adopt_child_observation_gate`
  acquires under `state` is provably private until `state.insert`.
- **Install half short-circuits.** `with_handoff_gate` returns without acquiring
  when `current.shares_gate(gate)`, and both `admit_child_locked` branches
  (member and nested scope) guard `install_observation_gate_locked` behind the
  same pointer check. The no-re-homing invariant is grounded twice in the code:
  `adopt_observation_gate` refuses a scope with a live dynamic route
  (`cells/scope.rs:448`), and `admit_observation_gate`'s probe refuses every
  stage past `Reserved`.
- **Enumeration is complete.** All seven `state.lock()` sites across
  `driver/admission_control.rs` and `driver.rs` were checked: only `reserve` and
  the install transaction hold `state` around a gate acquisition. `remove`,
  `close`, `signal_fused_cancel` and `cancel_dynamic_reservation_parts` all
  `drop(state)` before their terminal work.
- **`take_payload` replaces `Acceptance` correctly** — `mailbox/cell.rs:869` is
  `#[must_use]` and its doc carries the moving-out argument.
- **Disposal-venue facts.** `dispose_critical` (`shelterwood-runtime/src/disposal.rs:348`)
  really does retain rather than finish inline under exhausted worker creation,
  and `RetainedExit::drop` / `RetainedRecordedOutcome::drop`
  (`cells/retained.rs:131,173`) route only the failed variants there. The
  detached-lane list matches its call sites (`spawn.rs:296` for blocking-offload
  captured state, `capability.rs:340`, `child.rs:1254/1264`, `scope.rs:889/1383`).
  `RawDisposal::dispose` is indeed the on-task contained funnel for offload
  futures, continuations, timers and completions — the old "offloads" entry in
  the detached list was the inaccuracy this PR fixes.
- **Tokio policy matches the tree.** No non-test `tokio` path exists anywhere in
  `crates/shelterwood/src` (only doctest attributes); `shelterwood-core` has no
  tokio dependency at all, so the new section does not widen the core boundary.
- **Micro-item dispositions confirmed.** `flake.nix` has no doubled "the" on
  current main; the README is no longer `include_str!`ed and the three included
  guide pages carry no repo-relative links.

### Fixed on-branch (485013a)

1. **Install-half comment was not self-grounding.** It asserted "every
   reservation was adopted onto this same gate" without citing why the request
   belongs to *this* root. The `Arc::ptr_eq(current, &control)` route check at
   `driver.rs:701` is what makes it true; the comment now names it, so the proof
   is derivable at the call site instead of assumed.
2. **Detached-disposal enumeration was still incomplete.** Rejected lifecycle
   events (`cells/scope.rs:713,741`) also meet the convention and are now listed;
   the raw-funnel clause now states *why* those values take the contained route
   (a destructor panic there is cleanup evidence exit classification folds in)
   rather than the looser "its disposal verdict is part of exit classification".
3. **SPEC §15.4 drifted from the newly-normative §5.5.** Its closing convention
   still said blocking destruction "goes to detached disposal (§5.5's venue)",
   which under-describes §5.5 now that the critical lane is normative there. It
   now names both lanes.
4. **§16.18 grammar.** "What this item does *not* assert is one common destructor
   venue" does not parse; restored "is where the destructor then runs" and
   rewrapped the block to the item's ~73-col width.
5. **Prose-home table.** The book appendix named `CLAUDE.md` as the home while
   this PR's code comments are the first to cite `AGENTS.md`. The row now names
   the real file and notes the symlink.

### Not changed / refuted

- **No `debug_assert!` guarding the install-half gate identity.** Tempting as a
  fail-closed pin, but the only place to put it is inside the doubled section
  holding both the root gate and `state`; a panic there poisons both mutexes,
  which is exactly the anti-pattern the lock rule's second convention forbids.
  Comment-level accounting is the right depth here.
- **The §5.5 exhaustion asymmetry is intentional, not a contradiction.** §5.5's
  new text says the critical lane keeps a job queued, while §6.5 says the
  detached lane's final fallback destroys synchronously. Both are accurate for
  their lane (`dispose_critical` → `retain_fallback_disposal`; detached →
  `job.finish()`), and the two sentences are attached to different carriers.
- **`RetainedExit::install` / `retain_owned` dropping an `Exit` inline** looked
  like a lock-rule violation on first read; it is not. Both branches drop only
  when an equal retained copy is already held (`Arc::ptr_eq` for
  `ExitKind::Failed`), so it is provably refcount traffic.
- **Runtime-naming re-adjudication** is a deliberate policy change from the
  PR-242-era "only shelterwood-runtime may name tokio, tests too" rule. It is
  consistent with the post-fold dev-dependency and with the actual tree; noting
  it as a decision, not a defect.
- The `/pi-sketch.md` ignore entry is inert for the Nix source filter and the
  dirty-tree check (the file was already untracked).

### Merge notes

CI green at `485013af3786c92de81c0d85a98750684dc389bb`
(run 32602095942). One mechanical caveat: I reviewed from a detached worktree, so
the local `agent/issue-427-lock-docs` checkout in `.worktrees/issue-427` is one
commit behind the remote — pull there before pushing anything else.

PR #443, opened 2026-08-22
Make cancellation and handler test probes fail closed

### LINKED ISSUE #418: Inert and vacuous test probes: watcher_count leak pin, in-handler asserts eaten by once-child Finished
From the 2026-08-22 post-fold full review (13-agent). These are the round's only P1 findings: three tests whose probe passes on a regressed tree. Two are recurrences of the swallowed-assertion class (an assert wherever the panic is caught is inert — supervision eats it and a once child still reports `Finished`).

## 1. Vacuous probe: `quiet_signal_wait_cancellation_keeps_one_watch_registration`

`crates/shelterwood-runtime/src/sync.rs:1576`. The test's invariant is that 10,000 cancelled `changed()` waits leave no accumulated registrations, but its probe is `signal.watcher_count()` (sync.rs:167), which reads `WatchShared::receivers` — an endpoint count nothing in the loop can perturb: `SignalWatcher::changed()` takes `&mut self` and never clones the receiver, and waiter registrations live in the separate `WaiterRegistry` (`WatchShared::waiters`, sync.rs:752). Deleting the registry removal in the wait future's `Drop` (sync.rs:927–932) leaks 10,000 `RegisteredWaker`s and the test still passes. Likely a stale pin from a pre-`WaiterRegistry` implementation — the probe did not move with its type.

**Fix:** assert the waiter-registry length (small accessor on `WatchShared`; other tests already reach `waiters`, e.g. sync.rs:1723). The sibling `cancelled_waits_do_not_consume_the_signal` (sync.rs:2034) has the same blind spot.

## 2. Inert in-handler assert: `mailbox_work_queued_at_fire_can_retract_an_elapsed_timer`

`crates/shelterwood/tests/events.rs:533`. The property — a "cancel" handled after a zero-deadline timer has elapsed can still retract the fired delivery — is judged only by `assert_eq!(context.clear_timer(&"deadline"), Ok(true))` **inside the handler** plus the final log check. If retraction regresses (`Ok(false)`), the assert panics in-handler; supervision converts it to a `Panicked` exit, the panic kills the incarnation before the un-retracted `Fired` delivery reaches `handle` (so `"fired"` is never logged), and the once child (`RestartCondition::Never` + `Retention::Remove`) still yields `StopReason::Finished` — exactly the pass condition. The regression the test exists to catch is eaten by its own assert.

**Fix:** record `clear_timer`'s return into the log/an atomic and judge it in the test body (the discipline offloads.rs:797-800 documents).

## 3. Inert in-handler assert: `raw_readiness_is_definition_metadata_evaluated_once`

`crates/shelterwood/tests/raw.rs:95`. `assert_eq!(context.readiness(), Readiness::Immediate)` inside `StatefulReadiness::run` is the only witness that the in-run accessor reports the resolved readiness. If the accessor regresses, `run` panics → once child → `Finished`; every body-side judgment (`STATEFUL_READINESS_CALLS == 1`, `Finished`) still holds. The sibling at raw.rs:128 is only *accidentally* fail-closed (restartable child → intensity trip diverges).

**Fix:** publish the observed readiness through a channel/atomic and assert it in the body.

## 4. Sweep the remaining in-handler asserts

Lower stakes, same shape, found in the same pass:
- `tests/events.rs:852` (`IntervalActor::handle`): masked the same way, but the clear is teardown, not the tested property — body-side `ticks == 2` still judges the property. Convert for hygiene.
- `tests/events.rs:913,932` (`ClearedIntervalActor`): fail closed but slowly — a panic strands the test on a gate wait, failing as a harness timeout instead of a diagnostic.
- `tests/defaults.rs:315` and the offloads.rs fixture-wiring generation asserts were verified fail-closed (body-side evidence breaks on panic); no change needed.

ralexstokes:
Closes #418.

## Summary

- replace the vacuous signal endpoint-count probe with the actual watch waiter-registry length, checking both registration and cancellation cleanup
- pin the same registration/removal lifecycle for cancelled latch waits
- publish timer-retraction results from actor handlers and assert them in the test body, including the interval sweep cases
- publish the raw context's resolved readiness from `run` and judge it after supervision completes

## Verification

- `./tools/dev cargo test -p shelterwood-runtime cancel`
- `./tools/dev cargo test -p shelterwood --test events`
- `./tools/dev cargo test -p shelterwood --test raw raw_readiness_is_definition_metadata_evaluated_once`
- `./tools/dev just ci` (911 nextest cases plus doctests, examples, docs, API reachability, external-consumer checks, and Nix formatting)

## Sweep disposition

The restartable raw-readiness assertion and the defaults/offload fixture-wiring assertions remain unchanged: as documented in #418, their existing body-side evidence already fails closed when the handler or fixture assertion panics.

### DIFF
--- a/crates/shelterwood-runtime/src/sync.rs
+++ b/crates/shelterwood-runtime/src/sync.rs
@@ -163,9 +163,14 @@ impl Signal {
         }
     }
 
+    /// The channel-wide waiter-registry length.
+    ///
+    /// This is deliberately not the endpoint count: `changed()` never clones
+    /// its receiver, so an endpoint probe cannot observe a registration a
+    /// cancelled wait failed to remove.
     #[cfg(test)]
     fn waiter_count(&self) -> usize {
-        self.inner.shared.waiter_count()
+        self.inner.shared.waiters.len()
     }
 }
 
@@ -759,11 +764,6 @@ struct WatchShared<T> {
 }
 
 impl<T> WatchShared<T> {
-    #[cfg(test)]
-    fn waiter_count(&self) -> usize {
-        self.waiters.len()
-    }
-
     /// Acquires the retained value, tolerating poisoning.
     ///
     /// `modify_silently` and `read_with` run a caller closure under this
@@ -1593,9 +1593,17 @@ mod tests {
             let mut changed = Box::pin(watcher.changed());
             let mut context = Context::from_waker(Waker::noop());
             assert!(changed.as_mut().poll(&mut context).is_pending());
-            assert_eq!(signal.waiter_count(), 1);
+            assert_eq!(
+                signal.waiter_count(),
+                1,
+                "a parked wait holds exactly one registration"
+            );
             drop(changed);
-            assert_eq!(signal.waiter_count(), 0);
+            assert_eq!(
+                signal.waiter_count(),
+                0,
+                "cancelling the wait removes its registration"
+            );
         }
     }
 
@@ -2051,18 +2059,34 @@ mod tests {
             let first_poll =
                 std::future::poll_fn(|context| Poll::Ready(fired.as_mut().poll(context))).await;
             assert!(first_poll.is_pending());
-            assert_eq!(latch.state.waiters.len(), 1);
+            assert_eq!(
+                latch.state.waiters.len(),
+                1,
+                "a parked wait holds exactly one registration"
+            );
             drop(fired);
-            assert_eq!(latch.state.waiters.len(), 0);
+            assert_eq!(
+                latch.state.waiters.len(),
+                0,
+                "cancelling the wait removes its registration"
+            );
         }
 
         let mut live_waiter = Box::pin(latch.fired());
         let first_poll =
             std::future::poll_fn(|context| Poll::Ready(live_waiter.as_mut().poll(context))).await;
         assert!(first_poll.is_pending());
-        assert_eq!(latch.state.waiters.len(), 1);
+        assert_eq!(
+            latch.state.waiters.len(),
+            1,
+            "the surviving wait is still registered"
+        );
         assert!(latch.fire());
-        assert_eq!(latch.state.waiters.len(), 0);
+        assert_eq!(
+            latch.state.waiters.len(),
+            0,
+            "the fire drains every registration before waking"
+        );
         assert!(matches!(
             timeout(Duration::from_secs(1), live_waiter).await,
             Timeout::Completed(())
--- a/crates/shelterwood-runtime/src/sync.rs
+++ b/crates/shelterwood-runtime/src/sync.rs
@@ -164,8 +164,8 @@ impl Signal {
     }
 
     #[cfg(test)]
-    fn watcher_count(&self) -> usize {
-        self.inner.receiver_count()
+    fn waiter_count(&self) -> usize {
+        self.inner.shared.waiter_count()
     }
 }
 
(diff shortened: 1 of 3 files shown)


ralexstokes:
## Review (dedicated reviewer, #418 round)

Verdict: **correct and complete for #418's scope.** Every converted probe was
mutation-checked against a committed tree; all five fail closed. No lock-rule,
unwind or drop-glue concerns — the one new critical section (`WaiterRegistry::len`)
holds the registry mutex over a `Vec::len` and runs no user code.

### Mutation checks (each mutation reverted afterwards)

| Mutation | Expected witness | Result |
| --- | --- | --- |
| empty `WatchWait::drop` body | `quiet_signal_wait_cancellation_removes_waiter_registration` | FAIL (`left: 1, right: 0`) |
| empty `LatchWait::drop` body | `cancelled_waits_do_not_consume_the_signal` | FAIL (`left: 1, right: 0`) |
| `Timers::remove` returns `false` after a hit | retraction + interval probes | FAIL both (`[Ok(false)]` vs `[Ok(true)]`) |
| `Timers::remove` returns `true` on a miss | `a_zero_period_interval_arming_clears_the_key_and_stops_ticks` | FAIL (`[Ok(true), Ok(true)]`) |
| `RawContext::readiness` returns `AfterInit` | `raw_readiness_is_definition_metadata_evaluated_once` | FAIL (`Some(AfterInit)`) |

The registry is a plain `Vec<RegisteredWaker>` with `swap_remove`, so `len()` is
exact — the new probe measures the thing #418 said the old one could not perturb.

### Sweep disposition — independently confirmed

- `restartable_raw_readiness_is_resolved_once_for_the_definition` (raw.rs:146) and
  `raw_readiness_override_does_not_evaluate_actor_readiness` (raw.rs:190) both
  **fail** under the readiness mutation above, and fail fast (~4ms), not as
  harness timeouts. Leaving them is fine. Residual, author's call: they are still
  fail-closed *by consequence* (restart-intensity divergence / a stalled readiness
  gate) rather than by published evidence, so a later change to either test's
  restart policy could quietly re-inert them. Not worth reopening this PR for.
- `defaults.rs:315` is fail-closed by construction: the assert precedes every
  `values.fetch_add`, so a panic leaves `values == 0` against the body's `== 2`.
- Spot-checked two more of the same shape that #418 did not name, both fail closed:
  `stopping_raw_context_rejects_timers_and_offloads_with_payload_recovery` publishes
  `checked` only after the last assertion, and the `events.rs:130` continuation
  assert suppresses the `"offload"` log entry that `assert_exact_entries` counts.

### Fixed on branch — baf0dd6

Consistency only, no behavior change:

- `Signal::waiter_count` now reaches `self.inner.shared.waiters.len()` directly, the
  way the latch tests already reach `latch.state.waiters.len()`; the extra
  `WatchShared::waiter_count` hop is gone. Its doc comment records *why* the endpoint
  count is the wrong probe — that omission is how the stale one survived the move to
  `WaiterRegistry`.
- The six new registry-length assertions carry messages naming which half of the
  registration lifecycle they pin, matching the existing sibling at sync.rs:1736 and
  the messages the events/raw conversions already carry.

### Refuted

- Dropping the `watcher_count`/`receiver_count` probe loses no coverage: it was the
  vacuous assertion, and `WatchSender::receiver_count` stays exercised through the
  `cells/observe.rs` gating and the `WatchReceiver::drop` debug assert.
- No stale references to the renamed test anywhere in the tree (docs, specs, book).
- The new `ClearTimerResults` fixtures cannot deadlock the actor: every body-side
  lock is taken after `system.wait()`.

CI: green on baf0dd6 (run 32602059918). Not merged.

PR #444, opened 2026-08-22
Harden core engine and supervisor coverage

### LINKED ISSUE #423: Core engine/supervisor: dead fail-closed branch in dispatch_exit; unpinned Ready-in-Stopping; exploration alphabet lacks exhaustiveness guard
From the 2026-08-22 post-fold full review. No bugs in the core state machine; the review's verdict is that the reducer is *not* over-elaborate — every suspected over-generality had a real second consumer or a confirmed rationale. Items below, most severe first.

## 1. Confirmed dead code: `dispatch_exit`'s fail-closed branch contradicts its own comment

`crates/shelterwood-core/src/engine.rs:267-273`. The function asserts `!matches!(exit.kind(), ExitKind::NeverStarted)` with an unconditional `assert!` (every profile), then immediately follows with `if matches!(exit.kind(), ExitKind::NeverStarted) { return ExitDispatch::Terminal; }`. The `if` is unreachable — the assert always fires first. Leftover from the `debug_assert!` + fail-closed pattern before the #392/#395 `assert!` promotion. Worse, the comment argues the `Terminal` verdict *would be wrong* ("would publish a `NeverStarted` terminal paired with `last_incarnation: Some(..)`, which SPEC §B.3 excludes"), so the surviving branch contradicts its own documentation. Delete the `if`, trim the paragraph to describe only the assert.

## 2. Coverage: `Ready` accepted from `Stopping` is load-bearing but pinned by nothing

`crates/shelterwood-core/src/supervisor.rs:553-555` accepts `Event::Ready` in `Active | Stopping`. The `Stopping` half is reachable: on a *forced* stop the driver deliberately skips disarming the readiness gate (`driver/child.rs:892-894`), so a ready signal can publish while the reducer holds `Stopping`. No targeted test exercises Ready-in-Stopping, and the exploration walk cannot catch a mutant narrowing the set to `Active` only: `check_r1_r2_r6`'s "only a readiness edge sets the bit" stays vacuously true when the bit is never set, and the R6 progress assertion is gated on `!unready`. This is the walk's blind class — a permissive acceptance whose absence manifests as startup-never-completes (progress), not a wrong state. Three-line targeted test: spawn, `StopStarted`, `Ready`, assert `initial_ready`.

## 3. Quality: exploration walk's event alphabet has no exhaustiveness guard

`crates/shelterwood-core/src/supervisor/tests/exploration.rs:139`. The state fingerprint destructures exhaustively so a new field can't compile unclassified, but the event `alphabet` has no equivalent: a new `Event` variant compiles silently and is never explored — the silent-pruning failure the module's own doc warns about. An exhaustive `match` over a canonical variant listing (the exit.rs idiom) closes it at zero runtime cost.

## 4. Coverage (minor)

- Empty-roster finish: `settle_finish` (supervisor.rs:509-524) finishing a draining scope with zero children via the vacuous `all_children_joined()` has no supervisor-level test (all three exploration configs use non-empty rosters; covered only at driver level, driver/tests/shutdown.rs:782).
- `ReadinessGate` spurious deadline on an unbounded wait: `(Waiting { deadline: None }, Deadline { signal_seen: false })` (engine.rs:571-575) resolves through the catch-all; untested — one assertion in the existing readiness test closes it.

## 5. Comment-worthy asymmetries (no code change urged)

- `StopLadder` escalated tidy-beat overflow (engine.rs:192,198): a `None` deadline within `ARMING_HEADROOM` of the clock ceiling parks `Escalated`/`AbortingFramework` forever; `force` rescues only `Cooperative` (:134), which has both a rescue and a test. Theoretical (Instant within ~1s of its representable ceiling) — a comment noting the accepted asymmetry suffices.
- Admit-during-drain: the reducer accepts `admit` while draining (supervisor.rs:351-386, 452-471) and the newcomer would never be stopped; the only guard is the façade's `handle_admission` gate (driver.rs:701-713). Documented boundary, and admission is excluded from the walk's alphabet by construction — record that the strongest verification artifact in this subtree is blind to it by design, and only the façade's admission tests hold the line.
- `IntensityState::charges` unbounded under `Intensity::new(u64::MAX, Duration::MAX)` (engine.rs:320-336) — configuration self-harm, noted for completeness.

ralexstokes:
Closes #423.

## Summary

- remove the unreachable `NeverStarted` fallback after `dispatch_exit`'s unconditional invariant assertion
- pin the progress-sensitive Ready-in-Stopping and empty-roster drain transitions with supervisor tests
- make the exploration event alphabet exhaustively matched so new event variants cannot be silently omitted
- pin the unbounded-readiness spurious-deadline no-op
- document the accepted stop-ladder clock-ceiling, admit-during-drain, and maximum-intensity configuration boundaries

## Issue disposition

1. **Dead `dispatch_exit` branch:** deleted the unreachable `NeverStarted -> Terminal` branch and trimmed the comment to describe the assert-only invariant. The existing panic test continues to pin the invariant.
2. **Ready accepted from Stopping:** added `ready_during_stopping_releases_the_initial_readiness_gate`, which drives `Spawned -> StopStarted -> Ready` and asserts that initial readiness is recorded.
3. **Exploration alphabet exhaustiveness:** added an exhaustive `Event` match beside the canonical alphabet construction. Adding an event variant now breaks this test module at compile time until the alphabet guard is consciously updated.
4. **Minor coverage:** added `draining_an_empty_roster_finishes_on_settlement` for both scope flavors, including exactly-once publication, and added the unbounded-wait spurious-deadline assertion to the readiness-gate test.
5. **Accepted asymmetries:** no behavioral changes. Inline comments now record that escalated/framework tidy-beat overflow has no force rescue near the clock ceiling; the supported facade, not the pure reducer or finite-state walk, owns admit-during-drain rejection; and `Intensity(max=u64::MAX, within=Duration::MAX)` can retain an unbounded charge window as degenerate configuration self-harm.

## Verification

- `./tools/dev cargo test -p shelterwood-core` (107 unit tests plus doctests)
- `./tools/dev cargo test -p shelterwood --test dynamic draining_scopes_reject_admission_and_treat_removal_as_absent -- --exact`
- `./tools/dev just ci` (nightly fmt/clippy, 913 nextest tests, doctests, examples, docs/book, API reachability, manifest/external-consumer checks, nixfmt)

### DIFF
--- a/crates/shelterwood-core/src/engine.rs
+++ b/crates/shelterwood-core/src/engine.rs
@@ -189,6 +189,15 @@ impl StopLadder {
                 } else {
                     grace
                 };
+                // Unlike an unrepresentable cooperative grace, an
+                // unrepresentable tidy beat has no force rescue: `force`
+                // only expedites `Cooperative`, and this ladder is already
+                // `Escalated`. The same applies to the framework tidy beat
+                // below once the ladder reaches `AbortingFramework`. This can
+                // park either phase only when `now` is within
+                // `Deadline::ARMING_HEADROOM` of `Instant`'s ceiling; accept
+                // that theoretical clock-boundary asymmetry rather than
+                // substituting an earlier public deadline.
                 self.deadline = Deadline::after(now, tidy_abort_beat(beat)).instant();
                 Some(StopAction::Escalate)
             }
@@ -255,22 +264,12 @@ pub fn dispatch_exit(
     // verdict — it "is never input to restart or intensity accounting".
     // Every supported producer terminalizes the membership directly and never
     // reaches dispatch, so arriving here is a framework bug. Fail before
-    // charging a restart for an incarnation that never ran.
-    //
-    // Failing closed is the lesser of two wrong outcomes, not a correct one:
-    // the sole dispatch caller holds a real incarnation, so a `Terminal`
-    // verdict here would publish a `NeverStarted` terminal paired with
-    // `last_incarnation: Some(..)`, which SPEC §B.3 excludes. `ExitDispatch`
-    // carries no incarnation, so the pairing is the caller's to choose and
-    // cannot be repaired from inside this function, so every profile stops
-    // before publishing an inconsistent terminal projection.
+    // charging a restart for an incarnation that never ran or publishing the
+    // inconsistent terminal projection that SPEC §B.3 excludes.
     assert!(
         !matches!(exit.kind(), ExitKind::NeverStarted),
         "NeverStarted is a membership outcome outside incarnation dispatch"
     );
-    if matches!(exit.kind(), ExitKind::NeverStarted) {
-        return ExitDispatch::Terminal;
-    }
     if scope == ScopeMode::Draining || membership == MembershipStatus::Removing {
         return ExitDispatch::Terminal;
     }
@@ -283,6 +282,10 @@ pub fn dispatch_exit(
 
 #[derive(Debug)]
 pub struct IntensityState {
+    // This is deliberately one entry per restart still inside the policy
+    // window. `Intensity::new(u64::MAX, Duration::MAX)` can therefore retain
+    // charges without bound; that degenerate configuration is accepted as
+    // operator self-harm rather than adding a second, lossy accounting mode.
     charges: VecDeque<Instant>,
     total_restarts: TotalRestarts,
 }
@@ -1670,6 +1673,21 @@ mod tests {
             Some(ReadinessEffect::Disarmed)
         );
         assert!(!unsignaled_exit.needs_signal_watch());
+
+        let mut unbounded = ReadinessGate::new();
+        unbounded.step(ReadinessEvent::Configure {
+            readiness: crate::Readiness::Manual,
+            deadline: None,
+        });
+        assert_eq!(
+            unbounded.step(ReadinessEvent::Deadline {
+                now: deadline,
+                signal_seen: false,
+            }),
+            None,
+            "a spurious deadline cannot resolve an unbounded readiness wait"
+        );
+        assert!(unbounded.needs_signal_watch());
 
         let mut timed_out = ReadinessGate::new();
         assert_eq!(
--- a/crates/shelterwood-core/src/engine.rs
+++ b/crates/shelterwood-core/src/engine.rs
@@ -194,10 +194,11 @@ impl StopLadder {
                 // only expedites `Cooperative`, and this ladder is already
                 // `Escalated`. The same applies to the framework tidy beat
                 // below once the ladder reaches `AbortingFramework`. This can
-                // park either phase only when `now` is within
-                // `Deadline::ARMING_HEADROOM` of `Instant`'s ceiling; accept
-                // that theoretical clock-boundary asymmetry rather than
-                // substituting an earlier public deadline.
+                // park either phase only when `now` is within the beat plus
+                // `Deadline::ARMING_HEADROOM` of `Instant`'s ceiling — the
+                // beat is capped at ten milliseconds, so just over a second;
+                // accept that theoretical clock-boundary asymmetry rather
+                // than substituting an earlier public deadline.
                 self.deadline = Deadline::after(now, tidy_abort_beat(beat)).instant();
                 Some(StopAction::Escalate)
             }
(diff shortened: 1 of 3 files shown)


ralexstokes:
## Review — PR #444

Reviewed the whole diff against #423 and the surrounding code. **No correctness bugs.** The three claimed pins and the compile-time guard all hold under mutation; one stale spec citation and two soft claims are fixed on-branch.

### Verified by mutation, not by name

Committed the tree first, injected each mutant, reverted with `git checkout --`.

| Mutant | Result |
| --- | --- |
| `Event::Ready` acceptance narrowed to `IncarnationState::Active` only | `ready_during_stopping_releases_the_initial_readiness_gate` **fails**; the exploration walk still **passes** — confirms #423's claim that this is the walk's blind class |
| `all_terminal: !self.children.is_empty() && self.all_children_joined()` | `draining_an_empty_roster_finishes_on_settlement` **fails** |
| `settle_finish`'s `if self.finish_emitted { return; }` deleted | same test **fails** on the exactly-once assertion |
| `(Waiting { deadline }, Deadline { signal_seen: false })` widened to `deadline.is_none_or(..)` | `readiness_configuration_and_signal_deadline_race_are_engine_owned` **fails** on the new assertion |
| a new `Event::MutantProbe` variant added, with `apply`'s arm supplied | the *only* remaining compile error is `E0004` at `exploration.rs:182` — the guard is the thing that catches it, not `apply` |

Also confirmed the deleted branch was genuinely dead (`assert!` is unconditional in every profile; no `debug-assertions`-style escape), that the existing `#[should_panic]` test still pins the invariant, and that `dispatch_exit`'s sole caller (`driver/child.rs:1099`) holds no framework mutex, so the assert is not a poisoning site.

### Fixed on-branch — 6385151

1. **Stale SPEC citation (carried forward, not introduced).** The surviving comment cites `SPEC §B.3` for the `NeverStarted` / `last_incarnation: Some(..)` exclusion, but §B.3 is *Send/call errors*; the pairing lives in **§B.4 Events** (`Removed { last_incarnation: Option<Incarnation> } // None = never started`). Since this PR's own thesis is that a comment contradicting its code is the defect, the trimmed comment should not keep a citation that points at the wrong section. Corrected to §B.4 and named the pairing so the citation is checkable. It was the only `§B.x` citation in `crates/`, so no wider rot.

2. **The clock-ceiling note understated its own window.** `tidy_abort_beat` clamps to `[1ms, 10ms]` and so never returns zero, which means it always routes through `Deadline::at` — the parking condition is `now` within *the beat plus* `ARMING_HEADROOM` of the ceiling, not `ARMING_HEADROOM` alone. Reworded.

3. **`ready_during_stopping_...` asserted its conclusion but not its precondition.** The test drove `Spawned -> StopStarted -> Ready` and checked only `initial_ready`. A `StopStarted` that stopped transitioning would leave the child in `Active`, where `Ready` is accepted anyway — the test would keep passing under a name and doc comment claiming `Stopping` coverage. Verified: with `StopStarted` stubbed to a no-op the original test **passed**; with the added `child_state == Resident(Stopping)` and `!initial_ready` preconditions it **fails**.

### Refuted / no action

- **Exploration guard is not decorative.** Its runtime loop is a no-op, but the exhaustiveness is real and is not already provided by `apply`'s match — the probe above shows the guard is the sole remaining error once `apply` is satisfied, which is exactly the "added the arm, forgot the alphabet" case. All 14 `Event` variants are in fact present in the alphabet today, so the guard is not vouching for a list that is already incomplete.
- **`IntensityState` comment is accurate.** With `within = Duration::MAX` the `age > policy.within()` prune predicate never fires, so `charges` grows one entry per restart without bound; `Intensity::new(u64::MAX, Duration::MAX)` is constructible (only `ZeroDuration` is rejected).
- **The forced-stop premise checks out.** `driver/child.rs` disarms the readiness gate only under `if active.forced_outcome.is_none()`, so a forced stop leaves it `Waiting` and a later signal can publish `Ready` against a `Stopping` reducer state, exactly as the new test's doc says.
- **The two-flavor loop in the empty-roster test is not vacuous.** `finish_if_ready` ignores flavor once draining, but the `Ordered` arm additionally pins that `begin_drain` on an empty ordered roster emits no `StopChild` and leaves the stop cursor invariant-clean.
- **Cross-crate test name in `admit`'s doc** (`draining_scopes_reject_admission_and_treat_removal_as_absent`) can rot silently, but it resolves today (`crates/shelterwood/tests/dynamic.rs:1610`) and matches the established cross-citation practice in this codebase. Left as-is; not worth a check.

No lock-rule surface in this PR — it is pure-reducer code and tests, with no new `.lock()`, no user-value drop site, and no waker or drop-glue path.

### Verification

`cargo +nightly clippy -p shelterwood-core --all-targets --all-features -- -D warnings`, `cargo +nightly fmt --all -- --check`, and the full `shelterwood-core` suite (107/107) pass locally. CI **green** on `6385151` (run 32602008499).

No blockers.

PR #445, opened 2026-08-22
Harden scope publication invariants

### LINKED ISSUE #419: cells/scope: publish_drain diagnostic unwind can inline-drop a startup Exit; dead prune re-check; unchecked announced pairing
From the 2026-08-22 post-fold full review. No reachable bugs in `cells/scope.rs`; these are hardening/simplification/coverage items, most severe first.

## 1. Hardening: `publish_drain` can inline-drop a user-error-bearing startup during its diagnostic unwind

`crates/shelterwood/src/cells/scope.rs:595-627`. The closure's gate-consistency check (`shares_gate` over `terminal_disposals`) runs *before* `startup.take()`, so on refusal the raw `Option<Result<(), StartupError>>` stays in the outer local, and the trailing `assert!(published, ...)` unwinds with it still owned. `begin_drain_with_startup` (`driver/shutdown.rs:123-157`) can pass `Err(StartupError::StartupFailed(...))`, which recursively owns the triggering child's `Exit` with a type-erased user error — so the diagnostic panic destroys a user destructor inline on the unwinding driver thread: a hostile destructor makes it a double-panic abort, a blocking one wedges the unwind. Reachable only if a driver bug marks a member from a foreign gate — but failing cleanly on precisely that bug is the diagnostic's job, and the fix shape exists in-file: `terminalize_child` retains across the same kind of assertion ("Keep a retained owner across the residency assertion", scope.rs:761-764, pinned by `nonresident_terminal_exit_is_retained_before_the_residency_assertion`). Retain the startup's exit (e.g. `RetainedExit`) across the assert.

## 2. Simplification: `prune_child_locked`'s `resident_matches` re-check is tautological

`crates/shelterwood/src/cells/scope.rs:856-896`. The `position(...)` predicate and the `resident_matches` re-check are the identical expression (`projection().member.membership() == membership`), evaluated on the very element the position found, with the `current_children` guard held continuously between `position` and `remove` — nothing can interleave, and `membership()` is a `OnceLock` read. Both `if !resident_matches` branches (deferred panic at :883, `return false` at :885) are unreachable. Unlike `admit_child_locked`'s pop-check (which guards a real cross-statement window spanning the gate handoff), this "invariant" is the removal predicate itself. ~10 lines and one deferred-panic site can go.

## 3. Hardening: `admit_child_locked` marks `announced` through an unchecked `last_mut()`

`crates/shelterwood/src/cells/scope.rs:1389-1398`. The success path sets `announced = true` on whatever is last in `current_children`, silently no-oping if empty and marking the *wrong* resident if the "no residency mutation interleaves under the gate" convention ever broke — the very convention the refusal path diagnoses (`rejected_is_admission` + deferred panic at :1376-1386). The success path spans the widest window in the function (residency push → gate handoff → `Admitted` transition → `set_parent`) yet is the half without a diagnostic. Capture the push index (`let index = children.len(); children.push(...)`, then `children[index]` with a membership check) to make the pairing structural.

## 4. Simplification: `set_state_locked`'s transient-retained machinery is production-dead

`crates/shelterwood/src/cells/scope.rs:642-658`. `RetainedExit::retain_scope_state` only retains for `ScopeState::Stopped`, and no production caller routes `Stopped` through `set_state_locked` — `set_state_and_startup` passes Running/StartupFailed, `publish_drain` asserts Draining, and Stopped publication is owned by `publish_stopped_locked`, which has its own live copy of the same retain/write/surrender dance. Only the `#[cfg(test)]` `set_state` helper reaches it. Either route the helper through `publish_stopped_locked` and assert non-Stopped here, or fold the near-identical dance shared by `set_state_locked`, `set_startup_locked` (:912-941) and `publish_stopped_locked` (:1546-1586) into one helper.

## 5. Coverage: refusal-disposal venue and `set_parent` negative filter are mutation-invisible

- `transition_child_stage` (scope.rs:713) and `publish_child_restart` (:741) route a refused event's user `Exit` through `wakes.defer(dispose_detached(event))`. The existing test (`rejected_stage_transition_suppresses_its_lifecycle_publication`, :2088) checks only suppression; a mutation to plain `drop(event)` — losing the blocking-destructor convention — passes the suite. The member-layer twin *is* venue-pinned (`illegal_restart_transition_is_rejected_in_every_build`, member.rs:829); a `ThreadProbe`-carrying `Exited` refusal test at the scope layer closes the gap.
- `set_parent`'s filter (:358-360) — a consumed or already-finished pending shutdown is not forwarded on adoption — has only positive-path coverage (`pre_admission_restart_shutdown_is_published_when_the_scope_gets_a_parent`, driver/tests/startup.rs:48). A mutation deleting `!request.consumed` passes.

ralexstokes:
Closes #419.

## Summary

Hardens the scope observation paths identified by the post-fold review and adds mutation-sensitive coverage for the remaining disposal and shutdown-forwarding invariants.

## Numbered dispositions

1. **`publish_drain` diagnostic unwind:** fixed. A `RetainedExit` guard set is built before the raw startup local, kept across the foreign-gate assertion, and surrendered as raw refcount traffic after successful publication. The new `foreign_gate_drain_retains_startup_exit_across_the_diagnostic` test proves a nested user error is destroyed off the unwinding driver thread and the observation gate is not poisoned.
2. **`prune_child_locked` tautology:** simplified. Removed the duplicate membership comparison, unreachable deferred diagnostic, and unreachable false return. The resident is now selected once by the held-guard `position` predicate and removed directly.
3. **Unchecked admission announcement:** hardened. Admission captures its residency insertion index and marks only that indexed resident after verifying the membership. A missing or mismatched slot defers a post-unlock framework diagnostic and suppresses the `Added` edge instead of silently announcing the wrong resident.
4. **Production-dead transient state retention:** simplified. The test-only stopped-state helper now routes through `publish_stopped_locked`; the nonterminal writer rejects stopped state at its public entry and documents the internal invariant with a debug assertion. The dead `retain_scope_state` dance was removed from `set_state_locked` while the live terminal publisher remains unchanged.
5. **Mutation-invisible coverage:** closed. Both scope-level refusal APIs now have `ThreadProbe` tests proving rejected `Exited` events retain detached disposal. Parent adoption now has negative coverage for both a consumed pre-admission shutdown request and a defensively forged already-finished target, pinning both halves of the `set_parent` filter.

## Verification

- `./tools/dev cargo test --locked -p shelterwood --lib cells::scope::tests::` (26 passed)
- `./tools/dev cargo test --locked -p shelterwood --lib consumed_pre_admission_shutdown_is_not_published_when_the_scope_gets_a_parent`
- `./tools/dev cargo nextest run --locked -p shelterwood --all-features` (731 passed)
- `./tools/dev just ci` (915 nextest tests plus formatting, clippy, doctests, examples, docs, reachability, manifest, and external-consumer checks)

### DIFF
--- a/crates/shelterwood/src/cells/scope.rs
+++ b/crates/shelterwood/src/cells/scope.rs
@@ -571,7 +571,12 @@ impl ScopeCell {
 
     #[cfg(test)]
     pub(crate) fn set_state(&self, state: ScopeState) {
-        self.with_observation_gate(|txn| self.set_state_locked(state, &[], txn));
+        self.with_observation_gate(|txn| match state {
+            ScopeState::Stopped { reason } => {
+                self.publish_stopped_locked(txn, reason, None, None);
+            }
+            state => self.set_state_locked(state, &[], txn),
+        });
     }
 
     pub(crate) fn set_state_and_startup(
@@ -579,6 +584,10 @@ impl ScopeCell {
         state: ScopeState,
         startup: Result<(), StartupError>,
     ) {
+        assert!(
+            !matches!(state, ScopeState::Stopped { .. }),
+            "terminal scope state is published through publish_stopped_locked"
+        );
         self.with_observation_gate(|txn| {
             self.set_startup_locked(startup, txn);
             self.set_state_locked(state, &[], txn);
@@ -599,6 +608,15 @@ impl ScopeCell {
         terminal_disposals: &[Arc<MemberCell>],
     ) {
         assert!(matches!(state, ScopeState::Draining));
+        // Keep the failed startup's user error owned across the foreign-gate
+        // diagnostic below. The raw startup local is declared after these
+        // guards so unwind releases it first; the guard then transfers final
+        // destruction to critical disposal rather than running it on this
+        // driver thread.
+        let mut retained_startup = Vec::new();
+        if let Some(startup) = &startup {
+            RetainedExit::retain_startup_result(&mut retained_startup, startup);
+        }
         let mut state = Some(state);
         let mut startup = startup;
         let published = self.with_observation_gate(|txn| {
@@ -624,6 +642,11 @@ impl ScopeCell {
             published,
             "a drain entry may mark only a resident member on its observation gate"
         );
+        // A successful publication installed its own retained record copy.
+        // Surrender the diagnostic-only guards as raw refcount traffic.
+        for exit in retained_startup {
+            drop(exit.into_exit());
+        }
     }
 
     fn set_state_locked(
@@ -632,6 +655,10 @@ impl ScopeCell {
         terminal_disposals: &[Arc<MemberCell>],
         txn: &mut ObservationTxn<'_>,
     ) {
+        debug_assert!(
+            !matches!(state, ScopeState::Stopped { .. }),
+            "terminal scope state is published through publish_stopped_locked"
+        );
         // The marker slice is part of the state-writer signature so the #270
         // regression guard cannot be reordered at a caller. Every marker is
         // stored before the `Draining` record write whose release edge makes
@@ -639,8 +666,6 @@ impl ScopeCell {
         for member in terminal_disposals {
             member.set_terminal_disposal_pending(true);
         }
-        let mut transient_retained = Vec::new();
-        RetainedExit::retain_scope_state(&mut transient_retained, &state);
         if matches!(state, ScopeState::Draining | ScopeState::StartupFailed)
             && let Some(route) = self.dynamic_route_in(txn)
         {
@@ -650,12 +675,6 @@ impl ScopeCell {
             record.state = state.clone();
             record.refresh_retained_exits();
         });
-        // The scope record and emitted lifecycle event each install their own
-        // guards. Avoid an extra disposal submission for this call-local
-        // unwind guard.
-        for exit in transient_retained {
-            drop(exit.into_exit());
-        }
         txn.pulse(&self.observation.record);
         txn.pulse(&self.member.record);
         self.emit_locked(txn, LifecycleEventKind::ScopeState { state });
@@ -865,7 +884,6 @@ impl ScopeCell {
         let Some(resident) = resident else {
             return false;
         };
-        let resident_matches = resident.projection().member.membership() == membership;
         // Mirror the announcement: a resident an unwound admission installed
         // never published `Added`, and SPEC §3.2's pairing is both edges or
         // neither. It is still withdrawn and disposed.
@@ -874,22 +892,11 @@ impl ScopeCell {
             membership,
             last_incarnation: resident.projection().member.record().last_incarnation,
         });
-        // Queue the framework diagnostic ahead of every other effect. The
-        // accumulator keeps the first panic, so this is the one ordering under
-        // which an invariant failure cannot be displaced by an effect that
-        // panics for an unrelated reason. Every deferred-diagnostic site in
-        // the tree uses it.
-        if !resident_matches {
-            txn.defer(|| panic!("pruning must remove the requested resident membership"));
-        }
         // The projection can carry the last member/mailbox owner. Put it in
         // the transaction before the fallible publication path so unwind also
         // retires it only after the observation gate is released. The detached
         // handoff deliberately makes final member teardown asynchronous.
         txn.defer(move || runtime::dispose_detached(resident));
-        if !resident_matches {
-            return false;
-        }
         if let Some(event) = event {
             self.emit_locked(txn, event);
         }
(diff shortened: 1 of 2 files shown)


ralexstokes:
## Review

Reviewed all five numbered dispositions against #419, the lock rule, and the SPEC §3.2 pairing rules. **One real defect found and fixed on-branch (23a39df); one hardening gap in the new admission diagnostic, also fixed; one structural finding filed as #455.** Everything else holds up, and all five claimed pins are mutation-sensitive.

### Fixed on this branch — 23a39df

**1. `publish_drain` surrendered its diagnostic guards raw on a path that has no provable co-owner.** (P2; Codex flagged the same site independently, and I confirmed it by audit before reading that comment.)

`published` only means the foreign-gate check passed — it does not mean the record installed the startup. When `record.startup` is already `Some`, `set_startup_locked` rejects the incoming result and its own guards retire through `txn.defer(move || drop(incoming))` → `RetainedExit::drop` → `dispose_critical`. That job is *concurrent*. The post-commit `drop(exit.into_exit())` loop then races it for the last reference, and if the disposal worker wins, the driver thread runs the user error's destructor inline — the exact class this PR set out to close.

Fix: `set_startup_locked` now returns whether it installed, and `publish_drain` surrenders raw only on that verdict; otherwise the guards retire through critical disposal like every other unowned copy.

Not accompanied by a test, deliberately. The race is heavily biased toward the safe ordering — the surrender loop runs microseconds after commit, before the disposal worker has typically even started — so a `ThreadProbe` pin **passes on the broken code**. I verified that empirically (5/5 passes against unfixed `main` of this branch) rather than shipping a probe that fails open.

**2. The new success-path residency diagnostic could drop a last mailbox owner under the gate.** On the `!announced_is_admission` path the frame still owns `projection`, and the only way to reach that path is residency no longer holding this admission — i.e. exactly the case where the lookup clone *can* be the last `Arc<MemberCell>` owner. `return false` then unwinds it under the observation gate, which is what the refusal pop two blocks above is careful not to do. Now routed through `txn.defer(move || runtime::dispose_detached(projection))` ahead of the diagnostic, mirroring the refusal path.

**3. Doc accuracy (two comments).** `admit_child_locked`'s "a refusal publishes nothing" contract no longer described the new diagnostic's `false`, which is returned *after* gate handoff, parent wiring and the `Admitted` transition have landed (unobservable, since the same block queues a panic) — now stated. And the `set_state` fixture's `Stopped` arm now goes through the live terminal publisher, so a fixture publishing `Stopped` twice or over a stronger reason is silently suppressed by stop precedence instead of overwriting — worth a line on a test helper, since that is a fail-open shape for a future test author.

### Filed — #455

`RetainedExit::into_exit()`'s co-owner proof is conventional: 19 surrender sites, each justified by a comment. I audited all 19. The safe ones share a rule none of them states — **a surrender is safe exactly when it runs before the transaction commits**, because the co-owner is either the record the same critical section just wrote or a `RetainedExit` still queued in the effect list. `publish_drain` was the only post-commit surrender in the tree and the only wrong one. Filed with three candidate structural encodings.

### Verified, not changed

- **Disposition 1's test is mutation-sensitive.** Removing the retention makes `foreign_gate_drain_retains_startup_exit_across_the_diagnostic` fail on the thread-id assert. Drop order is right too: the raw `startup` local is declared after the guards, so unwind releases it as refcount traffic first.
- **Disposition 2's simplification is sound.** `position` and the deleted `resident_matches` really are the same expression on the same element under one continuously held guard. `remove` moves the `ResidentChild` out of the `Vec` and out of the guard's block, so nothing is destroyed under the residency mutex.
- **Disposition 3's index is stable.** `admit_observation_gate` / `adopt_descendant_observation_gates_locked` only touch the *adoptee's* residency, never the admitting parent's, so nothing can shift `residency_index` between the push and the announce.
- **Disposition 4's `debug_assert` is not load-bearing.** The removed transient dance protected a `Stopped` state's nested exit across `close_admission`. All three entries to `set_state_locked` now hard-refuse `Stopped` (`publish_drain`'s `assert!`, `set_state_and_startup`'s new `assert!`, `set_state`'s exhaustive match), so the release build cannot reach it. The new `assert!` also fires before the gate is taken — verdict, release, panic.
- **Disposition 5's four pins all fail under mutation.** `drop(event)` for `dispose_detached` in `transition_child_stage` → fails; same in `publish_child_restart` → fails; deleting `!request.consumed` → `consumed_pre_admission_shutdown_...` fails; deleting `request_is_pending` → `adoption_does_not_forward_...` fails. No coverage was lost in the rename of `rejected_stage_transition_suppresses_its_lifecycle_publication` — the suppression assertions are retained alongside the new venue assert.
- **The other 18 `into_exit()` sites are correct.** `scope.rs:836/1055/1096/1122`, `observe.rs:170/412`, `member.rs:609/616` and the rest all surrender pre-commit with a live co-owner (a written record, a public value, or a still-queued deferred effect).
- **SPEC §3.2 pairing stays exact.** The new diagnostic suppresses `Added` and leaves the resident unannounced, so `prune_child_locked` and `clear_residents_locked` publish no `Removed` for it.

CI green on 23a39df.

PR #446, opened 2026-08-22
Harden core waker and exit support

### LINKED ISSUE #424: Core waker/exit support: exit_kind_eq wildcard, WakerAction/WakerEffect mirror, untested retire_with and ready-edge containment
From the 2026-08-22 post-fold full review. The register/wake handshake, ready-edge retirement, drop-glue venues and panic-containment lattice were traced adversarially — no bugs. Items below, most severe first.

## 1. Quality: `exit_kind_eq`'s wildcard arm defeats exhaustiveness

`crates/shelterwood-core/src/exit.rs:492-507` ends with `_ => false`. A future `ExitKind` variant would compile silently and compare unequal to itself under `Exit::eq`, breaking the "structural for every variant except Failed" contract with no compiler signal — in a file that elsewhere forces exhaustive matching (the `StopReason` dance at exit.rs:1223-1229). Restructure as an outer exhaustive match on `left` with per-arm `if let` on `right`, or match `(left, right)` listing all same-variant arms plus a `mem::discriminant` mismatch arm.

## 2. Simplification: mirrored `WakerAction`/`WakerEffect` enums

`crates/shelterwood-core/src/waker.rs:27-39`. The two enums are the same four variants, differing only in carrying the `Waker`; `WakerEffects::push` (:74-81) is a mechanical four-arm re-match that must be extended in two places per new action. A `struct WakerEffect { waker: Waker, action: WakerAction }` (or `Vec<(Waker, WakerAction)>`) with the match moved into `flush` removes the mirror with no semantic change.

## 3. Simplification: byte-identical duplicate test + `retire`/`retire_with` asymmetry

- `waker_proxy.rs:485-497` (`re_registering_the_same_caller_never_reaches_its_clone_vtable`) and `:557-566` (`registration_clones_the_caller_after_unlock`) have literally identical bodies; either copy alone asserts both properties. Delete one or differentiate.
- `ProxiedPoll::retire_with` (waker_proxy.rs:112-118) contains the proxy drop in the accumulator while `retire` (:102-108) drops it bare; the only panic the drop can raise (poisoned-mutex `.expect`) would already have fired on the preceding `proxy.retire` — so the containment is dead weight or its absence is an inconsistency; make them match. The `take()` + `if let Some(x) = &opt` + explicit `drop(opt)` dance also reads more elaborate than `if let Some(proxy) = self.proxy.take() { ... }`.

## 4. Coverage: the cross-crate `retire_with`/`WakerAction::Run` seam is untested at its own layer

`waker_proxy.rs:239-243` + `waker.rs`. The property the CLAUDE.md ruling rests on — the caller-supplied `fn(Waker)` is *queued* under the leaf mutex and *invoked* only after unlock — is asserted nowhere: core's `ReentrantDrop`-style probes cover only `DropInline`, `waker.rs` has no test module, and the runtime crate doesn't pin it either (the façade timer tests cover only the composed behavior). A reentrant `fn(Waker)` that `try_lock`s the proxy state (mirroring the existing `ReentrantWake` probes) pins it in one short test.

## 5. Coverage: ready-edge hostile-destructor containment and re-arm-after-retire

- `ProxiedPoll::retire_ready` (waker_proxy.rs:86): the #398 ruling — a hostile caller-waker destructor panicking during synchronous retirement, panic discarded so the ready value (which may own a user payload) returns intact — has no test; all fixtures use benign counting vtables, so `discard_panic(panics.take())` never runs with a live payload. One panicking-drop waker polled to ready pins both halves.
- Re-arm after retirement: after `ProxiedPoll::retire` (e.g. `DisposingReceiver::close`), a later `poll` re-probes with noop and mints a fresh proxy while the primitive may still hold the old proxy's waker (`crates/shelterwood/src/mailbox/capability.rs:127-139` then :188-200). Trace is sound; no test pins the retire-then-poll-again sequence.
- `proxied_sleep.rs:89-94`: the documented "empty the caller slot first, cancel the timer second" ordering is pinned only by end state, which a swapped order also satisfies; a raw timer whose `Drop` calls `proxy.wake_by_ref()` (the existing `WindowWake` technique) observes the order via the caller's wake count.

## 6. Docs: two comments that under-describe the mechanism

- `wake_by_ref` (waker_proxy.rs:291-295) says an empty slot means the wake landed in `register`'s clone window — but the slot is also empty on the *ordinary* path, and the flag is left set there too. Confirmed consequence: after **every** genuine delivered wake, the next `register` takes one extra clone, installs, reads the stale flag, and immediately takes its own waker back out — one spurious wake + one spurious poll cycle + a second clone per genuine wake, the common case, not a race artifact. Deliberate #398 trade (staleness after task migration is indistinguishable), but the comment should say the flag is set unconditionally *because staleness is indistinguishable*; if the cost ever matters on the reply hot path, `register` could clear the flag without installing and `wake_by_ref(current)` after unlock.
- `ProxiedPoll` top doc (waker_proxy.rs:14-17) omits its one behavioral requirement on `target`: a wake delivered to the noop probe waker is recovered only because the immediate re-poll observes readiness level-wise. True for every framework-owned primitive in use; currently implicit.

## 7. Table-test nit

`exit.rs:1077` (`classification_precedence_is_table_driven`) discards `_discarded` in every row, so the losing-half return of `classify_exit` — the reason the retaining wrappers exist — is pinned only transitively; cheap to assert per row.

ralexstokes:
## Summary

- make `ExitKind` equality exhaustive on the left-hand variant and assert every discarded classification result in the precedence table
- represent deferred waker effects as one `{ waker, action }` pair and make the two `ProxiedPoll` retirement paths symmetric
- add core-level adversarial coverage for cross-crate callback retirement, ready-edge hostile destruction, retire/re-arm behavior, and timer teardown ordering
- document the level-wise probe requirement and the deliberately conservative wake-record behavior

## Issue dispositions

1. Replaced the wildcard equality arm with an exhaustive outer `ExitKind` match.
2. Removed the mirrored `WakerEffect` enum; queued effects now pair a `Waker` with its `WakerAction`.
3. Removed the byte-identical clone-vtable test and simplified both retirement methods to the same take-and-retire shape.
4. Added a reentrant `retire_with` test that proves the supplied function runs after the proxy leaf mutex is unlocked.
5. Added tests for contained panicking ready-edge destruction with payload preservation, fresh-proxy re-arming after retirement, and caller-slot-before-timer cancellation ordering.
6. Expanded the `ProxiedPoll` and `wake_by_ref` documentation to state the level-wise readiness requirement and ordinary-path recovery cost.
7. Extended every classification table row with its expected discarded exit.

## Validation

- `./tools/dev cargo test -p shelterwood-core`
- `./tools/dev just ci`

Closes #424

### DIFF
--- a/crates/shelterwood-core/src/exit.rs
+++ b/crates/shelterwood-core/src/exit.rs
@@ -565,19 +565,21 @@ impl PartialEq for Exit {
 impl Eq for Exit {}
 
 fn exit_kind_eq(left: &ExitKind, right: &ExitKind) -> bool {
-    match (left, right) {
-        (ExitKind::Completed, ExitKind::Completed)
-        | (ExitKind::NeverStarted, ExitKind::NeverStarted) => true,
-        (ExitKind::Failed(left), ExitKind::Failed(right)) => Arc::ptr_eq(&left.0, &right.0),
-        (ExitKind::Panicked { message: left }, ExitKind::Panicked { message: right }) => {
-            left == right
+    match left {
+        ExitKind::Completed => matches!(right, ExitKind::Completed),
+        ExitKind::Failed(left) => {
+            matches!(right, ExitKind::Failed(right) if Arc::ptr_eq(&left.0, &right.0))
         }
-        (
-            ExitKind::ReadinessTimedOut { deadline: left },
-            ExitKind::ReadinessTimedOut { deadline: right },
-        ) => left == right,
-        (ExitKind::Aborted { phase: left }, ExitKind::Aborted { phase: right }) => left == right,
-        _ => false,
+        ExitKind::Panicked { message: left } => {
+            matches!(right, ExitKind::Panicked { message: right } if left == right)
+        }
+        ExitKind::ReadinessTimedOut { deadline: left } => {
+            matches!(right, ExitKind::ReadinessTimedOut { deadline: right } if left == right)
+        }
+        ExitKind::Aborted { phase: left } => {
+            matches!(right, ExitKind::Aborted { phase: right } if left == right)
+        }
+        ExitKind::NeverStarted => matches!(right, ExitKind::NeverStarted),
     }
 }
 
@@ -998,6 +1000,7 @@ mod tests {
                 None,
                 Cancellation::NotObserved,
                 Exit::completed(Cancellation::NotObserved),
+                None,
             ),
             (
                 "cancellation overrides recorded completion",
@@ -1011,6 +1014,7 @@ mod tests {
                     },
                     Cancellation::Observed,
                 ),
+                Some(Exit::completed(Cancellation::Observed)),
             ),
             (
                 "recorded failure",
@@ -1019,6 +1023,7 @@ mod tests {
                 None,
                 Cancellation::NotObserved,
                 Exit::failed(failure.clone(), Cancellation::NotObserved),
+                None,
             ),
             (
                 "late cancellation preserves recorded failure",
@@ -1027,6 +1032,10 @@ mod tests {
                 Some(GracePhase::AfterGrace),
                 Cancellation::Observed,
                 Exit::failed(failure.clone(), Cancellation::Observed),
+                Some(Exit::aborted(
+                    GracePhase::AfterGrace,
+                    Cancellation::Observed,
+                )),
             ),
             (
                 "join panic overrides recorded failure",
@@ -1042,6 +1051,7 @@ mod tests {
                     },
                     Cancellation::NotObserved,
                 ),
+                Some(Exit::failed(failure.clone(), Cancellation::NotObserved)),
             ),
             (
                 "readiness timeout overrides cancellation",
@@ -1053,6 +1063,10 @@ mod tests {
                     ExitKind::ReadinessTimedOut { deadline },
                     Cancellation::Observed,
                 ),
+                Some(Exit::aborted(
+                    GracePhase::AfterGrace,
+                    Cancellation::Observed,
+                )),
             ),
             (
                 "join panic overrides recorded completion",
@@ -1068,6 +1082,7 @@ mod tests {
                     },
                     Cancellation::NotObserved,
                 ),
+                Some(Exit::completed(Cancellation::NotObserved)),
             ),
             (
                 "earlier recorded panic wins a tie",
@@ -1085,6 +1100,10 @@ mod tests {
                     },
                     Cancellation::Observed,
                 ),
+                Some(Exit::panicked(
+                    Some("drop panic".to_owned()),
+                    Cancellation::Observed,
+                )),
             ),
             (
                 "earlier recorded abort wins a tie",
@@ -1098,6 +1117,10 @@ mod tests {
                     },
                     Cancellation::Observed,
                 ),
+                Some(Exit::aborted(
+                    GracePhase::AfterGrace,
+                    Cancellation::Observed,
+                )),
             ),
             (
                 "cancelled join without a hard abort fails closed to within grace",
@@ -1111,6 +1134,7 @@ mod tests {
                     },
                     Cancellation::Observed,
                 ),
+                None,
             ),
             (
                 "hard abort phase is ignored when the join completed normally",
@@ -1119,6 +1143,7 @@ mod tests {
                 Some(GracePhase::AfterGrace),
                 Cancellation::Observed,
                 Exit::completed(Cancellation::Observed),
+                None,
             ),
             (
                 "join cancellation supplies a missing outcome",
@@ -1132,6 +1157,7 @@ mod tests {
                     },
                     Cancellation::Observed,
                 ),
+                None,
             ),
             (
                 "missing outcome fails closed",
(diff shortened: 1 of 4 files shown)


ralexstokes:
## Review

Reviewed the full diff against #424's dispositions, the lock rule, and the panic-containment lattice. **No correctness defects found.** Every behavioural claim the PR makes is real — I verified each by mutating the production code and watching the named test fail (mutation, run, `git checkout --`, on a committed tree):

| Claim | Mutation injected | Result |
|---|---|---|
| `exit_kind_eq` is now exhaustive (#424 item 1) | added an `ExitKind::MutationProbe` variant | `E0004` at `exit.rs:568`, the new outer match — the wildcard's silent-compile hole is closed |
| the classification table pins the discarded half (item 7) | `classify_exit` returns `None` for the discarded exit | fails on the first two-sided row |
| ready-edge hostile destruction is *contained*, not resumed (item 5) | `retire_ready` resumes `panics.take()` instead of discarding | test propagates `hostile caller-waker destructor` |
| `retire_with`'s effect runs **after** unlock (item 4) | inlined `WakerProxy::retire_with` so `effects.flush` happens under the guard | `try_lock` → `WouldBlock`, and the accumulator assert catches it (not fail-open) |
| retire fully un-arms, so a later poll re-arms a *fresh* proxy (item 5) | `ProxiedPoll::retire` uses `as_ref()` instead of `take()` | `!proxied.is_parked()` fails |
| slot-first / timer-second retirement ordering (item 5) | swapped the two halves of `ProxiedSleep::retire` | `WakeOnDropTimer`'s drop-time proxy wake reaches the caller, `wakes` reads 1 |

The `retire_with` test deserves a specific note: it is *not* fail-open. A wake invoked under the guard panics inside `PanicAccumulator::run`, so it would be swallowed — the test's `assert!(panics.take().is_none())` beside the wake count is what makes the probe closed. Good.

Also checked and clean: the `WakerEffect` fold is semantically identical (`push` and `flush` cover the same four actions); `Exit::eq`'s `Failed` arm still compares by `Arc::ptr_eq`, and the table's `failure.clone()` rows share provenance so the new `assert_eq!` on the discarded half is meaningful, not trivially true; the reworked `wake_by_ref_counting` correctly borrows through `ManuallyDrop` while `wake_counting` consumes, and no existing `drops`/`clones` assertion is weakened by the change; no doc, book or SPEC text referenced the removed `WakerEffect` enum.

### Fixed on the branch

**d3edaca** — item 6's wake-record disposition was documented but not pinned. Clearing the record on the delivery path — the optimisation the new comment explicitly rules out — passed every existing waker-proxy test at the time I checked, so the "deliberate trade" rested on prose alone. Added `a_delivered_wake_leaves_its_record_set_for_the_next_registration`, which holds the *ordinary* (non-racing) delivery case beside the race that `a_window_wake_consuming_the_previous_caller_still_reaches_the_new_one` already covers; it fails under that mutation. Same commit restates the comment's cost sentence in the vocabulary `register` already uses for it ("which the `Future` contract permits — a lost wake is not"), replacing "in exchange for making staleness indistinguishable from safety", which reads backwards.

### Filed

**#454** — while checking the containment the PR removed from `ProxiedPoll::retire_with`, I confirmed the issue's analysis is right (the drop wrap was dead weight: the `.lock().expect()` it followed would have fired first), but it surfaced a real asymmetry. `ProxiedSleep::retire` wraps its `ProxiedPoll::retire` call in the accumulator and documents why — "this path also runs during unwind. Keep every cleanup step inside the accumulator so a bookkeeping defect cannot turn a caller panic into an abort" — while the two sibling drop-glue seams (`shelterwood-runtime::ProxiedPoll::drop` and `DisposingReceiver::drop` → `retire_reply_waker`) reach the same `.expect` bare. Unreachable today (the leaf mutex is unpoisonable by construction) and it predates this PR, so it does not belong here; #454 records the three options including making the panic source structurally absent.

### Refuted / non-issues

- **Removing `panics.run(|| drop(proxy))` from `retire_with` weakens unwind safety.** It does not, materially: after `retire` the slot is empty, so `WakerProxy::drop` runs only its own empty-effects flush plus `Arc` bookkeeping, and the sole panic it could raise (`.lock().expect`) would already have fired uncontained one line earlier. `retire`/`retire_with` are now genuinely symmetric. Residual observation carried to #454.
- **`ProxiedPoll::retire`'s new `take()` shape could strand a caller waker if the proxy drops before the effects flush.** No — `WakerProxy::drop` re-runs `retire` on an already-emptied slot, and the caller waker is by then owned by the caller's sink, which flushes with no proxy mutex held.
- **`ready_retirement_…`'s `ManuallyDrop` dance could double-panic.** No — `drop_panicking_drop` panics on the *first* destruction only, so the proxied clone panics (contained) and the test's own reclaim of the original is quiet.
- **The `ProxiedPoll` doc's new level-wise requirement could be stale for the re-poll cycle.** It is correctly scoped: the no-op probe only runs when no proxy is installed, i.e. once per arming cycle, which is exactly what the sentence describes.

### CI

`ci` **pass** against head `d3edaca6931e2efcbf2c004b90c55ea4d1e07b59` (run 32602145412, confirmed by head SHA). Nothing blocks merging.

PR #447, opened 2026-08-22
Simplify raw timer delivery and pin offload publication

### LINKED ISSUE #425: Raw context/timers: fold period into TimerMessage::Interval; dead local_stop select arm; disposal-venue doc misstatement
From the 2026-08-22 post-fold full review. No bugs in the raw context/timers; the selector, fired-batch retry, and offload epilogue all traced clean. Items below, most severe first.

## 1. Simplification: fold `period` into `TimerMessage::Interval`

`crates/shelterwood/src/raw/context/timers.rs:20-23` (`TimerMessage`), :43 (`period: Option<Duration>` on `TimerEntry`), :288-291 and :336-338 (two `unreachable!` arms), plus `raw/context.rs:1238-1251` (`deliver_timer`). A one-shot is `Once` + `period: None`; an interval is `Interval` + `period: Some` — nothing type-enforces the pairing, both `unreachable!`s exist solely to re-derive it, and `deliver_timer` performs two `locate_arming` walks per one-shot delivery (`rearm_interval` → `OneShot`, then `take_due_once` re-locates the same arming). `Interval { message, clone, period }` makes the invalid pairing unconstructible, deletes both `unreachable!`s, and lets delivery resolve the entry once. Purely internal.

## 2. Simplification: dead `local_stop.fired()` arm in `wait_for_event`

`raw/context.rs:1271-1274`. `local_stop` is fired only from the actor's own task (`stop()` at :639, `finish_callback_init` at :667; the driver only observes it — driver/child.rs:64,636). `recv` checks `is_fired()` at :847 before reaching `wait_for_event`, and while the task is parked no other task can fire the latch — so the `select_two(local_stop.fired(), delivery)` arm can never be the waking arm; it costs one waiter registration/deregistration per idle park. Delete it, or comment that it is deliberate armor against a future cross-task fire (right now a reader cannot tell removal is safe — it is).

## 3. Docs: timers module doc misstates the disposal venue

`raw/context/timers.rs:1-6` ("so both always retire through `RawDisposal` rather than on the actor task"), echoed at :282-283. `RawDisposal::dispose` (`raw/disposal.rs:148-152`) runs the drop **inline on the calling thread — the actor task** — behind `catch_panic`, retaining the payload. The guarantee actually provided is containment + retention, not off-task disposal. In a codebase whose lock-rule vocabulary distinguishes the detached lane (`dispose_detached`) from inline contained drops, this conflation would mislead a venue audit. Two-line doc fix.

## 4. Docs: `run_blocking`'s availability while stopping is deliberate but undocumented

`raw/context.rs:794-807`. Every other intake API (`continue_with`, `set_timeout`, `set_interval`, `offload`, `offload_scoped`) rejects once `is_stopping() || !accepting`; `run_blocking` alone has no gate (and `actor.rs:97-102` passes through without a stage check while its siblings check stage). Sound — the `Blocking` future is caller-owned, not ledger-tracked, so freeze/teardown never depends on it — but the asymmetry is invisible. `run_blocking`'s doc should state it remains available during stop/drain and why that is safe, so the missing gate reads as a decision.

## 5. Coverage: no cross-variant timer-key replacement test

Same-variant replacement and zero-period interval clearing are pinned, but nothing replaces a live interval with a one-shot on the same key (or vice versa) and verifies the old cadence stops and a superseded arming already captured in a fired batch is skipped. `TimerStore::replace` is variant-agnostic so risk is low, but this is the one replacement axis with no pin — and item 1's refactor touches exactly this seam.

## 6. Coverage: offload `finished_published` pairing is comment-only

`raw/offload.rs` (`fire_finished`, ~line 250): the Release/Acquire pairing that gates ledger reclamation says in-code "No test pins that pairing — only this comment does." Consistent with the repo's history of promoting load-bearing comments to tests (#332), a targeted test is warranted, though no reachable failure was constructed from weakening it without also violating the comment.

ralexstokes:
## Summary

- encode interval periods inside `TimerMessage::Interval`, make invalid timer/period pairings unconstructible, and resolve timer delivery with one arming lookup
- remove the actor-task-only local-stop waiter from `RawContext::wait_for_event` and update the timer-waker topology probe
- correct the raw timer disposal venue documentation while preserving SPEC §6.5's incarnation-owned contained funnel, document its blocking-destructor worker-pinning caveat, and explain why `run_blocking` remains available during teardown
- cover cross-variant timer replacement after a fired batch captures the superseded arming, including termination of the old interval cadence
- pin the `finished_published` Release/Acquire contract with a cross-thread wake-effect publication witness

## Testing

- `./tools/dev cargo test -p shelterwood raw::context::timers::tests --lib`
- `./tools/dev cargo test -p shelterwood raw::context::tests --lib`
- `./tools/dev cargo test -p shelterwood raw::offload::tests::finished_publication_releases_wake_effect_to_reclaimer --lib`
- `./tools/dev cargo test -p shelterwood --test events`
- `./tools/dev cargo test -p shelterwood --test timer_waker_proxy raw_recv_timer_cancellation_does_not_stall_unrelated_timer_traffic -- --exact`
- `./tools/dev just ci`

Closes #425.

### DIFF
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -88,9 +88,11 @@ macro_rules! context_common_forwarders {
         /// Starts blocking work tied to actor shutdown and returned-future drop.
         ///
         /// Available from [`StopContext`] too — deliberately, unlike
-        /// continuations, timers, and offloads. Awaiting the returned
-        /// [`Blocking`] resumes a panic raised inside the closure at the
-        /// await point.
+        /// continuations, timers, and offloads. The returned [`Blocking`] is
+        /// caller-owned rather than tracked in the incarnation resource
+        /// ledger, so intake freeze and teardown never depend on its
+        /// completion. Awaiting it resumes a panic raised inside the closure
+        /// at the await point.
         ///
         /// Cancellation is cooperative; a hard-aborted operation's OS thread
         /// detaches and may outlive this actor incarnation.
--- a/crates/shelterwood/src/raw/context.rs
+++ b/crates/shelterwood/src/raw/context.rs
@@ -33,7 +33,7 @@ use super::{
 #[cfg(test)]
 use super::offload::OffloadPoll;
 
-use timers::{ArmingOrder, IntervalRearm, TimerMessage, TimerStore};
+use timers::{ArmingOrder, TimerMessage, TimerStore};
 
 type DeferredMessage<M> = Box<dyn FnOnce() -> M + Send + 'static>;
 /// An operation rejected because the actor incarnation is already stopping.
@@ -694,7 +694,7 @@ impl<M: Send + 'static> RawContext<M> {
         if self.is_stopping() || !self.resources.accepting {
             return Err(Rejected::new((key, message)));
         }
-        self.replace_timer(key, TimerMessage::Once(message), after, None);
+        self.replace_timer(key, TimerMessage::Once(message), after);
         Ok(())
     }
 
@@ -718,9 +718,12 @@ impl<M: Send + 'static> RawContext<M> {
         }
         self.replace_timer(
             key,
-            TimerMessage::Interval(message, Clone::clone),
+            TimerMessage::Interval {
+                message,
+                clone: Clone::clone,
+                period,
+            },
             period,
-            Some(period),
         );
         Ok(())
     }
@@ -786,7 +789,10 @@ impl<M: Send + 'static> RawContext<M> {
     ///
     /// Unlike `continue_with`, timers, and offloads, this operation has no
     /// stopping gate: it deliberately keeps working from stop paths, because
-    /// it is the one resource operation teardown code may still need.
+    /// it is the one resource operation teardown code may still need. The
+    /// returned [`Blocking`] is caller-owned rather than tracked in the
+    /// incarnation resource ledger, so intake freeze and teardown never wait
+    /// on it or depend on its completion.
     /// Awaiting the returned [`Blocking`] resumes a panic raised inside the
     /// closure at the await point (distinct from the runtime-teardown
     /// cancellation panic below).
@@ -927,20 +933,13 @@ impl<M: Send + 'static> RawContext<M> {
         }
     }
 
-    fn replace_timer<K>(
-        &mut self,
-        key: K,
-        message: TimerMessage<M>,
-        after: Duration,
-        period: Option<Duration>,
-    ) where
+    fn replace_timer<K>(&mut self, key: K, message: TimerMessage<M>, after: Duration)
+    where
         K: Hash + Eq + Send + 'static,
     {
         let now = runtime::now();
         let deadline = crate::deadline::Deadline::after(now, after).instant();
-        self.resources
-            .timers
-            .replace(key, deadline, message, period);
+        self.resources.timers.replace(key, deadline, message);
     }
 
     fn start_offload<F, T, C>(
@@ -1254,18 +1253,7 @@ impl<M: Send + 'static> RawContext<M> {
     }
 
     fn deliver_timer(&mut self, arming: ArmingOrder) -> Option<M> {
-        match self.resources.timers.rearm_interval(arming, runtime::now()) {
-            IntervalRearm::Missing => return None,
-            IntervalRearm::OneShot => {}
-            IntervalRearm::Interval(message) => return Some(message),
-        }
-
-        Some(
-            self.resources
-                .timers
-                .take_due_once(arming)
-                .expect("a due one-shot timer remains registered"),
-        )
+        self.resources.timers.deliver_due(arming, runtime::now())
     }
 
     fn next_timer_deadline(&self) -> Option<Instant> {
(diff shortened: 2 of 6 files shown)


ralexstokes:
## Review: dedicated reviewer pass

Reviewed the full diff against #425's six items, the lock rule, and `specs/`.
**No correctness bugs found.** Two findings fixed on-branch in 6bb64a7; CI green
on that SHA.

### Verified clean

- **`deliver_due`'s single-walk refactor is equivalent.** `locate_arming` miss →
  skip, interval → clone-then-rearm, one-shot → unlink + contained key disposal,
  matching the old `rearm_interval`/`take_due_once` pair arm for arm. The
  `swap_remove` index stays valid because nothing mutates the bucket between
  `locate_arming` and `unlink`, and the `arm_deadline` insert cannot duplicate a
  pair because `take_due` already removed it.
- **Retry semantics preserved.** A panicking user `Clone` still unwinds with
  `entry.deadline` at its prior value and the arming uncommitted in the installed
  `ReadyBatch` (`next_ready` commits only after `deliver_timer` returns), so
  `recv`'s documented "next receive retries the same arming" still holds.
- **Lock rule.** `TimerStore` sits behind no mutex; the only user code it runs
  (`Hash`/`Eq`, `Clone`, key/message destructors) runs with no lock held, and
  `RawDisposal::dispose` takes the panic-slot lock only *after* `catch_panic(drop)`
  returns. The `Interval` arm after `unlink` correctly disposes rather than
  `unreachable!`-ing with user ownership on the stack.
- **The `local_stop` select-arm deletion is safe.** `Latch::fire` on that latch has
  exactly two non-test call sites (`stop`, `finish_callback_init`), both requiring
  `&mut RawContext`, which `wait_for_event` holds for its whole pending window;
  `driver/child.rs` only observes it. `wait_for_event` has exactly one caller
  (`recv`), which checks the latch immediately above. No lost-wake window.
- **The topology probe is not fail-open.** `raw_recv_timer_cancellation_...`
  asserts the observed clone count *equals* 4, so a wrong ordinal fails rather
  than silently attaching the hostile destructor to another primitive.
- **`run_blocking`'s doc claim is accurate** — it registers nothing in
  `RawResources`; the `Blocking` future is entirely caller-owned.
- **The publication witness is sound.** `finished_published` really is the only
  synchronizes-with edge the consumer touches (it never reads the latch or the
  waiter registry), and sampling `woke` before `join()` keeps the join edge out.
  Mutating `FINISHED_PUBLISH_ORDERING` to `Relaxed` fails the test.

### Fixed (6bb64a7)

1. **Coverage regression: the interval rearm's entry-field write lost its only
   mutation-visible pin.** Rewriting `interval_rearm_overflow_makes_the_live_entry_dormant`
   away from `entry_mut(...).deadline` (deleted with `entry_mut`) means deleting
   `entry.deadline = deadline;` from `deliver_due` now **passes the entire
   `timers::tests` module** — verified by mutation on the committed tree. That write
   is load-bearing: a stale entry deadline makes `unlink` remove the wrong
   `(deadline, arming)` pair and strand a phantom `next_deadline()` for an arming
   nothing can deliver. Added `a_rearmed_interval_keeps_its_entry_and_deadline_index_in_step`,
   which pins it observably (rearm → remove key → `next_deadline()` is `None`) and
   fails under that mutation. It also gives the ordinary interval-rearm path its
   first store-level pin — the new cross-variant test only exercises a *superseded*
   arming.

2. **SPEC §6.5's new venue bullet was the only "Tokio" mention in `specs/`.** The
   normative text deliberately keeps the adapter unnamed and says "the executor"
   (e.g. §6.5's thread-exhaustion text at :765); the concrete naming is the
   rustdoc guides' register, not the spec's. Reworded to "pins the executor thread
   running that disposal". Flagging because the phrasing came verbatim from #425's
   adjudication comment — revert if the concreteness was intended there.

3. Recorded (comment only) that `FINISHED_PUBLISH_ORDERING`/`FINISHED_OBSERVE_ORDERING`
   are the *only* permitted spelling of that pair — see the residual below.

### Noted, not changed

- **The ordering pin is evadable at the use site.** The behavioral half is
  fail-open on a strong-memory host (the test says so), so the real pin is
  `assert_eq!(FINISHED_PUBLISH_ORDERING, Ordering::Release)` — which a future edit
  inlining `Ordering::Relaxed` at the store would slip past without tripping
  `dead_code` (the test module still uses the consts). Given the standing "no loom"
  ruling there is no cheaper structural fix; documented in the const comment
  instead. Accepted limitation, not a defect in this PR.
- **The user-facing half of the blocking-destructor caveat** (a blocked timer
  key/message destructor pins an actor's executor thread) is absent from
  `docs/shutdown-and-resources.md`, which covers only the detached lane's version.
  Deliberately not filed: #425's adjudication already ruled §6.5 is the home.
- `interval_rearm_overflow_...` losing the direct `entry.deadline == None` read is
  otherwise unobservable (a stale overflow deadline is not in the `BTreeSet`
  either way), so item 1's new test is the right place for that invariant.
- `deliver_timer` is now a one-line passthrough to `deliver_due`; left as-is,
  it keeps `runtime::now()` on the context side.

PR #448, opened 2026-08-22
Harden driver exit and fail-closed paths

### LINKED ISSUE #422: Driver: handle_exit holds a raw classified Exit across fallible calls; comment gaps; untested fail-closed arms
From the 2026-08-22 post-fold full review. No reachable bugs in the driver; items below, most severe first.

## 1. Hardening: `handle_exit` holds a raw classified `Exit` across fallible calls

`crates/shelterwood/src/driver/child.rs:1063-1143`. Between `classify_exit_retaining` and the point the exit is re-retained (`begin_terminal_disposal`) or published (`publish_child_restart`), the sole owner of a possibly-failed user error sits raw on the driver stack across `reduce`, `dispatch_membership_status` (which locks the dynamic-state mutex with `.expect("dynamic-state mutex poisoned")`), and two `assert!`s. A panic in that window runs the user error's destructor as unwind drop-glue on the driver thread — the exact class `RetainedExit` exists to prevent, and which `begin_terminal_disposal`'s own comment forbids ("no path may unwind or return it directly on the driver thread"). Every panic source in the window is unreachable-by-construction today, so this is hardening: wrap in `RetainedExit` at classification and unwrap at the two consumption points to make the property structural instead of argued.

## 2. Quality: undocumented `ancestor_shutdown_seen = true` in the self-shutdown arm

`crates/shelterwood/src/driver.rs:1347-1350`. The `Pending::Shutdown` arm silently sets `ancestor_shutdown_seen` for nested scopes, permanently suppressing any future `Pending::AncestorShutdown` observation. Correct only because (a) the ancestor arm's sole action, `begin_drain(ShutdownRequested)`, has already run idempotently, and (b) `construction_is_suppressed` (events.rs:217-225) consults the ancestor latch directly, not the flag. Neither dependency is stated at the assignment site; one comment prevents a future reader from "fixing" it or adding a third consumer of the flag.

## 3. Simplification: duplicated lifecycle-cause derivation in `handle_admission`

`driver.rs:701-714`. The rejection predicate (`is_none_or(!ptr_eq) || is_draining() || startup_failed()`) is immediately re-evaluated member-by-member to derive `cause`. Compute the cause once (an `if let`-chain producing `Option<NotAdmittingCause>`, rejecting on `Some`) — removes the double lifecycle sampling and the implicit sync requirement between the two evaluations.

## 4. Simplification: `DynamicState.entries` dual declaration

`driver/admission_control.rs:181-184`. The `#[cfg(not(test))]`/`#[cfg(test)]` field-visibility split exists only to let tests poke `entries`; a `#[cfg(test)] pub(super) fn entries_mut()` accessor keeps one declaration.

## 5. Coverage: two deliberately unreachable fail-closed arms are untested

- Closed-lane arm `ScopeWake::Message(None) | ControlMessage(None)` (driver.rs:1303-1321): production-unreachable (every sender is retained by the scope or a registered child), and no test constructs it. Its behavior — retained shutdown completion instead of a panic that would unwind queued admission wakers outside the contained epilogue — is load-bearing if the retention invariant ever regresses. A unit test dropping the senders around a hand-built `ScopeRuntime` pins it cheaply.
- The `handle_admission` promote/admit invariant arm (driver.rs:827-854): the diagnose-then-`complete_lost` ordering (framework diagnostic recorded before the response waker runs, containment on an unwinding stack) is untested — same class as the tested `Obligation` unwind-containment tests (storage.rs:124), so a fixture-level test is feasible.

## Verified clean (recording the negatives)

The arbitration drain race (`retain_woken_event`, events.rs:105) is correctly ordered and pinned; `pending` remains a local declared after `scope` in `run_scope_incarnation` (driver.rs:1210), respecting the AdmissionRequest drop-order hazard; every driver-side retained-fold path wraps failed payloads before refusals and invariant panics; `ScopeRuntime::drop` teardown ordering and both lost-wake audits check out; the gate → dynamic-state → member/residency nesting is held consistently with removals carried out after unlock.

ralexstokes:
## Summary

- keep the selected child `Exit` in `RetainedExit` from classification until terminal disposal or successful restart publication
- derive dynamic-admission rejection cause once, document the nested self-shutdown consumption flag, and replace `DynamicState.entries`' test-only visibility split with `entries_mut()`
- exercise the real driver wake path after both primary and control sender-retention invariants are deliberately broken

## Issue dispositions

1. `handle_exit` now wraps the classified result immediately. Terminal dispatch unwraps only into `begin_terminal_disposal`, whose first operation re-retains it. Restart dispatch clones through the retained carrier, asserts successful publication, and releases the local raw copy only after the member record and lifecycle event own retained copies. `handle_exit_retains_the_selected_failure_across_dispatch_panics` poisons dynamic membership dispatch after classification and verifies the selected application error is destroyed off the driver thread.
2. The nested `Pending::Shutdown` assignment now documents both dependencies: the ancestor arm would only repeat the same idempotent `begin_drain(ShutdownRequested)`, while construction suppression samples the ancestor latch directly.
3. `handle_admission` samples the lifecycle once and produces one `Option<NotAdmittingCause>` before the rejection branch, preserving the existing Draining > StartupFailed > NoLiveIncarnation precedence.
4. `DynamicState.entries` has one declaration. Unit fixtures use a `#[cfg(test)] entries_mut()` accessor.
5. `closed_event_lanes_publish_a_retained_shutdown_completion` uses a hand-built `ScopeRuntime`, removes the actual retained sender for each lane in turn, and verifies a retained `ShutdownRequested` completion reaches the contained scope epilogue. The promote/admit invariant was already covered on current `main` by `refused_dynamic_admission_panics_after_control_plane_unlock`; this PR preserves that evidence while moving its state inspection through `entries_mut()`. That test verifies the framework diagnostic remains primary over a hostile response waker, both control-plane locks are released, and the lost response completes fail-closed.

## Validation

- `./tools/dev cargo test -p shelterwood --lib driver::tests` (162 passed)
- `./tools/dev just ci` (913 tests plus doctests, examples, rustdoc/API reachability, external consumer, and Nix formatting)

Closes #422

### DIFF
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -698,20 +698,21 @@ impl ScopeRuntime {
             request.complete(Err(ReserveError::NotAdmitting(NotAdmittingCause::Terminal)));
             return;
         };
-        if self
+        let lifecycle = self.supervisor.lifecycle();
+        let not_admitting = if lifecycle.is_draining() {
+            Some(NotAdmittingCause::Draining)
+        } else if lifecycle.startup_failed() {
+            Some(NotAdmittingCause::StartupFailed)
+        } else if self
             .dynamic
             .as_ref()
             .is_none_or(|current| !Arc::ptr_eq(current, &control))
-            || self.supervisor.lifecycle().is_draining()
-            || self.supervisor.lifecycle().startup_failed()
         {
-            let cause = if self.supervisor.lifecycle().is_draining() {
-                NotAdmittingCause::Draining
-            } else if self.supervisor.lifecycle().startup_failed() {
-                NotAdmittingCause::StartupFailed
-            } else {
-                NotAdmittingCause::NoLiveIncarnation
-            };
+            Some(NotAdmittingCause::NoLiveIncarnation)
+        } else {
+            None
+        };
+        if let Some(cause) = not_admitting {
             let (definition, removed) = self.root.with_observation_gate(|txn| {
                 cancel_dynamic_reservation_parts(&self.root, &control, &request.slot, txn)
             });
@@ -1083,6 +1084,86 @@ async fn run_scope(plan: ScopePlan, role: ScopeRole) -> RetainedStopReason {
     RetainedStopReason::new(run_scope_incarnation(plan, role, epoch).await)
 }
 
+async fn wait_for_scope_wake(
+    scope: &mut ScopeRuntime,
+    signal: &mut runtime::WatchReceiver<crate::cells::MemberRecord>,
+    event_receiver: &mut runtime::UnboundedMpscReceiver<DriverEvent>,
+    dynamic_event_receiver: Option<&mut runtime::UnboundedMpscReceiver<DriverEvent>>,
+    pending: &mut Vec<(ArbitrationClass, Pending)>,
+) -> Option<StopReason> {
+    let ancestor_shutdown = scope
+        .role
+        .ancestor()
+        .filter(|_| !scope.ancestor_shutdown_seen)
+        .map(|latches| latches.framework_shutdown.clone());
+    let ancestor_abort = scope
+        .role
+        .ancestor()
+        .filter(|_| !scope.ancestor_abort_seen)
+        .map(|latches| latches.abort.clone());
+    let ancestor_command = async move {
+        let shutdown = async move {
+            if let Some(shutdown) = ancestor_shutdown {
+                shutdown.fired().await;
+            } else {
+                std::future::pending::<()>().await;
+            }
+        };
+        let abort = async move {
+            if let Some(abort) = ancestor_abort {
+                abort.fired().await;
+            } else {
+                std::future::pending::<()>().await;
+            }
+        };
+        let _ = runtime::select_two(shutdown, abort).await;
+    };
+    match runtime::wait_scope(
+        runtime::ScopeWait {
+            signal: signal.changed(),
+            parent_shutdown: ancestor_command,
+        },
+        event_receiver,
+        dynamic_event_receiver,
+        scope.deadlines.next_deadline(),
+    )
+    .await
+    {
+        runtime::ScopeWake::Signal | runtime::ScopeWake::ParentShutdown => {}
+        runtime::ScopeWake::Message(None) | runtime::ScopeWake::ControlMessage(None) => {
+            // Every lane sender is retained by the scope runtime or one of its
+            // registered children until the driver loop exits. A closed
+            // receiver would otherwise make this select arm permanently ready
+            // and spin. Fail closed by returning a retained shutdown
+            // completion: a framework panic here would unwind the other
+            // receiver and any queued admission response wakers outside
+            // ScopeRuntime's contained epilogue.
+            let reason = StopReason::ShutdownRequested;
+            let root_exit = scope
+                .role
+                .is_root()
+                .then(|| RetainedExit::new(stop_reason_root_exit(&reason)));
+            scope.completion = Some(ScopeCompletion {
+                reason: RetainedStopReason::new(reason.clone()),
+                root_exit,
+            });
+            return Some(reason);
+        }
+        runtime::ScopeWake::Deadline => {
+            // A producer becoming ready at the same instant owns the tie over
+            // its deadline. Give tasks woken by that clock edge one turn to
+            // publish their retained readiness latch before collecting due
+            // registrations.
+            runtime::yield_now().await;
+        }
+        runtime::ScopeWake::Message(Some(event))
+        | runtime::ScopeWake::ControlMessage(Some(event)) => {
+            retain_woken_event(event, pending);
+        }
+    }
+    None
+}
+
 async fn run_scope_incarnation(
     mut plan: ScopePlan,
     role: ScopeRole,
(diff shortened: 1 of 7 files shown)


ralexstokes:
## Review

Reviewed the full diff against the lock rule, unwind/drop-glue safety, `RetainedExit` handling, and test non-vacuity. All five #422 dispositions check out. Two commits pushed, one issue filed, several claims verified by mutation.

### Verified by mutation (committed tree, mutation reverted after each)

- **`handle_exit_retains_the_selected_failure_across_dispatch_panics` is not vacuous.** Reverting the `RetainedExit::new(classify_exit_retaining(..))` wrap makes it fail with `assert_ne` (`left: ThreadId(2), right: ThreadId(2)`), and the injected panic lands at `driver/removal.rs:19` — `removal_latched`'s `.expect("dynamic-state mutex poisoned")`, i.e. inside `dispatch_membership_status`, genuinely post-classification. That mattered: had the panic landed *before* classification, the test would have passed on `main` too, because `RetainedRecordedOutcome` disposes off-driver on its own.
- **`closed_event_lanes_publish_a_retained_shutdown_completion` reaches the arm on both lanes.** Replacing the arm body with `panic!` fails the test; splitting the arm so only `ScopeWake::ControlMessage(None)` panics also fails it, so the `closed_control = true` half is not just re-running the primary case.
- **Retention claim on the restart branch.** `MemberTransition::RestartScheduled` sets `record.last_exit` and `transition_locked` refreshes the record's retained guards (`cells/member.rs:585-616`), so `drop(exit.into_exit())` after `assert!(published)` really is refcount traffic. The two `exit.as_exit().clone()` arguments are also strictly safer than `main`'s raw clones: the carrier outlives `publish_child_restart`, so a panic inside it can only unwind refcount traffic.
- **Terminal branch.** `exit.into_exit()` → `begin_terminal_disposal`'s `RetainedExit::new` is a panic-free argument window (left-to-right evaluation over `key`, the exit, `Some(incarnation)`, `startup`).

### Fixed on this branch

- `74cc0e0` — **the restart half of the item-1 window was untested.** Disposition 1 claims both consumption points, but only the Terminal one had coverage. Added `a_refused_restart_publication_retains_the_selected_failure`: corrupts the member stage to `Admitted` so `RestartScheduled` is illegal for the cell reducer, `publish_child_restart` refuses, and `assert!(published)` fires with the exit still in the carrier. Mutation-checked — reverting the wrap fails it on the same `assert_ne`.
- `4bcbdce` — `#[must_use]` on `wait_for_scope_wake`. It stages the fail-closed completion on the scope *before* returning the reason, so a caller that dropped the `Option` would keep looping on the same permanently-ready closed lane — the spin the arm exists to prevent. `Option` carries no `must_use` of its own, so this was a silent-regression seam in the newly extracted signature.

### Filed, not fixed here — #458

The retained chain that this PR establishes at `handle_exit` continues through `PendingTerminal.exit` and is **broken again at the far end of the same lane**:

- `handle_construction_disposed` (`child.rs:1325`) does `let exit = exit.into_exit();` and then holds the raw copy across `terminalize_child`, `set_terminal_disposal_pending`, `membership_status`, `flush_supervisor_effects`, `self.children[key]` (panics on a missing key) and `prune_terminal`, destroying it on the driver thread at the `Removing` early return or at function exit. The safety argument is that `terminalize_child` installed a record copy — but its `bool` is discarded, and a *losing* terminalization (record already `Terminal`) installs none; `ScopeCell::terminalize_child` retires only its own clone through `wakes.defer(drop(exit))`, so which handle is last becomes a race with the disposal job. That is the strong-count reasoning the project rules out.
- `fail_startup` (`startup.rs:117-132`) widens it: `StartupFailure { cause: Child { exit } }` is assembled raw and dropped raw on the `supervisor_fail_startup() == None` early return.

Filed rather than folded in because it needs a reachability adjudication (is a losing terminalization reachable with `pending_terminal` installed?) and because the fix changes drop timing — the losing branch would gain a `dispose_critical` job on a path that has none today. Not a regression from this PR; same hardening class as item 1.

#458 also records the optional follow-up of giving `begin_terminal_disposal` a `RetainedExit` parameter, with the trade-off that it would turn `refused_terminal_disposal_retires_its_exit_off_the_driver` into a type-level tautology — a design call rather than a review fix.

### Refuted / checked clean

- **Disposition 3 preserves precedence.** New order is Draining > StartupFailed > `!ptr_eq` → `NoLiveIncarnation`; the old predicate/cause pair produced the identical rejection set and the identical cause for every combination, including draining-*and*-stale-control.
- **Disposition 5's "already covered on `main`" claim holds.** `refused_dynamic_admission_panics_after_control_plane_unlock` really does pin the promote/admit arm: framework diagnostic primary over the hostile `PanicWake`, both control-plane locks unpoisoned, arena insertion and entry promotion both committed, and the lost response resolving to `NotAdmitting(Terminal)`.
- **`wait_for_scope_wake` extraction is faithful.** Argument evaluation order, the `Deadline` yield, `signal.changed()` borrow and future-temporary drop points all match the inlined original; `pending` stays a caller local declared after `scope`, so the `AdmissionRequest` drop-order hazard is untouched.
- **The closed-lane fixture closes the lane it claims to.** `for_test` gives `disposal_events` its own channel, so replacing `scope.events` (resp. the `DynamicControl`) really does drop the last sender; a lane that stayed open would hang rather than pass.
- **"retained shutdown completion" in the test name.** `stop_reason_root_exit(ShutdownRequested)` is `Exit::completed(Observed)`, so the `RetainedExit` there is inert — but the name refers to the `RetainedStopReason` completion, which is type-level. Fine as written.
- **`entries_mut()` on read-only probes.** Slightly odd (`let mut state` for a `contains_key`), but a second `#[cfg(test)] entries()` accessor would partly re-create the dual declaration item 4 removed. Leaving as is.
- No new lock sites, no user code under a framework mutex, no `strong_count` probes anywhere in the diff.

### Local verification

`cargo nextest run -p shelterwood --lib` → 364/364 pass; `cargo clippy -p shelterwood --all-targets` clean. CI green on `4bcbdce` (run 32602359806, `headSha` confirmed).

PR #449, opened 2026-08-22
Tighten observation cell invariants

### LINKED ISSUE #420: Observation cells: attach_mailbox terminal-branch indirection, unreachable close-wins branch, unverifiable *_locked gate-identity convention
From the 2026-08-22 post-fold full review. No bugs in `cells/observe.rs` / `gate.rs` / `member.rs`; items below, most severe first.

## 1. Simplification: `attach_mailbox`'s terminal branch manufactures a synthetic losing race

`crates/shelterwood/src/cells/member.rs:634-680`. Attaching to an already-terminal member installs the control + teardown, then calls `terminalize_locked` with a *clone* of the stored winning exit purely to reach the teardown-draining code. In production that call is guaranteed to take the losing path (mailbox-Terminal implies record-Terminal under the gate; the other combination is only constructible via the test-only `stage_terminal_before_mailbox`), so it mutates-and-refreshes the record as a no-op and defers a `RetainedExit` whose drop submits a critical-disposal job (potential native thread start) for a clone that is provably non-last. A direct "take and defer the teardown" step does the same work without the synthetic race or the disposal job. If the indirection is kept as single-path defense, a comment saying the losing path is the expected production route would help.

## 2. Simplification: close-wins is enforced twice; the coalesce branch is unreachable

`crates/shelterwood/src/cells/observe.rs:581-591` vs :732 and :768. `SnapshotPublication::coalesce`'s `if self.closed { return newer.projection; }` branch is unreachable: both staging entry points (`SnapshotHub::publish` and `SnapshotHub::close`) decline via `txn.snapshot_hub_will_close(...)` before staging anything onto a hub with a staged close. Its comment presents it as the load-bearing rule, but the `will_close` guards are what fire — and the two mechanisms differ subtly (a coalesced publish would not set `published`; the `will_close` decline retires the producer without touching the staged close). Either delete the branch and let the guards carry the rule, or document it as defense-in-depth so a reader knows which layer is authoritative.

## 3. Quality: the `*_locked` gate-identity convention is unverifiable

`member.rs` (`transition_locked`:579, `terminalize_locked`:696, `update_locked`:546) and the hub methods in `observe.rs` trust by convention that the passed `&mut ObservationTxn` holds *this* member's/hub's current gate; `ObservationTxn` carries no gate identity, so nothing — not even a debug assertion — can check it. Callers in `plan.rs:165` and `scope.rs` route txns to members whose gate identity is an invariant of admission/adoption choreography several files away. A debug-only `ObservationGate` reference on the txn (with `shares_gate` asserts in the `*_locked` entry points, `detached()` opting out) would turn the deepest convention in this module into a checked one. Current callers all verified correct.

## 4. Coverage: lifecycle lag-fold and the `pending` fallback

`observe.rs:829-834, 864-874`. The `pending` field's `BroadcastReceive::Item` arm exists only for a broadcast misbehavior the pinned tokio version plus the `is_power_of_two` const assert (:26) already rule out — dead in practice and untested. Separately, the *reachable* half — folding a ring `Lagged(overflow)` into an in-flight explicit-lag marker (`dropped.saturating_add(overflow)`, :867) — has no test: `a_retained_event_after_explicit_lag_remains_subject_to_later_overflow` (:1467) deliberately consumes the marker before overflowing. A slow consumer plus subtree-removal forwarding loss (`cells/scope/projection.rs:289-292`) reaches it in production. One test that arms an explicit lag, overflows the ring, and asserts the single combined `Lagged { dropped }` count.

## 5. Coverage: `with_handoff_gate`'s adoption retry iteration is unpinned

`member.rs:515-544`. The observation-side retry is pinned (`pre_admission_observer_retries_after_gate_handoff`, driver/tests/observation.rs:1215) and adoption blocking behind an in-flight edge too (:1254), but the adoption/admission retry loop itself — the installed gate changed between `report_capture` and `current.lock()`, forcing a second `GateCapture::Adoption` iteration — is not. The probe machinery exists precisely to make this testable; use `replace_observation_gate` while an adopter is parked on the old gate.

## 6. Simplification: triplicated hub-install choreography

`observe.rs:600-642, 680-705, 754-764`. `SnapshotPublication::install`, the receiverless refresh in `subscribe`, and `close`'s init path each hand-roll the same `modify_silently` + `retired` + `generation_exhausted` + closed-check dance with different pulse/mint decisions. One helper parameterized by mint-and-pulse policy collapses three subtle blocks and makes the deliberate asymmetries declarations instead of emergent differences. Related: `RetainedLifecycleEvent::into_public` (:150-165) and `RetainedScopeSnapshot::into_public/into_parts` (:398-415) are the same `Arc::try_unwrap`-else-clone shape written twice.

## 7. Fold scar: `observe.rs` self-references through the façade path

`observe.rs:662, 723, 749, 929, 947, 980, 995` use `&mut crate::cells::ObservationTxn<'_>` and :18 `use crate::cells::RetainedExit`, where siblings (`gate.rs`, `member.rs`) use `super::`. Cosmetic pre-fold leftover that misstates the module relationship.

ralexstokes:
Closes #420.

## Summary

- drains an already-terminal member's newly attached mailbox teardown directly, without manufacturing a losing terminalization or cloning the winning exit
- gives `ObservationTxn` a debug-only gate identity and checks member writers plus snapshot/lifecycle publication boundaries against the current resident-tree gate
- makes the transaction-level `snapshot_hub_will_close` checks authoritative, and consolidates snapshot installation/retirement/generation/pulse choreography behind one policy-driven helper
- folds the duplicate retained-`Arc` extraction shape into one helper, uses sibling module paths throughout `observe.rs`, and reduces `SnapshotReceiver::borrow_latest_and_closed` to `pub(crate)`
- pins explicit lifecycle loss plus broadcast overflow into one leading saturated lag marker

## Issue dispositions

1. `attach_mailbox`: replaced the synthetic losing `terminalize_locked` call with direct post-unlock teardown finishing. `attaching_to_a_terminal_member_does_not_republish_its_record` pins the absence of the synthetic record edge, and the existing reentrant-wake test now uses a production-reachable already-terminal member.
2. close-wins: removed the unreachable `self.closed` coalesce branch. `publish` and `close` remain the authoritative pre-staging guards; `publication_after_staged_close_is_not_built` covers that route.
3. gate identity: `ObservationTxn` retains a debug-only reference to its acquired gate, while `detached()` opts out for isolated hub unit tests. `update_locked`, legal/rejected `transition_locked` calls, `terminalize_locked`, and every scope snapshot/lifecycle boundary assert shared identity. The new check also found two tests with impossible "resident on a foreign gate" fixtures; both now perform real admission first. The diagnostic method and its panic regression are themselves debug-only, keeping release builds warning- and dead-code-clean.
4. lifecycle lag: added `explicit_lag_and_ring_overflow_fold_into_one_leading_marker`, covering the `saturating_add` fold of explicit loss plus ring overflow and the retained suffix. The `pending` item fallback is deliberately retained as total defense: with Tokio 1.53.1 and the power-of-two capacity const assertion, `len() > capacity` must yield `Lagged`, so an `Item` there is not constructible through the pinned implementation.
5. handoff retry: current `main` already contains `adoption_retries_when_the_captured_child_gate_has_already_been_replaced`, which parks adoption on the old gate, replaces the installed gate, observes the second `GateCapture::Adoption`, and verifies successful admission. This PR keeps and runs that exact regression.
6. repeated choreography: `install_snapshot` owns closed-state refusal, displaced/uninstalled retirement, optional generation mint, deferred exhaustion panic, and optional pulse. Call sites declare transaction-install versus receiverless-refresh mint/pulse policy; `SnapshotHubState::new` unifies open/closed initialization. `unwrap_or_clone` removes the duplicate retained-`Arc` extraction.
7. fold scar: production and unit-test references in `observe.rs` now resolve through `super`/local imports rather than the façade path.
8. follow-up visibility: `borrow_latest_and_closed` is now `pub(crate)` and its obsolete doc-hidden boundary ruling is removed. The rustdoc JSON reachability gate reports clean.

## Validation

- `./tools/dev cargo test -p shelterwood --lib` (364 passed)
- `./tools/dev cargo check --workspace --release`
- `./tools/dev cargo test -p shelterwood --lib --release` (363 passed; the debug-only gate-identity regression is intentionally absent)
- `./tools/dev cargo +nightly clippy -p shelterwood --all-targets -- -D warnings`
- `./tools/dev just ci` (914 nextest tests; doctests, examples, rustdoc/API reachability, external consumer, and Nix formatting all pass)

### DIFF
--- a/crates/shelterwood/src/cells/gate.rs
+++ b/crates/shelterwood/src/cells/gate.rs
@@ -69,14 +69,18 @@ impl ObservationGate {
 /// from stranding already-committed wakes.
 pub(crate) struct ObservationTxn<'a> {
     guard: Option<MutexGuard<'a, ()>>,
+    #[cfg(debug_assertions)]
+    gate: Option<&'a ObservationGate>,
     effects: Vec<Box<dyn FnOnce()>>,
     snapshots: Vec<SnapshotPublication>,
 }
 
 impl<'a> ObservationTxn<'a> {
-    pub(super) fn new(guard: MutexGuard<'a, ()>) -> Self {
+    pub(super) fn new(gate: &'a ObservationGate, guard: MutexGuard<'a, ()>) -> Self {
         Self {
             guard: Some(guard),
+            #[cfg(debug_assertions)]
+            gate: Some(gate),
             effects: Vec::new(),
             snapshots: Vec::new(),
         }
@@ -86,9 +90,28 @@ impl<'a> ObservationTxn<'a> {
     pub(crate) fn detached() -> Self {
         Self {
             guard: None,
+            #[cfg(debug_assertions)]
+            gate: None,
             effects: Vec::new(),
             snapshots: Vec::new(),
         }
+    }
+
+    /// Checks that a locked cell writer received the transaction for its
+    /// current resident-tree gate.
+    ///
+    /// Unit tests for isolated hub mechanics use [`Self::detached`], which
+    /// deliberately carries no identity and therefore opts out.
+    pub(crate) fn debug_assert_gate(&self, gate: &ObservationGate) {
+        #[cfg(debug_assertions)]
+        if let Some(held) = self.gate {
+            debug_assert!(
+                held.shares_gate(gate),
+                "a locked observation writer requires its current tree gate"
+            );
+        }
+        #[cfg(not(debug_assertions))]
+        let _ = gate;
     }
 
     pub(crate) fn defer(&mut self, operation: impl FnOnce() + 'static) {
@@ -215,7 +238,7 @@ mod tests {
     fn observation_txn_commit_unlocks_before_effects_and_drains_once() {
         let gate = ObservationGate::new();
         let order = Arc::new(Mutex::new(Vec::new()));
-        let mut txn = ObservationTxn::new(gate.lock());
+        let mut txn = ObservationTxn::new(&gate, gate.lock());
         for value in [1, 2] {
             let gate = gate.clone();
             let order = Arc::clone(&order);
@@ -266,7 +289,7 @@ mod tests {
         );
 
         let payload = catch_unwind(AssertUnwindSafe(|| {
-            let mut txn = ObservationTxn::new(gate.lock());
+            let mut txn = ObservationTxn::new(&gate, gate.lock());
             txn.pulse(&hostile_sender);
             txn.pulse(&later_sender);
         }))
@@ -295,7 +318,7 @@ mod tests {
             let effects = Arc::clone(&effects);
             let gate = gate.clone();
             move || {
-                let mut txn = ObservationTxn::new(gate.lock());
+                let mut txn = ObservationTxn::new(&gate, gate.lock());
                 let first = Arc::clone(&effects);
                 let first_gate = gate.clone();
                 // An assertion here would be inert: `PanicAccumulator`
--- a/crates/shelterwood/src/cells/gate.rs
+++ b/crates/shelterwood/src/cells/gate.rs
@@ -77,6 +77,8 @@ pub(crate) struct ObservationTxn<'a> {
 
 impl<'a> ObservationTxn<'a> {
     pub(super) fn new(gate: &'a ObservationGate, guard: MutexGuard<'a, ()>) -> Self {
+        #[cfg(not(debug_assertions))]
+        let _ = gate;
         Self {
             guard: Some(guard),
             #[cfg(debug_assertions)]
@@ -102,16 +104,14 @@ impl<'a> ObservationTxn<'a> {
     ///
     /// Unit tests for isolated hub mechanics use [`Self::detached`], which
     /// deliberately carries no identity and therefore opts out.
+    #[cfg(debug_assertions)]
     pub(crate) fn debug_assert_gate(&self, gate: &ObservationGate) {
-        #[cfg(debug_assertions)]
         if let Some(held) = self.gate {
             debug_assert!(
                 held.shares_gate(gate),
                 "a locked observation writer requires its current tree gate"
             );
         }
-        #[cfg(not(debug_assertions))]
-        let _ = gate;
     }
 
     pub(crate) fn defer(&mut self, operation: impl FnOnce() + 'static) {
--- a/crates/shelterwood/src/cells/member.rs
+++ b/crates/shelterwood/src/cells/member.rs
@@ -644,13 +644,20 @@ impl MemberCell {
                     MemberMailbox::Terminal {
                         control, teardown, ..
                     } => {
+                        // Unreachable under the gate — `terminalize_locked`
+                        // only stores a teardown alongside a control, which the
+                        // arm above already rejects — and kept as total
+                        // defense against ever draining one twice.
                         if teardown.is_some() {
                             rejected = Some(mailbox);
                             None
                         } else {
-                            let teardown = mailbox.prepare_termination(txn);
+                            // The stored `teardown` field stays empty: the
+                            // drain below is the only reader this one can have,
+                            // and it owns it before the mailbox guard drops.
+                            let drained = mailbox.prepare_termination(txn);
                             *control = Some(mailbox);
-                            teardown
+                            drained
                         }
                     }
                 }
@@ -903,6 +910,11 @@ mod tests {
         );
     }
 
+    /// Pins the invariant the direct teardown drain has to keep, not the
+    /// rewrite itself: the losing `terminalize_locked` it replaced published
+    /// no record edge either, so no test can separate the two shapes. The
+    /// teardown half is pinned by
+    /// `attach_to_a_terminal_member_finishes_record_before_mailbox_wake`.
     #[test]
     fn attaching_to_a_terminal_member_does_not_republish_its_record() {
         let scope = isolated_scope("root", ScopeFlavor::Ordered);
@@ -922,7 +934,7 @@ mod tests {
                 .as_mut()
                 .poll(&mut Context::from_waker(Waker::noop()))
                 .is_pending(),
-            "terminal attachment drains teardown without a synthetic losing transition"
+            "attaching to an already-terminal member publishes no second record edge"
         );
     }
 
--- a/crates/shelterwood/src/cells/member.rs
+++ b/crates/shelterwood/src/cells/member.rs
@@ -820,6 +820,7 @@ mod tests {
     use super::*;
     use crate::cells::test_support::{ThreadProbe, isolated_scope};
 
+    #[cfg(debug_assertions)]
     #[test]
     #[should_panic(expected = "a locked observation writer requires its current tree gate")]
     fn locked_member_writer_checks_transaction_gate_identity() {
--- a/crates/shelterwood/src/cells/member.rs
+++ b/crates/shelterwood/src/cells/member.rs
@@ -332,17 +332,6 @@ impl MemberCell {
         self.record.watcher()
     }
 
-    #[cfg(test)]
-    pub(crate) fn stage_terminal_before_mailbox(&self, exit: Exit) {
-        let mut mailbox = self.mailbox.lock().expect("member mailbox mutex poisoned");
-        assert!(matches!(*mailbox, MemberMailbox::Unattached));
-        *mailbox = MemberMailbox::Terminal {
-            control: None,
-            exit: RetainedExit::new(exit),
-            teardown: None,
-        };
-    }
-
     /// Reports terminality or drain-entry terminal-disposal intent.
     ///
     /// `Acquire` pairs with the `Release` store in
@@ -427,7 +416,7 @@ impl MemberCell {
             report_capture();
             let guard = gate.lock();
             if gate.shares_gate(&self.current_observation_gate()) {
-                let mut txn = ObservationTxn::new(guard);
+                let mut txn = ObservationTxn::new(&gate, guard);
                 return operation
                     .take()
                     .expect("member observation operation runs exactly once")(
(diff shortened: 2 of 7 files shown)


ralexstokes:
## Review

Read against #420's seven items, the lock rule, and the gate-ordering exemptions in `CLAUDE.md`. **No correctness bugs found.** Every claimed pin that could be checked by mutating the code was checked; results below. Fixes are on `41764bb`, CI green on that SHA.

### Verified by mutation (committed tree, mutation reverted after each run)

| Claim | Mutation | Result |
| --- | --- | --- |
| Item 4: the new test pins the `dropped.saturating_add(overflow)` fold | replaced the fold with `let _ = overflow;` | **FAILS** (`left: Lagged { dropped: 5 }`, `right: 8`) — real pin |
| Item 2: `publish`'s `snapshot_hub_will_close` is the authoritative close-wins guard, so `coalesce`'s branch can become a `debug_assert!` | deleted `txn.snapshot_hub_will_close(self)` from `SnapshotHub::publish` | **FAILS** `publication_after_staged_close_is_not_built` — the premise of the new assert is genuinely covered |
| Item 5: `adoption_retries_when_the_captured_child_gate_has_already_been_replaced` pins the `with_handoff_gate` retry iteration | made the post-acquire re-check unconditional | **FAILS** (times out waiting for the second `GateCapture::Adoption`) — real pin, item 5 is satisfied by the existing test |
| Item 1: the new unit test pins the removal of the synthetic losing terminalization | restored the old `terminalize_locked` indirection verbatim | **PASSES** — see below |

Also confirmed the `close`-side `snapshot_hub_will_close` guard is uncovered (removing it leaves all 364 lib tests green). That is fine and I did not change it: a same-transaction double close is unconstructible, because `close_observation_locked` is `SnapshotHub::close`'s only caller and its `observation.closed` flag short-circuits the second entry. So the `coalesce` assert's premise rests on the `publish` half, which *is* pinned. Worth knowing rather than acting on.

### Findings

**1. `attaching_to_a_terminal_member_does_not_republish_its_record` does not discriminate the change it is attached to (fixed: comment only).** The losing `terminalize_locked` it replaced also left `published == false` and therefore also published no record edge, so the test passes verbatim on the pre-rewrite shape. It is still a worthwhile invariant pin — it just isn't a regression for this edit, and nothing can be: the rewrite has no observable delta (that is exactly why #420 filed it as a *simplification*). I also checked the teardown half: replacing `teardown.finish()` with `drop(teardown)` keeps `attach_to_a_terminal_member_finishes_record_before_mailbox_wake` green, because `MailboxTeardown::drop` runs `finish_framework` as a fail-safe — so that test pins "the teardown is drained inside the transaction", which is the property that matters, not the venue. Added a doc comment saying which half each test carries and corrected the assertion message, which claimed to observe the absence of the synthetic transition.

**2. Two of the new gate-identity checks are vacuous (fixed: removed).** `subscribe_snapshots` and `subscribe_lifecycle` asserted `debug_assert_gate(&self.current_observation_gate())` on a transaction they obtained from `self.with_observation_gate` one line earlier — which loops until `gate.shares_gate(&self.current_observation_gate())` holds, over the same `MemberCell` the assert re-reads. Nothing can change the installed pointer in between, because doing so requires the gate we hold. The check could never fire, and leaving it implies a boundary that is actually structural. Removed both and left a comment stating that only a `*_locked` writer receiving someone else's transaction has an identity left to verify. The five load-bearing ones stay: `update_locked`, `transition_locked`, `terminalize_locked`, `publish_snapshot_chain_locked` (self *and* each ancestor), `emit_locked`, `close_observation_locked` — those do get cross-scope transactions (`terminalize_child` → nested `publish_stopped_locked`/`close_observation_locked`, the ancestor chain, `admit_observation_gate` → `transition_locked` under the parent gate after the handoff install).

**3. `unwrap_or_clone` reimplements `Arc::unwrap_or_clone` (fixed).** Stable since 1.76, well under the 1.97.1 MSRV, and identical semantics (`try_unwrap().unwrap_or_else(|arc| (*arc).clone())`). Item 6 is about collapsing a duplicated shape; collapsed onto std instead of introducing a third local copy.

**4. `install_snapshot`'s `closed` parameter reads like it can clear a closure (fixed: comment).** The receiverless refresh passes `false` and the helper then does `state.closed = closed` unconditionally. It is safe only because the `if state.closed { return; }` above it means the write always re-asserts a flag that was already `false`. That is a load-bearing invariant of the newly *shared* helper, so it is now stated on the helper rather than left to be re-derived per call site.

**5. `attach_mailbox`'s shadowed `teardown` (fixed: renamed + commented).** The arm binds `teardown: &mut Option<…>` and then shadows it with `let teardown = mailbox.prepare_termination(txn)`, which reads like an assignment to the field but deliberately leaves the field empty. Renamed the local to `drained` and recorded that the `teardown.is_some()` refusal is now unreachable rather than merely unreached (`terminalize_locked` only ever stores a teardown alongside a control, which the preceding arm rejects).

**6. Trailing-comma style on the two new residency fixtures (fixed).** `ResidentProjection::new(Arc::clone(&member), None,)` — rustfmt tolerates it, but every other call site in the tree omits it. Also added a line to the `scope.rs` fixture saying *why* admission is now needed (residency is what puts the member on the root's gate), since the diff otherwise reads as an unexplained setup change.

### Refuted / checked and fine

- **`install_snapshot` is behaviour-preserving** across all three folded sites — I diffed the branches by hand: `mint_generation` keeps the `self.published` short-circuit; the receiverless refresh keeps mint-yes/pulse-no; effect ordering (uninstalled → retired → exhaustion panic → pulse) is unchanged; `state.closed` writes are no-ops on the refresh path. One improvement rides along that the description does not mention: `subscribe` no longer clones `initial` into the hub on the initialising path and then defers a drop of the original — it moves it in. Strictly fewer `Arc` clones and one fewer deferred drop.
- **Lock rule.** No new user-code-reachable work under a lock. `debug_assert_gate` reads an `RwLock` leaf already read under the gate elsewhere (`publish_drain`, `adopt_observation_gate`), and the `#[cfg(debug_assertions)]` at the *call sites* — not just inside the method — is what keeps the release build from paying `current_observation_gate()`'s read-and-clone; worth keeping in that shape. The `coalesce` and `attach_mailbox` panics follow the documented "compute the verdict, `debug_assert!` where releasing is impossible" convention, and the gate tolerates poisoning by design.
- **Gate ordering.** Audited every `*_locked` writer call site for the new asserts firing on a legitimate path: `plan.rs:491`, `driver.rs:782/798`, `admission_control.rs:438`, `scope.rs:672/686/711/740/798/1015/1358/1570/1644`. All reach members that are either `self.member` under `self.with_observation_gate`, residents (gate shared by the admission invariant), or reserved slots that adopted at `reserve` time (`admission_control.rs:294`). The one shape that could strand a member on a stale gate — a dynamic reservation not yet in residency when its scope is re-homed — is already ruled out by the "no running scope is subsequently re-homed" argument on `adopt_descendant_observation_gates_locked`.
- **`ObservationTxn::detached()` opting out is not a production fail-open** — it is `#[cfg(test)]`, so every production transaction carries an identity.
- **`borrow_latest_and_closed` → `pub(crate)`**: all eleven call sites are in-crate; release/`--all-targets` check and the reachability lane are green.
- **The `pending` `BroadcastReceive::Item` fallback staying untested** is the right call, and the description's reasoning (pinned Tokio + the power-of-two const assert make it unconstructible) matches what the code can reach.

### Not fixed, noted only

`transition_locked` and `terminalize_locked` have no gate-identity regression of their own — `locked_member_writer_checks_transaction_gate_identity` covers `update_locked` only. Since all three delegate to the same `ObservationTxn::debug_assert_gate`, three near-identical `should_panic` tests would pin the call site rather than the mechanism, so I left it. Flagging in case you want them anyway.

Nothing filed; nothing here rises to a separate issue. **CI: green on `41764bb05d4fe5f9b255bdd5f59d8c7f48ef2619`** (run `32602247001`). No merge blockers from my side.

ralexstokes:
## Combined-tree conflict with #445 — adjudicated and fixed on `2502e4e`

After `update-branch` pulled the merged #442–#447/#452 round in, combined-tree CI failed:

```
shelterwood cells::scope::tests::rejected_restart_publication_disposes_its_lifecycle_exit_off_thread
panicked at cells/gate.rs:110: a locked observation writer requires its current tree gate
```

Reproduced locally on `3d104c4`.

**Adjudication: the fixture is wrong, not the assert.** `rejected_restart_publication_disposes_its_lifecycle_exit_off_thread` arrived from #445 while this branch was open. It builds its subject with `child_member(&root, "invalid")`, which mints a `MemberCell` and never enters it into residency — so the member still carries the private `ObservationGate` that `MemberCell::new` gave it. `root.publish_child_restart` then takes the *root's* gate and hands that transaction to `member.transition_locked`, i.e. a locked writer driven by a transaction for a different tree.

That state is unconstructible in production: the driver publishes a restart only for a resident child, and residency (`admit_child_locked` → `admit_observation_gate`) is exactly what installs the shared tree gate. So this is the third instance of the "resident on a foreign gate" fixture the new check has found — the same class as the two the PR description already lists, just added concurrently rather than pre-existing. The assert stays load-bearing and unchanged.

**Fix:** the same repair its sibling `rejected_stage_transition_disposes_its_lifecycle_exit_off_thread` already carries on this branch — admit the member first, then force `record.stage` back to `Reserved` so the reducer still refuses the `Reserved → Restarting` projection transition, now under a real gate. `subscribe_lifecycle` stays after the admission so the `Added` edge is not in the stream. All four of the test's assertions still hold and still mean what they did: the refusal (`!publish_child_restart`), the untouched `Reserved` stage, the empty lifecycle stream, and — the point of #445's version — the `ThreadProbe` reporting a disposal thread different from the caller's.

Worth noting for the rest of the round: the merge itself was clean, and #445's rename/extension of the sibling test correctly kept the admission fixture #449 had added to it. This was purely the one test #445 introduced after that fixture pattern was established.

**Validation:** `cargo nextest run -p shelterwood` → 737/737 green; `cargo +nightly clippy -p shelterwood --all-targets -D warnings` clean; fmt clean. **CI green on `2502e4eb8a71704dd3d7197417b0dc53a685f983`** (run `32604099569`), PR now `MERGEABLE`/`CLEAN`.

Issue #429, opened 2026-08-22
Review 2026-08-22 tracking: post-fold full review (zero production bugs; 11 finding groups)


ralexstokes:
Tracking issue for the 2026-08-22 full review — the first after the #368 crate fold landed (PR #416). Thirteen parallel reviewers covered every source module, both test suites, the public-API/implementation fit, cross-module lock ordering, and docs/spec/tooling drift.

**Headline: zero reachable production bugs.** The round's only P1 findings are three test defects (probes that pass on a regressed tree). Everything else is hardening on unreachable-today panic paths, simplification, coverage, and doc drift.

## Issues filed, by priority

**P1 — test defects**
- #418 — inert/vacuous test probes: `sync.rs:1576` watcher_count vs `WaiterRegistry`, in-handler asserts at `tests/events.rs:533` and `tests/raw.rs:95` eaten by the once-child `Finished` path (recurrence of the swallowed-assertion class), plus a small sweep list.

**P2 — production hardening (unreachable today; each defeats a guarantee the code exists to provide)**
- #419 — cells/scope: `publish_drain` diagnostic unwind can inline-drop a user-error-bearing startup `Exit`; tautological `prune_child_locked` re-check; unchecked `last_mut()` announced pairing in admission; venue/negative-filter coverage.
- #421 — mailbox: `accept_locked` mismatch fallback non-total for Latest (would poison the mutex it exists to protect); structural `MailboxEffects` emptiness gate; `MailboxKind`/`ResolvedMailbox` dup; untested `bind` guard cluster.
- #422 — driver: `handle_exit` raw-`Exit` window across fallible calls; `ancestor_shutdown_seen` comment; untested fail-closed arms.
- #423 — core engine/supervisor: confirmed dead code in `dispatch_exit` (assert-then-unreachable-if, branch contradicts its own comment); unpinned Ready-in-`Stopping` (exploration-walk blind class); alphabet exhaustiveness guard.
- #427 — lock-rule accounting + doc drift: dynamic-state mutex held *around* gate acquisition on two admission paths (safe via two unstated invariants; CLAUDE.md accounting needed); CLAUDE.md names the deleted `Acceptance` type; tokio-in-façade-tests policy re-adjudication.

**P3 — simplification / coverage / docs**
[...]

PR #450, opened 2026-08-22
Harden runtime completion and teardown seams

### LINKED ISSUE #426: Runtime adapter: opaque poll-after-ready panic, BlockingJob Drop expect vs DisposalJob rationale, missing ActorWork pins
From the 2026-08-22 post-fold full review. No bugs; `dispose_critical`'s keep-the-job obligation, the `DisposingReceiver` mid-flight-send ownership, and the `blocking_pool_accepted` strong-count sample were all traced against the pinned tokio 1.53.1 source and check out. Items below, most severe first.

## 1. Quality: poll-after-ready panics with tokio's opaque `"called after complete"`

`crates/shelterwood-runtime/src/sync.rs:539` (`OneShotReceiver::poll_receive`) and :567-573 (the `Err(SENDING)|Err(SENT)` branch of `close_and_poll_receive`). Verified in pinned tokio 1.53.1 (`sync/oneshot.rs:1291`): `Receiver::poll` panics once it has yielded `Ready`, while `try_recv`/`close` stay total. Nothing in the wrapper's state word guards this: after `Ready(Some(value))` the state is `SENT`, so a re-poll re-enters tokio's poll and panics with a message naming neither shelterwood nor the seam. `DisposingReceiver` inherits it and is slightly worse: after ready retirement the proxy is uninstalled, so a re-poll re-probes with the noop waker and hits the same panic. Every current caller is `poll_fn(...).await`-driven and cannot re-poll — unreachable today — but the type is exposed through `ErasedOneShotReceiver::poll_receive`, where a façade regression would surface as tokio's panic instead of a framework diagnostic. A `SENT`-and-delivered fused state (or a "poll at most once past ready" doc on both receivers) makes the failure mode framework-owned. `DisposingReceiver::drop` is unaffected (`close_and_take` uses only the total pair).

## 2. Quality: `BlockingJob`'s mutex `.expect` in `Drop` contradicts `DisposalJob`'s stated rationale

`crates/shelterwood-runtime/src/spawn.rs:288` (also :247, :274), including inside `Drop for BlockingJob` — while `DisposalJob::lock_state` (`disposal.rs:72-76`) deliberately tolerates poison, with the comment that the critical drop path "must stay panic-free because it can run inside an unwind" (disposal.rs:94-99). The BlockingJob mutex guards only a `take()`, so it is unpoisonable in practice — but that is exactly the invariant the DisposalJob comment refused to rest on, and `BlockingJob::Drop` runs on Tokio teardown and can run during an unwind; a poisoned `.expect` there is a double panic → abort. One-line hardening: `PoisonError::into_inner` for shape-consistency, or a comment stating why the expect is acceptable here and not there.

## 3. Maintainability: `WaiterRegistry` family missing from the venue-split cross-citations

`crates/shelterwood-runtime/src/waker_proxy.rs:19-27` documents the per-seam adjudication of pending-retirement venues (inline for the mailbox reply receiver, disposal lane for one-shot/join/offload), but the third family — `LatchWait::drop` / `WatchWait::drop` (`sync.rs:274-279`, :927-932`), which destroy a registered caller-waker clone inline on the holder's thread — is not mentioned. Evidently deliberate (the registry never hands the waker to an external primitive, so it sits outside the closed #409 class, and tests pin the inline venue), but a venue-consistency auditor has to rediscover the ruling. One sentence in the `WaiterRegistry` doc (sync.rs:16-21) or the venue note.

## 4. Coverage: `ActorWork` has no direct tests

`spawn.rs:101-142`. Abort-on-drop, abort idempotence, and join-after-abort → `JoinOutcome::Cancelled` are exercised only transitively through façade offload/context tests — the exact area of past cancellation bugs (#349 oneshot-drop abort, #350 offload-cancel strand). Two cheap in-file pins (drop cancels the task; `join` after `abort` classifies as `Cancelled`, not `Panic`) keep the adapter's own contract mutation-visible. Minor siblings with zero direct coverage: `UnboundedMpscSender::send`'s closed-lane `Err(T)` path, `JitterRng`.

## 5. Maintainability: 1.53.1-specific behavior in test_support lacks a degradation note

`test_support.rs:121-148` (`submit_during_blocking_pool_shutdown`) depends on tokio 1.53.1 *running* a queued blocking task while draining `shutdown_background`. `blocking_pool_accepted` (spawn.rs:331-350) documents its own pin plus fail-safe degradation direction; the helper doesn't — a future tokio that dropped queued blocking tasks would make its dependent regression tests hang on `recv_timeout` rather than fail crisply. Acceptable under the hard `=1.53.1` pin; add it to the upgrade re-audit list.

## 6. Micro (optional)

`spawn_blocking_fallback_with`'s two-arm match (spawn.rs:201-209) both just drop; it exists to host ownership commentary that could live on a single `let _ = spawn(worker);`. (`close_and_poll_receive` being production-dead is already self-adjudicated in-code; `Signal`/`SignalWatcher`'s 1:1 wrapping is load-bearing narrowing — no action on either.)

ralexstokes:
Closes #426.

## What changed

1. `OneShotReceiver` now records every terminal receive edge and rejects later polls with a Shelterwood-owned diagnostic instead of leaking Tokio's opaque `called after complete` panic. The state covers direct polling, `close_and_poll_receive`, `close_and_take`, test-only `try_receive`, consuming `receive`, `DisposingReceiver`, staged publication, and the erased mailbox adapter.
2. `BlockingJob` centralizes poison-tolerant pending-state recovery for `run`, the acceptance sample, and—critically—`Drop`. A poison regression proves teardown remains panic-free and actually retires the captured operation.
3. `WaiterRegistry` now records why latch/watch cancellation retires registered caller-waker clones inline, distinct from external-primitive proxy retirement.
4. `ActorWork` has direct abort-on-drop and idempotent-abort/join-cancelled tests. The adjacent unbounded sender return path and `JitterRng` bounds also receive direct pins.
5. `submit_during_blocking_pool_shutdown` now names the Tokio 1.53.1 queued-task premise, its exact-version support, and the required upgrade re-audit/failure behavior.
6. The optional `spawn_blocking_fallback_with` match remains explicit: its two arms document different ownership evidence (the handle is detached on success; the consumed closure leaves the authoritative job for disposal on error), so collapsing it would erase the reason the helper exists.

## Follow-up findings from #429

- `FrameworkPlain` is now genuinely sealed with a private supertrait, and a compile-fail doctest pins that foreign response types cannot select inline destruction. Because the intentional admission/removal response types live in the separate façade crate, those cold cancellation paths now use the existing detached-disposal constructor rather than weakening the seal with a public unsafe opt-in.
- `modify_silently` and `read_with` explicitly state that their closures run under the watch value mutex and may only manipulate plain framework-owned data—no user code, user-value drops, blocking, panics, or re-entry.

## Adversarial review follow-up

The fresh review replaced the original unsafe-public marker design with actual sealing, extended one-shot completion tracking across the take/consume/erased paths, and strengthened poison recovery coverage to observe captured-operation retirement.

## Verification

- `./tools/dev cargo test -p shelterwood-runtime` (80 unit tests plus sealing doctest)
- `./tools/dev cargo test -p shelterwood` (full façade unit, integration, and doc suites)
- `./tools/dev just ci` (918 nextest tests plus formatting, clippy, docs, examples, reachability, external-consumer, and Nix checks)
- `git diff --check`

Kept as a draft until the tracking workflow marks the reviewed PR ready.

### DIFF
--- a/crates/shelterwood-runtime/src/mailbox.rs
+++ b/crates/shelterwood-runtime/src/mailbox.rs
@@ -133,6 +133,8 @@ mod tests {
 
     use shelterwood_core::{ErasedOneShotClose, ErasedOneShotReceiver, ErasedValue};
 
+    use crate::sync::ONESHOT_REPOLL_PANIC;
+
     use super::{TokioOneShotReceiver, mailbox_runtime};
 
     struct CountWake(Arc<AtomicUsize>);
@@ -222,8 +224,6 @@ mod tests {
 
     #[test]
     fn adapter_oneshot_rejects_poll_after_close_and_take_with_framework_diagnostic() {
-        const REPOLL: &str = "shelterwood one-shot receiver polled after completion";
-
         let runtime = mailbox_runtime();
         let (sender, mut receiver) = runtime.oneshot();
         sender
@@ -245,7 +245,7 @@ mod tests {
         .expect_err("an erased receiver cannot poll after terminal take");
         assert_eq!(
             payload.downcast_ref::<&'static str>().copied(),
-            Some(REPOLL)
+            Some(ONESHOT_REPOLL_PANIC)
         );
     }
 
--- a/crates/shelterwood-runtime/src/mailbox.rs
+++ b/crates/shelterwood-runtime/src/mailbox.rs
@@ -120,6 +120,7 @@ impl MailboxSignalWatcher for TokioSignalWatcher {
 #[cfg(test)]
 mod tests {
     use std::{
+        panic::{AssertUnwindSafe, catch_unwind},
         sync::{
             Arc,
             atomic::{AtomicUsize, Ordering},
@@ -216,6 +217,35 @@ mod tests {
         assert_eq!(
             *value.downcast::<u8>().expect("value type is preserved"),
             11
+        );
+    }
+
+    #[test]
+    fn adapter_oneshot_rejects_poll_after_close_and_take_with_framework_diagnostic() {
+        const REPOLL: &str = "shelterwood one-shot receiver polled after completion";
+
+        let runtime = mailbox_runtime();
+        let (sender, mut receiver) = runtime.oneshot();
+        sender
+            .send(Box::new(12_u8))
+            .unwrap_or_else(|_| panic!("the adapter receiver is live"));
+        let value = receiver
+            .as_mut()
+            .close_and_take()
+            .expect("close-and-take recovers the published erased value");
+        assert_eq!(
+            *value.downcast::<u8>().expect("value type is preserved"),
+            12
+        );
+
+        let mut context = Context::from_waker(Waker::noop());
+        let payload = catch_unwind(AssertUnwindSafe(|| {
+            let _ = receiver.as_mut().poll_receive(&mut context);
+        }))
+        .expect_err("an erased receiver cannot poll after terminal take");
+        assert_eq!(
+            payload.downcast_ref::<&'static str>().copied(),
+            Some(REPOLL)
         );
     }
 
--- a/crates/shelterwood-runtime/src/spawn.rs
+++ b/crates/shelterwood-runtime/src/spawn.rs
@@ -698,14 +698,6 @@ mod tests {
         }
     }
 
-    struct ActorDropNotice(mpsc::Sender<()>);
-
-    impl Drop for ActorDropNotice {
-        fn drop(&mut self) {
-            let _ = self.0.send(());
-        }
-    }
-
     struct PanicWake(Arc<AtomicUsize>);
 
     impl Wake for PanicWake {
@@ -830,7 +822,7 @@ mod tests {
         let (started, started_rx) = tokio::sync::oneshot::channel();
         let (dropped, dropped_rx) = mpsc::channel();
         let work = super::spawn_actor_work(async move {
-            let _notice = ActorDropNotice(dropped);
+            let _notice = RecordingDrop(dropped);
             let _ = started.send(());
             std::future::pending::<()>().await;
         });
@@ -841,7 +833,7 @@ mod tests {
         tokio::time::timeout(Duration::from_secs(1), async {
             loop {
                 match dropped_rx.try_recv() {
-                    Ok(()) => break,
+                    Ok(_) => break,
                     Err(mpsc::TryRecvError::Empty) => tokio::task::yield_now().await,
                     Err(mpsc::TryRecvError::Disconnected) => {
                         panic!("the aborted actor work did not drop its task state")
@@ -884,7 +876,7 @@ mod tests {
     #[test]
     fn blocking_job_recovers_poison_for_pending_checks_and_drop() {
         let (captured_dropped, captured_dropped_rx) = mpsc::channel();
-        let captured = ActorDropNotice(captured_dropped);
+        let captured = RecordingDrop(captured_dropped);
         let (completion, _receiver) = super::oneshot();
         let job = super::BlockingJob::new(move || drop(captured), completion);
         let injected = panic::catch_unwind(panic::AssertUnwindSafe(|| {
--- a/crates/shelterwood-runtime/src/spawn.rs
+++ b/crates/shelterwood-runtime/src/spawn.rs
@@ -3,7 +3,7 @@ use std::{
     future::{Future, poll_fn},
     ops::RangeBounds,
     panic::resume_unwind,
-    sync::{Arc, Mutex},
+    sync::{Arc, Mutex, MutexGuard, PoisonError},
     task::Poll,
 };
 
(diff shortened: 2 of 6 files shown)


ralexstokes:
## Review

Reviewed against #426, CLAUDE.md's lock rule, and the teardown/unwind classes
this round exists for. **No correctness bugs found.** Every behavior the
description claims is pinned was mutation-checked; one terminal edge was not,
and is now.

### Verified by mutation (each mutation caught by exactly the claimed test)

| mutation | caught by |
| --- | --- |
| drop `completed = true` in `poll_receive`'s ready arm | `oneshot_repoll_uses_a_framework_owned_diagnostic` |
| drop it in `close_and_poll_receive`'s non-`Pending` arm | same |
| drop it in `try_receive`'s `Ok` arm | same |
| drop it in `close_and_take` | `adapter_oneshot_rejects_poll_after_close_and_take_...` + the sync twin |
| `lock_pending` back to `.expect(...)` | `blocking_job_recovers_poison_...` — and it fails as **SIGABRT** ("thread caused non-unwinding panic"), i.e. the test reproduces the exact abort class the fix removes |
| `Drop for ActorWork` made a no-op | `dropping_actor_work_aborts_its_task` |
| blanket `impl<X> Sealed for X` | the `compile_fail` seal doctest (which does run — `#[doc(hidden)]` does not suppress doctest collection) |

Also checked directly, in pinned tokio 1.53.1 source: `Receiver::try_recv`
sets `self.inner = None` on `Ok` **and** on `Err(Closed)`, returning early only
for `Empty`, and `close_and_take` always reaches one of those two after
`close()`. So marking `close_and_take`/`try_receive` terminal is exactly right,
and marking a bare `close()` non-terminal is right too (tokio keeps that
receiver pollable — the test's last block is the passing twin).

Re-poll reachability re-walked for every caller of the four receive edges
(`spawn.rs:178`, `task.rs:319`, `tree/admission.rs`, `reply.rs::poll_reply`,
façade `mailbox/capability.rs`, driver tests): all are single-shot `poll_fn`
or consuming futures, and `poll_reply` only reaches `close_and_poll_receive`
from the `Pending` arm, so the new `assert!` cannot fire on a working program.
`ProxiedPoll::poll`'s probe/re-poll pair is also safe — the probe only re-polls
after a `Pending`, which never sets `completed`.

### Fixed on the branch — 2154d3b

1. **`try_receive`'s `Err(Closed)` terminal edge was mutation-invisible.**
   Removing `self.completed = true` from that arm left all 80 tests green, so
   the one edge where an exhausted-but-empty take would fall back to tokio's
   opaque `called after complete` was unpinned. Added it to
   `oneshot_repoll_uses_a_framework_owned_diagnostic`, plus the bare-close twin
   that must stay pollable. Mutation-checked.
2. **Completion contract documented where it lives.** A doc comment on the
   `completed` field (which tokio edges consume the receiver, and that `close`
   is deliberately not one of them) and a sentence on `close_and_poll_receive`
   saying every non-`Pending` outcome is terminal — the wrapper is stricter
   than tokio there (the `Ok(_)` CAS arm closes without a `try_recv`), which
   was previously only implicit.
3. **Single-sourced the re-poll diagnostic.** `mailbox.rs`'s test had a third
   copy of the message literal; it now uses `sync`'s `ONESHOT_REPOLL_PANIC`.
4. **Reused the existing drop probe.** The new `ActorDropNotice` restated
   `test_support::RecordingDrop`, which is already imported in `spawn`'s tests
   (and is what the neighbouring `failed_native_fallback_isolates_capture_...`
   uses). Dropped the new struct.

### Filed — #457

`FrameworkPlain` / `DisposingReceiver::new_framework` now has **zero
production callers**. The seal is the right fix for #429, but since the only
two markers lived in the façade, removing them leaves the marker trait, its
`()`/`Result` impls, the `Sealed` module, the seal doctest and the inline-drop
fast path alive only for one test — and nothing warns (`pub` items under
`#![allow(unreachable_pub)]`, and the reachability probe only walks the
supported façade API). #457 asks the round to choose deliberately: delete it,
or give it a runtime-owned user.

### Refuted / non-issues

- **Venue change is safe.** `tree/admission.rs` moving admission and removal
  responses to the detached lane costs one disposal submission per abandoned
  future. Both `Removal` and `PendingAdmission` are only constructed by the
  public `Tree` API and drop on user tasks, never under the observation gate,
  so this adds no submission to a locked path. Outside a runtime the fallback
  is one shared queued thread, not one per value.
- **`receive(self)`'s assert fires at first poll, not at call.** That is the
  usual async-fn body deferral, but here it is correct: an unpolled `receive()`
  future should not panic, and `self` is consumed so no re-poll is reachable.
- **Hard `assert!` rather than `debug_assert!`** is right, against the repo's
  usual diagnostic-only convention: tokio panics in release too, so a
  `debug_assert` would leave release builds with the opaque message — which is
  the entire point of #426 item 1. No mutex is held at either call site.
- **`WaiterRegistry` inline retirement claim checked, not just read.**
  `LatchWait::drop`/`WatchWait::drop` route through
  `drop_registered` → `PanicAccumulator`, so a hostile caller-waker destructor
  is contained during an unwind. I extended the new sentence to name the
  accepted cost (a slow destructor stalls the abandoning waiter alone), so it
  reads as the same per-seam adjudication as the reply receiver's.
- **`jitter_rng_respects_requested_bounds`** pins `fastrand` more than
  shelterwood, but it would still catch an inclusive/exclusive slip in the
  one-line delegation. Kept.
- **Item 6 (`spawn_blocking_fallback_with`'s two-arm match) left explicit** —
  agreed; the arms carry different ownership evidence.
- `test_support`'s 1.53.1 degradation note and the `BlockingJob` poison
  recovery both land exactly where #426 asked.

CI: green on `2154d3b`.

PR #451, opened 2026-08-22
Harden mailbox transitions and disposal coverage

### LINKED ISSUE #421: Mailbox: accept_locked mismatch fallback is not total for Latest; make the effects emptiness gate structural; untested bind guards
From the 2026-08-22 post-fold full review. No reachable bugs in the mailbox module; items below, most severe first.

## 1. Hardening: `accept_locked`'s incarnation-mismatch fallback is not total for Latest mailboxes

`crates/shelterwood/src/mailbox/cell.rs:1590-1596` + :328-350. The mismatch arm is written to be total: it returns `AcceptTransition::Full(message)` and defers the invariant panic to the post-unlock effects flush, precisely so the diagnostic never poisons the mailbox mutex. But `submit`'s `Full` handler parks, and `park`'s `Bound(Available)` arm asserts `let Some(MailboxKind::Queue(capacity)) = self.kind else { unreachable!(...) }` — for a **Latest** mailbox the fallback therefore lands on an always-on `unreachable!` raised *inside* the transaction, which poisons `MailboxCell::state` for every later caller and displaces the deferred first-wins diagnostic. Not reachable today (the incarnation is read from the same binding in the same critical section), so this is a defense-in-depth inconsistency: the fallback's own comment ("the by-value user message returns through the caller's post-unlock path") is false for one of the two mailbox kinds. Cheapest fix: a dedicated transition (like `Exhausted`) that `submit`/`try_send` complete without parking.

## 2. Maintainability: make `MailboxEffects::drop`'s emptiness gate structural

`cell.rs:612-644`. The early-return checks seven fields by hand, and `MailboxEffectBatch` restates the same seven plus two capability handles. A new effect field added to `MailboxEffects` but forgotten in the gate is *silently discarded* whenever it is the only effect set — exactly the class the transaction shapes exist to prevent, and nothing fails to compile. Extract the effect payload into one shared struct (owned by `MailboxEffects`, moved wholesale into the batch, single `is_empty()`) so both track new fields by construction. Same spirit as #333.

## 3. Simplification: `MailboxKind` is a verbatim duplicate of `ResolvedMailbox`

`cell.rs:80-84` vs `crates/shelterwood-core/src/policy.rs:841-845`. Same two variants, same payloads, both `Copy + Eq + Debug`; `configure` (cell.rs:1408-1411) exists largely to map one onto the other, and the mismatch panic Debug-prints the internal twin. Storing `ResolvedMailbox` in `MailboxState.kind` deletes the enum and the mapping; the façade already names `ResolvedMailbox` in the `MailboxControl::configure` signature.

## 4. Simplification: `WithdrawalDisposition → WakerAction` mapping triplicated verbatim

`cell.rs:1268-1276, 1290-1298, 1314-1322`. All three arms of `withdraw`'s outcome match contain the identical five-line disposition→action block feeding the identical `state.waker.take(...)`. A method `WithdrawalDisposition::action(&self, runtime) -> WakerAction` collapses ~24 lines and removes drift risk (the Accepted/Terminated arms exist only as keep-a-future-regression-total hardening, so drift there would be invisible).

## 5. Coverage: `MailboxControl::bind`'s four guard branches have zero tests

`cell.rs:1424-1445`. No test exercises: (1) bind-after-terminal silent ignore ("terminality wins that race permanently" — the driver relies on it), (2) "mailbox must be configured before its first bind", (3) "must close the prior incarnation before rebinding", (4) "bind token is foreign or was already consumed". Each is one mutation from vanishing invisibly (deleting the `token.claim` check passes the suite). Related untested corner: `configure` called twice mints a second token sharing the already-consumable permit (mailbox.rs:54-60, cell.rs:1412-1414); after the first bind consumes it the second token can never bind — neither documented nor tested.

## 6. Coverage: exhaustion paths pin "after unlock" but not the detached venue

`cell/tests.rs:1543,1579` — the waiter-identity and accepted-sequence exhaustion tests pin only destroyed-after-unlock (`LockCheckingMessage`), not the detached-disposal *venue* CLAUDE.md claims for mailbox payloads; a regression dropping the exhausted message inline on the caller passes. The close-flush test at :1680 already shows the `ThreadRecordingMessage` pattern to reuse. Also `cell/tests.rs:1781`: the final "parked message destroyed outside the mailbox mutex" assertion observes a drop the test's own `withdraw` performs — trivially true; trim or re-aim.

## 7. Quality: receive-path exhaustion is asymmetric with the bind path, undocumented

`cell.rs:1687-1689` vs :1467-1483. `bind` freezes on promotion exhaustion (with a comment); `receive` → `promote_waiters` leaves `Bound(Full)`, so each subsequent receive re-runs promotion and re-raises the post-unlock exhaustion panic while parked waiters are re-observed. Both are process-fatal diagnostics, so not a bug — but a sentence at `promote_waiters` saying why receive does not freeze (or making it freeze) prevents a future "fix" to either side.

## 8. Small scars

- `cell.rs:1487-1493`: `let MailboxKind::Queue(capacity) = kind else { unreachable!() };` followed by `let _ = capacity;` — bind `Queue(_)` instead.
- `mailbox.rs:28-29` + `timer.rs:8-9`: the `timer` module is `#[cfg(test)]`-gated at declaration *and* wraps its body in `#[cfg(test)] mod tests` — one gate is redundant.
- `cell/tests.rs:528-531`: assertion message names `IncludeFrozen`/`LiveOnly`, pre-fold names for `ReceiveMode::Drain`/`LiveThrough` — misleading if it fires.
- Test-only: five near-identical ~40-line unsafe `RawWakerVTable` Arc-probe scaffolds (`futures.rs:1013-1058, 1225-1266`, `deadline.rs:330-363`, `reply.rs:226-266`, `timer.rs:23-62`) differ only in the drop-side observation; one parameterized raw scaffold in `test_support.rs` deletes ~150 lines of repeated `unsafe` and concentrates the safety comments.

ralexstokes:
Closes #421.

## Summary

- make the accepting-incarnation mismatch a dedicated non-parking transition, with rejected user payload disposal completed before the post-unlock diagnostic
- move all mailbox effect state through one `MailboxEffectPayload`, store `ResolvedMailbox` directly, and centralize withdrawal-disposition-to-waker-action mapping
- cover every `MailboxControl::bind` guard and document/test repeated compatible configuration: returned tokens alias one one-shot permit, so only the first can bind
- pin waiter-id and accepted-sequence exhaustion to the detached disposal venue; remove the assertion that only observed the test's own inline withdrawal
- document why receive-time promotion exhaustion intentionally leaves the live mailbox full instead of freezing it
- remove the remaining small scars: redundant queue binding, duplicate timer test gate, stale receive-mode assertion names, and five duplicated raw-waker scaffolds
- add direct `Mailbox::Latest` coverage proving a panicking displaced-payload destructor runs after unlock, remains contained, and does not lose the replacement

## Validation

- `./tools/dev cargo test -p shelterwood mailbox::cell::tests`
- `./tools/dev cargo test -p shelterwood mailbox::`
- `./tools/dev just ci`

### DIFF
--- a/crates/shelterwood/src/mailbox.rs
+++ b/crates/shelterwood/src/mailbox.rs
@@ -167,7 +167,9 @@ pub(crate) trait MailboxTermination: Send {
 pub(crate) trait MailboxControl: fmt::Debug + Send + Sync {
     /// Installs the declaration-time mailbox policy before the first bind.
     /// Reconfiguration may only repeat the same resolved policy; a mismatch
-    /// panics after the mailbox lock has been released.
+    /// panics after the mailbox lock has been released. Tokens returned by
+    /// repeated compatible configuration share one one-shot permit, so only
+    /// the first token presented can bind; later aliases are rejected.
     fn configure(
         &self,
         mailbox: ResolvedMailbox,
--- a/crates/shelterwood/src/mailbox/cell.rs
+++ b/crates/shelterwood/src/mailbox/cell.rs
@@ -319,16 +319,23 @@ impl<M> MailboxState<M> {
                 return waiters.park(operation);
             }
             MailboxBinding::Bound(BoundState::Available(incarnation)) => {
-                let Some(ResolvedMailbox::Queue(capacity)) = self.kind else {
-                    unreachable!("only a capacity-bound queue can park while bound")
-                };
                 // Diagnostic-only under the mailbox mutex: `Available` parks
                 // only after the accepting transition filled this configured
-                // queue, or after that transition's own incarnation-mismatch
-                // fallback, which has already recorded its invariant panic on
-                // the effects sink. The Full transition remains total without
-                // this check, and no test expects the diagnostic panic.
-                debug_assert_eq!(self.queue.len(), capacity.get());
+                // queue, so a latest mailbox never arrives here -- its
+                // incarnation mismatch now leaves through its own non-parking
+                // transition instead. Releasing the guard to raise the verdict
+                // is impossible mid-transition and an always-on panic would
+                // poison the mailbox for every later caller, so record it and
+                // keep the transition total: parking preserves the operation's
+                // user message under either mailbox kind. No test expects this
+                // diagnostic.
+                debug_assert!(
+                    matches!(
+                        self.kind,
+                        Some(ResolvedMailbox::Queue(capacity)) if self.queue.len() == capacity.get()
+                    ),
+                    "only a filled capacity-bound queue can park while bound"
+                );
                 let incarnation = *incarnation;
                 let mut waiters = WaiterQueue::default();
                 if !waiters.park(operation) {
@@ -579,15 +586,28 @@ impl<M> Default for MailboxEffectPayload<M> {
 
 impl<M> MailboxEffectPayload<M> {
     fn is_empty(&self) -> bool {
-        // `isolate_displaced` only selects how a nonempty displaced set is
-        // flushed; by itself it represents no work.
-        !self.pulse
-            && self.displaced.is_empty()
-            && self.wakers.is_empty()
-            && self.returned.is_none()
-            && !self.accepted_sequence_exhausted
-            && self.rejected_bindings.is_empty()
-            && self.invariant_panic.is_none()
+        // Destructured rather than field-accessed: a new effect field is then
+        // a compile error here instead of an effect silently discarded
+        // whenever it is the only one this transition set.
+        let Self {
+            pulse,
+            displaced,
+            // `isolate_displaced` only selects how a nonempty displaced set is
+            // flushed; by itself it represents no work.
+            isolate_displaced: _,
+            wakers,
+            returned,
+            accepted_sequence_exhausted,
+            rejected_bindings,
+            invariant_panic,
+        } = self;
+        !pulse
+            && displaced.is_empty()
+            && wakers.is_empty()
+            && returned.is_none()
+            && !accepted_sequence_exhausted
+            && rejected_bindings.is_empty()
+            && invariant_panic.is_none()
     }
 }
 
@@ -736,6 +756,11 @@ impl<M: Send + 'static> MailboxEffectBatch<M> {
         }
         payload.wakers.flush(&mut panics);
         if !payload.isolate_displaced {
+            // A live latest transition displaces at most the one value it
+            // replaced, so per-element containment here is totality for a
+            // future multi-displacement caller rather than a tested behavior;
+            // `bind`, the only caller that can accumulate several, isolates
+            // them into a `MailboxPayload` above instead.
             for envelope in payload.displaced.drain(..) {
                 panics.run(|| drop(envelope));
             }
@@ -1748,7 +1773,7 @@ pub(super) enum WithdrawalDisposition {
 }
 
 impl WithdrawalDisposition {
-    fn action(&self, runtime: &Arc<dyn MailboxRuntime>) -> WakerAction {
+    fn action(self, runtime: &Arc<dyn MailboxRuntime>) -> WakerAction {
         match self {
             Self::Inline => WakerAction::DropInline,
             Self::Isolated => WakerAction::Dispose(Arc::clone(runtime)),
--- a/crates/shelterwood/src/mailbox/cell.rs
+++ b/crates/shelterwood/src/mailbox/cell.rs
@@ -1,8 +1,7 @@
 use std::{
     collections::{BTreeMap, VecDeque},
     fmt,
-    num::NonZeroUsize,
-    ops::Deref,
+    ops::{Deref, DerefMut},
     sync::{
         Arc, Mutex, MutexGuard,
         atomic::{AtomicBool, Ordering},
@@ -75,12 +74,6 @@ impl ReceiveMode {
             Self::LiveThrough(limit) => sequence <= limit,
         }
     }
-}
-
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-enum MailboxKind {
-    Queue(NonZeroUsize),
-    Latest,
 }
 
 pub(super) struct Envelope<M> {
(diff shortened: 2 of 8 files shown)


ralexstokes:
## Review

Reviewed the full diff against #421, the lock rule, and the documented lock
order. **No reachable bugs.** The headline fix is correct and the coverage
additions are real pins, not fail-open probes — I mutation-checked six of them
(below). Four findings, all fixed on-branch in 8abdeab; one follow-up filed.

### Verified by mutation (committed tree, mutation injected, reverted after)

| Mutation | Result |
| --- | --- |
| `reject_incarnation_mismatch` disposes inline instead of via the lane | `latest_incarnation_mismatch_uses_the_nonparking_post_unlock_path` FAILS |
| `accept_locked` mismatch returns `Full(message)` again (the pre-PR shape) | **does not compile** — `AcceptTransition::IncarnationMismatch` becomes dead code under `-D dead-code` |
| `bind`'s `token.claim` guard deleted | `bind_rejects_a_foreign_token_without_consuming_the_valid_one` + `repeated_configuration_tokens_share_one_consumable_permit` FAIL |
| `bind`'s terminal early return deleted | `bind_after_terminal_is_silently_ignored` FAILS |
| `accept_locked` drops the displaced envelope inline under the mutex | `latest_displacement_contains_a_panicking_payload_destructor` + `live_latest_displacement_drops_after_unlock_and_replacement_visibility` FAIL |
| `submit`'s two exhaustion arms drop instead of dispose | both `*_exhaustion_disposes_the_message_on_the_detached_lane` FAIL |

The `IncarnationMismatch` transition being *structurally* required (dead-code
denial, not a test) is the strongest of these — the fallback cannot silently
regress to the parking shape.

### Findings fixed (8abdeab)

1. **Stale comment left the removed hazard documented as live.** `park`'s
   `Bound(Available)` arm still said it is entered "after that transition's own
   incarnation-mismatch fallback, which has already recorded its invariant panic
   on the effects sink." Both halves are now false: the mismatch leaves through
   its own non-parking transition and records nothing on the sink. Rewritten to
   state the one remaining route (the accepting transition filled the configured
   queue).

2. **The `unreachable!` #421 item 1 named is still there, just unreachable.**
   Item 1's objection was not only that the fallback routed into `park`, but that
   `park`'s `let Some(ResolvedMailbox::Queue(capacity)) = self.kind else {
   unreachable!(..) }` is an always-on panic *raised inside the transaction*,
   poisoning `MailboxCell::state` for every later caller. Removing the route
   leaves the construct. That is exactly the case CLAUDE.md's convention covers —
   "where releasing is impossible, `debug_assert!` instead, as
   `MailboxState::take_waiters` does". Folded the two diagnostics (`unreachable!`
   + `debug_assert_eq!` on the queue length) into one `debug_assert!` with a total
   fallback: parking preserves the operation's user message under either mailbox
   kind, so the transition stays total without poisoning. No test expects the
   diagnostic, same as `take_waiters`.

3. **The effects-emptiness gate was still not structural.** `MailboxEffectPayload`
   correctly removes the *batch* restatement (item 2's main complaint), but
   `is_empty()` remained a hand-written seven-field conjunction inside the same
   struct — adding an effect field still compiled fine and was still silently
   discarded whenever it was the only effect set. (The manual `Default` impl
   catches a new field; `is_empty` did not.) `is_empty` now destructures `Self`
   with no `..`, so a new field is `E0027: pattern does not mention field`.
   Confirmed by adding a probe field: pre-fix it compiled, post-fix it does not.

4. **Nits.** `WithdrawalDisposition::action` takes `self` by value like
   `ReceiveMode::accepts` (both are `Copy`). Added a note at the non-isolated
   displaced loop recording why its per-element containment is untestable — see
   below.

### Recorded, not fixed

- **"remains contained" is over-claimed in the PR description.** Removing
  `panics.run(..)` from the non-isolated displaced loop passes the whole mailbox
  suite. That is not a gap in the test: on the non-isolated path `displaced`
  holds at most one envelope by construction — `accept_locked` pushes one per
  transition, and the only caller that can accumulate several
  (`promote_waiter_queue`'s Latest arm) is reachable only from `bind`, which sets
  `isolate_displaced()` and routes them through `MailboxPayload`, whose own `Drop`
  already contains per element. So the accumulator there is totality for a future
  multi-displacement caller, and I documented it as such rather than adding a
  test for an unreachable shape. "runs after unlock" and "does not lose the
  replacement" *are* pinned (mutation above).
- **`submit`/`try_send`'s `IncarnationMismatch` arms have no test.** Unavoidable:
  `transaction.status()` and `accept_locked` read the same `binding` under the
  same guard, so the mismatch is unreachable through the public path. The new
  test is correctly white-box; the arms themselves stay uncovered.
- **`MailboxEffects: Deref/DerefMut → MailboxEffectPayload`** is not a lock-rule
  regression — the fields were directly reachable on the sink before, so nothing
  newly takeable is exposed, and it matches `MailboxTxn: Deref → MailboxState`.

### Filed

- #456 — the `probe_waker` consolidation only covered the mailbox module's five
  scaffolds; ~12 more hand-rolled `RawWakerVTable` Arc probes remain in
  `task.rs`, `driver/tests/system_join.rs`, the integration tests, and
  core/runtime. Ranked by cost there; out of scope for this PR.

### Lock rule / lock order

Re-derived every touched critical section. `reject_incarnation_mismatch` runs
strictly after `transaction.finish` releases the guard and flushes, and disposes
before it panics — correct order per CLAUDE.md's "compute the verdict, release,
then panic" plus the detached-disposal convention for user payloads.
`disposition.action(&self.runtime)` clones an `Arc` under the doubled
mailbox→operation section, which is refcount traffic with a provable other owner.
`MailboxEffectPayload` moves wholesale out of the sink in `Drop` after the guard
is emptied — unchanged. Lock order member mailbox → `MailboxCell::state` →
`SendOperation::state` is untouched.

CI green at 8abdeab (run 32602277761). Not merging.

PR #452, opened 2026-08-22
Clean up static builder API and test fixtures

### LINKED ISSUE #428: Public API and test-harness cleanups: static ReserveError split, lifecycle-scan helper, waker fixture consolidation
From the 2026-08-22 post-fold full review. The public surface itself passed cleanly (export list exactly curated, every exported item exercised, must_use audit green, api-reachability/external-consumer probes match the post-fold shape). Remaining cleanups, most impactful first.

## 1. Static builders share the dynamic `ReserveError`

`crates/shelterwood/src/tree/builders.rs:84-184` — `Tree`/`DynamicTree::add_*`/`reserve_*` return `Result<_, ReserveError>` (`cells/admission.rs:9`), whose variants include `NoRuntime`, `RemovalInProgress`, and `NotAdmitting` with five causes — none producible by a pre-spawn static builder (only `EmptyId`/`DuplicateId`/`IdentityExhausted` are reachable there; `spawn_builder` checks the runtime separately via `BuildError::NoRuntime`). Under the invalid-states-unconstructible philosophy, static callers are forced to match arms that cannot occur. A 3-variant static error (~25 lines, honesty gained; the break is free) or an explicit doc note.

## 2. Integration-harness dedup: "scan lifecycle for `Exited(id)`" hand-rolled at ~13 sites

Across 7 files: `tests/drain.rs:416-424` (`exited_actor`), `reply_waker_containment.rs:86-104` (which also re-implements the lag-marker panic `common/lifecycle.rs::next_event` already owns), plus disposal.rs, lifecycle.rs (×4), observation.rs, offloads.rs, subtrees.rs. A `next_exit_of(events, id)` in `common/lifecycle.rs` (beside `last_panic_message`, the same shape) deletes all of them.

## 3. Waker-fixture consolidation and a name collision

- `tests/timer_waker_proxy.rs` (BLOCKING_DROP/COUNTING/ORDINAL vtables, lines 77/126/197) and `tests/mailbox_future_drop_containment.rs` (PANICKING_DROP/ORDINAL, lines 86/191) each build their own raw-waker families while `common/waker.rs` holds only `hostile_waker`; the two `ORDINAL_WAKER_VTABLE`s differ in mechanics but overlap in purpose. (Adjacent: the five ~40-line unsafe Arc-probe scaffolds in the mailbox white-box tests, tracked in the mailbox issue.) Related prior art: #413.
- Two unrelated fixtures are both named `DropProbe` (`tests/disposal.rs:29` thread-reporting, `tests/dynamic.rs:50` bool flag) — rename even if not unified.
- Unbounded wait outlier: `driver/tests/shutdown.rs:491` is the only `gate.wait_entered().await` not wrapped in `timeout(DRIVER_PROGRESS_WAIT, …)` like its siblings; a regression there hangs to the harness timeout instead of the promised diagnostic.

## 4. Façade micro-cleanups

- `src/actor.rs:151`: `Context::new(raw, stage)` accepts `DeliveryStage::Initializing` but no caller ever passes it (`new_initializing` owns that stage); the invalid combination is representable, blast radius one private constructor.
- `src/actor.rs:264`: `clear_timer` returns `Result<bool, Rejected<()>>`; `Rejected<()>` carries nothing — asymmetric with `RawContext::clear_timer -> bool` (`raw/context.rs:729`). Defensible (drain-stage rejection stays diagnosable like its siblings); leave unless someone trips on it.
- Formatting outliers rustfmt tolerates: attributes placed above doc comments (`tree/system.rs:164`, :120) and missing blank lines between items (`system.rs:23-24`, `builders.rs:31-32`, `slots.rs:337-338`, :417-418, :516-517).

## Recorded non-findings (so they aren't re-litigated)

`plan.rs` is the right size (43% tests; the four `SlotCell` terminalize variants map 1:1 onto lock-rule venues with distinct callers); the one-line re-export shims pull their weight (they mint the `crate::exit::`-style internal namespaces); `SnapshotReceiver::borrow_latest_and_closed` is the *only* doc-hidden façade-reachable item and carries its ruling in-code (note: method-level, hence invisible to the E0432 root-path probe); `Reply<T>` deliberately lacks `#[must_use]` (dropping is a protocol action); the definition families' setter boilerplate is already collapsed by `common_options_setters!` and encodes genuinely different option subsets.

ralexstokes:
Closes #428.

## What changed

- Split pre-spawn declaration failures into the exhaustive three-variant `StaticReserveError` (`EmptyId`, `DuplicateId`, `IdentityExhausted`). `Tree` and `DynamicTree` builders now expose only those reachable outcomes; live `DynamicScopeRef` reservation and admission keep the broader `ReserveError`. Updated API conformance coverage, error docs, crate docs, and SPEC §9/§12 accordingly.
- Added `common::next_exit_of` and migrated every current simple “scan until this child id exits” fixture (seven sites across drain, disposal, offloads, and reply-waker containment). The remaining `Exited` matches intentionally stay local because they judge ordering with `Started`/`RestartScheduled`/`Removed`, match membership or incarnation rather than id, count multiple children, or drain a complete stream; routing those through an exit-only helper would discard evidence the tests assert.
- Consolidated the counting and ordinal hostile raw-waker families in `tests/common/waker.rs`. Timer cancellation supplies its blocking action to shared ordinal fixtures, while mailbox completion supplies its panic action; explicit wake-or-drop and drop-only modes preserve each original vtable's exact semantics while removing both duplicated ordinal vtables and the timer-local counting/blocking families.
- Renamed the unrelated dynamic fixture to `FlagDropProbe`, and bounded the lone shutdown `wait_entered` outlier with `DRIVER_PROGRESS_WAIT`.
- Replaced `Context::new(raw, stage)` with dedicated live and frozen-prefix constructors, leaving initialization constructible only through `new_initializing` and its owned initialization boundary.
- Kept `Context::clear_timer -> Result<bool, Rejected<()>>`: the unit payload is intentionally empty, but the rejected result preserves the same diagnosable frozen-prefix refusal as the other context operations.
- Normalized the item spacing and doc/attribute placement identified in the issue.
- Extended the external-consumer installable-seam probe to cover the private `MailboxEffectQueue` and `MailboxEffectSink` names called out in the tracking follow-up.

## Adversarial review

A fresh reviewer found and fixed three issues before readiness:

- The first shared ordinal fixture made the mailbox test's drop-only panic run from consuming `wake` too. The fixture now has explicit modes: timer ordinal actions retain wake-or-drop behavior, while the blocking-drop and mailbox-destructor probes remain drop-only.
- The SPEC's summary paragraph still listed only two static reserve failures. It now names all three `StaticReserveError` variants.
- Replacing static builder signatures had removed `ReserveError` from the API conformance test. The test now pins both error types and explicitly checks the live `DynamicScopeRef` reservation signatures remain `ReserveError`.

The reviewer also mutation-tested the mailbox fixture: temporarily resuming the captured ready-edge waker panic produced the intended SIGABRT double-panic failure, while the restored production boundary passes.

## Verification

- `./tools/dev cargo test -p shelterwood --test timer_waker_proxy` (5 passed)
- `./tools/dev cargo test -p shelterwood --test mailbox_future_drop_containment` (5 passed)
- `./tools/dev cargo test -p shelterwood --test api_trait_conformance` (7 passed)
- affected lifecycle suites: reply-waker containment (1), drain (8), disposal (41), offloads (32), all passed
- focused static-builder and shutdown regressions passed
- initial implementation and post-review `./tools/dev just ci` runs passed: nightly fmt/clippy, 911 nextest tests, doctests, every example, rustdoc/mdBook, API reachability, core manifest, external consumer (including both new seam probes), and Nix formatting

### DIFF
--- a/crates/shelterwood/docs/errors.md
+++ b/crates/shelterwood/docs/errors.md
@@ -58,10 +58,12 @@ response wait.
 
 ## The rest of the error surface, briefly
 
-- **Declaration and admission** — `ReserveError` (duplicate or in-removal
-  id, scope not admitting: `NotAdmittingCause`), `BuildError` (root
-  lowering), `PolicyError` (invalid policy data, rejected at
-  construction).
+- **Declaration and admission** — `StaticReserveError` (empty or duplicate
+  id, identity exhaustion) for pre-spawn tree declarations; `ReserveError`
+  (those common id failures plus in-removal ids, unavailable runtime, and a
+  scope not admitting: `NotAdmittingCause`) for live dynamic admission;
+  `BuildError` (root lowering); `PolicyError` (invalid policy data, rejected
+  at construction).
 - **Startup** — `StartupError` from `wait_started`; `StartOrShutdownError`
   pairs the startup failure with any rollback timeout report;
   `StartupFailure`/`StartupFailureCause` describe a nested scope's failed
--- a/crates/shelterwood/src/actor.rs
+++ b/crates/shelterwood/src/actor.rs
@@ -280,10 +280,19 @@ impl<A: Actor> fmt::Debug for Context<'_, A> {
 }
 
 impl<'a, A: Actor> Context<'a, A> {
-    fn new(raw: &'a mut RawContext<A::Msg>, stage: DeliveryStage) -> Self {
+    fn new_live(raw: &'a mut RawContext<A::Msg>) -> Self {
         Self {
             raw,
-            stage,
+            stage: DeliveryStage::Live,
+            owns_init_boundary: false,
+            actor: PhantomData,
+        }
+    }
+
+    fn new_draining(raw: &'a mut RawContext<A::Msg>) -> Self {
+        Self {
+            raw,
+            stage: DeliveryStage::DrainingFrozenPrefix,
             owns_init_boundary: false,
             actor: PhantomData,
         }
@@ -697,7 +706,7 @@ impl<A: Actor> RawActor for Handler<A> {
 
         while let Some(message) = raw.recv().await {
             let handled = {
-                let mut context = Context::<A>::new(raw, DeliveryStage::Live);
+                let mut context = Context::<A>::new_live(raw);
                 actor.handle(message, &mut context).await
             };
             if let Err(error) = handled {
@@ -709,8 +718,7 @@ impl<A: Actor> RawActor for Handler<A> {
             MailboxShutdown::Drain => {
                 while let Some(message) = raw.try_recv() {
                     let handled = {
-                        let mut context =
-                            Context::<A>::new(raw, DeliveryStage::DrainingFrozenPrefix);
+                        let mut context = Context::<A>::new_draining(raw);
                         actor.handle(message, &mut context).await
                     };
                     if let Err(error) = handled {
--- a/crates/shelterwood/src/cells/admission.rs
+++ b/crates/shelterwood/src/cells/admission.rs
@@ -1,8 +1,25 @@
-//! Dynamic membership admission errors and removal outcomes.
+//! Static declaration and dynamic admission errors, plus removal outcomes.
 
 use std::fmt;
 
 use shelterwood_core::ChildId;
+
+/// A pre-spawn child declaration or reservation error.
+///
+/// Static builders cannot observe runtime admission state: only id validation,
+/// duplicate detection, and membership identity exhaustion are reachable.
+#[derive(Clone, Debug, Eq, PartialEq, thiserror::Error)]
+pub enum StaticReserveError {
+    /// The child id was empty.
+    #[error("child id must not be empty")]
+    EmptyId,
+    /// A declared membership already occupies the id.
+    #[error("child id `{0}` is already declared")]
+    DuplicateId(ChildId),
+    /// The scope can mint no further membership identities.
+    #[error("membership identity space is exhausted")]
+    IdentityExhausted,
+}
 
 /// A child reservation or dynamic admission error.
 #[derive(Clone, Debug, Eq, PartialEq, thiserror::Error)]
--- a/crates/shelterwood/src/driver/tests/shutdown.rs
+++ b/crates/shelterwood/src/driver/tests/shutdown.rs
@@ -488,7 +488,10 @@ async fn hard_force_preserves_the_first_terminal_observer_panic_across_its_fallb
         .await
         .dispatch(&mut scope);
     }
-    gate.wait_entered().await;
+    assert!(matches!(
+        crate::runtime::timeout(DRIVER_PROGRESS_WAIT, gate.wait_entered()).await,
+        crate::runtime::Timeout::Completed(())
+    ));
     let arrived_completion = crate::runtime::timeout(DRIVER_PROGRESS_WAIT, async {
         while scope.disposal_event_receiver.is_empty() {
             crate::runtime::yield_now().await;
--- a/crates/shelterwood/src/lib.rs
+++ b/crates/shelterwood/src/lib.rs
@@ -123,7 +123,8 @@
 //! [`ScopeRef`] addresses an ordered scope; [`DynamicScopeRef`] adds
 //! runtime admission and removal, whose outcomes are [`Admission`],
 //! [`Removal`], [`RemoveOutcome`], [`ReserveError`], and
-//! [`NotAdmittingCause`]. Scope-level state is [`ScopeState`] and
+//! [`NotAdmittingCause`]. Pre-spawn declarations return
+//! [`StaticReserveError`]. Scope-level state is [`ScopeState`] and
 //! [`MembershipStatus`].
 //!
 //! ## Tasks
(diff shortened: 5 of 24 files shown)


ralexstokes:
## Review — PR #452

Reviewed the full diff (24 files) against #428's intent, ran targeted suites, and mutation-checked the two claims that matter (the shared mailbox waker fixture, and the new dynamic-signature pin). One real finding, fixed on-branch in `2606dd5`; the rest is refuted or recorded.

### Finding 1 (fixed) — the consolidated blocking fixture aimed at the wrong clone

`blocking_drop_waker` in `tests/timer_waker_proxy.rs` used to fire on the **first drop of any** clone; the port replaced that with `action_ordinal_drop_waker(1, …)`. Ordinal 1 is not the timer's waker. `Deadlined::poll` (`mailbox/deadline.rs:143-220`) polls the operation first — registering its clone — and `poll_timer` second, so ordinal 1 is the *send operation's* caller clone and ordinal 2 is the timer proxy's stored caller, exactly as `TIMER_CALLER_WAKER_ORDINAL` already names it in the mailbox test.

Instrumenting `run_ordinal_action` to report every retirement confirms the drift is real:

```
PROBE retire ordinal=2 thread=ThreadId(6) name=Some("tokio-rt-worker")
PROBE retire ordinal=1 thread=ThreadId(8) name=Some("tokio-rt-worker")
```

The timer clone retires *first*, so the pre-PR "first drop wins" fixture blocked ordinal 2. After the port the blocker sat on the send operation's retirement instead — a different (also lane-routed, hence still passing) seam than the one the test's name and its `#398 ruling 3` premise paragraph describe.

Not fail-open: targeting a nonexistent ordinal makes the test fail loudly at `entered_rx.recv_timeout` rather than pass vacuously (verified with target 9 → `a framework-owned caller-waker clone reaches retirement: Timeout`). But the test was no longer pinning the property it claims. Fixed by naming `TIMER_CALLER_WAKER_ORDINAL = 2` locally, with the registration-order reasoning the deleted vtable comment used to carry.

### Finding 2 (fixed) — the id-emptiness rule was restated

`BuilderCore::reserve` inlined `id.as_str().is_empty()` rather than going through `checked_id`, leaving two copies of the rule that can drift. Folded onto one `non_empty_id` predicate, mirroring the `mint_reserved_slot`/`mint_reserved_slot_inner` split this PR already introduced; `checked_id` keeps its `ReserveError` face for the dynamic path.

### Finding 3 (fixed, doc) — `next_exit_of` first-vs-last

`last_panic_message` documents *why* it drains to closure (a restarting child cannot pass by matching an earlier incarnation). `next_exit_of` stops at the first match, which is right for all seven migrated sites but is the opposite discipline, so it now says so and points at the other helper.

### Verified, not changed

- **Mailbox fixture is not vacuous after the swap.** Replacing the completion-path `discard_panic(panics.take())` in `Deadlined::poll` with `resume_unwind` makes `recv_ready_after_parking_contains_the_retired_timer_waker` abort (`panic in a destructor during cleanup` → SIGABRT); restored, it passes.
- **The new dynamic-reservation pin is real.** Retyping one annotation in `_assert_live_dynamic_reservation_errors` to `StaticReserveError` produces E0308, so the uncalled closure genuinely type-checks the live `DynamicScopeRef` signatures. (A `fn`-pointer coercion, the idiom used elsewhere in that file, can't work here — `impl Into<ChildId>` in argument position isn't turbofishable — so the closure is the right tool.)
- **Ordinal fixture semantics otherwise match the originals.** `action_on_wake=true` reproduces the timer vtable's `wake → drop` delegation; `false` reproduces both the old blocking-drop vtable's inert `wake` and the mailbox vtable's drop-only panic. `run_ordinal_action` takes the action out of the mutex and releases the guard *before* invoking it, so a blocking or panicking action neither runs under the fixture's lock nor poisons it — the lock rule holds in the fixture too. One-shot-via-`take` vs one-shot-via-`AtomicBool` is not observable: only one clone ever carries a given ordinal, and a raw waker is consumed by exactly one of `wake`/`drop`.
- **No coverage lost in the `next_exit_of` migrations.** The two `offloads.rs` sites and `reply_waker_containment.rs` dropped their explicit `timeout(POLL_TIMEOUT/5s, …)` wrappers, but `next_item` bounds every `recv` at the same `POLL_TIMEOUT` and panics on both lag markers and stream closure, so the hand-rolled `Lagged`/`None` arms are preserved rather than dropped. `disposal.rs`'s two sites and `drain.rs::exited_actor` were mechanical.
- **The "intentionally stayed local" claim holds.** Every remaining `LifecycleEventKind::Exited` match in `tests/` judges membership, incarnation, exit *counts*, or `Started`/`RestartScheduled` ordering — none is a plain scan-until-this-id.
- **Static/dynamic error split is complete.** No `.md` or `.rs` outside the dynamic paths still points a pre-spawn builder at `ReserveError`; `book/src/dynamic-scopes.md`'s two references are correctly dynamic. SPEC §9/§12 and `docs/errors.md` read coherently against the code.

### Refuted / non-findings

- **`Context::new_live` + `new_draining` as two near-identical constructors.** That is the point — a single `new(raw, stage)` is what made `Initializing` + `owns_init_boundary: false` representable. `new_initializing` remains the only door to the owned boundary.
- **`clear_timer -> Result<bool, Rejected<()>>` kept.** #428 flagged it as defensible-and-leave; the rustdoc already explains that retraction stays available to a live handler after `stop` and is refused only in frozen-prefix drain, which is what makes the empty rejection worth its shape.
- **`dispose_rejected<D, E>` losing its `ReserveError` bound.** `E` is only returned, never disposed; `FrameworkPlain` matters for the admission-response receivers, which are unchanged, so `StaticReserveError` correctly does not carry the marker.
- **`#[must_use]` on `System` moving below its doc comment.** Semantically identical, and it makes the placement match the rest of the file — the lint is not lost (`Result` carries its own `must_use` on the reserve surface regardless).

### Verification

`cargo +nightly fmt --all` clean, `clippy -p shelterwood --all-targets --all-features -D warnings` clean, `just fmt` check clean. Targeted: `timer_waker_proxy` (5), `mailbox_future_drop_containment` + `reply_waker_containment` + `drain` + `tasks` (28), and a 42-test reservation-focused filter — all green. GitHub CI passes on `2606dd5`.

Nothing filed; nothing blocking merge.

PR #461, opened 2026-08-22
Harden raw teardown result retention

### LINKED ISSUE #432: Raw epilogue holds the completed exit result bare across teardown: teardown panic + hostile returned error aborts (Actor-reachable)
External-review claim (2026-08-22 round), **CONFIRMED by repro** — and broader than filed: reachable through the supported `Actor` API, not just raw. Repro committed on local branch `verify-teardown` (`crates/shelterwood/tests/raw_teardown_result_containment.rs`): 1 control passes, 2 repro tests SIGABRT (raw path and `Actor` path).

## Mechanism

`crates/shelterwood/src/raw/definition.rs:258-267` holds the completed exit result as a plain local (`Option<ExitResult>`) while the epilogue runs five fallible teardown steps — mailbox freeze (:268-272), resource freeze (:273), retained resource-disposal panic (:274/:280), `join_resources` (:278), `drop_raw` (:282), `drop_actor` (:285). Each step is contained into `cleanup_panic`, but `raw/definition.rs:293-296` then calls `resume_preferred_panic_outside_unwind` — deliberately the non-containing variant (`shelterwood-core/src/panic.rs:82-93`), so the epilogue cannot fall through to the `.expect()` at :297. On `result = Some(Err(user_error))` + any teardown panic, the coroutine unwinds with the user error still owned by the local.

Ownership: `ExitError` is `Arc<ExitErrorInner>` (`exit.rs:118`) just minted by the user's callback — **no framework copy exists yet**. The first retention point is one hop later: `RetainedRecordedOutcome` applied at `driver/child.rs:62` inside `ReportCompletion::fill`, whose own comment states the principle being violated here ("would otherwise run the application error's destructor inline on the teardown thread"). The `CatchUnwindFuture` at `driver/child.rs:590` cannot help — the second panic happens inside the first unwind, below the boundary.

Abort backtrace (from the repro) pins the frames: `HostileError::drop` → `Arc<ExitErrorInner>::drop_slow` (last owner) → `drop_glue::<Option<Result<(), ExitError>>>` (the `result` local, definition.rs:258) → "panic in a destructor during cleanup … aborting."

## Reachability

All safe user code: a callback returns `Err(e)` where `e`'s destructor panics, **and** any teardown step panics (the repro uses the actor state's own `Drop`; a hostile timer-key/message destructor works identically since `RawDisposal` records it into the panic slot that becomes `cleanup_panic`). The `Actor` path is affected because `impl<A: Actor> RawActor for Handler<A>` (`actor.rs:473`) routes `fail_after_teardown` (:575-582) into this same epilogue, and `drop_actor` at :285 destroys `A`. Even absent a panicking destructor, a *blocking* user-error destructor stalls the actor's Tokio worker during the unwind.

## Severity / remedy

**P1** (external review said Critical; needs two hostile ingredients, but both are in the declared threat model and the outcome is a process abort). The codebase treats exactly this shape as a defect everywhere else (`RetainedExit`, `RetainedRecordedOutcome`, the `child.rs:58-62` comment). Suggested remedy shape: give `result` a `RetainedExitResult`-style carrier whose `Drop` routes a failed result to `runtime::dispose_critical` (mirroring `RetainedExit`), with `into_result()` at :297 on the normal path — converting the unwind-time drop from a user destructor into a submission.

Part of the external-review verification round tracked on #429. Relates to #422 (the driver-side `handle_exit` raw-`Exit` window is the same shape one layer up, currently unreachable there).

ralexstokes:
## Summary

- retain a completed raw `ExitResult` across the fallible incarnation teardown epilogue
- route a failed result to critical disposal when teardown unwinds, while preserving normal downstream ownership
- cover both `RawActor` and supported `Actor` paths with hostile-error/teardown-panic regressions
- record the new carrier in the lock-rule ownership accounting

## Testing

- `./tools/dev cargo test -p shelterwood --test raw_teardown_result_containment`
- `./tools/dev just ci`
- `./tools/dev just ci-nix`

Closes #432.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -68,6 +68,11 @@ rests on:
   `reconcile_recorded_outcomes_retaining` and
   `classify_disposal_panic_retaining`; the raw `shelterwood-core` folds hand
   both halves back and are for core's own tests.
+  `RetainedExitResult` closes the earlier raw-incarnation window: the completed
+  callback result stays in that carrier across the fallible teardown epilogue,
+  so an epilogue panic transfers a failed result to critical disposal instead
+  of destroying its application error during the unwind. The normal path takes
+  the result back before returning it to the driver.
   `ScopeCell::clear_residents_locked`, `prune_child_locked` and
   `admit_child_locked` route `Arc<MemberCell>`s through
   `runtime::dispose_detached` after unlock: the last member owner can also be
--- a/crates/shelterwood/src/cells/retained.rs
+++ b/crates/shelterwood/src/cells/retained.rs
@@ -5,7 +5,7 @@ use shelterwood_core::{
     Exit,
     engine::ScopeState,
     exit::{
-        Cancellation, ExitKind, GracePhase, JoinOutcome, RecordedOutcome, StartupError,
+        Cancellation, ExitKind, ExitResult, GracePhase, JoinOutcome, RecordedOutcome, StartupError,
         StartupFailure, StartupFailureCause, StopReason, classify_disposal_panic, classify_exit,
         reconcile_recorded_outcomes,
     },
@@ -171,6 +171,42 @@ impl Drop for RetainedRecordedOutcome {
         };
         if outcome.is_failed() {
             runtime::dispose_critical(outcome);
+        }
+    }
+}
+
+/// A completed incarnation result retained across a fallible teardown epilogue.
+///
+/// The failed variant owns a type-erased application error that no framework
+/// copy has reached yet: the first retention point downstream is
+/// [`RetainedRecordedOutcome`], one hop after the incarnation future returns.
+/// The raw epilogue runs several contained teardown steps while holding the
+/// completed result and then resumes any surviving panic, so the carrier is
+/// dropped during that unwind — and destroying a user error there is a second
+/// panic inside the first one's cleanup, which aborts the process.
+/// [`Self::into_result`] on the normal return path preserves ordinary
+/// downstream ownership.
+pub(crate) struct RetainedExitResult(Option<ExitResult>);
+
+impl RetainedExitResult {
+    pub(crate) fn new(result: ExitResult) -> Self {
+        Self(Some(result))
+    }
+
+    pub(crate) fn into_result(mut self) -> ExitResult {
+        self.0
+            .take()
+            .expect("retained exit result was already taken")
+    }
+}
+
+impl Drop for RetainedExitResult {
+    fn drop(&mut self) {
+        let Some(result) = self.0.take() else {
+            return;
+        };
+        if let Err(error) = result {
+            runtime::dispose_critical(error);
         }
     }
 }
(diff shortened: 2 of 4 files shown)


ralexstokes:
Fresh adversarial review: no findings. I traced the completed ExitResult through every fallible raw-epilogue step, teardown unwind, task cancellation, normal handoff into RecordedOutcome, and critical-disposal failure paths. The carrier is armed before mailbox/resource teardown, its Drop transfers only failed results to dispose_critical, and the normal path disarms it exactly once before driver ownership. The RawActor and supported Actor regressions both prove the hostile error is destroyed off the actor thread while preserving the actor-drop panic. Fresh checks: targeted regression test passed; ./tools/dev just ci passed (952 tests plus fmt, clippy, doctests, examples, docs, API reachability, and manifest checks). Worktree remained clean, so I made no speculative changes.

ralexstokes:
## Review

Verified the fix is real and load-bearing, then addressed three findings directly on the branch (34f3089).

**Fix correctness — confirmed.** Mutation-checked by replacing the carrier's `Drop` body with an inline `drop(result)`: both new tests SIGABRT (`panic in a destructor during cleanup`), and pass with the disposal restored. The retention window is complete — the carrier is armed at the only point the result can first exist, and `into_result()` runs after `resume_preferred_panic_outside_unwind`, so every teardown-panic path drops it during the unwind. It also covers the hard-abort case: `result` is declared after `owner`, so the carrier's drop glue runs *before* `RawIncarnationOwner::drop` can resume a panic of its own. Traced the downstream hop too — the returned value reaches `RecordedOutcome::returned` → `ReportCompletion::fill` with no intervening yield point, so no second bare window opens.

### Findings addressed

1. **The carrier lived apart from its family.** `RetainedExit`, `RetainedRecordedOutcome` and `RetainedStopReason` all live in `cells/retained.rs` — which is already a cross-cutting home, not a cells-only one (`RetainedRecordedOutcome`'s consumer is `driver/child.rs`). `RetainedExitResult` is the fourth member of exactly that family and the AGENTS.md paragraph this PR amends describes them as one. Moved it there and exported it through the existing `pub(crate) use retained::*`; `raw/definition.rs` now imports it and drops its `runtime::self` import.

2. **No unit coverage for the carrier itself.** Each sibling carries a two-direction pair — one pinning the disposal venue, one pinning that the take-back path keeps ordinary caller drop timing. The PR's integration tests pinned only the unwind direction, and only indirectly (via the abort). Added the matching pair using the established `ThreadProbe`/`TEST_WAIT` support. Mutation-checked both: `drop(result)` in `Drop` fails the first, and a cloning `into_result` fails the second.

3. **The containment probe compared against the wrong thread.** `let actor_thread = std::thread::current().id()` captured the *test* thread; it coincides with the teardown thread only because `#[tokio::test]` defaults to the current-thread flavour. Under a multi-thread flavour the `assert_ne!` would have passed vacuously. The actors now record the thread that actually ran their destructor, and the assertion compares the disposal thread against that.

### Verification

- `./tools/dev just ci` — green (954 tests)
- mutation checks above, each on a committed tree

Merging.

PR #462, opened 2026-08-22
Remove unused framework-plain receiver seam

### LINKED ISSUE #457: FrameworkPlain / DisposingReceiver::new_framework has no production caller after #450
Fallout from the #426 hardening round (PR #450), noted in review rather than
expanded into that PR.

#450 seals `FrameworkPlain` with a private supertrait
(`crates/shelterwood-runtime/src/sync.rs`), which is the right fix for the
#429 finding: a foreign crate could previously mark a user-owning response
type as plain and select inline destruction in drop glue. Because the two
real markers lived in the façade (`ReserveError`, `RemoveOutcome` in
`crates/shelterwood/src/cells/admission.rs`), sealing the trait meant those
two impls had to go, and `crates/shelterwood/src/tree/admission.rs` moved the
admission and removal response receivers to `DisposingReceiver::new`
(detached lane).

That is a safe and cheap trade for those cold cancellation paths. The residue
is that the whole marker seam is now unreachable from production code:

- `DisposingReceiver::new_framework` — only caller is
  `sync::tests::framework_disposing_receiver_drops_abandoned_values_inline`.
- `FrameworkPlain`, its `()` / `Result<T, E>` impls, the `framework_plain::Sealed`
  module, and the `compile_fail` seal doctest exist only to support that one
  test.

Nothing warns: the items are `pub` in a crate with `#![allow(unreachable_pub)]`,
and the façade reachability probe only walks the supported public API, so this
seam will rot silently.

Two ways out, both fine — this is a taste call the round should make
deliberately rather than by omission:

1. **Delete it.** `DisposingReceiver::new` covers every current need; the
   inline fast path buys one avoided lane submission on paths that only run
   when a caller abandons an admission or removal future. Removing
   `new_framework`, `FrameworkPlain`, `framework_plain`, the doctest and the
   test is a net simplification, and it removes the only reason a reviewer
   has to reason about "is this response type really plain?".
2. **Give it a runtime-owned user.** If the inline venue is wanted for a seam
   the runtime crate itself owns, add that caller and keep the seal. Nothing
   in the current adapter needs it.

Related: the project's own taste note that a marker/newtype must enforce
something at a call site or not exist.

ralexstokes:
Closes #457.\n\nRemoves the now-production-unreachable FrameworkPlain marker family, DisposingReceiver::new_framework, its compile-fail documentation, and its sole supporting test. DisposingReceiver::new remains the single production constructor and continues routing abandoned values through detached disposal.\n\nVerification: ./tools/dev cargo test -p shelterwood-runtime; ./tools/dev just ci.

### DIFF
--- a/crates/shelterwood-runtime/src/sync.rs
+++ b/crates/shelterwood-runtime/src/sync.rs
@@ -676,41 +676,6 @@ impl<T> OneShotReceiver<T> {
     }
 }
 
-mod framework_plain {
-    pub trait Sealed {}
-
-    impl Sealed for () {}
-
-    impl<T: Sealed, E: Sealed> Sealed for Result<T, E> {}
-}
-
-/// Marker for framework-owned response values whose destructor runs no user
-/// code.
-///
-/// [`DisposingReceiver::new_framework`] drops an abandoned published value
-/// inline in holder drop glue instead of submitting it to the blocking
-/// disposal lane, so this bound is what keeps that fast path from ever
-/// reaching a destructor that can block, panic, or re-enter. Implement it
-/// only for plain framework enums and their compositions; a type that can
-/// own user data — an `Exit`, a message, a type-erased error — must go
-/// through [`DisposingReceiver::new`] instead.
-///
-/// This trait is sealed: response types owned by another crate must use
-/// [`DisposingReceiver::new`], even when that crate currently knows their
-/// variants are plain. That keeps the authority to select inline destruction
-/// in the same runtime crate that owns the disposal implementation.
-///
-/// ```compile_fail
-/// struct ForeignResponse;
-/// impl shelterwood_runtime::FrameworkPlain for ForeignResponse {}
-/// ```
-#[doc(hidden)]
-pub trait FrameworkPlain: framework_plain::Sealed + Send + 'static {}
-
-impl FrameworkPlain for () {}
-
-impl<T: FrameworkPlain, E: FrameworkPlain> FrameworkPlain for Result<T, E> {}
-
 /// One-shot receive state that keeps user value destruction off framework and
 /// holder drop glue.
 ///
@@ -737,20 +702,6 @@ pub struct DisposingReceiver<T> {
 impl<T: Send + 'static> DisposingReceiver<T> {
     pub fn new(inner: OneShotReceiver<T>) -> Self {
         Self::with_dispose(inner, dispose_detached::<T>)
-    }
-}
-
-impl<T: FrameworkPlain> DisposingReceiver<T> {
-    /// Creates a receiver whose abandoned value is plain framework data.
-    ///
-    /// This is an implementation seam for internal response enums. Unlike
-    /// [`Self::new`], dropping a published but unclaimed value does not
-    /// submit work to the blocking disposal lane; the [`FrameworkPlain`]
-    /// bound is what keeps that inline drop from ever reaching a user-owned
-    /// destructor.
-    #[doc(hidden)]
-    pub fn new_framework(inner: OneShotReceiver<T>) -> Self {
-        Self::with_dispose(inner, drop)
     }
 }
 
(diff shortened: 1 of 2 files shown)


ralexstokes:
## Review

Reviewed the diff against #457's adjudication, then swept the workspace (crates, `tests/`, doctests, `examples/`, `specs/`, `book/`, `tools/`) for anything that leaned on the removed seam. The PR implements the adjudicated shape exactly — `FrameworkPlain`, its `()`/`Result<T, E>` impls, `framework_plain::Sealed`, the `compile_fail` seal doctest, `DisposingReceiver::new_framework`, and the supporting test are all gone, and nothing else moves.

### The central question: gate or dead weight?

**Dead weight, confirmed.** The marker enforced nothing at any call site as of this PR's base. `git log -S new_framework` gives the whole history: the only two call sites ever to exist were `crates/shelterwood/src/tree/admission.rs:56` and `:240` at `a13d024`, and `4dda6eb` ("Seal inline disposal and complete one-shot tracking") already converted both to `DisposingReceiver::new`. From that commit onward the sealed trait's only remaining consumer was the test this PR deletes, which constructed a probe type solely to exercise the constructor being removed.

So the house rule applies cleanly: a marker must enforce something at a call site or not exist. Nothing loses a type-level guarantee here — the guarantee it offered (inline destruction is safe) was one no live code was asking for.

**Behavior is unchanged.** All five live `DisposingReceiver` constructions (`task.rs:291`, `spawn.rs:175`, `tree/admission.rs` ×2, plus the façade's separate capability-module type) already used `new`/detached disposal. No abandonment path changes venue.

### Fixed and pushed (`dac1e24`)

Two leftovers the deletion left behind:

1. `crates/shelterwood/src/tree/admission.rs:56-58` and `:249-250` — comments added by `4dda6eb` explaining *why* the detached lane was chosen over "the runtime's inline-disposal marker … deliberately sealed in its own crate". With the marker gone there is no alternative to choose, so the prose pointed a reader at a design that no longer exists. Removed both.
2. `crates/shelterwood-runtime/src/sync.rs:708` — `with_dispose` existed only to share field initialization between the two constructors. Folded into `new`, now the only one. The erased `dispose: fn(T)` field stays: `Drop` cannot name `T: Send + 'static` itself, which is the load-bearing erasure the struct doc already explains.

`./tools/dev just ci` green on the pushed head (exit 0; 953 tests, clippy `-D warnings`, rustdoc `-D warnings`, API reachability + prelude containment clean, core-manifest and external-consumer probes clean, nixfmt clean).

### Examined and refuted

- **Lock rule.** The deletion removes only an inline-drop fast path nothing selects. What survives is `dispose_detached`, the *stronger* venue (blocking-worker disposal), so no destruction newly happens under a lock. `Drop for DisposingReceiver` is untouched and still runs its `PanicAccumulator` with no framework lock held. Nothing in CLAUDE.md's lock-rule accounting cites `FrameworkPlain` — the marker was a venue *optimization* gate, not a lock-safety gate.
- **Reachability probes.** `FrameworkPlain` appears in no allowlist. It lived in `shelterwood-runtime` (exported only through `lib.rs:24`'s `pub use sync::*` and `#[doc(hidden)]`) and was never reachable from the façade, so `tools/external-consumer/src/main.rs`'s framework-only import list never named it; `check-core-manifest.sh` pins manifest deps, not item names. Nothing needed updating.
- **Stale prose elsewhere.** After the two comment removals, `FrameworkPlain` / `new_framework` / `framework_plain` occur zero times in the tree. `specs/`, `book/` and `CLAUDE.md` never mentioned the seam.
- **Deleted test coverage.** `framework_disposing_receiver_drops_abandoned_values_inline` pinned only the deleted constructor's venue; it asserted nothing still true. The sibling `disposing_receiver_*` tests continue to pin the detached venue and the ready-edge waker containment. The `compile_fail` doctest sealed a trait that no longer exists.
- **Orphaned imports.** `cell::Cell` was correctly dropped from the test imports and is unused elsewhere in that module; clippy `-D warnings` on `--all-targets` confirms.

### Follow-ups (out of scope, not filed)

- `crates/shelterwood-runtime/src/lib.rs:24` re-exports `sync` with a glob, so any `pub` item in `sync.rs` becomes runtime-crate public API by default — that is how a single-call-site implementation seam ended up publicly exported. Tightening that glob to an explicit list would make the next such seam visible at review time rather than at audit time.

PR #463, opened 2026-08-23
Make waker-proxy locking poison tolerant

### LINKED ISSUE #454: Waker-proxy retirement: the `.lock().expect()` accumulator wrap is applied at one of three unwind-reachable call sites
Noticed while reviewing #446 (issue #424). Not a reachable bug — the proxy
mutex is unpoisonable by construction — but the defence-in-depth posture is
currently applied at exactly one of the three drop-glue call sites, which
makes it read as an oversight in the other two.

`ProxiedSleep::retire` (`crates/shelterwood-core/src/proxied_sleep.rs:79-99`)
states the policy explicitly:

```rust
// The proxy mutex guards framework-owned data and is documented
// unpoisonable, but this path also runs during unwind. Keep every
// cleanup step inside the accumulator so a bookkeeping defect cannot
// turn a caller panic into an abort.
panics.run(|| timer_poll.retire(action, &mut effects));
```

That wrap covers the whole of `ProxiedPoll::retire`, including
`WakerProxy::retire`'s `.lock().expect("waker proxy mutex poisoned")` and the
subsequent `WakerProxy` drop.

The two sibling seams reach the same `.expect` from drop glue with no
accumulator around it:

- `shelterwood-runtime::ProxiedPoll::drop`
  (`crates/shelterwood-runtime/src/waker_proxy.rs:52-63`) → `retire_detached`
  → `ProxiedPoll::retire_with` → `WakerProxy::retire_with` → `.lock().expect(…)`.
- `DisposingReceiver::drop` → `retire_reply_waker`
  (`crates/shelterwood/src/mailbox/capability.rs:158-164`) →
  `ProxiedPoll::retire` → `.lock().expect(…)`.

In both, a panic there escapes drop glue; if the frame is already unwinding
that is a double panic and an abort — precisely the outcome `ProxiedSleep`'s
comment says the wrap exists to prevent.

Note this asymmetry predates #446: the `panics.run(|| drop(proxy))` that PR
removed from `ProxiedPoll::retire_with` wrapped only the proxy drop, never the
`.lock()` that would have fired first, so removing it changed nothing material
(the issue-#424 disposition analysing it as dead weight is correct).

Options, in increasing scope:

1. Do nothing and say so: add one sentence to `ProxiedSleep::retire`'s comment
   noting the wrap is local belt-and-braces, not a rule the other seams are
   failing to meet.
2. Wrap uniformly at the two drop-glue call sites.
3. Make it structural: have `WakerProxy` acquire its own guard through a helper
   that cannot panic (`lock().unwrap_or_else(PoisonError::into_inner)`), on the
   grounds that a poisoned leaf holding only framework-owned data has nothing
   to protect. That removes the panic source rather than containing it, and
   would let all three sites drop the ceremony.

Related to #427 (lock-rule accounting and doc drift).

ralexstokes:
## Summary

- route every WakerProxy registration guard acquisition through one poison-recovering helper
- remove ProxiedSleep retirement accumulator ceremony that existed only for the former panicking lock acquisition
- document the framework-only recovery invariant and test register, wake, retire, and drop after deliberate poisoning

## Validation

- ./tools/dev cargo test -p shelterwood-core waker_proxy
- ./tools/dev cargo test -p shelterwood-core
- ./tools/dev just ci

Closes #454

### DIFF
--- a/crates/shelterwood-core/src/proxied_sleep.rs
+++ b/crates/shelterwood-core/src/proxied_sleep.rs
@@ -78,12 +78,10 @@ impl ProxiedSleep {
 
     fn retire(&mut self, action: WakerAction, panics: &mut PanicAccumulator) {
         let mut effects = WakerEffects::default();
-        let timer_poll = &mut self.timer_poll;
-        // The proxy mutex guards framework-owned data and is documented
-        // unpoisonable, but this path also runs during unwind. Keep every
-        // cleanup step inside the accumulator so a bookkeeping defect cannot
-        // turn a caller panic into an abort.
-        panics.run(|| timer_poll.retire(action, &mut effects));
+        // Proxy retirement only moves framework-owned bookkeeping into the
+        // effects sink, and its leaf-lock acquisition recovers poison. The
+        // caller-owned waker remains inside the accumulator-backed flush.
+        self.timer_poll.retire(action, &mut effects);
         effects.flush(panics);
 
         // Slot first, timer second: once the caller waker is gone, cancelling
--- a/crates/shelterwood-core/src/proxied_sleep.rs
+++ b/crates/shelterwood-core/src/proxied_sleep.rs
@@ -80,7 +80,9 @@ impl ProxiedSleep {
         let mut effects = WakerEffects::default();
         // Proxy retirement only moves framework-owned bookkeeping into the
         // effects sink, and its leaf-lock acquisition recovers poison. The
-        // caller-owned waker remains inside the accumulator-backed flush.
+        // proxy's own drop then finds the slot already emptied, so no caller
+        // destructor runs here either; the caller-owned waker stays inside
+        // the accumulator-backed flush below.
         self.timer_poll.retire(action, &mut effects);
         effects.flush(panics);
 
--- a/crates/shelterwood-core/src/waker_proxy.rs
+++ b/crates/shelterwood-core/src/waker_proxy.rs
@@ -316,7 +316,7 @@ mod tests {
         mem::ManuallyDrop,
         panic::{AssertUnwindSafe, catch_unwind},
         sync::{
-            Arc, Weak,
+            Arc, TryLockError, Weak,
             atomic::{AtomicBool, AtomicUsize, Ordering},
         },
         task::{Context, Poll, RawWaker, RawWakerVTable, Wake, Waker},
@@ -364,10 +364,13 @@ mod tests {
     impl Drop for ReentrantDrop {
         fn drop(&mut self) {
             let state = self.0.upgrade().expect("the proxy remains live");
-            let _guard = state
-                .caller
-                .try_lock()
-                .expect("a displaced waker drops after the proxy mutex is released");
+            let _guard = match state.caller.try_lock() {
+                Ok(guard) => guard,
+                Err(TryLockError::Poisoned(error)) => error.into_inner(),
+                Err(TryLockError::WouldBlock) => {
+                    panic!("a displaced waker drops after the proxy mutex is released")
+                }
+            };
         }
     }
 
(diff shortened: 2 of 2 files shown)


ralexstokes:
## Review of #463 (issue #454, option 3)

Reviewed the whole diff against `main` (`proxied_sleep.rs`, `waker_proxy.rs`), the adjudicated shape on #454, and the three seams the issue names. Two fixes pushed on the branch; `just ci` green at `c44eea5` (955 tests).

### Shape conformance

Implements option 3 as ruled: one private `WakerProxyState::registration()` helper doing
`lock().unwrap_or_else(PoisonError::into_inner)`, every acquisition routed through it, and the one site that carried the ceremony (`ProxiedSleep::retire`) drops it. That matches the house pattern already used at `shelterwood-runtime/src/sync.rs:856` (`WatchShared::value`), `spawn.rs:244` (`BlockingJob::lock_pending`) and `shelterwood/src/cells/gate.rs:34`. The other two seams the issue lists keep their accumulators for the right reason — they still flush a *user* waker effect (`runtime/src/waker_proxy.rs:57-64`, `mailbox/capability.rs:158-164`), which is not the ceremony #454 was about.

### 1. Can recovery resume on torn state? No — verified field by field

The guard covers exactly `Registration { caller: WakerSlot, woken: bool }` (`waker_proxy.rs:139-146`). Every critical section under it does one of three things: compare waker pointers (`WakerSlot::will_wake`), set/clear a `bool`, or move a `Waker` between the slot and an effects sink **constructed before the guard**. No user code, no allocation of the sink, no callback. The only in-section panic source left is `Vec::push`'s capacity/alloc path in `WakerEffects::push`, which aborts rather than unwinds on the platform allocator.

Even granting a hypothetical unwind there, the reachable residues are one-directional and conservative:

* `wake_by_ref` (`:296-311`) sets `woken` **then** empties the slot, so a tear can leave `woken = true` with the slot already empty — one spurious poll for the next registrant, which `Future` permits.
* `register` (`:200-222`) installs before it reads-and-clears `woken`, so a tear leaves the flag set — again a spurious wake.
* The converse — a cleared record beside a removed waker, i.e. the lost wake the whole handshake exists to prevent — has no code path: the set and the take are in the same critical section with nothing between them.

I folded that derivation into the `registration()` doc (it previously said only "contains only framework-owned state", which slightly overstates the case: the slot does hold a caller-owned `Waker`; the point is that no *invariant spans a panic point*).

### 2. Does it swallow evidence? Refuted as a finding

A `debug_assert!(!is_poisoned)` inside `registration()` would fire in precisely the case the PR exists to remove — poisoned leaf reached from drop glue during an unwind — turning the diagnostic into the abort. That is not "fail-closed in addition to a debug abort"; it is the abort. The three existing poison-recovery helpers in this repo carry a prose rationale and no diagnostic, and this one now does too. Nothing else is silenced: no verdict, no user callback, no panic payload is dropped by this change.

### 3. Lock rule: clean

The helper adds nothing under the guard — `PoisonError::into_inner` is a move of the same guard. No framework lock is taken beneath the leaf, and every caller-waker clone/wake/drop still leaves through `WakerEffects` after unlock (pinned by the reentrancy probes, which panic if `try_lock` reports `WouldBlock`).

### 4. Confirmed finding: the new test's premise was fail-open (fixed, `71272ee`)

`poisoned_proxy_leaf_remains_usable_across_every_transition` asserted only `poisoned.is_err()`. Every step after it also passes on an *unpoisoned* proxy, so if the injected panic had ever escaped before the guard was acquired the test would have gone green while testing nothing. Not hypothetical — mutation-checked: deleting the `lock()` from the injection closure (leaving the bare `panic!`) **passed** on `fce3bb4`. Fixed by asserting `proxy.state.caller.is_poisoned()`; that mutation now fails.

Also in `71272ee`:

* **`retire_with` was uncovered.** The test walked `register`/`wake_by_ref`/`retire`/drop glue, but not the cross-crate seam — which is the one `shelterwood-runtime`'s `ProxiedPoll::drop` reaches the leaf through, i.e. the first of the two sites #454 was filed about. Added, with a `ReentrantWake` wake-count assertion so it carries positive evidence rather than just "did not panic". The three retirement steps now map one-to-one onto the three seams.
* **Two sibling reentrancy probes were left poison-intolerant.** The PR fixed `ReentrantDrop::drop`'s `try_lock` to recover poison but not `ReentrantWake::wake` (`:345`) or `clone_reentrant` (`:398`), which still did `.try_lock().expect(…)`. On a poisoned-but-free mutex `try_lock` returns `Err(Poisoned)`, so extending the poison test over either — which this review did — would have failed with the misleading message "runs after the proxy mutex is released". Folded all three onto one `assert_proxy_released` helper where only `WouldBlock` means a guard is held. Mutation-checked: making the helper treat any `Err` as held fails the poison test.

### 5. The behaviour under test really is the abort class #454 names

Restoring `.lock().expect("waker proxy mutex poisoned")` on the strengthened branch does not merely fail the test — it produces `panic in a destructor during cleanup / thread caused non-unwinding panic. aborting.` (SIGABRT). So the drop-glue tail added in `fce3bb4` is load-bearing, and nextest's per-test isolation keeps it an ordinary `#[test]`.

### 6. `ProxiedSleep::retire` unwrapping is sound (comment hardened, `c44eea5`)

Removing `panics.run(…)` there also removed containment for the `WakerProxy` drop that `ProxiedPoll::retire` performs after taking the slot. That is safe, but for a reason worth writing down rather than re-deriving: the caller's `retire` has already emptied the slot into *its* sink, and nothing can refill it (the proxy is owned through `&mut ProxiedPoll`; `wake_by_ref` only takes), so `WakerProxy::drop`'s own sink is provably empty and runs no caller destructor. Added that clause to the comment the PR rewrote.

## Follow-ups (not folded in)

1. `WakerProxy::drop`'s doc says the caller waker's vtable runs "contained, but synchronously". Only half true, and I verified it with a scratch test: on a **non-unwinding** thread `WakerEffects::Drop` flushes into a temporary `PanicAccumulator` whose own `Drop` *resumes* the payload, so a hostile destructor panic propagates out of the drop glue; during an unwind it is discarded. Not reachable as a defect (every framework caller retires beforehand with its own accumulator), but the word "contained" invites the wrong assumption.
2. `replacement_drops_the_displaced_caller_after_unlock` and `proxy_drop_retires_the_caller_after_unlock` assert only from inside `ReentrantDrop::drop` — if the waker were leaked instead of dropped, both pass silently. Pre-existing house pattern; a drop counter on the helper would close it for all four users at once.

PR #464, opened 2026-08-23
Share raw-waker probes across tests

### LINKED ISSUE #456: Test-only: extend the shared raw-waker probe beyond the mailbox module
Follow-up from the #421 round (PR #451).

#421 item 8 called out five near-identical hand-rolled `RawWakerVTable` Arc-probe
scaffolds inside `crates/shelterwood/src/mailbox/`. PR #451 collapsed all five into
one parameterized `probe_waker(on_clone, on_drop)` in
`crates/shelterwood/src/mailbox/test_support.rs`, deleting ~150 lines of repeated
`unsafe` and concentrating the safety comments in one place.

The same duplication class survives elsewhere in the workspace. Ranked by how
cheaply `probe_waker` could absorb them:

1. **Same crate, same `pub(crate)` reach — no new machinery needed.**
   - `crates/shelterwood/src/task.rs:480` (`PANICKING_DROP_WAKER_VTABLE`)
   - `crates/shelterwood/src/driver/tests/system_join.rs:44` (`HOSTILE_WAKER_VTABLE`)
   Both are clone-passthrough + observable-drop probes, exactly `probe_waker`'s
   shape. Moving `probe_waker` out of `mailbox::test_support` into a crate-level
   test-support module would let both drop their `unsafe`.

2. **Integration tests.** `crates/shelterwood/tests/common/waker.rs:37` already is
   a shared integration-side helper, but `tests/timer_waker_proxy.rs` (three
   vtables) and `tests/mailbox_future_drop_containment.rs` (two) each restate
   their own. `common/waker.rs` is the natural home for an integration-side twin
   of `probe_waker`; the ordinal-recording and blocking-drop variants are still
   just `on_drop` closures.

3. **`shelterwood-core` / `shelterwood-runtime`.** `waker_proxy.rs`,
   `proxied_sleep.rs`, `sync.rs`, `spawn.rs`, `timer.rs` hold seven more. These
   cannot reuse the facade helper across the crate boundary, and at least
   `REENTRANT_CLONE_VTABLE` needs a clone-side probe that re-enters the proxy —
   representable as an `on_clone` closure, but it would want its own per-crate
   copy of the scaffold rather than a shared one.

Not a correctness item: every scaffold is currently correct, and each was reviewed
when it landed. The argument is the one #421 item 8 made — every hand-written
vtable is an independent `unsafe` refcount contract that a reviewer has to
re-derive, and the shared probe makes each regression state only the behavior it
observes.

Suggest doing (1) at minimum, (2) if it stays mechanical, and leaving (3) as a
deliberate no.

ralexstokes:
Closes #456.

Moves the unit-test probe_waker scaffold to crate-level test support and migrates mailbox, task, and SystemRun join probes. Adds the integration-side twin, migrates the remaining mailbox-future scaffold and shared hostile-waker fixture, and retains the already-centralized specialized ordinal/counting integration helpers. The core/runtime tier remains deliberately unchanged per adjudication.

Fresh adversarial review restored the mailbox fixture's original panic contract for consuming and by-reference wakes and added direct coverage for clone, wake, wake-by-reference, drop, and panicking-wake ownership.

Verification: ./tools/dev just ci (956 tests plus docs, examples, API reachability, external consumer, and nixfmt).

### DIFF
--- a/crates/shelterwood/src/driver/tests/system_join.rs
+++ b/crates/shelterwood/src/driver/tests/system_join.rs
@@ -1,62 +1,25 @@
 use std::{
     mem::ManuallyDrop,
     sync::Arc,
-    task::{Context, Poll, RawWaker, RawWakerVTable, Waker},
+    task::{Context, Poll, Waker},
 };
 
 use super::support::*;
-use crate::{cells::RetainedStopReason, policy::ScopeFlavor};
+use crate::{cells::RetainedStopReason, policy::ScopeFlavor, test_support::probe_waker_with_wake};
 
 #[derive(Default)]
 struct HostileWakeState {
     woken: tokio::sync::Notify,
 }
 
-unsafe fn clone_hostile_waker(data: *const ()) -> RawWaker {
-    // SAFETY: every pointer using this vtable came from an Arc of the
-    // matching type. ManuallyDrop preserves the reference represented by
-    // `data`; the returned raw waker owns only the new clone.
-    let state = ManuallyDrop::new(unsafe { Arc::<HostileWakeState>::from_raw(data.cast()) });
-    RawWaker::new(
-        Arc::into_raw(Arc::clone(&state)).cast(),
-        &HOSTILE_WAKER_VTABLE,
-    )
-}
-
-unsafe fn wake_hostile_waker(data: *const ()) {
-    // SAFETY: wake consumes the Arc reference represented by this raw waker.
-    let state = unsafe { Arc::<HostileWakeState>::from_raw(data.cast()) };
-    state.woken.notify_one();
-}
-
-unsafe fn wake_by_ref_hostile_waker(data: *const ()) {
-    // SAFETY: ManuallyDrop preserves the reference represented by `data`.
-    let state = ManuallyDrop::new(unsafe { Arc::<HostileWakeState>::from_raw(data.cast()) });
-    state.woken.notify_one();
-}
-
-unsafe fn drop_hostile_waker(data: *const ()) {
-    // SAFETY: drop consumes the Arc reference represented by this raw waker.
-    drop(unsafe { Arc::<HostileWakeState>::from_raw(data.cast()) });
-    panic!("injected SystemRun join caller-waker drop panic");
-}
-
-static HOSTILE_WAKER_VTABLE: RawWakerVTable = RawWakerVTable::new(
-    clone_hostile_waker,
-    wake_hostile_waker,
-    wake_by_ref_hostile_waker,
-    drop_hostile_waker,
-);
-
 fn hostile_waker() -> (ManuallyDrop<Waker>, Arc<HostileWakeState>) {
     let state = Arc::new(HostileWakeState::default());
-    let raw = RawWaker::new(
-        Arc::into_raw(Arc::clone(&state)).cast(),
-        &HOSTILE_WAKER_VTABLE,
+    let wake_state = Arc::clone(&state);
+    let waker = probe_waker_with_wake(
+        || {},
+        move || wake_state.woken.notify_one(),
+        || panic!("injected SystemRun join caller-waker drop panic"),
     );
-    // SAFETY: `raw` owns one Arc reference and its vtable maintains that
-    // ownership across clone, wake, and drop.
-    let waker = unsafe { Waker::from_raw(raw) };
     (ManuallyDrop::new(waker), state)
 }
 
--- a/crates/shelterwood/src/lib.rs
+++ b/crates/shelterwood/src/lib.rs
@@ -226,6 +226,8 @@ mod raw;
 mod runtime;
 mod scope;
 mod task;
+#[cfg(test)]
+mod test_support;
 mod tree;
 
 pub use actor::{Actor, ActorDef, ActorOnceDef, Context, Handler, StopContext};
--- a/crates/shelterwood/src/mailbox.rs
+++ b/crates/shelterwood/src/mailbox.rs
@@ -24,8 +24,6 @@ mod errors;
 mod futures;
 mod reply;
 #[cfg(test)]
-mod test_support;
-#[cfg(test)]
 mod timer;
 pub(crate) use cell::*;
 pub use errors::*;
--- a/crates/shelterwood/src/mailbox/cell/tests.rs
+++ b/crates/shelterwood/src/mailbox/cell/tests.rs
@@ -15,9 +15,9 @@ use crate::{
     mailbox::{
         ActorIdentity, ActorRef, ChildId, Incarnation, MailboxControl, MailboxReceiver,
         SendErrorKind,
-        test_support::{mint_actor_incarnation, mint_actor_membership},
     },
     policy::{ResolvedDefaults, ResolvedMailbox},
+    test_support::{mint_actor_incarnation, mint_actor_membership},
 };
 
 use super::MailboxCell;
--- a/crates/shelterwood/src/mailbox/deadline.rs
+++ b/crates/shelterwood/src/mailbox/deadline.rs
@@ -251,7 +251,7 @@ mod tests {
         task::{Context, Poll, Wake, Waker},
     };
 
-    use crate::mailbox::test_support::probe_waker;
+    use crate::test_support::probe_waker;
 
     /// Stays pending on its first elapsed poll, modelling an atomic
     /// completion transition that won without publishing its payload yet.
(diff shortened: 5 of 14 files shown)


ralexstokes:
## Review of #464 (head `58eeb3a`, reviewed on a clean worktree)

Scope: the whole diff (`origin/main...HEAD`), the #456 adjudication, and every
call site of the five bespoke probes this PR folds away. Because the dominant
risk in a helper-consolidation PR is coverage that disappears without a test
going red, the review was driven by mutation rather than by reading.

### Coverage equivalence, helper by helper

The shared vtable's four entries were compared against each replaced vtable
operation-by-operation (side effect, and whether the operation consumes,
preserves, or clones the raw reference).

| Replaced probe | Old vtable | New spelling | Verdict |
| --- | --- | --- | --- |
| `src/mailbox/test_support.rs::probe_waker` (→ `src/test_support.rs`) | clone→`on_clone`; wake→silent consume; wake_by_ref→no-op; drop→`on_drop` | `probe_waker` = `probe_waker_with_wake(on_clone, \|\| {}, on_drop)` | **Equivalent** (strict superset: wake is now routable, and defaults to the old silent no-op for all five existing callers — `deadline.rs`, `futures.rs`, `reply.rs`, `timer.rs`, `task.rs`) |
| `src/task.rs::panicking_drop_waker` | clone→∅; wake→silent consume; wake_by_ref→no-op; drop→consume + `panic!` | `probe_waker(\|\| {}, \|\| panic!(…))` | **Equivalent** |
| `src/driver/tests/system_join.rs::hostile_waker` | clone→∅; wake→consume + `notify_one`; wake_by_ref→`notify_one`, no consume; drop→consume + `panic!` | `probe_waker_with_wake(\|\| {}, notify_one, panic!)` | **Equivalent** — and this is the helper that forced the `on_wake` plumbing in `58eeb3a`; the first commit alone would have silently dropped the notify |
| `tests/common/waker.rs::hostile_waker` | clone→∅; wake→silent consume; wake_by_ref→no-op; drop→consume + `panic_any(&'static str)` | `probe_waker(\|\| {}, panic_any(drop_panic))` | **Equivalent**, payload type preserved (`&'static str`, which `reply_waker_containment.rs` downcasts) |
| `tests/mailbox_future_drop_containment.rs::panicking_drop_waker` | clone→∅; wake→consume + `panic!("…wake panic")`; wake_by_ref→`panic!` w/o consume; drop→consume, **first** drop panics | `probe_waker_with_wake(\|\| {}, panic!, first-drop-panics)` | **Equivalent** — the once-only latch moves from a bespoke `FirstWakerDropPanics` payload into the `on_drop` closure, but both live behind the single `Arc` shared by every clone, so "the first destruction of any clone panics" is unchanged |
| `tests/common/waker.rs::{counting,ordinal,ordinal_drop}_waker` | unchanged | unchanged | **Deliberately unchanged, justified**: `ordinal_waker` needs *per-clone* identity (each clone mints its own ordinal), which a probe whose callbacks live in one shared `Arc` cannot express. `timer_waker_proxy.rs` was already on these helpers on `main`, so tier 2's second scaffold needed no migration. |

Ordering nuance worth recording, not a finding: three old vtables destroyed
their `Arc` *before* panicking; the shared `drop_probe`/`wake_probe` hold it
across the callback and release it during the unwind. That is inherent (the
callback lives inside the `Arc`), it is the shape `main` already shipped in
`src/mailbox/test_support.rs`, and none of the captured states have panicking
destructors, so no new double-panic edge exists.

The two probe copies (`src/test_support.rs` vs `tests/common/waker.rs`) are
byte-identical modulo doc comments — a faithful restatement, and the twin is
explicitly what #456 tier 2 asked for, so the duplication is adjudicated, not
a "test doubles should delegate" violation.

### Mutation checks

"Pinned" claims verified by mutating the implementation on a committed tree,
never by name-checking a test.

1. **Caller waker never reaches the external primitive.** `ProxiedPoll::poll`
   → poll the target with the caller's own `Context`.
   Red: `driver::tests::system_join::{system_run_wait,system_run_shutdown}_proxies_a_pending_driver_join_caller_waker`,
   `task::tests::staged_one_shot_wait_cannot_double_panic_at_completion_delivery` (**SIGABRT**),
   `reply_delivery_waker_proxy::successful_recv_cannot_double_panic_with_hostile_waker_and_value_drops` (**SIGABRT**),
   `reply_delivery_waker_proxy::successful_call_contains_reply_caller_waker_retirement`,
   `mailbox::reply::tests::{close,ready_race,timeout_arbitration_value}_…`,
   `mailbox::timer::tests::drop_glue_retires_the_caller_waker_on_the_disposal_lane`.
   That is four of the five migrated helpers going red at once.
2. **Inline waker retirement stops being contained.**
   `WakerEffects::flush`: `WakerAction::DropInline => panics.run(|| drop(waker))`
   → `drop(waker)`.
   Red: `mailbox_future_drop_containment::{recv_drop,call_reply_phase_drop,recv_ready_after_parking}_…` (**SIGABRT**),
   `mailbox::reply::tests::close_retires_the_reply_caller_waker_and_contains_a_hostile_destructor`.
   This is the mutation the fifth migrated helper answers.
3. **The proxy loses the caller's wake.** `WakerProxyState::wake_by_ref`:
   `WakerAction::Wake` → `WakerAction::DropInline`.
   Red (hang → TIMEOUT): both `system_join` tests. Confirms the `on_wake`
   restoration in `58eeb3a` is load-bearing and not decorative.
4. **The new probe self-tests are live, not inert.** Suppressing `on_clone`
   and `on_wake` inside `src/test_support.rs` turns both of
   `test_support::tests::{probe_waker_routes_clone_wake_and_drop_without_leaking,
   panicking_consuming_wake_still_retires_its_raw_reference}` red. Their
   `catch_unwind` asserts on the returned `Result`, not inside the closure, so
   nothing is swallowed.

### Finding confirmed (fixed on the branch)

**The integration-side twin had no self-coverage, and its consumers cannot
detect it going inert.** `crates/shelterwood/tests/common/waker.rs:38-42`

Every consumer of the twin's `hostile_waker` —
`reply_delivery_waker_proxy.rs:60,127`, `result_delivery_waker_proxy.rs:90`,
`admission_removal_waker_proxy.rs:32,75`, `reply_waker_containment.rs:61` — is
an abort-class regression that judges only "the process survived". Nothing
asserts that the injection happened.

Demonstrated rather than argued: replacing the twin's `drop_probe` body with
`let _ = &probe;` (callback silently skipped) leaves
`reply_delivery_waker_proxy` and `admission_removal_waker_proxy` **fully
green** — the hostile destructor is gone from every one of them and no test
notices. The PR concentrates all four suites' injection behind one shared
vtable while adding self-tests only to the crate-internal copy, so this is the
consolidation's own new failure mode.

Fixed in `9c926fc`: the crate-internal probe's two self-tests are mirrored into
the existing `tests/common/main.rs` fixture target (the established home for
`common/` self-pins), plus a third that pins `hostile_waker`'s named `&'static str`
drop payload. Re-running the same mutation now turns all three red while the
consumer suites stay green — which is exactly the signal that was missing.

### Refuted

- *The shared probe is more permissive than what it replaced.* No: the table
  above is operation-by-operation equivalence, and mutations 1–3 show four of
  the five migrated helpers still go red on the implementation behaviors they
  name.
- *`wake_by_ref` newly firing `on_wake` weakens the old no-op probes.* The
  default `on_wake` is `|| {}`, so every 2-argument `probe_waker` caller keeps
  a literal no-op there.
- *Duplicating `WakerProbe` between the crate and `tests/common` restates a
  test double.* #456 tier 2 asks for exactly that twin; the copies are
  code-identical, and the drift risk is now pinned by `9c926fc`.
- *`ordinal_waker`/`counting_waker` should have been folded onto the probe.*
  They carry per-clone ordinal identity the shared probe structurally cannot
  express; `timer_waker_proxy.rs` was already using them on `main`.
- *`tests/mailbox_future_drop_containment.rs`'s once-only drop latch changed
  scope by moving into a closure.* Both the old field and the new capture live
  behind the single shared `Arc`.

- `probe_waker` (2-arg) is re-exported from `tests/common/mod.rs:26` but no
  integration test uses it directly — only `hostile_waker` does, internally.
  It is invisible today under `#![allow(unused_imports)]`; worth dropping from
  the re-export list if no second consumer appears.
- `mailbox_future_drop_containment::send_timeout_…` and
  `call_acceptance_phase_…` survive mutation 2 because their retirement goes to
  the disposal lane, and they survive mutation 1 too. Their venue is pinned
  elsewhere (`mailbox::timer::tests::drop_glue_…`, `timer_waker_proxy.rs`), but
  the two tests themselves are weaker than their siblings against a
  containment regression. Unrelated to this PR.

### CI

`./tools/dev just ci` green on `9c926fc` (`CARGO_BUILD_JOBS=2`, `nice -n 10`,
`--test-threads=2`), including docs, examples, API reachability, prelude
containment, core-manifest, external-consumer and nixfmt.

PR #466, opened 2026-08-23
Bind provisional memberships to their child IDs

### LINKED ISSUE #436: shelterwood-core: adopt_or_mint_membership accepts a foreign provisional Membership, breaking cross-ID/cross-scope incomparability
External-review claim (2026-08-22 round), **CONFIRMED — and understated** (tokens become fully *equal*, not just comparable). Severity **P3**, not the review's High: façade-unreachable, framework's own single production call site is well-formed by construction. Demonstrated with a live external consumer crate using only safe, non-doc-hidden public core API.

## Mechanism

`shelterwood-core/src/identity.rs:55-72` (`Membership`) promises "Tokens for different child ids or owning scopes are incomparable and return `false`", and the promise is normative (SPEC.md:228-231: "Equality is exact identity"). But `adopt_or_mint_membership` (identity.rs:431-449) accepts *any* copied `Membership` as the provisional for *any* id: the `Entry::Vacant` arm does `entry.insert(FenceCounter::from_fence(provisional.0))` with no check that the provisional was minted for that id or scope.

Demonstrated: seeding two ids (or two scopes) from one provisional makes subsequently minted tokens **equal across domains** (`m_a2 == m_b2 -> true`) and ordered across domains (`supersedes -> true`), breaking the `Eq`/`Hash` key contract that SPEC.md:476/3408-3410 and docs/shutdown-and-resources.md:40 tell users to build registries on. `Incarnation::supersedes` inherits it (gates on membership equality, identity.rs:95).

## Adjudication

- Façade-unreachable, verified: `shelterwood/src/lib.rs:13` keeps `mod identity` private; only `ChildId`/`Membership`/`Incarnation` escape. The single production caller (`plan.rs:394`) passes id and provisional from the same slot.
- **But the framework-only ruling doesn't cleanly cover it**: core marks `capability`, `proxied_sleep`, `waker`, `waker_proxy` `#[doc(hidden)]`; `identity` (core lib.rs:15) is plainly public and documented. The crate-level disclaimer is the only guard, and no CI lane audits core's own public surface (`tools/check-external-consumer.sh` probes the façade root only; justfile:31-38 says core "needs no separate walk"). This finding sits exactly in that acknowledged tooling gap.
- The PR #329 ruling ("Membership is a process-wide key") cuts *against* the code: cross-context comparability is intended as identity equality — precisely what the hole breaks.

## Remedies (cheapest first)

1. `#[doc(hidden)]` the minting authority (`ScopeIdentity`, `MintedMembership`, `MembershipReconciliation`, `IncarnationCounter`, poisoned counters), joining the existing framework-only family, while `ChildId`/`Membership`/`Incarnation` stay plainly public. Aligns the marker with the ruling that already governs it.
2. Structural (invalid-states-unconstructible): make the provisional a non-`Clone` `ProvisionalMembership` carrying the `ChildId` it was minted for, so `adopt_or_mint_membership` reads the id from the token (linear ownership, same shape as the existing `#[must_use] MembershipReconciliation`). Closes the cross-ID half by construction; the cross-scope half additionally needs donor relinquishment, for which remedy 1 is the pragmatic answer.

Also worth recording: this is the second observation this round that core's public surface has no audit lane (see also #428's note that the E0432 probe can't see method-level items).

Part of the external-review verification round tracked on #429.

ralexstokes:
## Summary

- hide the `shelterwood-core` membership-minting authority family from generated documentation while leaving `ChildId`, `Membership`, and `Incarnation` plainly public
- introduce a non-`Clone` `ProvisionalMembership` that carries its minting child ID and is consumed by stable-scope reconciliation
- retain and consume the provisional authority through the production declaration-lowering path
- cover the core API shape, embedded-ID reconciliation, and existing adoption/exhaustion behavior

## Testing

- `./tools/dev cargo test --locked -p shelterwood-core --test identity_authority`
- `./tools/dev cargo test --locked -p shelterwood-core identity::tests`
- `./tools/dev cargo test --locked -p shelterwood plan::tests::failed_override_lowering_evicts_adopted_lineages_from_the_override`
- `./tools/dev just ci`

Closes #436.

### DIFF
--- a/crates/shelterwood-core/src/identity.rs
+++ b/crates/shelterwood-core/src/identity.rs
@@ -306,6 +306,17 @@ pub struct IncarnationCounter {
 }
 
 /// Linear authority to reconcile one declaration-time membership.
+///
+/// The token carries the [`ChildId`] its lineage was minted for, and
+/// [`ScopeIdentity::adopt_or_mint_membership`] reads the id from it rather
+/// than from a second argument: donating a lineage to a *different* id is
+/// therefore unconstructible, not merely unused. Deliberately not `Clone` —
+/// one minted lineage may seed at most one stable id, so a copyable token
+/// would restore the cross-domain donation the binding removes. The
+/// remaining cross-*scope* half (two provisionals for one id, adopted into
+/// two stable scopes) cannot be closed by construction and rides on the
+/// framework-only ruling that keeps this whole minting family
+/// `#[doc(hidden)]`.
 #[derive(Debug)]
 #[doc(hidden)]
 pub struct ProvisionalMembership {
@@ -348,6 +359,14 @@ impl MintedMembership {
                 generations: PoisonedCounter::new(),
             },
         }
+    }
+
+    /// Returns the child id this lineage was minted for.
+    ///
+    /// Keeping the id inside the grant is what lets a member cell derive its
+    /// own id from its identity instead of accepting the two separately.
+    pub fn id(&self) -> &ChildId {
+        &self.provisional.id
     }
 
     #[cfg(any(test, feature = "test-util"))]
@@ -450,6 +469,9 @@ impl ScopeIdentity {
     /// successor. Terminalization evicts the lineage, so an ordinary later
     /// remove-and-re-add or post-restart rebuild donates a fresh, incomparable
     /// identity instead.
+    ///
+    /// The reconciled id comes from the [`ProvisionalMembership`] itself, so
+    /// the lineage can only ever be donated to the id it was minted for.
     pub fn adopt_or_mint_membership(
         &mut self,
         provisional: ProvisionalMembership,
--- a/crates/shelterwood-core/src/identity.rs
+++ b/crates/shelterwood-core/src/identity.rs
@@ -143,6 +143,7 @@ impl Fence {
 /// `u64::MAX` is poison and is never returned. Once the last usable value has
 /// been minted, no successor can be minted.
 #[derive(Clone, Debug)]
+#[doc(hidden)]
 pub struct PoisonedCounter {
     current: u64,
 }
@@ -204,6 +205,7 @@ impl Default for PoisonedCounter {
 
 /// A thread-safe [`PoisonedCounter`] with the same fail-closed domain.
 #[derive(Debug)]
+#[doc(hidden)]
 pub struct AtomicPoisonedCounter(AtomicU64);
 
 impl AtomicPoisonedCounter {
@@ -297,16 +299,26 @@ impl FenceCounter {
 
 /// A membership and the generation counter that can mint only its incarnations.
 #[derive(Debug)]
+#[doc(hidden)]
 pub struct IncarnationCounter {
     membership: Membership,
     generations: PoisonedCounter,
 }
 
+/// Linear authority to reconcile one declaration-time membership.
+#[derive(Debug)]
+#[doc(hidden)]
+pub struct ProvisionalMembership {
+    id: ChildId,
+    membership: Membership,
+}
+
 /// An inseparable identity grant: the counter for an incarnation lineage is
 /// allocated exactly once as its first membership method returns.
 #[derive(Debug)]
+#[doc(hidden)]
 pub struct MintedMembership {
-    membership: Membership,
+    provisional: ProvisionalMembership,
     incarnation_counter: IncarnationCounter,
 }
 
@@ -317,6 +329,7 @@ pub struct MintedMembership {
 /// reconciliation is never optional to consume.
 #[derive(Debug)]
 #[must_use]
+#[doc(hidden)]
 pub enum MembershipReconciliation {
     /// The stable scope adopted the provisional lineage unchanged.
     Adopted,
@@ -327,9 +340,9 @@ pub enum MembershipReconciliation {
 }
 
 impl MintedMembership {
-    fn new(membership: Membership) -> Self {
+    fn new(id: ChildId, membership: Membership) -> Self {
         Self {
-            membership,
+            provisional: ProvisionalMembership { id, membership },
             incarnation_counter: IncarnationCounter {
                 membership,
                 generations: PoisonedCounter::new(),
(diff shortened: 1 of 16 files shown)


ralexstokes:
## Review of #466 (head `7af8c81`), fixes pushed as `a5bf46d`

Reviewed: the full diff, identity minting/reservation/adoption/cancellation end to end
(`ScopeIdentity::mint_membership` → `mint_reserved_slot_inner` → `MemberCell::new` →
`BuilderCore::lower` → `ScopeCell::adopt_or_mint_membership` → `rebase_membership` /
`terminalize`), the lock-rule accounting, and every `MemberCell::new` call site. No
correctness bug found; four fixes to the PR's own new shapes are pushed.

### Both adjudicated remedies from #436 are present — verified, not name-checked

**Remedy 1 (marker).** `#[doc(hidden)]` lands on `PoisonedCounter` (identity.rs:146),
`AtomicPoisonedCounter` (:208), `IncarnationCounter` (:302), `ProvisionalMembership`
(:310), `MintedMembership` (:319), `MembershipReconciliation` (:332) and `ScopeIdentity`
(:399). I enumerated every top-level `pub` item in `identity.rs`: after the change the
only plainly-public ones are `ChildId`, `Membership`, `Incarnation` — exactly the split
the adjudication asked for, and `pub mod identity` correctly stays visible to carry them.

**Remedy 2 (structural).** `ProvisionalMembership` is non-`Clone`, carries the `ChildId`
it was minted for, and `adopt_or_mint_membership` destructures the id out of the token
(identity.rs:465) instead of taking it as a second argument. The cross-ID half is now
unconstructible: the fields are private and the only constructor is the private
`MintedMembership::new`, which is fed the same id the lineage was minted under at both
sites (`mint_membership` :443, and `entry.key().clone()` in the occupied arm :459).

### Verified by mutation (each reverted)

1. Added `#[derive(Clone)]` to `ProvisionalMembership` → `tests/identity_authority.rs`
   fails to compile, E0283 at `assert_not_impl!(ProvisionalMembership: Clone)`. The
   linearity fence has teeth.
2. Diverted the token's id inside `adopt_or_mint_membership`
   (`ChildId::from(format!("{}!", id.as_str()))`) → **both**
   `identity::tests::adopted_membership_orders_a_direct_mint_and_later_rebuild` and the
   new `identity_authority::provisional_membership_selects_its_minting_child_id` FAIL.
   (The cruder mutation — ignoring the id entirely — doesn't even compile: `dead_code`
   fires on the unread field.)

### Findings fixed on the branch (`a5bf46d`)

1. **Invalid states unconstructible, one layer up.** Making the grant carry its id left
   `MemberCell::new(id, identity)` (`cells/member.rs:253` at PR head) accepting the id
   *beside* the grant. Since adoption now reconciles under the token's id, a member whose
   `id()` disagreed with the id it would be adopted under was representable again —
   exactly the shape #436 asked to close by construction. Added
   `MintedMembership::id()`, changed `MemberCell::new` to read the id from the grant, and
   dropped the duplicate argument at all 52 call sites (2 production, the rest fixtures).
   No fixture changed meaning; 956 tests pass.
2. **Rationale not recorded at the new shapes.** `ProvisionalMembership` said only
   "Linear authority to reconcile one declaration-time membership" — nothing about what
   non-`Clone` buys, or that the cross-*scope* half deliberately stays open and rides on
   remedy 1. Both now documented at the type, plus a line on
   `adopt_or_mint_membership` saying the id comes from the token.
3. **Lock-rule ordering was correct but unexplained.** `take_provisional_membership`
   deliberately splits into two statements so the guard is released before the verdict
   can panic (the chained form would keep the temporary alive to end of statement and
   poison the mutex). Comment added so a future "simplification" doesn't collapse it.
4. **Test-file honesty.** `supported_identity_values_remain_plain_public_types` does not
   pin "plain public" — `#[doc(hidden)]` is a documentation marker no test here can
   observe. Added a module comment stating precisely what the file pins and what it
   cannot, referencing the deferred core audit lane.

### Refuted concerns

- **Lock rule / doubled-gate section.** No operation is added there. The new
  `MemberCell::provisional_membership` mutex is a leaf touched only in `new` and
  `take_provisional_membership`, and the sole caller is `BuilderCore::lower`
  (`plan.rs:400`), which holds no observation gate; argument evaluation releases the
  guard *before* `ScopeCell::adopt_or_mint_membership` takes `child_identity`, so no
  nesting is introduced. Inside `child_identity` the added work is a `ChildId`
  (`Arc<str>`) clone and, in the occupied arm, the dropped passed-in key — plain
  framework data with no user destructor, covered by the framework-owned-data and
  Arc-traffic exemptions. `DynamicControl::reserve`'s state-mutex → child-identity-mutex
  edge is unchanged.
- **An id bound twice / a provisional outliving its reservation.** `lower` consumes
  `BuilderCore` by value, `reserve` dedupes ids, so each `MemberCell` is adopted at most
  once; `rebase_membership`'s `OnceLock` already fences a second rebase.
  `take_provisional_membership`'s `expect` is an internal invariant, not a reachable
  panic.
- **Cancelled / failed adoption leaving a bound-but-absent member.** On `Exhausted` the
  token is consumed and `lower` returns; `terminalize` still evicts through
  `adopting_root`, and slots past the failure point never had their token taken, so their
  eviction stays the documented fence-mismatch no-op. Dropping an unconsumed
  `ProvisionalMembership` destroys only a `ChildId` and a `Membership` — no user value,
  so no disposal lane is owed and `#[must_use]` would be wrong here.
- **SPEC §3.2 `Added`/`Removed` pairing, unannounced residents.** Untouched — the PR
  reaches neither residency, admission, nor the observed child set.
- **Vacuous tests.** None. Both new tests assert in the test body; the `assert_not_impl!`
  fence is compile-time. Both were mutation-checked above.
- **Cross-scope half still open.** Two provisionals for one id from one declaration
  builder can still seed two stable scopes (the exhaustion test at identity.rs:740
  constructs exactly that shape). This is the residual the adjudication explicitly
  assigned to remedy 1, not a regression.
- **`assert_not_impl!` duplication** with `crates/shelterwood/tests/api_trait_conformance.rs:25`
  is verbatim and unavoidable — core cannot depend on the façade, and sharing would mean
  `#[macro_export]`ing from core's root.

1. `MemberCell::take_incarnation_counter` (`cells/member.rs:328`) panics with the
   `incarnations` guard still held: the chained `.lock().expect(..).take().expect(..)`
   keeps the temporary alive to end of statement, so the assertion poisons the mutex for
   every later caller — the CLAUDE.md "compute the verdict, release, then panic" rule. I
   scanned production code for the shape and this is the only instance (the other hits
   are all inside `mod tests`). Pre-existing, so left alone.
2. Core still has no public-surface audit lane, so remedy 1 is unpinned by construction —
   already recorded on #436 as a deferred probe. A rustdoc-JSON walk over
   `shelterwood-core` asserting the doc-hidden family would close it.
3. Every `MemberCell` now carries a `Mutex<Option<ProvisionalMembership>>` that only
   builder slots ever take; roots, standalone tasks and dynamically admitted members hold
   it dead. Harmless today, but if it ever matters the token belongs on `SlotCell` beside
   `take_definition`.

CI: `./tools/dev just ci` passes on `a5bf46d` (exit 0, 956 tests).

PR #468, opened 2026-08-23
Retain exits through terminal disposal

### LINKED ISSUE #458: Driver: the retained-exit chain is broken again in handle_construction_disposed and fail_startup
From the review of PR #448 (#422 item 1). #448 closes the raw-`Exit` window in
`handle_exit` by wrapping the classified exit in `RetainedExit` from
classification until terminal disposal or a committed restart publication. The
retained chain then continues through `PendingTerminal.exit`, and is broken
again at the far end of the same lane.

## 1. `handle_construction_disposed` holds the terminal exit raw across the rest of the path

`crates/shelterwood/src/driver/child.rs:1325-1345`.

```rust
let exit = exit.into_exit();
self.terminalize_child(key, exit.clone(), terminal.exited_incarnation, startup);
member.set_terminal_disposal_pending(false);
if self.supervisor.membership_status(key) == MembershipStatus::Removing {
    self.flush_supervisor_effects();
    return;
}
if terminal.startup == StartupDisposition::Aborted
    && !self.supervisor.lifecycle().is_draining()
{
    self.fail_startup(key, exit);
}
if self.children[key].options.retention == crate::Retention::Remove {
    self.prune_terminal(key);
}
```

`terminal.exit` arrives as a `RetainedExit`, and `into_exit()` releases that
protection for the remainder of the function. The raw local then spans
`terminalize_child`, `set_terminal_disposal_pending`, `membership_status`,
`flush_supervisor_effects`, an arena index (`self.children[key]`, which panics
on a missing key) and `prune_terminal`, and is destroyed on the driver thread
at the `Removing` early return or at function exit.

The safety argument is that `terminalize_child` installed an equivalent
retained copy in the member record, making the local drop refcount traffic.
That argument is neither stated nor guaranteed:

- the return value is discarded. `ScopeCell::terminalize_child` returns `false`
  for a *losing* terminalization — the record is already `Terminal(_)`, so no
  record copy of *this* exit is installed. It retires only its own clone
  through `wakes.defer(move || drop(exit))` (`cells/scope.rs:767-774`), which
  submits `dispose_critical`. Whether the driver's raw local or that job holds
  the last handle is then a race, i.e. exactly the strong-count reasoning the
  project rules out.
- an invariant panic in the window (the arena index) unwinds the raw local on
  the driver stack — the class #422 item 1 exists to close.

Open adjudication: is a losing terminalization reachable *here*, with
`pending_terminal` installed? `begin_terminal_disposal` guards on supervisor
state (`joined`/`is_disposing`), not on the member record's stage, and the
member record can be terminalized from cells-layer paths outside the driver
(`discharge_child_terminality`, parent fallback teardown). If it is
unreachable, the fix is still the cheap one below plus a comment; if it is
reachable, the losing exit is currently a user destructor on the driver thread.

Shape: keep `exit` as `RetainedExit`, pass `exit.as_exit().clone()` to
`terminalize_child`, capture its `bool`, and release the guard with
`drop(exit.into_exit())` only on the `true` branches — mirroring
`ScopeCell::terminalize_child`'s own `cells/scope.rs:798-801` comment. Note the
drop-timing consequence: on the losing branch the guard retires through
`dispose_critical`, which is one extra blocking-pool job on a path that has
none today.

## 2. `fail_startup` builds the structured failure raw

`crates/shelterwood/src/driver/startup.rs:117-132`. `StartupFailure { cause:
Child { exit } }` is assembled from the raw `exit` parameter and then

```rust
let Some(state) = supervisor_fail_startup(&mut self.supervisor) else {
    return;
};
```

drops that failure — and the user error inside it — on the driver thread when a
later child loses the startup verdict. `let child = &self.children[key];` above
it can also panic with the raw exit live. Same guard as (1): it is only
refcount traffic while the member record's copy exists.

## 3. Optional: `begin_terminal_disposal` could take the carrier

`child.rs:1220` takes `exit: Exit` and re-retains as its first operation, so
`handle_exit`'s `exit.into_exit()` at the call site opens a (provably
panic-free) raw argument window. Taking `RetainedExit` would make the property
structural rather than argued, at four call sites (`child.rs:736`,
`child.rs:917`, `startup.rs:140`, `tests/shutdown.rs:297`).

Trade-off worth deciding explicitly: it would turn
`refused_terminal_disposal_retires_its_exit_off_the_driver`
(`driver/tests/shutdown.rs:291`) into a type-level tautology, trading a
behavioural pin for a compile-time one.

## Not affected

`discharge_child_terminality` (`child.rs:120-160`) also holds a raw classified
exit across fallible calls, but its `classify_exit_retaining(None,
JoinOutcome::Cancelled, None, Observed)` can never produce `ExitKind::Failed`,
so no user error is at risk there.

ralexstokes:
Closes #458

## Summary
- carry RetainedExit across terminal-disposal entry, construction disposal, and startup arbitration
- release the carrier as raw refcount traffic only after terminal publication wins
- preserve the refused-disposal destructor-thread regression while adapting it to the structural carrier boundary

## Verification
- ./tools/dev cargo test -p shelterwood driver::tests::
- ./tools/dev just ci

### DIFF
--- a/crates/shelterwood/src/driver/child.rs
+++ b/crates/shelterwood/src/driver/child.rs
@@ -1344,26 +1344,29 @@ impl ScopeRuntime {
         member.set_terminal_disposal_pending(false);
         if self.supervisor.membership_status(key) == MembershipStatus::Removing {
             self.flush_supervisor_effects();
-            if terminalized {
-                // The retained member record outlives this early return.
-                drop(exit.into_exit());
+        } else {
+            if terminal.startup == StartupDisposition::Aborted
+                && !self.supervisor.lifecycle().is_draining()
+            {
+                self.fail_startup(key, &exit);
             }
-            return;
+            if self.children[key].options.retention == crate::Retention::Remove {
+                self.prune_terminal(key);
+            }
         }
-        if terminal.startup == StartupDisposition::Aborted
-            && !self.supervisor.lifecycle().is_draining()
-        {
-            self.fail_startup(key, &exit);
-        }
-        if self.children[key].options.retention == crate::Retention::Remove {
-            self.prune_terminal(key);
-        }
+        // Both routes above are fallible, so the guard is released once, here.
+        // A winning publication installed an equivalent retained copy in the
+        // terminal member record, which the `member` handle keeps alive past
+        // this statement, so surrendering the raw copy is refcount traffic.
+        //
+        // A losing publication is unreachable today: #458 established that by
+        // whole-workspace enumeration rather than by any local guard, so no
+        // test pins the other branch and a probe for it would be a
+        // strong-count race by construction. It is retained defensively — the
+        // guard simply falls out of scope, and `RetainedExit::drop` retires
+        // the user error through critical disposal at the cost of one extra
+        // blocking-pool job.
         if terminalized {
-            // The terminal member record owns an equivalent retained copy.
-            // Keep the guard through every later fallible operation above,
-            // then surrender it as raw refcount traffic. A losing publication
-            // leaves the guard intact so its drop retires the exit through
-            // critical disposal.
             drop(exit.into_exit());
         }
     }
--- a/crates/shelterwood/src/driver/child.rs
+++ b/crates/shelterwood/src/driver/child.rs
@@ -733,7 +733,7 @@ impl ScopeRuntime {
             // Exhaustion is a terminal outcome, not an exceptional cleanup
             // path. Join retained-definition disposal before terminality,
             // retention, removal completion, or ordered-scope progression.
-            self.begin_terminal_disposal(key, exit, None, startup);
+            self.begin_terminal_disposal(key, RetainedExit::new(exit), None, startup);
             return;
         };
 
@@ -914,7 +914,12 @@ impl ScopeRuntime {
             // A never-ran child and a child stopped between restart
             // incarnations share the same post-disposal terminal route. Hard
             // shutdown still detaches disposal through `hard_forced` below.
-            self.begin_terminal_disposal(key, exit, None, StartupDisposition::NotAborted);
+            self.begin_terminal_disposal(
+                key,
+                RetainedExit::new(exit),
+                None,
+                StartupDisposition::NotAborted,
+            );
         }
     }
 
@@ -1108,7 +1113,7 @@ impl ScopeRuntime {
             membership_status,
         ) {
             ExitDispatch::Terminal => {
-                self.begin_terminal_disposal(key, exit.into_exit(), Some(incarnation), startup);
+                self.begin_terminal_disposal(key, exit, Some(incarnation), startup);
             }
             ExitDispatch::ScheduleRestart => {
                 let sample =
@@ -1220,14 +1225,15 @@ impl ScopeRuntime {
     pub(super) fn begin_terminal_disposal(
         &mut self,
         key: ChildKey,
-        exit: Exit,
+        exit: RetainedExit,
         exited_incarnation: Option<Incarnation>,
         startup: StartupDisposition,
     ) {
-        // Retain before every refusal and invariant verdict. A failed Exit can
-        // own hostile user error drop glue, so no path may unwind or return it
-        // directly on the driver thread.
-        let mut exit = Some(RetainedExit::new(exit));
+        // Keep the caller's guard through every refusal and invariant verdict.
+        // A failed Exit can own hostile user error drop glue, so no call-site
+        // argument window or local path may unwind or return it directly on
+        // the driver thread.
+        let mut exit = Some(exit);
         if !self.supervisor.contains(key)
             || self.supervisor.is_disposing(key)
             || self.supervisor.joined(key)
@@ -1322,8 +1328,12 @@ impl ScopeRuntime {
         } else {
             StartupDisposition::NotAborted
         };
-        let exit = exit.into_exit();
-        self.terminalize_child(key, exit.clone(), terminal.exited_incarnation, startup);
+        let terminalized = self.terminalize_child(
+            key,
+            exit.as_exit().clone(),
+            terminal.exited_incarnation,
+            startup,
+        );
         // Keep the marker installed until terminal publication has committed.
         // A concurrent shutdown sampler then sees either pending cleanup or a
         // terminal member, never the gap between those two representations.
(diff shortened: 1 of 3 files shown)


ralexstokes:
## Review: retained-exit chain through terminal disposal

Reviewed the full diff against #458's 2026-08-22 adjudication, walked every hop a failed child `Exit` takes from classification to release, and verified the behavioural claims by mutation rather than by name. `./tools/dev just ci` green (exit 0) at `d74a8d0`.

**Verdict: the PR implements the ruling exactly** — item 1+2 (carry `RetainedExit` through `handle_construction_disposed` and `fail_startup`, pass `as_exit().clone()`, capture the `bool`, release only on winning branches) and item 3 (`begin_terminal_disposal` takes the carrier, all 4 call sites). Two documentation/structure fixes pushed; no correctness defect found.

### Hop table

| # | Hop | Site | Status |
|---|-----|------|--------|
| 1 | classification fold | `child.rs:1071` | retained (#448) |
| 2 | dispatch → terminal entry | `child.rs:1113` | **retained (this PR)** — was `exit.into_exit()` |
| 3 | exhaustion / inactive-stop entries | `child.rs:736`, `child.rs:919` | **retained at the call site (this PR)**; the `record().last_exit` read is a raw local covered by the member record's own guard set |
| 4 | refusal + `DisposalStarted` invariant window | `child.rs:1234-1266` | retained (`Option<RetainedExit>`) |
| 5 | parked state | `PendingTerminal.exit`, `child.rs:178` | retained (#448) |
| 6 | disposal-panic fold | `child.rs:1312` | retained (`classify_disposal_panic_retaining`) |
| 7 | publication argument | `child.rs:1331-1336` | raw clone, **caller-retained (this PR)**; callee re-retains at `child.rs:663`, cells layer at `scope.rs:799` |
| 8 | record/mailbox install | `member.rs:723/733/758` | retained |
| 9 | marker clear / status read / effects flush | `child.rs:1344-1346` | **retained (this PR)** |
| 10 | startup arbitration | `startup.rs:117-131` | **retained (this PR)**; the `StartupFailure` raw clone is caller-retained, and scope state / startup result re-retain via `RetainedExit::retain_startup_result` |
| 11 | prune + `self.children[key]` arena index | `child.rs:1352-1353` | **retained (this PR)** |
| 12 | release | `child.rs:1368` | raw drop gated on `terminalized`; a loser keeps the carrier → `dispose_critical` |

No raw hop remains in the lane.

### Lock-rule verification

Every drop point of the carrier, and its destination:

- `begin_terminal_disposal` — three refusal returns, the `is_disposing` assert, `children.get_mut` miss, and the already-pending refusal all unwind/return `Option<RetainedExit>` → `dispose_critical`. Once installed into `PendingTerminal`, the option is `None` and its drop is a no-op.
- `handle_construction_disposed` — panics in `terminalize_child` (`resume_panic` at `child.rs:249`), `set_terminal_disposal_pending`, `membership_status`, `flush_supervisor_effects`, `fail_startup`'s arena index, `self.children[key]`, and `prune_terminal` all unwind the carrier → `dispose_critical`. **Unwind path confirmed closed.**
- Winning release at `child.rs:1368`: provably non-last. `MemberStage::Terminal` has exactly one write site (`member.rs:758`), inside `terminalize_locked`, which sets the mailbox terminal state and the record stage in the same call — so record-not-Terminal ⟹ mailbox-not-Terminal ⟹ `terminalize_locked`'s `losing_exit` is `None` ⟹ *this* exit is what lands in `record.stage`, `record.last_exit` and `refresh_retained_exits()`. Drop order also holds: `member` is bound before `exit`, so the record copy is alive across the explicit release, and neither `prune_terminal` nor `reclaim_child` can destroy the cell while that `Arc` is held.
- Under-lock traffic in `ScopeCell::terminalize_child` (`scope.rs:833/836`) is a clone and a drop of a provably non-last handle while the gate is held — the record copy was installed one line earlier. Unchanged by this PR.

The PR does remove a real (invariant-guarded, hence unreachable) hazard, worth stating: on `main`, a panic in `terminalize_child` *before* publication — `expect("terminalized child remains registered")` at `child.rs:675`, or `expect("a supervised terminal child must remain in parent residency")` at `scope.rs:872` — unwound `handle_construction_disposed`'s raw local with no record copy behind it. That is a user destructor mid-unwind, i.e. a potential double-panic abort. It is now structurally impossible.

### Empirical checks (mutation, on the committed tree)

1. **The adapted refusal test is not a type-level tautology.** Injected `if let Some(exit) = exit.take() { drop(exit.into_exit()); }` into `begin_terminal_disposal`'s refusal branch → `refused_terminal_disposal_retires_its_exit_off_the_driver` **FAILS**: `assertion left != right failed: a refusal cannot destroy the user error on the driver, left: ThreadId(2), right: ThreadId(2)`. The ruling's trade held: the test still probes the disposal artifact, not the type.
2. **The `if terminalized` gate is pinned by nothing.** Replaced it with `main`'s unconditional `drop(exit.into_exit())` at both sites → full `-p shelterwood` suite **751/751 green**. That is consistent with #458's ruling that a losing publication is unreachable, and it is not fixable with a test: on `main` the losing/panicking path always has **two** live handles (the driver's raw local plus the deferred `RetainedExit` already headed to `dispose_critical`), so any destruction-thread probe is exactly the strong-count race the project rules out. I chose to document that rather than land a flaky test — see the fix below.

### Findings fixed on the branch — `d74a8d0`

1. **Duplicated release site with a comment that misstates the reason.** `child.rs` released the guard from two places, and the `Removing` branch justified it with "The retained member record outlives this early return" — the record's *lifetime* is not the argument; ownership of an *equivalent retained copy* is. Folded the early return into an `else` branch so the guard is surrendered from a single point with one correct justification. Control flow is exactly equivalent (the early return had no work after it).
2. **The adjudication was nowhere in the artefact.** #458's whole content is that a losing publication is unreachable *by whole-workspace enumeration, with no local guard*, and that retention is wanted anyway. Neither the code comment, the PR body, nor the one-line commit message said so, so a future reader would read the losing branch as live and would not know why no test covers it. The comment now records the unreachability, that it rests on global enumeration rather than a local guard, that a probe would be a strong-count race, and the accepted cost (one extra blocking-pool job).

### Refuted / checked, no action

- **Honesty of the framing.** Commit message, PR body and code comments claim only structural retention; nothing claims a live bug was fixed, and retention is not dropped on unreachability grounds. Matches the ruling.
- **`fail_startup` taking `&RetainedExit`.** The reference does not itself retain, but it makes the caller's carrier a precondition of the signature, and the raw `StartupFailure` clone is covered for the whole call including the `supervisor_fail_startup` early return and the `&self.children[key]` index. Correct as written.
- **`discharge_child_terminality`** (`child.rs:120-160`) — still genuinely out of scope: `classify_exit_retaining(None, Cancelled, None, Observed)` cannot produce `ExitKind::Failed`.
- **`terminal.startup` vs the recomputed `startup` local** at `child.rs:1348` — intentional per the comment at `child.rs:1316-1324`, not a slip introduced here.
- No `specs/` or `CLAUDE.md` change needed: the lock-rule section already names "pending terminal disposal" under `RetainedExit`.

1. `ScopeRuntime::terminalize_child` (`child.rs:653`) still takes `exit: Exit` and re-retains at `663`. It now has exactly one production caller, and that caller holds a carrier — so the last argument window in this lane could be closed structurally by taking `&RetainedExit`. It would rewrite `missing_terminal_resource_retires_its_exit_before_panicking` (`shutdown.rs:318`) and touch ~8 white-box tests, which is beyond #458's ruling.
2. Same for the cells-layer `ScopeCell::terminalize_child` (`scope.rs:789`), which has many callers.
3. `child.rs:735` and `child.rs:917` read `record().last_exit` into a raw local before wrapping. Safe (the record's guard set covers it), but a `RetainedExit`-returning record accessor would remove the pattern entirely.

PR #469, opened 2026-08-23
Harden send-path caller-waker unwind handling

### LINKED ISSUE #433: Send-path caller-waker unwind gaps: withdraw_send_with resumes over an owned message; SendOperation::poll drops the replacement waker un-sunk
Two external-review claims (2026-08-22 round), both **CONFIRMED by SIGABRT repro** and sharing one shape: a caller-waker destructor allowed to unwind out of a frame that owns either a recovered user message or a half-completed framework transition. The codebase already names and contains this shape at `deadline.rs:187-198`, `deadline.rs:208-217`, `futures.rs:378-385`, and `cell.rs:1367-1394`; these are the gaps in that sweep. Repros committed on local branch `verify-mailbox` (`ccb88f3`, four probes in `mailbox/futures.rs` tests; the two `repro_*` tests are abort-class — remove/gate before merging anything).

## 1. `withdraw_send_with` resumes a waker-drop panic over an owned recovered message (send_timeout abort)

`crates/shelterwood/src/mailbox/futures.rs:414-437`. The function builds `result` from `withdrawal.take_outcome()` (:420), then `withdrawal.finish()` (:435) drops the `Withdrawal`, whose `_waker_effects: WakerEffects` field flushes under a fresh `PanicAccumulator` (`shelterwood-core/src/waker.rs:97-101`) — and that accumulator's drop **resumes** the caught hostile waker-drop panic when the stack is not already unwinding (`panic.rs:118-125, 60-71`). The panic escapes `finish()` while `result` still owns the recovered `SendError<M> { message }`; a panicking message destructor is then a panic-in-drop-during-cleanup → **process abort**. Reachable on the plain product path: `Deadlined::poll` (`mailbox/deadline.rs:219-220`) → `poll_deadlined(..., TimeoutArbitration)` → `withdraw_send_with(self, WithdrawalDisposition::Inline)` (`futures.rs:458`). Only the `Inline` disposition is affected (`call`'s pre-acceptance path uses `Isolated` and is pinned; `short_circuit` has no waker).

Not an accepted outcome by the project's own rule: `Deadlined::poll` contains the identical hazard 40 lines away and documents why (`deadline.rs:187-198` — "the swallowed diagnostic is the accepted cost"; #398 ruling: poll path = contained inline). `withdraw_send_with`'s own doc (:407-413, "a hostile waker destructor can neither divert nor destroy it") is falsified by this path. **Severity: High** (needs hostile waker drop + hostile message drop). Remedy: run the flush under `catch_panic`, route the recovered message to isolated disposal on the panic branch, then contain (matching `deadline.rs:197`).

## 2. `SendOperation::poll` never consumes the replacement waker in its completing arms (abort with NO hostile message)

`crates/shelterwood/src/mailbox/cell.rs:210` — `poll(&self, replacement: Option<Waker>, current: &Waker)`; the `Accepted` (:221-223) and `Terminated` (:224-235) arms never consume `replacement`, so it drops bare at end of frame (post-unlock — no lock-rule violation — but never routed through `WakerEffects`/`WakerAction`). The clone window is `futures.rs:326-348` (waker cloned outside the operation lock). Three reproduced consequences when an accept/terminate lands in that window and the caller waker's `drop` panics:

- **Acceptance half:** poll unwinds with the waker panic instead of returning `Ok(incarnation)` for a send that *was* accepted.
- **Termination half (probe):** `message.take()` at `cell.rs:227` empties the operation state, then the replacement's drop panics — the message is **leaked in the return place** (not returned, not destroyed).
- **Termination half (real task shape) — worse than the review filed:** the unwind runs `SendFuture::drop`, which withdraws a `Parked` send whose operation is now `Terminated { message: None }` and hits the **framework's own unguarded `panic!` at `cell.rs:1337`** ("a terminal operation must retain its message") — panic-in-drop-during-cleanup → **abort with a benign message**. `cell.rs:1333-1339` is the one invariant site in `withdraw` lacking the `std::thread::panicking()` guard its sibling at `cell.rs:1367-1394` applies.

**Severity: High** — the abort variant needs only one hostile ingredient (a panicking waker destructor, squarely in the declared threat model) plus a genuine concurrent race the caller does not control. Remedy: in `SendOperation::poll`, dispose of `replacement` *before* mutating state/taking the message in the completing arms (so a panic leaves the operation consistent and `SendFuture::drop` can still withdraw), or better, route it through the existing `WakerSlot`/`WakerEffects` sink; independently, add the `thread::panicking()` guard at `cell.rs:1333-1339`.

Part of the external-review verification round tracked on #429. Relates to #421 (mailbox structural hardening).

ralexstokes:
Closes #433.

## Summary

- contain inline withdrawal waker-destruction panics before returning a recovered send message
- retire a replacement caller waker that loses the send clone window before observing a completed operation or taking its terminal message
- contain the terminal-message invariant during an existing `SendFuture` unwind
- preserve isolated-withdrawal disposal diagnostics outside the adjudicated inline poll-path containment seam
- add safe regressions for timeout, termination, acceptance, the defensive invariant path, and diagnostic scope; no abort-class probes are retained

## Verification

- `./tools/dev cargo test -p shelterwood mailbox::futures::tests` (19 passed)
- `./tools/dev just ci` (958 nextest tests plus formatting, clippy, doctests, examples, docs, API checks, external consumer, and nixfmt)

### DIFF
--- a/crates/shelterwood/src/mailbox/cell.rs
+++ b/crates/shelterwood/src/mailbox/cell.rs
@@ -211,13 +211,17 @@ impl<M> SendOperation<M> {
         let result = loop {
             let mut state = self.state.lock().expect("send operation mutex poisoned");
             let result = match &mut state.outcome {
-                // The replacement was cloned outside this lock. Completion can
-                // win that window, leaving a caller waker which must be retired
-                // before a terminal message is moved into the return value. A
-                // hostile destructor is contained at this ready seam for the
-                // same reason as reply and timer retirement: after the handoff,
-                // unwinding would destroy a user value in the poll frame.
-                OperationOutcome::Accepted(_) | OperationOutcome::Terminated { .. }
+                // The replacement was cloned outside this lock. Every outcome
+                // but `Waiting` refuses it, so it must be retired before this
+                // frame either moves a terminal message into the return value
+                // or raises one of the defensive diagnostics below. Both would
+                // otherwise destroy the caller waker mid-unwind -- beside a
+                // user value in the first case, inside an existing panic in the
+                // second. A hostile destructor is contained at this ready seam
+                // for the same reason as reply and timer retirement.
+                OperationOutcome::Accepted(_)
+                | OperationOutcome::Terminated { .. }
+                | OperationOutcome::Withdrawn
                     if replacement.is_some() =>
                 {
                     None
@@ -256,14 +260,14 @@ impl<M> SendOperation<M> {
 
             // Stage the clone through the structural waker sink, then drain it
             // with no operation lock held and before re-reading the stable
-            // completing outcome. `flush` catches its vtable panic; taking and
+            // non-waiting outcome. `flush` catches its vtable panic; taking and
             // discarding the payload prevents `PanicAccumulator::drop` from
             // resuming it over the eventual by-value result.
             let mut staged = WakerSlot::default();
             staged.replace(
                 replacement
                     .take()
-                    .expect("a completing clone window retains its replacement waker"),
+                    .expect("a refused clone window retains its replacement waker"),
                 &mut effects,
             );
             staged.take(WakerAction::DropInline, &mut effects);
--- a/crates/shelterwood/src/mailbox/cell.rs
+++ b/crates/shelterwood/src/mailbox/cell.rs
@@ -202,22 +202,33 @@ impl<M> SendOperation<M> {
 
     pub(super) fn poll(
         &self,
-        replacement: Option<std::task::Waker>,
+        mut replacement: Option<std::task::Waker>,
         current: &std::task::Waker,
     ) -> OperationPoll<M> {
         // Effects precede the guard, so even an unwind drops the guard before
         // invoking a displaced RawWaker vtable.
         let mut effects = WakerEffects::default();
-        let result = {
+        let result = loop {
             let mut state = self.state.lock().expect("send operation mutex poisoned");
-            match &mut state.outcome {
+            let result = match &mut state.outcome {
+                // The replacement was cloned outside this lock. Completion can
+                // win that window, leaving a caller waker which must be retired
+                // before a terminal message is moved into the return value. A
+                // hostile destructor is contained at this ready seam for the
+                // same reason as reply and timer retirement: after the handoff,
+                // unwinding would destroy a user value in the poll frame.
+                OperationOutcome::Accepted(_) | OperationOutcome::Terminated { .. }
+                    if replacement.is_some() =>
+                {
+                    None
+                }
                 OperationOutcome::Accepted(incarnation) => {
-                    Ok(OperationPoll::Accepted(*incarnation))
+                    Some(Ok(OperationPoll::Accepted(*incarnation)))
                 }
                 OperationOutcome::Terminated {
                     message,
                     final_incarnation,
-                } => message.take().map_or_else(
+                } => Some(message.take().map_or_else(
                     || Err("a terminal operation retains its message until observed"),
                     |message| {
                         Ok(OperationPoll::Terminated {
@@ -225,19 +236,40 @@ impl<M> SendOperation<M> {
                             final_incarnation: *final_incarnation,
                         })
                     },
-                ),
+                )),
                 OperationOutcome::Waiting { .. } => {
-                    if let Some(replacement) = replacement {
+                    if let Some(replacement) = replacement.take() {
                         state.waker.replace(replacement, &mut effects);
-                        Ok(OperationPoll::Pending)
+                        Some(Ok(OperationPoll::Pending))
                     } else if state.waker.will_wake(current) {
-                        Ok(OperationPoll::Pending)
+                        Some(Ok(OperationPoll::Pending))
                     } else {
-                        Ok(OperationPoll::NeedsWakerClone)
+                        Some(Ok(OperationPoll::NeedsWakerClone))
                     }
                 }
-                OperationOutcome::Withdrawn => Err("a withdrawn send future was polled"),
+                OperationOutcome::Withdrawn => Some(Err("a withdrawn send future was polled")),
+            };
+            drop(state);
+            if let Some(result) = result {
+                break result;
             }
+
+            // Stage the clone through the structural waker sink, then drain it
+            // with no operation lock held and before re-reading the stable
+            // completing outcome. `flush` catches its vtable panic; taking and
+            // discarding the payload prevents `PanicAccumulator::drop` from
+            // resuming it over the eventual by-value result.
+            let mut staged = WakerSlot::default();
+            staged.replace(
+                replacement
+                    .take()
+                    .expect("a completing clone window retains its replacement waker"),
+                &mut effects,
+            );
+            staged.take(WakerAction::DropInline, &mut effects);
+            let mut panics = PanicAccumulator::default();
+            effects.flush(&mut panics);
+            crate::runtime::discard_panic(panics.take());
         };
         drop(effects);
         match result {
@@ -1377,6 +1409,18 @@ impl<M: Send + 'static> MailboxCell<M> {
             Ok(transition) => transition,
             Err(message) => {
                 transaction.finish(());
+                if std::thread::panicking() {
+                    // `withdraw` is reached from `SendFuture` drop glue. If a
+                    // concurrent poll already consumed a terminal message and
+                    // then unwound, repeating this invariant panic would abort
+                    // the process. Preserve the primary unwind and return an
+                    // empty carrier whose remaining effects still flush after
+                    // every framework guard has been released.
+                    return Withdrawal {
+                        outcome: None,
+                        _waker_effects: waker_effects,
+                    };
+                }
                 panic!("{message}");
             }
         };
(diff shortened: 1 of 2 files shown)


ralexstokes:
## Review: send-path caller-waker unwind handling

Reviewed the full diff against `main` (`crates/shelterwood/src/mailbox/{cell,futures}.rs`), #433's two claims and its adjudicated remedy (WakerSlot/WakerEffects sink route + `thread::panicking()` guard at the terminal-message invariant), the lock rule in `CLAUDE.md`, SPEC §15.4/§16.18, and the `deadline.rs` / `DisposingReceiver::close` precedents the containment is modelled on. Everything below is mutation-checked, not name-checked.

**Verdict: the PR implements the ruled shape correctly.** One residual in the same function, fixed on-branch.

### Panic-point audit — every hostile-waker vtable site on the send path

| Site | Lock held? | Verdict |
|---|---|---|
| `context.waker().clone()` in `SendFuture::poll` (futures.rs:481) | none | **Safe.** Unwind leaves the operation `Waiting{message: Some}`; `SendFuture::drop` withdraws and routes the message to isolated disposal, and the withdrawal's waker action is `Dispose`, which cannot panic. |
| `WakerSlot::replace` displaced-waker drop, `Waiting` arm (cell.rs:243) | released at `drop(state)` before `drop(effects)` | **Safe** (pre-existing; venue pinned by `replacing_a_send_waker_runs_its_vtable_outside_the_operation_lock`). Escaping panic owns no user value in that frame. |
| Refused replacement retired in the `Accepted`/`Terminated` arms (cell.rs:216-275) | none — `drop(state)` precedes staging *and* flush | **Fixed by this PR.** Contained via `flush(&mut panics)` + `discard_panic(panics.take())`, so `PanicAccumulator::drop` cannot resume it over the by-value `OperationPoll::Terminated { message }`. Matches `deadline.rs:187-198`'s adjudicated inline poll-path venue. |
| Refused replacement in the **`Withdrawn`** arm | n/a | **Was still un-sunk** — see finding below. Fixed. |
| `withdrawal.finish()` under `Inline`, `withdraw_send_with` (futures.rs:570-577) | none (`MailboxTxn` already finished) | **Fixed by this PR.** `catch_panic` + `discard_panic` around `finish()` keeps `result`'s recovered `SendError<M>` alive. |
| `withdrawal.finish()` under `Isolated` (futures.rs:576, and `CallOperation`'s pre-acceptance path at futures.rs:775) | none | **Safe, diagnostic correctly preserved.** The action is `WakerAction::Dispose`, and `dispose_detached`'s last-resort `job.finish()` (`disposal.rs:313`, `catch_panic(\|\| drop(value))` + `discard_panic`) is total — `MailboxRuntime::dispose` cannot panic in production. The new test injects a panicking `dispose` impl, which is a test-only condition. Containing it here would only swallow a genuine framework bug. |
| `SendFuture::drop` → `withdrawal.finish()` (futures.rs:517) | none | **Safe** (pre-existing, `Dispose` action). |
| `MailboxCell::withdraw` invariant break during an existing `SendFuture` unwind (cell.rs:1416) | mailbox txn finished first | **Fixed by this PR.** Mutation-confirmed abort-class: see below. |

### Finding confirmed and fixed on-branch (264c8e5)

**`SendOperation::poll`'s `Withdrawn` arm still dropped the replacement waker un-sunk, over its own `panic!`** (`crates/shelterwood/src/mailbox/cell.rs:216-224` as merged). The new guard covered `Accepted` and `Terminated` only. With a leftover replacement, `Withdrawn` broke to `Err` → `drop(effects)` (empty) → `panic!("a withdrawn send future was polled")`, and the still-live `replacement: Option<Waker>` was then destroyed by that panic's cleanup — a panic in a destructor during cleanup, i.e. the exact abort geometry this PR removes, and literally the second half of the issue title ("drops the replacement waker un-sunk").

Not reachable on a product path (`SendFutureState::Done` refuses a second poll before `SendOperation::poll` is entered), so this is defensive-path hardening, not a live bug — but the fix is one pattern alternative and it makes the rule uniform: *every* outcome that refuses the clone retires it through the sink before the frame can return a value **or** raise a diagnostic. Also reworded the two comments and the (unreachable) `expect` to say "refused" rather than "completing".

Also on the same commit:
- **`retiring_a_clone_window_replacement_runs_outside_the_operation_lock`** — the new staged retirement's *venue* was unpinned. The existing `OperationLockProbe` cannot be reused here: it `panic!`s inside the destructor, and the PR's new containment swallows exactly that (the [swallowed-assertion class](https://github.com/ralexstokes/shelterwood/issues/418)). The new probe publishes `operation.state.try_lock().is_ok()` into a mutex and the **test body** judges it. This is SPEC §16.18's "lock-held probe … park a send and complete it" oracle for the new path.
- **`a_withdrawn_clone_window_retires_its_replacement_before_the_diagnostic`** — pins the fix above; the send is dropped *inside* the `catch_unwind`, so it also re-exercises the `thread::panicking()` guard.
- Hoisted the four copies of the parked-operation extraction into `parked_operation`.

### Mutation checks (all on the committed tree, reverted with `git checkout --`)

| Mutation | Result |
|---|---|
| Delete the `Accepted \| Terminated if replacement.is_some()` arm | `acceptance_in_the_waker_clone_window_contains_replacement_drop` FAIL; `termination_in_the_waker_clone_window_...` FAIL |
| `if std::thread::panicking()` → `if false` (cell.rs:1416) | `terminal_message_invariant_is_contained_during_an_existing_unwind` **SIGABRT** ("panic in a destructor during cleanup") |
| `take_outcome_if_present` → `Some(self.take_outcome())` | `terminal_message_invariant_..._existing_unwind` **SIGABRT** |
| Drop the `Inline` containment in `withdraw_send_with` | `send_timeout_contains_its_inline_waker_panic_before_returning_the_message` **SIGABRT** |
| Contain the `Isolated` branch too | `isolated_withdrawal_preserves_a_disposal_submission_diagnostic` FAIL |
| Drop `\| OperationOutcome::Withdrawn` from my widened guard | `a_withdrawn_clone_window_retires_its_replacement_before_the_diagnostic` **SIGABRT** |
| `drop(state)` → `let _state = state` (flush under the operation lock) | `retiring_a_clone_window_replacement_runs_outside_the_operation_lock` FAIL |

Every test in the PR is load-bearing; none is vacuous, and no assertion sits inside a swallowed frame — the two hostile destructors only `record_disposal` + `panic!`, and every judgement is made in the test body after the fact.

### Refutations (concerns raised and dismissed with evidence)

- **"The retry loop can spin or re-read a mutated outcome."** No. Iteration 1 consumes `replacement` via `take().expect(...)`, so iteration 2's guard is false and every arm yields `Some` — exactly two iterations, bounded. And `Accepted`/`Terminated`/`Withdrawn` are absorbing: `accept`/`terminate` transition only out of `Waiting`, and `withdraw` needs `&mut SendFuture`, which the polling task holds. The only further mutation is the `message.take()` the re-read itself performs.
- **"The discarded replacement loses a wakeup."** No. It is discarded only on outcomes where `poll` returns `Ready`, so no wake is owed. On `Waiting` the replacement is still installed under the same critical section that publishes a completion, so the pre-existing no-lost-wake argument at futures.rs:452-459 is untouched.
- **"The `Isolated` path has the same abort geometry."** No — `runtime.dispose` is total in production (see the audit table).
- **Lock order.** `SendOperation::poll` takes only the operation mutex and is called only from `SendFuture::poll`, with no mailbox guard above it; `MailboxCell::withdraw` still takes mailbox-then-operation. Neither the new flush nor the new early return runs with any guard held. No change to `MailboxCell::state → SendOperation::state`.
- **specs/.** SPEC §15.4/§16.18 already state exactly what this lands (the waker slot "exposes no operation that returns or replaces a waker without an effects sink"; §16.18's lock-held probe). No spec text is contradicted and none needs amending.

1. `Withdrawal::take_outcome`'s `expect` (cell.rs:1846) is now reachable-in-principle against the new empty carrier: `withdraw_send_with` still uses the infallible accessor, so a `withdraw` on a `thread::panicking()` stack that hits an invariant would double-panic there. Both conjuncts require polling a `SendTimeout`/`CallFuture` from inside a destructor during an unwind *and* an already-broken operation state, so it is not reachable today — but the carrier's `None` state is really a drop-glue-only shape and might be better expressed as such (e.g. `withdraw` returning a venue-typed carrier) than as an `Option` two accessors share.
2. Retiring a waker straight into a sink needs the `WakerSlot::default()` / `replace` / `take` dance (cell.rs:266-273) because `WakerEffects::push` is private. A `WakerEffects::retire(waker, action)` in `shelterwood-core` would collapse eight lines to one and delete the unreachable `expect`; it keeps the structural property (a sink is still mandatory) but widens the doc-hidden core surface, so it's a judgement call worth making deliberately rather than inside this PR.
3. `SendOperation`'s drop glue destroys `state.waker` bare (no sink), which is outside the `WakerSlot` contract. Unreachable today — every path that can destroy the operation empties the slot first (`accept`/`terminate`/`withdraw`), including the new panicking early return — but it is the one remaining un-sunk `Waker` destruction in the module.

### CI

`./tools/dev just ci` green on `264c8e5` (960 nextest tests, fmt, clippy, doctests, examples, docs, API reachability, prelude containment, core manifest, external consumer, nixfmt). Pushed to `agent/issue-433-send-waker-unwind`.

PR #470, opened 2026-08-23
Benchmark public mailbox operations

ralexstokes:
## Summary

- add batched public-API benchmarks for `send`, `try_send`, and `call`
- cover zero-byte, 64-byte, and 4 KiB payloads across bounded queue capacities
- name the send arms for their backpressure regime — `send_buffered` (capacity
  holds the batch, no sender parks) versus `send_backpressured` (capacity below
  the batch, so consumer scheduling is inside the timed region). Capacity picks
  which operation is measured; it is not a tuning knob over one operation.
- compare current-thread and four-worker Tokio runtimes
- exercise one, four, and sixteen short-lived producer tasks, with spawn/join
  cost explicit in the benchmark name and documentation
- stop the clock at the last accepted send. `Mailbox::Queue` is FIFO, so a
  reply barrier inside the timed region would wait for the handler to run the
  whole batch; the barrier instead drains the mailbox *between* iterations, and
  it still runs every iteration because that is what guarantees an empty queue
  at the next iteration's start.
- pre-build the payload buffers outside the timed region. The message value
  itself is necessarily constructed inside it, and `ActorRef::call` boxes its
  constructor — for `payload_4096` that is a 4 KiB allocation and memcpy per
  call. That is the operation's cost, not harness overhead, and the source says
  so at the call site.
- run the group at a coarse-guardrail sampling budget (20 samples, 0.5 s
  warm-up, 2 s measurement) with a 10% noise floor and a 1% significance level,
  keeping the 48-case suite under three minutes inside `just bench`

Stacked on #467.

## Validation

- `./tools/dev just ci`
- `./tools/dev just bench-check`
- `./tools/dev cargo bench --locked --manifest-path tools/benchmarks/Cargo.toml --bench mailbox -- --test`
- `./tools/dev nix build .#checks.x86_64-linux.benchmark-check --no-link --no-update-lock-file`

### DIFF
--- a/tools/benchmarks/Cargo.lock
+++ b/tools/benchmarks/Cargo.lock
@@ -140,6 +140,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tinytemplate",
+ "tokio",
  "walkdir",
 ]
 
@@ -189,6 +190,12 @@ name = "either"
 version = "1.18.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "252afb9ae5eaa683babdc6a068b3f5726eb19e05070c731f9b2a23a7c3e8ed34"
+
+[[package]]
+name = "fastrand"
+version = "2.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "da7c62ceae207dd37ea5b845da6a0696c799f85e97da1ab5b7910be3c1c80223"
 
 [[package]]
 name = "find-msvc-tools"
@@ -460,11 +467,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "shelterwood"
+version = "0.1.0"
+dependencies = [
+ "shelterwood-core",
+ "shelterwood-runtime",
+ "thiserror",
+]
+
+[[package]]
 name = "shelterwood-benchmarks"
 version = "0.0.0"
 dependencies = [
  "criterion",
+ "shelterwood",
  "shelterwood-core",
+ "tokio",
 ]
 
 [[package]]
@@ -472,6 +490,15 @@ name = "shelterwood-core"
 version = "0.1.0"
 dependencies = [
  "thiserror",
+]
+
+[[package]]
+name = "shelterwood-runtime"
+version = "0.1.0"
+dependencies = [
+ "fastrand",
+ "shelterwood-core",
+ "tokio",
 ]
 
 [[package]]
@@ -536,6 +563,27 @@ checksum = "be4d6b5f19ff7664e8c98d03e2139cb510db9b0a60b55f8e8709b689d939b6bc"
 dependencies = [
  "serde",
  "serde_json",
+]
+
+[[package]]
+name = "tokio"
+version = "1.53.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "202caea871b69668250d242070849eb495be178ed697a3e98aebce5bc81a0bed"
+dependencies = [
+ "pin-project-lite",
+ "tokio-macros",
+]
+
+[[package]]
+name = "tokio-macros"
+version = "2.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "78773a2a397f451582ce068015985c33193cf6dea8b74d2a639fe457b2f07b0e"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 3.0.3",
 ]
 
 [[package]]
--- a/tools/benchmarks/Cargo.toml
+++ b/tools/benchmarks/Cargo.toml
@@ -13,6 +13,12 @@ publish = false
 name = "core"
 harness = false
 
+[[bench]]
+name = "mailbox"
+harness = false
+
 [dev-dependencies]
-criterion = "=0.8.2"
+criterion = { version = "=0.8.2", features = ["async_tokio"] }
+shelterwood = { path = "../../crates/shelterwood" }
 shelterwood-core = { path = "../../crates/shelterwood-core" }
+tokio = { version = "=1.53.1", features = ["rt", "rt-multi-thread", "time"] }
--- a/tools/benchmarks/README.md
+++ b/tools/benchmarks/README.md
@@ -23,6 +23,17 @@ Criterion accepts a name filter after `--`. For example:
   --manifest-path tools/benchmarks/Cargo.toml -- core/deadline
 ```
 
+The suites separate deterministic structural costs from executor-backed public
+operations:
+
+- `core` covers the supervision reducer and deadline queue.
+- `mailbox` covers batched send, fail-fast send, request/reply, and concurrent
+  producer tasks over both current-thread and four-worker Tokio runtimes.
+
+`concurrent_send_tasks` deliberately includes task spawn and join overhead. It
+represents the common application shape of short-lived producer tasks rather
+than claiming to isolate mailbox mutex contention alone.
+
 The core supervisor `startup` and `drain` cases apply one child's transition
 sequence and then settle before advancing to the next child. They deliberately
 stress incremental, level-triggered settlement; they do not model the driver's
--- a/tools/benchmarks/README.md
+++ b/tools/benchmarks/README.md
@@ -29,10 +29,35 @@ operations:
 - `core` covers the supervision reducer and deadline queue.
 - `mailbox` covers batched send, fail-fast send, request/reply, and concurrent
   producer tasks over both current-thread and four-worker Tokio runtimes.
+  Four details govern how to read it:
+  - The timed region of every batched send case ends at the last accepted
+    send. `Mailbox::Queue` is FIFO, so a reply barrier inside that region
+    would have to wait for the handler to run the whole batch, reporting
+    `enqueue + handler + wakeups + a round trip` under the name of a send and
+    burying the difference between two send APIs in the shared drain. The
+    barrier still runs once per iteration, between measurements, so every
+    iteration starts from an empty queue.
+  - Capacity selects which operation runs rather than tuning one operation,
+    so the arms are named for their regime. `send_buffered` holds the whole
+    batch and parks no sender; `send_backpressured` parks every send past its
+    capacity, so its timed region legitimately includes consumer scheduling.
+    The two are not comparable, and the backpressured arms are the noisy ones.
+  - `call` is measured whole. Boxing the message constructor, creating the
+    reply channel, and arming and retiring a real deadline wheel entry are
+    inside the timed region because the operation cannot be performed without
+    them; no deadline ever expires there.
+  - The group runs at a reduced sampling budget (20 samples, 0.5 s warm-up,
+    2 s measurement) with a 10% noise floor and a 1% significance level. That
+    keeps the 48-case suite near three minutes inside `just bench` and keeps
+    routine spread on the backpressured arms out of the change reports.
 
 `concurrent_send_tasks` deliberately includes task spawn and join overhead. It
 represents the common application shape of short-lived producer tasks rather
-than claiming to isolate mailbox mutex contention alone.
+than claiming to isolate mailbox mutex contention alone. Under the
+`current_thread` group there is no real contention to isolate: the producers
+interleave cooperatively on a single thread, and only the `multi_thread_4`
+group exercises parallel senders. Its mailbox capacity sits below the batch
+size in both groups, so it is a backpressured arm as well.
 
 The core supervisor `startup` and `drain` cases apply one child's transition
 sequence and then settle before advancing to the next child. They deliberately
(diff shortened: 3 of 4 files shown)


ralexstokes:
## Measurement snapshot

Criterion baseline `ryzen-9955hx-2026-08-23`, collected from cumulative commit
`4625d66` on an AMD Ryzen 9 9955HX with CPUs 0-3 pinned. The run used `rustc
1.97.1`, 1 s warm-up, 2 s measurement, and 30 samples. The host was shared and
using the `powersave` governor, so treat these as comparative data from one run,
not universal thresholds.

Each point below is throughput for a fully drained 256-message batch:

| Runtime / workload | 64-byte payload | 4 KiB payload |
| --- | ---: | ---: |
| current-thread `send`, capacity 64 | 3.325 M msg/s | 0.907 M msg/s |
| current-thread `try_send` | 3.726 M msg/s | 1.071 M msg/s |
| current-thread `call` | 0.570 M call/s | 0.340 M call/s |
| 4-worker `send`, capacity 64 | 3.452 M msg/s | 1.494 M msg/s |
| 4-worker `try_send` | 3.919 M msg/s | 1.386 M msg/s |
| 4-worker `call` | 0.256 M call/s | 0.120 M call/s |

The 4 KiB multi-thread call case was noisy: its batch estimate was 2.13 ms with
a 1.57-2.68 ms 95% interval. I would not use that one as a hard baseline yet.

For 64-byte payloads, `concurrent_send_tasks` (including task spawn/join) was:

| Producer tasks | Current-thread | 4-worker |
| ---: | ---: | ---: |
| 1 | 3.295 M msg/s | 3.207 M msg/s |
| 4 | 3.247 M msg/s | 2.750 M msg/s |
| 16 | 3.091 M msg/s | 2.515 M msg/s |

Two useful signals from this machine: cross-thread scheduling dominates tiny
call/reply work, while the multi-thread runtime improves buffered 4 KiB sends
by about 65%. Short-lived producer-task overhead also becomes visible as the
producer count rises, especially on the multi-thread runtime.

ralexstokes:
## Review

Compiles, passes nightly clippy `-D warnings` and the fmt gate, and all 48 cases pass `--test`. No crash-class or silent-drop defect — `try_send` cannot currently see a `Full` and would panic loudly if it did. The findings are measurement validity and doc accuracy, and the headline one is backed by measured numbers rather than inspection.

Measurements below: `taskset -c 0-3 nice -n 15`, short samples, otherwise-idle machine.

### 1. `send` / `try_send` / `concurrent_send_tasks` do not measure the named operation — the drain dominates. Medium-high
`benches/mailbox.rs:118`, `:142`, `:199`. Every timed closure ends with `black_box(barrier(actor).await)`, and `Mailbox::Queue` is FIFO (`shelterwood-core/src/policy.rs:803`), so the barrier reply cannot arrive until the handler has run for all 256 messages. The timed region is `256 × enqueue + 256 × handler + wakeups + one full round trip`, not the send path.

The numbers confirm the mechanism: on `multi_thread_4/payload_0`, **`try_send` (64.8 µs) and `send/1024` (64.9 µs) are indistinguishable** — the entire difference between a parking-capable send future and a fail-fast enqueue disappears into the shared drain. On `current_thread/payload_4096` the delta is 42 µs out of 287 µs, so the named operation is ~15% of the reported number and a 2× regression in `try_send`'s enqueue path would move the figure by single-digit percent.

This is the contract line at README.md:47 ("Each benchmark name identifies the exact work inside the timed region"). Either the names and README should say batch round-trip throughput, or the drain should leave the timed region — `iter_custom` stopping the clock at the last accepted send and barriering afterward.

### 2. The `capacity` parameter silently changes what `send` measures — medium
`mailbox.rs:104` sweeps capacity `[1, 64, 1024]` against a fixed `BATCH = 256`. At capacity 1 every send after the first parks and needs a wake; at 1024 none park. Measured `multi_thread_4/payload_0`: `send/1` = 575 µs vs `send/1024` = 64.9 µs, an **8.8× spread**. `mailbox/*/send/1` is a park/wake scheduling benchmark wearing the name `send`, and neither the file nor README.md:31-33 says so.

Related, and a practical problem: the backpressured cases are the noisy ones — `send/64` came out `[77.0, 98.2, 127.0] µs`, a 65% spread. At criterion defaults with no `significance_level`/`noise_threshold` tuning, these two will generate frequent false regression reports.

### 3. The barrier's own cost is charged to the 256 elements, with a runtime-dependent bias — low-medium
`mailbox.rs:92-98`. Each barrier is a full `ActorRef::call`: boxed constructor, reply channel, and a real 30 s deadline timer armed and retired (`crates/shelterwood/src/mailbox/futures.rs:285-305` → `deadline.rs:68-135`, constructing a `ProxiedSleep` and inserting/removing a wheel entry). One call is ~4 µs on `multi_thread_4/payload_0`, so the barrier is ~6% of that group's `try_send` figure but only ~1.2% of `current_thread/payload_4096`'s — a fixed overhead varying by 5× across groups README.md:50 invites you to read side by side. Arming 257 real timers per iteration is also in some tension with README.md:53.

### 4. "Construct messages outside timed regions" does not hold for `call` — low-medium
`mailbox.rs:159-162`. Only the payloads are built in setup. The message is constructed inside the timed region and `ActorRef::call` boxes the constructor (`futures.rs:299`), so for `payload_4096` each of the 256 calls pays a ~4 KiB heap allocation plus memcpy inside timing. That is inherent to the `call` signature and arguably legitimate to measure, but the PR body's claim is not accurate as written, and there is no allocation-free comparison point (`reply_channel` + `send` + `recv`) to separate the API's packaging cost from the round trip.

### 5. `try_send`'s no-`Full` invariant has one slot of headroom and is undocumented — low
`mailbox.rs:127-139`. `try_capacity = BATCH + 1 = 257` and `.expect(...)` panics on `Full`. This is currently safe — occupancy is `state.queue.len()` and the slot frees at dequeue, not handler completion (`crates/shelterwood/src/mailbox/cell.rs:1653`, `:1731`), and the barrier guarantees an empty queue at iteration start, so peak occupancy is 256. Verified empirically over ~5,800 iterations with no panic. But the invariant chain appears nowhere in the file and the margin is one slot. Note also this is the only configuration in the suite without backpressure, so `try_send` is not a clean fail-fast-vs-parking A/B against `send/1` or `send/64`.

### 6. `just bench` wall time — low
3 payload sizes × 2 runtimes × 8 cases = 48 benchmarks at criterion defaults, roughly +7 min on top of `core`. Not a CI concern (`justfile:78` runs only `bench-check`, which is `--no-run`), but README.md:11 presents `just bench` as the ordinary entry point.

### Clean
- **Payload allocation is genuinely outside timing.** Verified against criterion 0.8.2 source (`bencher.rs:604-650`): async `iter_batched` calls `setup()` before `measurement.start()` on both paths. Only leakage is the input `Vec`'s buffer deallocation when `for payload in messages` drains its `IntoIter` — 1 MiB for `payload_4096`, negligible against 245-287 µs.
- **The reply barrier does guarantee full drain.** FIFO plus barrier-after-all-sends; for `concurrent_send_tasks` all producer handles are joined at :196-198 first. The timer cannot stop with messages queued.
- **Throughput declarations are right.** Every one of the 8 benchmarks processes exactly 256 messages, including `call` and `concurrent_send_tasks` (256 divides by 1, 4, 16 so no producer is short-changed).
- **No runtime-setup leakage.** Builder, `Tree::new`/`add_actor`/`spawn`/`wait_started` and `shutdown` are all outside every timed closure, and teardown genuinely joins — `System::shutdown` awaits `shutdown_scope` and `join_driver()` (`driver.rs:94-102`), so README.md:49 holds and no sample leaks into the next.
- **`black_box` coverage adequate.** `send`'s `Incarnation` at :113-116 is dropped un-boxed, but the send crosses an `Arc<MailboxCell>` and a mutex in another crate with no LTO — inconsistent, not defective.
- **`concurrent_send_tasks` disclosure is substantially right** (README.md:35-37 plus the `_tasks` suffix). One caveat worth a line: under `current_thread` it exercises no real contention, only cooperative interleaving on one thread.

ralexstokes:
## Review fixes (`benches/mailbox.rs` + its README bullet)

Pushed as `cbcd052`. Every finding confirmed against the source before changing
anything; per-finding disposition below.

### 1. The drain dominates the named operation — **fixed, and measured**

Confirmed. `Mailbox::Queue` is FIFO (`shelterwood_core::policy::Mailbox::Queue`),
so the barrier reply cannot be produced until the handler has run every message
ahead of it.

The batched cases (`send`, `try_send`, `concurrent_send_tasks`) now run through
`iter_custom`: a shared `timed_batches` helper builds the input, times only the
batch, and then drains with an untimed barrier. The barrier still runs once per
iteration — it is what makes the next iteration start from an empty queue.
`call` keeps `iter_batched`, because a `call` batch has nothing to exclude.

Indicative spot-checks, **not** a baseline — shared machine, `taskset -c 4-7
nice -n 10`, before/after collected back to back:

| case | before | after | drain share removed |
| --- | ---: | ---: | ---: |
| `multi_thread_4/payload_0` send @1024 | 66.7 µs | 60.5 µs | 9% |
| `multi_thread_4/payload_0` `try_send` | 65.3 µs | 47.9 µs | 27% |
| `current_thread/payload_4096` send @1024 | 279.7 µs | 127.0 µs | 55% |
| `current_thread/payload_4096` `try_send` | 237.1 µs | 85.4 µs | 64% |

The separation the finding asked for is there: `try_send` vs buffered `send` was
2% apart on `multi_thread_4/payload_0` (65.3 vs 66.7) and is now 26% apart
(47.9 vs 60.5); on `current_thread/payload_4096` the ratio goes from 1.18x to
1.49x. The asymmetry is itself the evidence — `try_send` returns immediately and
was paying for the whole drain, while an awaited `send` overlaps more of it.

### 2. Capacity silently changes what `send` measures — **fixed**

Confirmed and named apart. Arms below `BATCH` are `send_backpressured/{1,64}`;
`send_buffered/1024` holds the whole batch and parks no sender. The README says
the two are not comparable and why. Group change-detection is retuned for the
noisy arms: `significance_level(0.01)`, `noise_threshold(0.10)`.

### 3. Barrier cost charged to the elements — **subsumed; residue named**

With finding 1 fixed, the barrier is outside every timed region, and `call` has
no barrier at all, so its cost is now 0% of every reported figure rather than
6%/1.2%. Residue: `call` still arms and retires one real wheel entry per call
inside its timed region (`Deadlined::poll` constructs a `ProxiedSleep` on first
poll for any non-zero budget). That is inherent to the operation — a real caller
passes a deadline — and it is now stated at the call site and in the README. It
does not conflict with the "real sleeps and deadline expiry" contract line: no
deadline is ever awaited to expiry here, the entry is armed and retired.

### 4. PR body claim about message construction — **corrected**

Confirmed false as written. Body rewritten: payload *buffers* are pre-built; the
message value is necessarily constructed inside the timed region and
`ActorRef::call` boxes the constructor (a 4 KiB allocation plus memcpy per call
at `payload_4096`). A source comment on the `call` case now lists everything the
signature forces into the region.

### 5. `try_send`'s one-slot headroom — **documented and re-derived**

The invariant chain is written at the call site: occupancy is the queue length
checked at accept (`accept_locked`), the slot frees at *dequeue* — `take_next`
pops the envelope before the handler runs, and `promote_waiter_queue` recomputes
availability from the queue length — and the barrier guarantees an empty queue at
iteration start. Re-derived against the finding-1 restructure: `timed_batches`
still drains every iteration, so peak occupancy is still `BATCH` = 256 against
`try_capacity` = 257. The comment says a restructure that stops draining
invalidates it.

### 6. `just bench` wall time — **fixed**

20 samples, 0.5 s warm-up, 2 s measurement for the mailbox groups. Measured: the
8-case `current_thread/payload_0` group now runs in 26 s, so the 48-case suite is
about 2.6 min against roughly 7 min at Criterion defaults. Documented in the
README bullet and at the constants.

### 7. `current_thread` `concurrent_send_tasks` has no real contention — **noted**

Added to the paragraph I own: under `current_thread` the producers interleave
cooperatively on one thread and only `multi_thread_4` exercises parallel senders.
Also noted there that its capacity (64) is below the batch size, so it is a
backpressured arm in both groups.

### Superseded figures in the measurement snapshot

The [measurement snapshot comment](https://github.com/ralexstokes/shelterwood/pull/470#issuecomment-5383387201)
above is superseded for the mailbox rows and was not re-collected:

- **`send`, capacity 64** rows (both runtimes, both payloads): the benchmark is
  renamed `send_backpressured/64` and its timed region no longer contains the
  drain. Both the name and the number are stale.
- **`try_send`** rows (both runtimes, both payloads): timed region rescoped; the
  spot-checks above show 27–64% of those figures was drain.
- **`call`** rows: still measuring the same thing, but collected under the old
  sampling configuration (30 samples / 1 s warm-up) rather than the new one.

The next published baseline should be collected fresh from this head on a quiet
machine; nothing above is a substitute for it.

### Verification

- `./tools/dev just bench-check` — clean (nightly clippy `-D warnings`, compile)
- `cargo bench ... --bench mailbox -- --test` — all 48 cases pass
- `cargo +nightly fmt --manifest-path tools/benchmarks/Cargo.toml -- --check` — clean

All runs `taskset -c 4-7 nice -n 10` on a shared machine.

PR #471, opened 2026-08-23
Make retained-exit surrender transactional

### LINKED ISSUE #455: RetainedExit::into_exit's co-owner proof is conventional: 19 surrender sites, one of which was wrong
From the PR #445 review (issue #419 round).

`RetainedExit::into_exit()` converts a guard back into a raw `Exit`, which
downgrades a *guaranteed* isolated destruction into ordinary inline drop
timing. Every call site therefore carries a hand-written proof that another
owner exists, in a comment:

- `cells/observe.rs:160` "The public event owns a raw clone corresponding to
  every guard, so retiring these copies inline is provably refcount-only."
- `cells/member.rs:608` "The deferred transition proves this clone cannot be
  last."
- `cells/scope.rs:834` "The terminal member record now owns the equivalent
  retained copy."
- ...and 16 more.

Auditing all 19 during the #419 review surfaced a single, uniform structural
rule the comments never state:

> A surrender is safe exactly when it runs **before** the `ObservationTxn`
> commits. The co-owner is either a record the same critical section just wrote
> or a `RetainedExit` still sitting in the transaction's deferred-effect queue.
> A surrender that runs **after** commit has no such co-owner: the deferred
> effect has already run and handed its clone to `runtime::dispose_critical`,
> which is a *concurrent* job. Racing that job for the last reference is how a
> user destructor gets back onto the framework thread.

`ScopeCell::publish_drain` was the only post-commit surrender in the tree, and
it was the only wrong one (fixed on that PR in 23a39df; Codex flagged the same
site independently). Nothing prevents the next one — `into_exit()` is a plain
method with no notion of where in a transaction's lifetime it is being called.

Shapes worth weighing:

1. Make the co-owner an argument rather than a comment — e.g. a
   `surrender_beside(&self, coowner: &Exit)` that debug-asserts `Arc::ptr_eq`
   on the failed payload, so the proof is at least checkable in debug builds.
2. Give the transaction the surrender: `txn.surrender(guards)` queues the raw
   drop into the same effect list, which makes "before commit" structural
   instead of positional, and removes the post-commit shape entirely.
3. Keep `into_exit()` only for the genuine hand-offs to user-owned values
   (`RetainedStopReason::into_public`, `RetainedLifecycleEvent::into_public`,
   `driver.rs:460`) and give the framework-internal surrenders a distinct,
   narrower name so the two classes stop sharing one method.

Low priority: the audit found no other live violation. It is filed because the
one violation was invisible to review for a full round and is not detectable by
any test — the race is heavily biased toward the safe ordering, so a
`ThreadProbe` pin passes on the broken code.

ralexstokes:
Closes #455.

## Summary

- Add `ObservationTxn::surrender`, the only framework-internal raw-release path for `RetainedExit`. It queues retained guards into the transaction's deferred-effect list, releases the observation gate first, and prioritizes surrenders ahead of ordinary effects that can hand a co-owner to concurrent disposal.
- Rename the former general `RetainedExit::into_exit` escape hatch to `into_user_owned` and leave it at only the three adjudicated production handoffs: `RetainedStopReason::into_public`, `RetainedLifecycleEvent::into_public`, and the root completion handoff in `driver.rs`.
- Propagate surrender collection through member/scope guard refresh, recursive snapshot construction and commit-time snapshot producers, so deduplicated guards cannot be released outside an open observation transaction.
- Carry `RetainedExit` itself through the driver terminal-disposal and terminal-publication seams. The newer restart-publication path now moves its guard into the cell transaction instead of surrendering after that transaction returns.
- Document the structural rule in `AGENTS.md` and add mutation-sensitive tests for concurrent-owner ordering plus prioritized surrender draining after unlock during unwind and a hostile surrender effect.

## Call-site audit

All 21 former `into_exit()` sites (the adjudicated 19 plus two later driver additions) were classified and migrated:

- **4 genuine user-owned handoffs:** 3 production sites use `into_user_owned`; the fourth is their drop-timing unit test.
- **13 framework-internal raw releases:** record/snapshot deduplication, direct snapshots, member transitions, scope startup/state/terminal publication, and restart publication now route through `ObservationTxn::surrender`.
- **3 framework carrier transfers:** driver terminalization, terminal-disposal entry, and post-disposal routing keep `RetainedExit` intact rather than extracting a raw `Exit` between framework layers.
- **1 retained fold:** disposal-panic classification extracts privately inside `cells::retained` and immediately re-wraps the selected/discarded halves; it is neither a surrender nor a user handoff.

`rg` now finds no `into_exit()` call or definition. `into_user_owned()` has exactly the three production handoffs named by the adjudication plus its unit test.

## Verification

- `./tools/dev cargo test --locked -p shelterwood --lib observation_txn_`
- `./tools/dev just ci` (956 nextest tests plus formatting, clippy, doctests, examples, docs, API reachability, manifest, external-consumer, and Nix formatting checks)

Fresh adversarial review added the unwind/panic-continuation regression in c26070c and found no remaining raw-conversion or lock-order defect.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,11 +18,14 @@ Two types implement the rule and are the shapes to reach for:
 
 - **`ObservationTxn`** (`crates/shelterwood/src/cells/gate.rs`) holds the
   observation-gate guard plus a deferred-effect list. `defer`/`pulse` queue
-  work; `commit` drops the guard *then* runs the queue through a
-  `PanicAccumulator`. Its `Drop` runs the same path during an unwind, so a
-  poisoned transaction cannot strand already-committed wakes. Every retained
-  control-plane writer takes the token, which makes an out-of-transaction
-  mutation unavailable by construction.
+  work; `surrender` queues framework-internal `RetainedExit` raw drops at the
+  front, so they run after unlock but before an ordinary effect can hand their
+  co-owner to concurrent disposal. `commit` drops the guard *then* runs the
+  queue through a `PanicAccumulator`. Its `Drop` runs the same path during an
+  unwind, so a poisoned transaction cannot strand already-committed wakes.
+  Every retained control-plane writer takes the token, which makes both an
+  out-of-transaction mutation and a post-commit surrender unavailable by
+  construction.
 - **`MailboxTxn`** (`crates/shelterwood/src/mailbox/cell.rs`) is the same idea for
   every mutable mailbox transition. It owns the state guard beside a
   `MailboxEffects` sink that collects signal pulses, waker wake/drop actions,
@@ -68,6 +71,11 @@ rests on:
   `reconcile_recorded_outcomes_retaining` and
   `classify_disposal_panic_retaining`; the raw `shelterwood-core` folds hand
   both halves back and are for core's own tests.
+  `RetainedExit::into_user_owned` is deliberately narrower than surrender: it
+  appears only at the three genuine public-ownership handoffs
+  (`RetainedStopReason::into_public`, `RetainedLifecycleEvent::into_public`,
+  and the root completion handoff in `driver.rs`). Framework-internal raw
+  refcount drops go through `ObservationTxn::surrender` instead.
   `RetainedExitResult` closes the earlier raw-incarnation window: the completed
   callback result stays in that carrier across the fallible teardown epilogue,
   so an epilogue panic transfers a failed result to critical disposal instead
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,11 +71,15 @@ rests on:
   `reconcile_recorded_outcomes_retaining` and
   `classify_disposal_panic_retaining`; the raw `shelterwood-core` folds hand
   both halves back and are for core's own tests.
-  `RetainedExit::into_user_owned` is deliberately narrower than surrender: it
-  appears only at the three genuine public-ownership handoffs
-  (`RetainedStopReason::into_public`, `RetainedLifecycleEvent::into_public`,
-  and the root completion handoff in `driver.rs`). Framework-internal raw
-  refcount drops go through `ObservationTxn::surrender` instead.
+  `RetainedExit::into_user_owned` is deliberately narrower than surrender, and
+  `pub(in crate::cells)` is what keeps it narrow: no driver-layer caller can
+  extract a raw `Exit` from a carrier at all, so a carrier crossing a driver
+  seam stays a carrier. Its two remaining users are
+  `RetainedLifecycleEvent::into_public`, the one genuine hand-off to a
+  user-owned value, and `RetainedStopReason::into_public`, whose two call
+  sites release a framework copy beside an installed co-owner and therefore
+  still rest on the older conventional proof. Every other framework-internal
+  raw refcount drop goes through `ObservationTxn::surrender`.
   `RetainedExitResult` closes the earlier raw-incarnation window: the completed
   callback result stays in that carrier across the fallible teardown epilogue,
   so an epilogue panic transfers a failed result to critical disposal instead
--- a/crates/shelterwood/src/cells/gate.rs
+++ b/crates/shelterwood/src/cells/gate.rs
@@ -330,6 +330,57 @@ mod tests {
     }
 
     #[test]
+    fn observation_txn_unwind_drains_panicking_surrender_prefix_after_unlock() {
+        let gate = ObservationGate::new();
+        let observed = Arc::new(Mutex::new(Vec::new()));
+        let payload = catch_unwind(AssertUnwindSafe({
+            let observed = Arc::clone(&observed);
+            let gate = gate.clone();
+            move || {
+                let mut txn = ObservationTxn::new(&gate, gate.lock());
+                let ordinary = Arc::clone(&observed);
+                let ordinary_gate = gate.clone();
+                txn.defer(move || {
+                    ordinary
+                        .lock()
+                        .expect("probe mutex remains healthy")
+                        .push((3, ordinary_gate.is_held()));
+                });
+
+                let first = Arc::clone(&observed);
+                let first_gate = gate.clone();
+                txn.defer_surrender(move || {
+                    first
+                        .lock()
+                        .expect("probe mutex remains healthy")
+                        .push((1, first_gate.is_held()));
+                    std::panic::panic_any("hostile surrender effect");
+                });
+
+                let second = Arc::clone(&observed);
+                let second_gate = gate.clone();
+                txn.defer_surrender(move || {
+                    second
+                        .lock()
+                        .expect("probe mutex remains healthy")
+                        .push((2, second_gate.is_held()));
+                });
+
+                std::panic::panic_any("primary panic");
+            }
+        }))
+        .expect_err("the original unwind reaches its boundary");
+
+        assert_eq!(payload.downcast_ref::<&str>(), Some(&"primary panic"));
+        assert_eq!(
+            *observed.lock().expect("probe mutex remains healthy"),
+            [(1, false), (2, false), (3, false)],
+            "surrender effects stay ahead of the ordinary suffix, all run after unlock, and a \
+             hostile surrender cannot strand later effects during unwind"
+        );
+    }
+
+    #[test]
     fn observation_txn_drop_runs_later_pulses_after_a_hostile_waker() {
         let gate = ObservationGate::new();
         let (hostile_sender, mut hostile_receiver) = runtime::watch(());
--- a/crates/shelterwood/src/cells/gate.rs
+++ b/crates/shelterwood/src/cells/gate.rs
@@ -208,6 +208,7 @@ impl MailboxEffectSink for ObservationTxn<'_> {
 #[cfg(test)]
 mod tests {
     use std::{
+        fmt,
         future::Future,
         panic::{AssertUnwindSafe, catch_unwind},
         sync::{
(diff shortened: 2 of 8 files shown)


ralexstokes:
## Review of #471 (head c26070c, fixes pushed as a96642f)

Reviewed the whole diff against the #455 adjudication, the lock rule in `AGENTS.md`, and the
surrounding surrender/co-owner call sites. Everything below is either confirmed by a mutation
or refuted with the code that refutes it.

### Both halves of the adjudication are present

- **Shape 2 (`txn.surrender`)** — `ObservationTxn::surrender` (`crates/shelterwood/src/cells/retained.rs:117`)
  on top of `defer_surrender` (`crates/shelterwood/src/cells/gate.rs:130`). Guards are queued at the
  head of the deferred-effect list, so "before commit" is positional-free: the only way to reach
  `RetainedExit`'s raw `Exit` from a framework release path is to hold the token.
- **Shape 3 (the split)** — `into_exit` is gone; `into_user_owned` is the narrow twin.

### Confirmed by mutation

| mutation | result |
|---|---|
| `defer_surrender` appends instead of inserting at `surrender_effects` | `observation_txn_surrenders_before_ordinary_deferred_disposal` FAILS (`left: ThreadId(2), right: ThreadId(2)`) and `observation_txn_unwind_drains_panicking_surrender_prefix_after_unlock` FAILS with `[(3,false),(1,false),(2,false)]` |
| `commit` drops the gate guard *after* the effect drain | 5 of 6 `observation_txn_*` tests FAIL, including both new ones |
| `ObservationTxn::surrender` drops the guards inline instead of calling `defer_surrender` | **passed all 385 lib tests** — see finding 3 |

The two new tests are non-vacuous: both assert in the test body, and the `expect`s inside the
deferred effects run through `PanicAccumulator` on a *non*-unwinding drop, so they resume and
fail the test rather than being swallowed.

### Findings, all fixed on the branch in a96642f

1. **`driver.rs:460` was not a user hand-off** — the adjudication listed "the root completion
   handoff in `driver.rs`" among the three genuine hand-offs to user-owned values, but
   `ScopeCompletion::root_exit` is already an `Option<RetainedExit>` and
   `finish_root_incarnation` immediately did `RetainedExit::new(exit)` again. So the site
   unwrapped a carrier only to re-wrap it, leaving the raw failed payload on the driver stack
   across `RetainedStopReason::new(reason)`'s own retention walk (arguments evaluate
   left-to-right). This is exactly the PR's own *"framework carrier transfers keep `RetainedExit`
   intact rather than extracting a raw `Exit` between framework layers"* category — the PR
   converted `terminalize_child` and `begin_terminal_disposal` that way and missed this one.
   `finish_root_incarnation` now takes `impl Into<RetainedExit>` and wraps before the reason
   (`crates/shelterwood/src/cells/scope.rs:1033`).

2. **`into_user_owned` was `pub(crate)`, i.e. still a crate-wide convention** — with finding 1
   fixed it has no caller outside `cells`, so it is now `pub(super)`
   (`crates/shelterwood/src/cells/retained.rs:40`). A driver-layer raw extraction is now
   unavailable by construction, not by call-site count. Mutation-checked: restoring the
   `driver.rs` call fails to compile with `error[E0624]: method into_user_owned is private`.

3. **The "after unlock" half of the new invariant was not pinned** — `surrender` dropping its
   guards inline, under the gate, passed the entire lib suite. The existing tests pin the
   *ordering* (surrender before ordinary effects) and pin `defer_surrender`'s venue, but nothing
   pinned that `ObservationTxn::surrender` uses that lane at all. Added
   `observation_txn_surrender_releases_the_raw_exit_after_unlock`
   (`crates/shelterwood/src/cells/gate.rs:366`): a deliberately sole-owner guard whose error
   payload reports `gate.is_held()` from its destructor, so the release venue becomes observable.
   It asserts both that nothing is released at the call site and that the release sees an
   unlocked gate. Fails under the mutation above.

4. **Documentation claim corrected** — `AGENTS.md` asserted the three remaining
   `into_user_owned` sites are "genuine public-ownership handoffs". Only
   `RetainedLifecycleEvent::into_public` (`cells/observe.rs:158`, reached from the two
   `LifecycleItem::Event` returns) is one. See follow-up 1 for the other. Text updated to say
   what is actually true, including the new visibility fence.

5. **Nit folded in** — `publish_drain`'s `startup_installed` outer flag became vestigial once the
   surrender moved inside the transaction (the `!startup_installed` branch only dropped a vec
   that `mem::take` had already emptied). Collapsed into a let-chain plus an unconditional drop.

### Refuted

- *Can a surrendered value be surrendered twice, or lost?* No. `surrender` moves the guards into
  the closure, and `commit` always runs from `Drop` — `panics.run` contains a hostile effect, so
  a panic in one surrender neither skips the rest of the queue nor strands the remaining
  `RetainedExit`s in that same closure: they drop as *carriers* during its unwind and retire
  through `dispose_critical`. The unwind path and the commit path are literally the same
  function.
- *Is the `surrendered: &mut Vec<RetainedExit>` out-param a hole?* It is not `#[must_use]`, so a
  future caller can forget to hand it to the transaction — but the failure mode is fail-safe: a
  dropped `Vec<RetainedExit>` retires through `dispose_critical`, i.e. a redundant disposal job,
  never a raw user destructor. Legal under the gate per the lock rule's `Arc`-traffic exemption.
- *Does `ScopeCell::snapshot` surrender its guards before the public co-owner exists?*
  No. `with_observation_txn_probed` evaluates `operation(&mut txn)` into the return place and
  only then drops `txn`, so the returned `Arc<ScopeSnapshot>` is alive across the flush.
- *Does moving snapshot install from `&mut Vec<effects>` to `&mut ObservationTxn` break the
  head/tail invariant?* No — `install` only calls `surrender` (head) and `defer` (tail), and
  `commit` appends the spent publication directly, so `effects[0..surrender_effects]` stays
  exactly the surrender prefix.
- *Is the head priority actually load-bearing in production, or only in the test?* Load-bearing.
  `MemberCell::terminalize_locked` queues `txn.defer(move || drop(losing_exit))` — a
  `RetainedExit` bound for `dispose_critical` — *before* `scope.rs:1131`'s
  `wakes.surrender([exit])` releases the raw peer of that very payload. Appending would race the
  disposal worker for the last reference on the committing thread.
- *`scope.rs:1103`/`1130`'s `drop(reason.into_public())` under the gate.* Checked, not a bug: if
  `publish_stopped_locked` published, the record owns a clone; if precedence rejected it, the
  `wakes.defer(move || { drop(state); drop(transient_retained); })` closure still holds one at
  flush time. Conventional, but sound. See follow-up 1.

### Follow-ups (not folded in, no issues filed)

1. `RetainedStopReason::into_public` (`cells/retained.rs:320`) is a misnomer: both call sites are
   `drop(reason.into_public())` under the observation gate (`cells/scope.rs:1103`, `1130`), i.e.
   framework-internal releases resting on exactly the hand-written co-owner proof #455 set out to
   delete — and the second proof (above) is subtler than any of the original 19. The natural
   shape is the one this PR just built: a `surrender(self, txn)` that hands the retained exits to
   the transaction and defers the raw reason. Left alone because the adjudication explicitly
   named this method as surviving.
2. `driver.rs:120`/`:177` call `finish_live_root_incarnation` with a raw `Exit` produced by
   `classify_retained_root_driver_join` (`Result<RetainedStopReason, Exit>`), which is re-wrapped
   inside. Same carrier-transfer shape as finding 1, one layer further out.
3. `commit`'s trailing `self.surrender_effects = 0` (`cells/gate.rs:192`) is unreachable state —
   `commit` only runs from `Drop`. Harmless defensive code; left as is.

### Verification

`./tools/dev just ci` on a96642f: **pass** — 957 nextest tests, fmt, clippy, doctests, examples,
docs, API reachability + prelude containment, core manifest, external consumer, nixfmt.

ralexstokes:
Rebased onto `origin/main` (`df2d019`), which now includes #468 ("Retain exits through terminal disposal"). One conflict resolution is semantic rather than textual, so recording it here.

**Where.** `ScopeRuntime::handle_construction_disposed` in `crates/shelterwood/src/driver/child.rs`.

**What #468 landed.** The terminal-disposal guard is kept across `terminalize_child` / `fail_startup` / `prune_terminal` and released once at the end:

```rust
let terminalized = self.terminalize_child(key, exit.as_exit().clone(), ...);
...
if terminalized {
    drop(exit.into_exit());
}
```

That release rests on a conventional co-owner proof — the terminal member record holds an equivalent retained copy, which the local `member` handle keeps alive — which is exactly the shape #455 set out to make unrepresentable. With `into_user_owned` narrowed to `pub(in crate::cells)`, no driver-layer caller can reach the raw `Exit` at all, so the statement cannot survive as written.

**Resolution.** #468's stronger signatures are kept verbatim (`begin_terminal_disposal(exit: RetainedExit)`, `fail_startup(exit: &RetainedExit)`, guard held across every fallible hop, single convergence point at the end). Two changes on top:

- `terminalize_child` is handed `exit.clone()` — a guarded clone — instead of `exit.as_exit().clone()`. No raw `Exit` crosses the driver/cell seam at all now, which is strictly tighter than #468's "no call-site argument window". That clone is the copy that retires as *refcount traffic*: `ScopeCell::terminalize_child` surrenders it via `wakes.surrender([exit])` inside the publishing transaction, where the terminal member record is its structural co-owner.
- The driver's own guard falls out of scope instead of being surrendered raw. The driver owns no `ObservationTxn` to surrender into, and minting an empty root transaction purely to queue a refcount decrement would be theater, so `RetainedExit::drop` is the venue.

**Behavioral delta.** Terminalizing a `Failed` child now costs one `dispose_critical` blocking-pool job that #468's `if terminalized` branch avoided. This is the venue #468 itself already documented and accepted for its (unreachable) losing branch — "the guard simply falls out of scope, and `RetainedExit::drop` retires the user error through critical disposal at the cost of one extra blocking-pool job" — now applied on both branches. It is the conservative direction: the guard is never released on an unproven co-owner. Consequently the `terminalized` binding is gone; #468's `if/else` restructure of the `Removing` branch is kept, since it still gives the single convergence point the guard retires at.

Nothing else conflicted. #455's two halves are intact: `ObservationTxn::surrender` / `defer_surrender` with the prioritized surrender prefix, and the `into_exit` → `into_user_owned` split at `pub(in crate::cells)` (`rg into_exit` is empty; the three remaining `into_user_owned` call sites are all inside `cells`). `./tools/dev just ci` is green on the rebased head (970 tests).

ralexstokes:
Rebased onto `origin/main` (`93cb20a`), which now includes #465 ("Fence root terminality on the driver join"). One conflict resolution is semantic rather than textual, so recording it here.

**Where.** `ScopeRuntime::drop`'s epilogue (`driver.rs`) and `ScopeCell::finish_root_incarnation` (`cells/scope.rs`).

**What conflicted.** This PR's third commit narrowed the raw-exit escape hatch by giving `finish_root_incarnation` an `impl Into<RetainedExit>` parameter and deleting `RetainedExit::into_exit` at the driver call site. #465 deleted that call site outright: `ScopeCompletion.root_exit` is gone, the epilogue now calls plain `finish_incarnation`, and `finish_root_incarnation` is demoted to `#[cfg(test)]`.

**Resolution.** Took #465's side in the epilogue — there is no longer a root-exit branch to narrow — and kept this PR's signature on the now test-only `finish_root_incarnation`, with its doc comment re-worded to drop the reference to a driver call site that no longer exists. Net effect: the escape hatch this PR was closing is closed twice over, since the only remaining caller is `mod tests`. `rg into_exit` is still empty and the three `into_user_owned` call sites remain inside `cells`.

`./tools/dev just ci` green on the rebased head: 978 tests, doctests, examples, docs, API reachability, prelude containment, core-manifest, external-consumer, nixfmt.

Issue #437, opened 2026-08-22
Design: AdmissionTxn — make the child-registration choreography's partial-effect edges structural


ralexstokes:
Design question from the second external-idea batch (2026-08-22, tracked on #429), accepted as the round's best structural suggestion — with a framing correction.

## Problem

Driver admission coordinates four state families separately: supervisor membership, runtime resources (driver storage/dynamic entries), dynamic state, and residency + observation announcement. The seams between them are where this round's partial-effect hazards live:

- #419 item 3 — the `announced` flag is paired to the pushed resident via an unchecked `last_mut()` across the widest window in `admit_child_locked` (residency push → gate handoff → `Admitted` transition → `set_parent`).
- #435 — the parent back-link is installed before a fallible step, escapes the `announced` gate entirely, and has no withdrawal edge on any path (healthy removal included).
- #427 item 1 — two admission-lane paths hold the dynamic-state mutex around gate acquisitions, safe only via two unstated invariants.
- Several deferred-panic diagnostics exist solely to detect cross-family drift (`rejected_is_admission`, the promote/admit invariant arm at driver.rs:827-854 — untested, per #422 item 5).

## Framing correction (why this is NOT "one transaction that does everything")

The current *ordering* is itself the containment design: residency is taken before the first fallible step so a bookkeeping panic leaves the graph owned and retiring through scope clear (CLAUDE.md's admission passage, SPEC §3.2/§5.5). And CLAUDE.md's doubled-section accounting explicitly warns against widening what runs under the gate. So the remedy is not to concentrate more work under one lock — it is to *formalize the existing choreography* the way `ObservationTxn` formalized gate effects: an `AdmissionTxn`-shaped owner that

1. records each installed effect (residency slot index, parent link, membership adoption, dynamic entry, announcement) as it lands,
2. makes the announced-flag pairing and parent-link pairing structural (indexed, not `last_mut()`; link paired with an explicit withdrawal edge or ordered after announcement — the #435 design fork lands here),
3. on unwind, reverses or retires exactly the recorded prefix (subsuming the deferred-panic drift diagnostics), and
4. gives the #427 invariants a place to be stated and debug-asserted instead of argued in review.

## Scope notes

- The gate-handoff retry and the doubled-section contents stay as they are — this is bookkeeping structure, not lock-order change.
- Success criterion per the PR #234 bar: the transaction must *delete* diagnostics/assertions (the tautological re-check class, the unchecked pairing), not add a layer beside them.
[...]

PR #473, opened 2026-08-23
Make admission announcement target its residency slot

ralexstokes:
Stack: 1 of 3 for #437. This PR intentionally leaves #437 open; the driver-side ledger in PR3 completes the signed-off design.

Summary:
- replace the raw residency index convention with a must-use ResidentSlot returned by push_unannounced
- make announce consume that exact index-and-membership token
- preserve residency-first unwind containment and the real rejected_is_admission diagnostic
- add interposed-resident and membership-mismatch regression tests

Verification:
- ./tools/dev just ci (956 tests plus doctests, examples, docs, API reachability, manifest and Nix formatting checks)
- mutation check: changing ResidentSlot::announce back to last_mut makes resident_slot_announces_its_index_when_a_later_resident_exists fail
- mutation check: removing the membership guard makes resident_slot_refuses_a_different_membership_at_its_index fail

### DIFF
--- a/crates/shelterwood/src/cells/scope.rs
+++ b/crates/shelterwood/src/cells/scope.rs
@@ -945,13 +945,6 @@ impl ScopeCell {
         terminalized.expect("a supervised terminal child must remain in parent residency")
     }
 
-    /// Installs residency without announcing it, as an admission that
-    /// unwound between its residency push and its `Added` publication does.
-    #[cfg(test)]
-    fn push_unannounced_resident_for_test(&self, child: ResidentProjection) {
-        self.current_children().push(ResidentChild::new(child));
-    }
-
     #[cfg(test)]
     pub(crate) fn prune_child(&self, member: &MemberCell) -> bool {
         self.with_observation_gate(|wakes| self.prune_child_locked(member, wakes))
@@ -2312,8 +2305,9 @@ mod tests {
         let mut lifecycle = root.subscribe_lifecycle();
 
         // Install residency exactly as an admission that unwound before its
-        // `Added` publication leaves it.
-        root.push_unannounced_resident_for_test(ResidentProjection::new(Arc::clone(&member), None));
+        // `Added` publication leaves it: the production push, with its slot
+        // dropped instead of announced.
+        drop(root.push_unannounced(ResidentProjection::new(Arc::clone(&member), None)));
         assert_eq!(root.resident_projections().len(), 1);
         assert!(root.snapshot().children.is_empty());
 
--- a/crates/shelterwood/src/cells/scope.rs
+++ b/crates/shelterwood/src/cells/scope.rs
@@ -101,6 +101,41 @@ impl ResidentChild {
     }
 }
 
+/// The exact residency slot installed by one in-flight admission.
+///
+/// The observation gate prevents a production admission from interleaving a
+/// second residency mutation, but the slot still carries its own index and
+/// membership. That makes the later announcement structurally address the
+/// resident whose `Added` edge is about to be published instead of relying on
+/// whichever child happens to be last.
+#[must_use = "an installed resident slot must be announced or withdrawn"]
+struct ResidentSlot<'a> {
+    children: &'a Mutex<Vec<ResidentChild>>,
+    index: usize,
+    membership: Membership,
+}
+
+impl ResidentSlot<'_> {
+    /// Marks this admission's exact resident as announced.
+    ///
+    /// The token and everything inspected under the residency mutex are plain
+    /// framework-owned data, so this adds no effect to the doubled observation
+    /// gate section's lock-rule accounting.
+    fn announce(self) -> bool {
+        let mut children = self
+            .children
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        children
+            .get_mut(self.index)
+            .filter(|resident| resident.projection().member.membership() == self.membership)
+            .is_some_and(|resident| {
+                resident.announced = true;
+                true
+            })
+    }
+}
+
 impl Drop for ResidentChild {
     fn drop(&mut self) {
         let Some(projection) = self.projection.take() else {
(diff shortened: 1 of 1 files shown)


ralexstokes:
## Review (PR 1 of the #437 stack)

Reviewed this PR's own diff only (`git diff origin/main...HEAD`, `cells/scope.rs`), against the signed-off #437 design, walking `admit_child_locked`'s refusal and unwind paths as well as the happy path, plus the `announced`/`Added` pairing readers (`prune_child_locked`, `clear_residents_locked`, snapshot producer) and the lock-rule accounting.

### Confirmed finding (fixed on the branch)

**The token was consumed on only one of admission's two residency edits.** At `crates/shelterwood/src/cells/scope.rs:108` (as of e9a6198) the attribute read `#[must_use = "an installed resident slot must be announced or withdrawn"]`, but `ResidentSlot` had no `withdraw`. On the refusal path the slot was bound and silently dropped while the resident was re-derived with a raw `self.current_children().pop()` plus a membership compare (`scope.rs:1459-1462`). Two consequences:

- the doc promised an operation the type did not have, and `#[must_use]` cannot fire on a bound local, so nothing enforced it;
- the refusal half was left "relying on whichever child happens to be last" — exactly the convention the PR's own doc comment says the slot exists to delete. The two halves could disagree about which resident this admission installed with nothing catching it.

Fixed in **5ccd628** by adding `ResidentSlot::withdraw` (`scope.rs:158`) and routing the refusal through it (`scope.rs:1480`). Semantics are preserved and the verdict is strictly stronger:

- it still pops the last entry, so residency cannot grow past a refusal even when the diagnostic fires;
- it hands the displaced `ResidentChild` back **by value** so it still reaches `txn.defer(dispose_detached(...))` after unlock — no user destructor moves under the residency mutex;
- `is_admission` now requires `index == len - 1` *and* the membership match, rather than only comparing the popped membership.

Carried along in the same commit:

- `announce` re-implemented the poison-tolerant lock by hand; the slot now holds `&ScopeCell` and reuses `ScopeCell::current_children` (`scope.rs:137`).
- `ResidentSlot` had been inserted between `impl ResidentChild` and `impl Drop for ResidentChild`; moved below so the drop glue stays with its type.
- Documented that dropping the token unconsumed is the *containment* path, not a leak — the `#[must_use]` wording otherwise reads as forbidding precisely the unwind behaviour `admit_child_locked` depends on.

**bad870e** folds the now-duplicated `#[cfg(test)] push_unannounced_resident_for_test` into the production `push_unannounced`; `pruning_an_unannounced_resident_publishes_no_removal_edge` now installs its resident through the real path with the slot dropped instead of announced, which is a closer model of the unwind it claims to reproduce.

### Verified by mutation (committed tree, reverted after each)

Author's two claims, both reproduced:
- `announce` addressing the last index instead of `self.index` → `resident_slot_announces_its_index_when_a_later_resident_exists` FAILS.
- `announce`'s membership filter removed → `resident_slot_refuses_a_different_membership_at_its_index` FAILS.

New pins for `withdraw`:
- `children.len() == self.index + 1` removed → `resident_slot_withdrawal_refuses_an_interposed_resident` FAILS.
- the membership compare removed → `resident_slot_withdrawal_refuses_a_different_membership_at_its_index` FAILS.

### Refuted / checked clean

- **Double announcement.** Impossible: `announce` and `withdraw` both take `self` by value, so one slot drives at most one residency edit. This is strictly better than the pre-PR raw `residency_index`, which was reusable.
- **Announced-but-absent / unannounced leaking into the observed set.** Unchanged by this PR. `prune_child_locked` (`scope.rs:955`) and `clear_residents_locked` (`scope.rs:1517`) both gate `Removed` on `announced`, and the snapshot producer skips unannounced residents. The one residual window — a panic between `announce()` and `emit_locked(Added)` leaving an announced resident whose `Added` never published — is inherent to the ordering the snapshot producer requires (the mark must precede the publication) and pre-dates this PR; `emit_locked`'s only fallible framework path, sequence exhaustion, returns rather than panics.
- **Stale index.** The only code between the push and the announce is `admit_observation_gate`, `set_parent` and the member handoff closure; none mutate this scope's `current_children`. The membership and last-entry checks are the diagnostics for a future violation, not a live hazard.
- **Lock rule.** `announce` and `withdraw` hold only `current_children`. `announce` flips a `bool` and reads a `OnceLock` membership — plain framework data. `withdraw` moves a `ResidentChild` out by value; the destination is the caller's `txn.defer`, which is the "moving a user value out" exemption, not a drop under the lock. Nothing was added to the doubled-gate section: `push_unannounced` runs before the child gate is acquired and both edits run after it is released.
- **Membership captured at push time** (a genuine behavioural difference from `main`, which read it after the transition). `MemberCell::membership()` is only mutable through `rebase_membership`, whose sole production caller is plan lowering (`plan.rs:404`) on a `Reserved` member before any admission, so the two reads cannot diverge.
- **Test vacuity.** All five `resident_slot_*` assertions are in the test body, not inside an actor handler or `catch_unwind`, and each is mutation-pinned above.

### Scope note (not a defect)

#419 items 2 and 3 — the tautological `prune_child_locked` re-check and the unchecked `last_mut()` announcement — were already gone from `main` (5321ba9 already carried the index + membership check). This PR's contribution is therefore the token and the regression coverage rather than a behaviour fix, which is how the description reads.

- Neither the `announce` nor the `withdraw` failure arm's deferred panic has a test; both are unreachable by construction under the gate, so covering them needs a test-only seam that corrupts residency mid-admission.
- #419 items 1, 4 and 5 (the `publish_drain` startup retention, `set_state_locked`'s production-dead transient machinery, and the refusal-disposal venue coverage gap) are untouched by this stack.

### CI

`./tools/dev just ci` green at bad870e — 959 tests, doctests, examples, docs, API reachability, prelude containment, core manifest, external consumer, nixfmt.

### Rebase note

I pushed **5ccd628** and **bad870e** on top of e9a6198, so **#474** and **#475** need a rebase onto the new head of this branch (rebase, not merge, per the stacked-branch rule).

PR #474, opened 2026-08-23
Sever nested observation links on retirement

### LINKED ISSUE #435: Nested scope's parent link escapes the announced gate: events from an unannounced subtree reach ancestors; link has no withdrawal edge
External-review claim (2026-08-22 round): **mechanism CONFIRMED by repro, severity downgraded** from High to Low–Medium, and the broken rule corrected — SPEC §3.2's Added/Removed pairing is NOT violated; the causal-precedence rule is. Repro committed on local branch `verify-admission` (`cells/scope.rs` tests module: `claim4_panicked_admission_parent_link_probe`, failing by design, plus a passing healthy-path control).

## Mechanism

`cells/scope.rs:351-357` — `set_parent` installs the parent link (poison-tolerant write) *before* taking the fallible control lock (`.expect("scope control mutex poisoned")`, :357). `admit_child_locked` calls it at :1339 and only marks the resident announced at :1396-1398 before `emit_locked(Added)` (:1399). So a panic in that window leaves an **unannounced resident with a live parent link** (nested-scope branch only; plain members install no parent link).

The `announced` flag gates only what the *parent* publishes about the child (snapshot skip `projection.rs:132`, Removed skip `scope.rs:1412-1418`, prune skip `scope.rs:872`). It does not gate what the *child pushes up*: `emit_locked` forwards to every ancestor hub via `ancestors_locked` (`projection.rs:226, 301-303`), so the orphan's later `ScopeState::Stopped` reaches a root that never saw its `Added`. Repro confirms: root lifecycle receives the event, snapshot generation bumps (content unchanged, benign).

## Trigger and severity

Second-order: no user code runs under `ScopeControl`, but a real poison chain exists — `publish_stopped_locked` holds the control guard across `member.terminalize_locked` (scope.rs:1563-1566), which `.expect()`s the member mailbox mutex (member.rs:708); a poisoned member-mailbox mutex therefore poisons the control mutex, and the next admission panics in `set_parent`. Framework-invariant chain, not user-reachable in one step — exactly the class the containment machinery exists for. Control-plane lane is unaffected (`key_for` returns `None` for the unknown membership, `driver/events.rs:232-236`).

**Severity Low–Medium.** Added/Removed pairing holds (the existing pin `panicked_resident_admission_lingers_until_scope_clear`, scope.rs:1937, still passes); the rule broken is SPEC.md:3549-3550 ("a subtree child's own `Added` precedes any event from inside it"); the spec's own reader protocol (SPEC.md:3594-3603) classifies the unknown-token event as discardable, confining harm to a from-the-start subscriber relying on causal precedence. Observation-plane only; no memory/liveness consequence.

## The structural question the review missed

The control test shows **the parent link has no withdrawal edge on any path** — a healthy admitted-then-cleared nested scope also forwards later events up a still-installed link (deliberate today, per `rejected_nested_admission_preserves_original_parent_and_gate`). The panic-specific delta is only that the token was *never introduced* by any `Added`. So the fix decision is: clear the link on removal, order `set_parent` after announcement, or make link installation infallible — each has different observable consequences and needs an adjudication against the CLAUDE.md admission narrative, which currently reads as complete containment for a half-wired resident but does not cover the back-link.

Part of the external-review verification round tracked on #429. Relates to #419 (scope admission hardening).

ralexstokes:
Stack: 2 of 3 for #437, based on #473.

Summary:
- sever the parent back-link only after prune or displacement selects its resident for retirement
- cover announced and unannounced panic-resident retirement while preserving rejected duplicate parentage
- amend SPEC B.4 so Removed is the former parent stream final word about a subtree
- invert both verify-admission probes and prove later events remain on the removed subtree local stream
- pin prune, explicit clear, and residency-replacement retirement independently, including poisoned parent-link recovery

Verification:
- ./tools/dev just ci (959 tests plus doctests, examples, docs, API reachability, manifest and Nix formatting checks)
- mutation check: replacing the common sever with a no-op fails the announced, unannounced, and poison-tolerance regressions
- mutation check: omitting only the prune-path sever fails the prune retirement case while the prior 386-test library suite remained green

Closes #435

### DIFF
--- a/crates/shelterwood/src/cells/scope.rs
+++ b/crates/shelterwood/src/cells/scope.rs
@@ -2204,71 +2204,119 @@ mod tests {
         );
     }
 
+    /// The venues that withdraw a resident from its scope. Every one of them
+    /// severs, so each pin below runs against all three.
+    #[derive(Clone, Copy, Debug)]
+    enum Retirement {
+        Prune,
+        Clear,
+        Replace,
+    }
+
+    impl Retirement {
+        fn retire(self, root: &Arc<ScopeCell>, member: &MemberCell) {
+            match self {
+                Self::Prune => assert!(root.prune_child(member)),
+                Self::Clear => root.clear_residents(),
+                Self::Replace => assert!(root.set_admitted_children(Vec::new())),
+            }
+        }
+    }
+
     #[test]
-    fn clearing_unannounced_nested_resident_severs_its_parent_link() {
+    fn retiring_an_unannounced_nested_resident_severs_its_parent_link() {
+        for retirement in [Retirement::Prune, Retirement::Clear, Retirement::Replace] {
+            let root = isolated_scope("root", ScopeFlavor::Ordered);
+            let nested = child_scope(&root, "nested", ScopeFlavor::Dynamic);
+            let mut root_lifecycle = root.subscribe_lifecycle();
+            let mut nested_lifecycle = nested.subscribe_lifecycle();
+
+            // `set_parent` installs the back-link before this poisoned control
+            // lock unwinds the admission, leaving an unannounced resident behind.
+            let poisoned = Arc::clone(&nested);
+            assert!(
+                catch_unwind(AssertUnwindSafe(move || {
+                    let _control = poisoned.control.lock().expect("control starts healthy");
+                    panic!("inject admission bookkeeping panic");
+                }))
+                .is_err()
+            );
+            catch_unwind(AssertUnwindSafe(|| {
+                let _ = root.admit_child(ResidentProjection::new(
+                    Arc::clone(&nested.member),
+                    Some(Arc::clone(&nested)),
+                ));
+            }))
+            .expect_err("poisoned parent wiring unwinds admission");
+
+            assert!(Arc::ptr_eq(
+                &nested
+                    .parent()
+                    .expect("the failed admission installed its link"),
+                &root
+            ));
+            retirement.retire(&root, &nested.member);
+            assert!(
+                nested.parent().is_none(),
+                "{retirement:?} severs an unannounced resident's parent link"
+            );
+            assert_eq!(
+                root_lifecycle.try_recv(),
+                Err(LifecycleTryRecvError::Empty),
+                "the failed admission published neither membership edge"
+            );
+
+            nested.set_state(ScopeState::Stopped {
+                reason: StopReason::Finished,
+            });
+            assert!(matches!(
+                nested_lifecycle.try_recv(),
+                Ok(LifecycleItem::Event(_))
+            ));
+            assert_eq!(
+                root_lifecycle.try_recv(),
+                Err(LifecycleTryRecvError::Empty),
+                "post-retirement events stay local to the removed subtree ({retirement:?})"
+            );
+        }
+    }
+
+    /// Retirement severs; destroying a resident child does not.
+    ///
+    /// A refused duplicate admission temporarily owns a second `ResidentChild`
+    /// for an already-resident subtree and then disposes it, so a sever in
+    /// `Drop` would detach a subtree that is still resident under its original
+    /// parent. `rejected_nested_admission_preserves_original_parent_and_gate`
+    /// walks that refusal path, but it retires the duplicate through *detached*
+    /// disposal — the assertion outruns the drop, so only this synchronous
+    /// destruction pins the venue.
+    #[test]
+    fn destroying_a_resident_child_leaves_the_subtree_parent_link_installed() {
         let root = isolated_scope("root", ScopeFlavor::Ordered);
         let nested = child_scope(&root, "nested", ScopeFlavor::Dynamic);
-        let mut root_lifecycle = root.subscribe_lifecycle();
-        let mut nested_lifecycle = nested.subscribe_lifecycle();
+        assert!(root.admit_child(ResidentProjection::new(
+            Arc::clone(&nested.member),
+            Some(Arc::clone(&nested)),
+        )));
 
-        // `set_parent` installs the back-link before this poisoned control
-        // lock unwinds the admission, leaving an unannounced resident behind.
-        let poisoned = Arc::clone(&nested);
+        drop(ResidentChild::new(ResidentProjection::new(
+            Arc::clone(&nested.member),
+            Some(Arc::clone(&nested)),
+        )));
+
         assert!(
-            catch_unwind(AssertUnwindSafe(move || {
-                let _control = poisoned.control.lock().expect("control starts healthy");
-                panic!("inject admission bookkeeping panic");
-            }))
-            .is_err()
-        );
-        catch_unwind(AssertUnwindSafe(|| {
-            let _ = root.admit_child(ResidentProjection::new(
-                Arc::clone(&nested.member),
-                Some(Arc::clone(&nested)),
-            ));
-        }))
-        .expect_err("poisoned parent wiring unwinds admission");
-
-        assert!(Arc::ptr_eq(
-            &nested
-                .parent()
-                .expect("the failed admission installed its link"),
-            &root
-        ));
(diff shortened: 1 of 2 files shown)


ralexstokes:
## Review — PR 2 of the #437 stack (`agent/issue-437-residency-slot`...`HEAD`)

Reviewed the PR's own diff only (`scope.rs` +192, `SPEC.md` +5). Confirmed it implements the shape signed off on #437: retirement of a `ResidentChild` severs the child's parent back-link before publishing `Removed` or retiring the graph, on prune / displacement / clear including the unannounced-panic lane, and never on rejection of an incoming duplicate.

### Findings — 2 confirmed, both mutation-invisible pins, both fixed on-branch (67fba7e)

**1. The care point's pin was fail-open (`cells/scope.rs:100-109` doc comment ↔ `cells/scope.rs:2412`).**
The "deliberately not part of `Drop`" invariant rested on `rejected_nested_admission_preserves_original_parent_and_gate`. Moving the sever into `ResidentChild::drop` leaves that test **green**:

```
=== M1: sever inside ResidentChild::drop ===
Summary [0.003s] 1 test run: 1 passed, 387 skipped   (x5 runs)
```

Mechanism (not a guess — verified): the refusal path retires the duplicate through `txn.defer(move || runtime::dispose_detached((projection, rejected)))` (`scope.rs:1492`), and `dispose_detached` funnels to the shared disposal thread (`shelterwood-runtime/src/disposal.rs:334`). The test's `nested.parent()` assertion outruns the drop. Inserting `sleep(300ms)` before it turns the same mutation red:

```
FAIL cells::scope::tests::rejected_nested_admission_preserves_original_parent_and_gate
  panicked at crates/shelterwood/src/cells/scope.rs:2446
```

*Fix:* added `destroying_a_resident_child_leaves_the_subtree_parent_link_installed`, which destroys a duplicate `ResidentChild` for an already-resident subtree **synchronously** and asserts the original link survives. Red under M1, no timing dependence. (This is a negative pin — it is green on the base by construction, since the base has no sever at all. Called out in the test's doc comment.)

**2. Prune's unannounced lane was unpinned (`cells/scope.rs:974`).**
The comment there says "An unannounced resident takes this path too", but gating *only* prune's sever on `announced` keeps the whole library suite green:

```
=== M6: prune severs only announced ===
Summary [1.070s] 388 tests run: 388 passed, 0 skipped
```

`clearing_unannounced_nested_resident_severs_its_parent_link` covered the clear venue only; the prune arm of `announced_nested_retirement_…` uses an announced resident.

*Fix:* hoisted the venue enum into a shared `Retirement::retire` helper and renamed the unannounced test to `retiring_an_unannounced_nested_resident_severs_its_parent_link`, running it across Prune / Clear / Replace. Red under M6.

### Mutation results after the fix

| mutation | outcome |
|---|---|
| `sever_parent` body → no-op (base behavior) | 3 FAIL: `announced_nested_retirement_…`, `retiring_an_unannounced_…`, `nested_retirement_tolerates_a_poisoned_parent_link` |
| drop `resident.sever_parent()` from `prune_child_locked` | FAIL `announced_nested_retirement_…` |
| drop the sever loop from `clear_residents_locked` | FAIL `announced_nested_retirement_…` |
| gate both severs on `announced` | FAIL `retiring_an_unannounced_…` |
| gate *only* prune's sever on `announced` | **was green**, now FAIL `retiring_an_unannounced_…` |
| `sever_parent` → `.expect()` instead of poison-tolerant | FAIL `nested_retirement_tolerates_a_poisoned_parent_link` |
| sever moved into `ResidentChild::drop` | **was green**, now FAIL `destroying_a_resident_child_…` |

### Refuted concerns (each checked, none real)

- **Severing under a lock is not a lock-rule violation.** `ScopeCell::sever_parent` (`scope.rs:405`) writes `None` over an `Option<Weak<ScopeCell>>` under a poison-tolerant leaf mutex. Dropping a `Weak` never runs `ScopeCell::drop` — at worst it deallocates when both counts hit zero — so no user value, callback or panic enters the critical section.
- **No ABBA against the parent-to-child gate order.** The new edge is *observation gate → child parent mutex*, which `set_parent` (`scope.rs:418`) already establishes; the parent mutex is a strict leaf (`set_parent`'s guard is a statement temporary and is released before it touches `control`). The two driver prune sites (`driver/removal.rs:56`, `:98`) call `prune_child_locked` while additionally holding the dynamic-state mutex, giving *gate → dynamic state → parent mutex* — still acyclic, since nothing is acquired under the parent mutex. `DynamicControl::reserve`'s doubled section is unaffected.
- **No mid-traversal race.** `ancestors_locked` / `emit_locked` (`cells/scope/projection.rs:230,271`) only run under the resident-tree observation gate, which an admitted nested scope shares with its parent; prune and clear take the same gate. And `request_shutdown_locked`'s `self.parent()` (`scope.rs:1242`) runs under the child's (= shared) gate too.
- **No lost wake or lost `Removed`.** Events emitted before the sever in the same transaction are already routed onto the ancestor hubs by `emit_locked` and are flushed in queue order at commit, so the sever cannot strand them. The `Removed` edge itself is published on the *parent* and walks the parent's own ancestors, so severing the child first cannot suppress it.
- **No half-severed graph on unwind.** `ResidentChild::sever_parent` can't panic: `projection()`'s `expect` holds for every live resident, the lock is poison-tolerant, and the `Weak` overwrite is infallible. It therefore adds no new panic source to the pre-existing window between `mem::take`/`remove` and `txn.defer(dispose_detached(...))` — the residents in that window are still displaced through detached disposal on both the normal and unwind paths.
- **Per-element containment is untouched.** Retirement does not change who drops a resident; `ResidentChild::drop` (`scope.rs:148`) still owns the boundary, so containment still holds at every nesting depth and on `ScopeCell`'s drop glue (no `impl Drop for ScopeCell` exists).
- **The sever is complete.** `set_parent` has exactly one caller (`admit_child_locked:1448`), and the only residency removals are prune (`:961`), the refusal pop (`:1484`, deliberately not severing), and clear's `mem::take` (`:1526`). No fourth retirement venue exists. The `announce()`-failure arm (`:1512`) leaves a link behind, but it is a deferred-panic diagnostic on an under-gate-unreachable state.
- **SPEC amendment is complete and localized.** `SPEC.md:3567-3571` is the only place the repo states the forwarding contract; `book/` does not restate it. Snapshot-chain propagation rides the same `ancestors_locked` walk, but a removed subtree is already out of the parent's snapshot, so the amendment needs no snapshot clause.
- **Clear-then-readmit within one `set_admitted_children` transaction is not a hazard.** It cannot occur: clearing performs no member transition, so a cleared membership is still `Admitted` and `admit_observation_gate`'s `would_accept` probe refuses it. Clear is a terminal venue in every real caller (`driver.rs:441`, `plan.rs:497`, `driver.rs:556` at startup).

- A retired subtree keeps the former parent's *observation gate* even after the back-link is severed, so it contends on the parent tree's gate for the rest of its life. Pre-existing and harmless (no path re-homes a removed subtree), but it is the one asymmetry left in "residency is the ownership edge for upward observation".
- `rejected_nested_admission_preserves_original_parent_and_gate` (on `main`) is still timing-dependent for its *parent-link* clause; its gate/residency clauses are sound. The new synchronous test covers the invariant, but making that one deterministic would need a disposal barrier the test support doesn't offer today.
- The poison-chain narrowing #435's verification surfaced (`publish_stopped_locked` holding the control guard across the member-mailbox `.expect`, `scope.rs:1667+`) is still open, as the issue notes — not a dependency of this PR.

### CI

`./tools/dev just ci` green on the pushed head (`67fba7e`): 959 workspace tests (389 in `shelterwood` lib), doctests, examples, docs, API reachability, prelude containment, core-manifest, external-consumer, nixfmt.

**Stack note:** #474 moved (`4d68cf9` → `67fba7e`), so #475 needs a rebase onto the new head before merge.

Issue #395, opened 2026-08-21
Remove load-bearing debug_asserts: promote, restructure, or certify diagnostic-only


ralexstokes:
Adjudicated 2026-08-21 across #392/#393: rather than adding a release-profile
test lane, remove the debug/release behavioral divergence at its source. A
`debug_assert!` may not be load-bearing — no test may depend on it firing and
no correctness property may rely on it — so every site either becomes an
always-on check or is certified diagnostic-only.

## Scope

~105 `debug_assert!` sites across 25 files (count as of 0bf8529, heaviest:
`shelterwood-cells/src/cells/scope.rs` ×20, `shelterwood-mailbox/src/cell.rs`
×14, `shelterwood/src/driver.rs` ×8, `shelterwood-core/src/engine.rs` ×8,
`shelterwood/src/raw/context.rs` ×8).

## Policy, per site

1. **Outside framework locks → promote to `assert!`.** The seven driver
   verdict sites from #382 are the archetype: they were deliberately placed
   outside every lock, so promotion is legal under the lock rule and closes
   #392's release gap (a refused transition becomes a loud fail-stop into the
   driver's containment instead of a silent half-applied operation).
2. **Under a lock, release possible → restructure to compute-verdict,
   release, then `panic!`** (the `MailboxCell::bind` pattern from the lock
   rule's conventions).
3. **Under a lock, release impossible → keep `debug_assert!`, certified
   diagnostic-only** (the `MailboxState::take_waiters` class). Certification
   means: mark the site as diagnostic, and confirm no test asserts on the
   panic and no downstream behavior differs when it is compiled out.
4. **Drop the three `#[cfg(debug_assertions)]` test gates** added in #380
   (`illegal_restart_transition_retires_its_user_error_off_thread`,
   `nonresident_terminal_exit_is_retained_before_the_residency_assertion`,
   `rejected_resident_admission_detaches_its_last_mailbox_owner`) as their
   guarding assertions become always-on, so the tests hold in both profiles.

Retention-before-assertion properties (#380's tests) must survive promotion:
where a `debug_assert!` today fires *after* a user value is routed to
retention/disposal, the promoted `assert!`/`panic!` must keep that ordering.

## Exit criteria

- Workspace release run (`cargo nextest run --release --workspace`) green
  with the test gates removed — run once at the end of the sweep as the
  proof, not added as a standing lane.
- Every remaining `debug_assert!` carries the diagnostic-only certification.
[...]

PR #475, opened 2026-08-23
Make dynamic admission installs unwind-safe

### LINKED ISSUE #437: Design: AdmissionTxn — make the child-registration choreography's partial-effect edges structural

ralexstokes:
Stack: 3 of 3 for #437, based on #474.

Summary:
- replace the function-local admission outcome enum and hand-written invariant arm with an owning AdmissionInstall ledger
- claim the child runtime, reserved entry, arena key, projection, and response obligation across installation
- retain rejected child ownership through locked never-started terminalization so a publication panic cannot run its fallback under the observation gate
- make Drop retire the projection after unlock and complete a lost response with panic containment
- debug-assert the admission gate identity invariant at ledger construction and record it in AGENTS.md
- add direct lost-response Drop coverage, a locked-terminalization unwind regression, and retain the hostile response-waker invariant test

Verification:
- ./tools/dev just ci (961 tests plus doctests, examples, docs, API reachability, manifest and Nix formatting checks)
- mutation check: emptying AdmissionInstall::drop makes dropping_admission_install_completes_the_lost_response fail

Closes #437

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -170,8 +170,9 @@ rests on:
   root gate while calling `admit_child_locked`, but the reserved member was
   already adopted onto that root gate and the root cannot be re-homed while
   its dynamic route is live, so the handoff check short-circuits without a
-  second gate acquisition. Changing either publication or re-homing invariant
-  requires re-deriving this exception.
+  second gate acquisition. `AdmissionInstall` debug-asserts that identity when
+  it takes ownership, before either guard is acquired. Changing either
+  publication or re-homing invariant requires re-deriving this exception.
   `WakerProxy`'s mutex (`crates/shelterwood-core/src/waker_proxy.rs`) sits at the
   other end of that order: it is a **leaf**. `wake_by_ref` acquires it from
   whatever thread drives the external primitive — the timer driver, a sender,
--- a/crates/shelterwood/src/cells/scope.rs
+++ b/crates/shelterwood/src/cells/scope.rs
@@ -395,6 +395,18 @@ impl ScopeCell {
         })
     }
 
+    /// Diagnostic for the dynamic admission install exemption.
+    ///
+    /// A reservation is adopted onto this scope's gate before publication and
+    /// the live dynamic route prevents the root from being re-homed. The
+    /// driver asserts that invariant when it creates the install ledger, so
+    /// the later handoff check cannot acquire another gate under dynamic
+    /// state.
+    pub(crate) fn shares_observation_gate_with(&self, member: &MemberCell) -> bool {
+        self.current_observation_gate()
+            .shares_gate(&member.current_observation_gate())
+    }
+
     fn current_children(&self) -> MutexGuard<'_, Vec<ResidentChild>> {
         self.observation
             .current_children
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -9,7 +9,6 @@ mod startup;
 
 use std::{
     collections::{BTreeMap, BTreeSet},
-    mem::ManuallyDrop,
     ops::{Index, IndexMut},
     sync::{Arc, OnceLock},
     time::Instant,
@@ -105,14 +104,18 @@ fn resident_projection(slot: &SlotCell) -> ResidentProjection {
 /// runs after both control-plane guards have gone: it retires the projection
 /// through detached disposal and completes the caller's response fail-closed.
 ///
-/// `request` is manual so that deleting this destructor cannot silently fall
-/// back through `AdmissionRequest`'s field drop. The ledger, rather than field
-/// order, is the structural owner of the lost-response contract.
+/// The ledger is the *primary* owner of the lost-response contract, not the
+/// only one: `request` stays a plain `Option<AdmissionRequest>` so its own
+/// `Obligation` still discharges the caller's promise on any path this
+/// destructor fails to cover, as SPEC §15.5 requires of every cross-task
+/// promise. Field order below is the Drop body's order, so even the bare
+/// fallback keeps terminality ahead of the removal completion and both ahead
+/// of the response.
 struct AdmissionInstall {
-    request: Option<ManuallyDrop<AdmissionRequest>>,
     projection: Option<ResidentProjection>,
     child: Option<Box<ChildRuntime>>,
     entry: Option<DynamicEntry>,
+    request: Option<AdmissionRequest>,
     key: Option<ChildKey>,
     entry_promoted: bool,
     entry_installed: bool,
@@ -127,7 +130,7 @@ impl AdmissionInstall {
         );
         Self {
             projection: Some(resident_projection(&request.slot)),
-            request: Some(ManuallyDrop::new(request)),
+            request: Some(request),
             child: Some(Box::new(child)),
             entry: None,
             key: None,
@@ -139,22 +142,20 @@ impl AdmissionInstall {
 
     fn request(&self) -> &AdmissionRequest {
         self.request
-            .as_deref()
+            .as_ref()
             .expect("an unfinished admission install owns its request")
     }
 
     fn request_mut(&mut self) -> &mut AdmissionRequest {
         self.request
-            .as_deref_mut()
+            .as_mut()
             .expect("an unfinished admission install owns its request")
     }
 
     fn take_request(&mut self) -> AdmissionRequest {
-        ManuallyDrop::into_inner(
-            self.request
-                .take()
-                .expect("an admission install releases its request once"),
-        )
+        self.request
+            .take()
+            .expect("an admission install releases its request once")
     }
 
     fn slot(&self) -> &Arc<SlotCell> {
@@ -296,8 +297,7 @@ impl Drop for AdmissionInstall {
         if let Some(entry) = self.entry.take() {
             panics.run(move || drop(entry));
         }
-        if let Some(request) = self.request.take() {
-            let mut request = ManuallyDrop::into_inner(request);
+        if let Some(mut request) = self.request.take() {
             panics.run(|| request.complete_lost());
             panics.run(move || drop(request));
         }
--- a/crates/shelterwood/src/driver.rs
+++ b/crates/shelterwood/src/driver.rs
@@ -9,6 +9,7 @@ mod startup;
 
 use std::{
     collections::{BTreeMap, BTreeSet},
+    mem::ManuallyDrop,
     ops::{Index, IndexMut},
     sync::{Arc, OnceLock},
     time::Instant,
@@ -33,7 +34,7 @@ use crate::{
     ScopeState, ShutdownStraggler, ShutdownTimeout, StartupFailure, StartupFailureCause,
     cells::{
         LifecycleEventKind, MemberCell, MemberStage, MemberTransition, NotAdmittingCause,
-        ReserveError, ResidentProjection, RetainedExit, RetainedRecordedOutcome,
+        ObservationTxn, ReserveError, ResidentProjection, RetainedExit, RetainedRecordedOutcome,
         RetainedStopReason, ScopeCell, ScopeControlEvent, StartupDisposition,
         classify_disposal_panic_retaining, classify_exit_retaining,
         reconcile_recorded_outcomes_retaining,
@@ -66,7 +67,7 @@ use shelterwood_core::supervisor::{
 };
 
 use admission_control::{
-    AdmissionRequest, DynamicControl, DynamicEntry, cancel_dynamic_reservation_parts,
+    AdmissionRequest, DynamicControl, DynamicEntry, DynamicState, cancel_dynamic_reservation_parts,
     reject_admission_after_disposal,
 };
 pub(crate) use admission_control::{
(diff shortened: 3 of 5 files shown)


ralexstokes:
## Review — #475 (stack 3/3, own diff vs `agent/issue-435-sever-parent` @ 4d68cf9)

Reviewed the ledger, the rewritten install section, the `insert_vacant` seam, the
AGENTS.md accounting and the three tests. Verdict: **the slice #437 signed off is
implemented and the mechanism is sound.** Two hardening fixes and one missing venue
pin are pushed as `d1238ce`; one design decision (`ManuallyDrop`) is flagged for
adjudication rather than changed unilaterally; the rest of my candidate concerns are
refuted below with evidence.

`./tools/dev just ci` on the pushed head: **PASS**, 962 tests + doctests, examples,
docs, API reachability, prelude containment, core-manifest, external-consumer, Nix fmt.

---

### 1. Per-step unwind audit of the install sequence

Everything below is enumerated from the point the ledger exists
(`driver.rs:976`) — deliberately outside both control-plane guards, so
`AdmissionInstall::drop` (`driver.rs:280`) always runs after the dynamic-state
mutex and the `ObservationTxn` guard are gone. I checked that frame ordering
holds: the txn lives in `with_observation_gate`'s frame, which is *deeper* than
`handle_admission`'s, so closure locals → txn drop (unlock + effects flush) →
`install` drop. Confirmed empirically by mutation M2 below.

| # | Step | If it unwinds here | Reserved member | Announcement | Response | Locks |
|---|---|---|---|---|---|---|
| 0 | `AdmissionInstall::new` gate debug-assert (`:119`) | ledger not yet built; `request`/`child` drop with their own obligations | reservation intact in state | none | Obligation fallback | none held ✔ |
| 1 | `state.lock()` / recheck read (`:988-995`) | entry untouched | reserved | none | ledger `complete_lost` | reservation not yet claimed ✔ |
| 2 | rejection arm: `claim_entry` (`:1001`) then `terminalize_never_started_locked` (`:1005`) — **the arm the hostile-seam test drives** | entry + runtime + projection + response all owned by the outer ledger | retired: entry drop completes the remover with `Removed`, projection retires detached | never announced | `complete_lost`, contained | `state` already dropped; gate released *before* Drop ✔ (was a violation on the previous shape — see M2/M3) |
| 3 | `claim_entry` on the healthy path (`:1011`) | as #2 | reservation retired rather than stranded `Reserved` (base left it in the map until scope teardown, blocking re-reservation) | none | `complete_lost` | ✔ |
| 4 | `insert_child` (`:1021`) — the one genuinely fallible step in the window | no arena key was recorded, so nothing is left in the arena | entry retires via ledger Drop | none | `complete_lost` | ✔ |
| 4b | `insert_child` returns `Err` (identity exhausted, `:1023-1035`) | runtime parked back in the ledger *across* the locked terminalization, then handed to the rejection | entry returned as `removed`, retired through `reject_admission_after_disposal` | none | `complete(Err(IdentityExhausted))` | ✔ |
| 5 | `record_key` / `promote_entry` / `install_entry` (`:1038-1040`) | structurally unreachable `Option` unwraps; the `assert!` at `:189` was the one *reachable-by-bug* panic under `state` → fixed, see F1 | entry retires via Drop | none | `complete_lost` | ✔ after F1 |
| 6 | `admit_child_locked` (`:1046`) | entry is already re-installed as `Resident`, so the id stays occupied (same as base); residency owns whatever prefix `admit_child_locked` installed and retires it at scope clear | ledger's projection copy retires detached; residency's own copy handled by `admit_child_locked` | `complete_lost` | ✔ |
| 7 | `finish()` invariant asserts (`:232-262`) | this *is* the old hand-written invariant arm, now the type's unwind path | projection retires detached | none (refusal published neither edge) | `complete_lost`, panic-contained, invariant stays primary | asserts run with **no** guard held — pinned by `refused_dynamic_admission_panics_after_control_plane_unlock`'s two `is_poisoned` assertions |
| 8 | commit-time effects flush inside `ObservationTxn::drop` | install already committed; ledger Drop disposes a provably non-last projection clone and completes lost — identical to base, where `request`'s Obligation did the same | — | — | — | ✔ |

Cross-cutting checks:
- **No double-install.** `install_entry` (`:216`) is the only writer, `Entry::Vacant`
  is the only accepting arm, and `entry_installed` gates the announcement.
- **No announced-without-`Removed`.** Announcement happens only inside
  `admit_child_locked`, which is only reached with the entry installed; every earlier
  unwind leaves the resident either absent or unannounced (SPEC §3.2 pairing preserved
  by #473/#474's slot + sever work).
- **Poisoning.** The dynamic-state mutex is `.expect()`ed by every caller, so a panic
  under it wedges the scope permanently. After F1, no panic site remains under it that
  isn't a structurally-unreachable `Option` unwrap of the same class the base already
  had (`entry_mut(id).expect("the matching reservation was just resolved")`). The
  observation gate is poison-tolerant by construction (`cells/gate.rs:31-34`).
- **Rejected/panicking projections reach `dispose_detached` after unlock.** `Drop`
  (`:281-283`) submits the projection first, inside a `PanicAccumulator`; `finish`
  (`:250`) and `into_rejection` (`:270`) instead drop it inline, correctly, because the
  still-owned `request.slot` provably retains the same member graph — so that drop is
  refcount traffic, not a possibly-last mailbox owner.

### 2. Re-derived lock-order exception (AGENTS.md "install half")

The exception rests on two invariants. **Neither is changed by this diff, and both
still hold.**

- *Publication.* This diff **does** change when the reservation sits in
  `DynamicState.entries`: it is now removed at `:1011` and re-inserted at `:1040`,
  where the base promoted in place. The window is entirely inside one
  continuously-held `state` guard, nothing inside it (`insert_child` → arena only,
  `promote`, `insert_vacant`) reads or re-enters dynamic state — it could not, or it
  would self-deadlock on a non-reentrant `std::sync::Mutex` — so no other thread can
  observe the gap. The externally-visible publication rule ("an exact remover sees
  either the reservation or a resident carrying its live arena key") is unchanged, and
  `refused_dynamic_admission_panics_after_control_plane_unlock` now pins the
  round-trip explicitly (`entries_mut().get(id).is_some_and(|e| !e.is_reserved())`
  after the unwind).
- *Re-homing.* Untouched. Re-derived independently: `admit_observation_gate` refuses
  any scope with a live dynamic route (`cells/scope.rs:527-530`), a live route needs a
  started driver, and a scope is parented before its driver starts — pinned by
  `driver::tests::observation::gate_handoff_rejects_a_scope_with_an_unadmitted_dynamic_reservation_without_rehoming`.
  Since `reserve` adopted the slot onto the root gate before publishing it, the slot's
  gate is still the root's gate at install time, so `with_handoff_gate`
  (`cells/member.rs:512-519`) takes its `current.shares_gate(gate)` short-circuit and
  **acquires no second gate under `state`**.

The new `debug_assert!` at `driver.rs:120` is the right encoding of exactly that
second invariant, and its venue is right: it reads both gate pointers *before* either
guard, so a fire cannot poison `state`. It is a racy read of the root's pointer, but
the property it checks is that re-homing is unconstructible, so there is no race to
lose. AGENTS.md's new sentence is accurate; `CLAUDE.md` is a symlink to it, so the
prose lands in both.

**Widening check:** the diff adds to the doubled-gate section only framework-plain
work — a `HashMap` remove/insert, an enum write, and one `Arc` clone pair
(`install.projection()`), all provably non-last. No user value is destroyed, no user
callback invoked, no panic raised (after F1) inside it. The exception does not need
re-deriving beyond the above.

### 3. Findings fixed and pushed (`d1238ce`)

**F1 — `record_key`'s `assert!` panicked under both control-plane guards.**
`driver.rs:189` (pre-fix). It is a diagnostic with a single caller following the only
key-producing step, but if it ever fired it would poison the dynamic-state mutex that
every `reserve`/`remove`/`close` caller `.expect()`s, wedging the scope's whole control
plane — precisely the case AGENTS.md rules on ("where releasing is impossible,
`debug_assert!` instead"). Downgraded to `debug_assert!` + plain assignment.

**F2 — `claim_entry` / `restore_child` had no guard at all, where the consequence is
strictly worse.** `driver.rs:173,181`. Both land under both guards, and both are silent
overwrites: an overwritten `DynamicEntry` drops its `Obligation<RemovalResponses>`
there, waking a parked remover's waker **under the observation gate**; an overwritten
`ChildRuntime` drops its terminality obligation, which re-enters that same gate
(`discharge_child_terminality` → `terminalize_never_started`) and deadlocks. Same
diagnostic treatment as F1, with the hazard named in a comment.

**F3 — the fused-cancel rejection arm had no venue coverage, and it is the arm this PR
actually repairs for the reservation.** `rejected_admission_retains_its_child_across_locked_terminalization`
covers the *runtime* half via the `!matches_reservation` branch, which by construction
claims no entry. The `matches_reservation && fused-fired` disjunct is the only arm that
claims the reservation *before* terminalizing — and on the previous shape a panic there
dropped the entry inside the gated closure, waking a parked remover under the gate.
Added `rejected_admission_wakes_its_removal_waiter_outside_the_gate`: it parks a
remover on the reserved entry with a waker that records `observation_gate().is_held()`,
fuses the latch from `MailboxControl::configure` (i.e. during conversion, past
`handle_admission`'s pre-conversion latch check and before the ledger exists), and
panics from the locked terminalization seam. `PanicTerminationMailbox` grew an optional
`fuse` for this; the existing test passes `fuse: None`.

### 4. Verified by mutation, not by name

Committed tree first, mutation second, `git checkout --` third, per house rules.

- **M1 — emptied `AdmissionInstall::drop`.** `dropping_admission_install_completes_the_lost_response`
  **and** `refused_dynamic_admission_panics_after_control_plane_unlock` both go red
  (`assertion failed: matches!(response…Terminal)`). Confirms the PR's own claim and,
  more usefully, confirms the old untested invariant arm is now pinned to the ledger's
  destructor. Note the mutation also produced `error: method complete_lost is never
  used` — the ledger is the sole owner of that contract, exactly as the doc claims.
- **M2 — closure-owned child** (hoist `install.take_child()` to the top of the gated
  closure so the runtime is a closure local again — the base shape).
  `rejected_admission_retains_its_child_across_locked_terminalization` fails with
  `left: Err(Timeout)`: the terminality fallback re-enters the held gate and deadlocks.
  So that test *would* have gone red on this PR's base, and its `recv_timeout` framing
  is the right way to catch it.
- **M3 — closure-owned entry** (restore `let removed = matches_reservation.then(…)` as a
  closure local, claimed only after terminalization). My new F3 test fails with
  `left: Some(true)` — the remover is woken **with the observation gate held**. That is
  the base-shape lock-rule violation, now pinned.
- Not mutation-checked, deliberately: the F1/F2 `debug_assert!`s. They are diagnostics
  for unreachable states and nothing may depend on them (#395 rule).

### 5. Design point flagged, not changed — `Option<ManuallyDrop<AdmissionRequest>>`

`driver.rs:108`. The doc is honest about why it's there: it makes M1 bite instead of
silently falling back through `AdmissionRequest`'s own `Obligation`. But the trade is
worth stating explicitly, because it cuts against SPEC §15.5 ("every cross-task promise
… is held as an owned value **whose destructor discharges it**, fail-closed … unwinding
alone MUST discharge every outstanding promise"). Unlike `WakerSlot`, which makes the
*unsafe* action unrepresentable, `ManuallyDrop` here makes the *safe fallback*
unavailable: it prevents no mistake, it converts one class of future mistake from
"benign, still correct" into "leaked `OneShotSender` → the caller's `Admission` future
never resolves at all", i.e. a hang, which is the one outcome §15.5 forbids. Nothing is
reachable today — all three exit paths (`finish`, `into_rejection`, `drop`) take the
request, and both consuming methods assert *before* they take it, so an assert failure
still routes through `Drop`.

If you want both properties, there is a version that keeps the destructor: make the
field a plain `Option<AdmissionRequest>` and pin the `Drop` body on the one effect field
order cannot produce — the projection's **disposal venue**. A plain field drop retires
`ResidentProjection` inline on the driver thread; the ledger retires it through
`dispose_detached`. A thread-id probe on that retirement fails under M1 just as loudly,
without giving up the fail-closed fallback. Your call — I did not change it, since it is
the PR's central structural decision and reverting it silently would have weakened your
own pin.

### 6. Refuted

- *"`drop(child)` in `Drop` destroys the user construction closure inline, against the
  detached-disposal convention."* No — `ChildRuntime::construction` is
  `runtime::Isolated<ChildConstruction>`, whose `Drop` calls `dispose_detached`
  (`shelterwood-runtime/src/disposal.rs:44-48`).
- *"The remove/re-insert weakens identity exclusion on the panic path."* No. The only
  step in the entry-removed window that can realistically panic is `insert_child`, and
  it records no arena key before it can, so a fire leaves nothing in the arena and
  correctly frees the id — strictly better than base, which stranded the reservation as
  `Reserved` until scope teardown. Every panic site *after* `install_entry` finds the
  entry already re-installed, i.e. identical to base.
- *"`request` is dropped inline on the driver thread in `Drop`, possibly as the last
  `Arc<SlotCell>` owner."* Same as base (the invariant arm dropped `request` inline
  too), and the projection submitted to detached disposal at the top of `Drop` holds
  the member graph.
- *"`finish()` lost the base's `if !std::thread::panicking()` guard around the
  diagnostic."* Correctly lost: `finish` is only reached on the non-panicking return
  path, and the containment moved into `Drop`'s accumulator, which is the correct venue.
  The hostile-waker test still shows the invariant staying primary.
- *"Removing then re-inserting could fire a removal completion."* `DynamicState::remove`
  is a bare `HashMap::remove` returning the entry by value; the entry is *moved* into
  the ledger, never dropped, so no obligation discharges inside the critical section.

### 7. Follow-ups (out of this diff, not folded in, no issues filed)

1. **Same hazard class, three untouched sibling sites.** A reserved entry is held as a
   plain local across `take_terminal_reservation` — which reaches the same hostile
   `prepare_termination` seam — while the observation gate is held:
   `cancel_dynamic_reservation_parts` (`admission_control.rs:535-545`, reached from
   `handle_admission`'s three early-rejection arms and `cancel_reservation`),
   `DynamicControl::remove`'s reserved branch (`:404-411`), and `close`'s loop
   (`:440-446`, where the whole remaining `entries` map is also live). A panic in the
   terminalization drops those entries under the gate and wakes parked removers there —
   exactly what F3 now forbids on the install lane. The ledger shape generalizes.
2. `DynamicState::insert_vacant`'s `Entry::Occupied` arm and the
   `if install.entry_installed` guard at `driver.rs:1045` are unreachable while the
   `state` guard spans remove→re-insert. They are coherent fail-closed structure, so I
   left them, but they are the two arms in the new code with no possible test.
3. `dropping_admission_install_completes_the_lost_response`'s
   `!root.observation_gate().is_poisoned()` is a weak assertion in that fixture (no gate
   is entered on that path at all, so it can only fail if the fallback both took the
   gate and panicked inside it). The equivalent assertion in
   `refused_dynamic_admission_panics_after_control_plane_unlock` *is* load-bearing.
4. On the unwind path the removal response and the admission response both complete
   before `Isolated<ChildConstruction>`'s detached disposal finishes, where the healthy
   path orders them via `dispose_definition_then` ("must … complete before any response
   releases ownership back to the caller"). Base was no better (arbitrary local drop
   order), so this is not a regression — but if that ordering is meant to be normative
   on failure paths too, the ledger's `Drop` is where it would go.

### 8. Stack

Rebase needed before merge: this branch is based on `agent/issue-435-sever-parent`
@ `4d68cf9`, and #474 has since moved to `67fba7e` ("test: pin the retirement-only
sever venue and the unannounced prune lane"). That commit touches only `cells/scope.rs`'s
`mod tests` (from line ~2181), while this PR's `cells/scope.rs` change is at ~372, so the
rebase should be textually clean. Repo rule: rebase, not merge.

ralexstokes:
Rebased onto `origin/main` (`62f4fa1`, now carrying #473 and #474) and applied the ruling on §5's flagged design point: **plain `Option<AdmissionRequest>` + field order, not `ManuallyDrop`** (`f03a6ac`).

**What changed.** The request field is plain again, so its own `Obligation` still discharges the caller's promise on any path the ledger's destructor fails to cover — the SPEC §15.5 property `ManuallyDrop` was trading away. To keep the fallback from being *merely* fail-closed, the fields are now declared in the destructor's own order (`projection`, `child`, `entry`, `request`), so even a bare field-drop keeps the child's terminality ahead of the entry's removal completion and both ahead of the response. Everything else in the PR is unchanged.

**On the venue pin proposed in review — it does not bite, and I checked rather than assumed.** Mutation: replace `panics.run(move || runtime::dispose_detached(projection))` with `panics.run(move || drop(projection))` in `AdmissionInstall::drop` (retire inline on the driver thread instead of detached). Result: **785 tests, 0 failures.** Two reasons, both structural:

- `ResidentProjection` is two `Arc`s and owns nothing exclusively, so a thread-id probe can only observe the venue when the ledger's copy is the *last* owner. In `dropping_admission_install_completes_the_lost_response` the fixture's dynamic state and reservation both co-own the member graph, so no destructor runs at ledger drop at all.
- Even in a fixture stripped down to sole ledger ownership, the projection's disposal job races the ledger's own inline `child`/`entry`/`request` drops for last-owner, so `assert_ne!(observed, current_thread)` would be a coin flip, not a pin. Serializing the lane to remove the race does not work either: inside a runtime `dispose_detached` dispatches to the **blocking pool**, not the single shared fallback thread, so occupying one job does not queue the next.

**What does fence the response half:** deleting `complete_lost()` from the destructor fails the build — `-D dead-code` fires because the ledger's `Drop` is its only caller. Weaker than a behavioural pin, but it is a real gate, and the deletion it guards is benign by construction now anyway, which is the point of the ruling.

Recording the residual honestly: with this shape, deleting the destructor's body is *semantically* a no-op except for the projection's disposal venue, and that venue has no observable pin. Worth a follow-up only if a general disposal-lane observation seam ever lands.

`./tools/dev just ci` green on the rebased + reworked head: **990 tests**, doctests, examples, docs, API reachability, prelude containment, core-manifest, external-consumer, nixfmt.

